PR #21, opened 2025-10-24
Base array and linear algebra functionality

### LINKED ISSUE #25: Base Array, Linear Algebra, and Distribution Functionality
- [x] Implement base array utility methods, defining what we mean by "vector", "matrix", "batch array", etc.
- [x] Implement linear operator class 
- [ ] Reorganize/update distributions classes, including implementing Gaussian distribution that works with linear operators
- [ ] Add unit tests and tutorial-style examples

arob5:
This is an early stage draft PR to share some initial ideas on standardizing our core scientific computing frameworks for this package. At present, we have decided to build using numpy arrays, but should keep in mind the idea of supporting jax or other array backends in the future. 

The current state of this code is very much in flux, and will likely change often. The initial files I recommend looking at are:
- `custom_types.py`: type hints to use for array and array-like objects. 
- `linalg/utils.py`: validation/coercion to scalars/vectors/matrices/square matrices. Have to make judgement calls on what we allow here - I'm open to changes. Other utilities will be added as this PR progresses. 
- `linalg/linop.py`: after looking through some other open-source projects, I was convinced that the "linear operator" abstraction is a very useful thing to have, and would fit nicely within probpipe. See scipy, pytorch, lineax, gpjax for examples of linear operator classes. This file currently contains my very early thoughts on a simple linear operator abstraction suitable for our purposes. 

Added unit tests:
- `tests/linalg/test_utils.py`
- `tests/linalg/test_linop.py`

Suggested Updates:
- [ ] Use `Array` and `ArrayLike` type aliases from `probpipe/custom_types.py`
- [ ] Change `Multivariate` to `RealVector` (or something like this)
- [ ] Currently `Multivariate.cov` is defined to return type Array; generalize to `Array | LinOp`
- [ ] Create a `distributions` sub-folder. Should this be `probpipe/distributions` or `probpipe/core/distributions`
- [ ] Standardize variable name for number of samples (currently there are different conventions); options: n_samples, num_samples, num_samp, n_samp, n, etc.
- [ ] In `Multivariate.sample()` have the number of samples default to 1.
- [ ] Have `Multivariate.log_density()` return shape `(n,)` (aligning with the `_ensure_vector()` default behavior). In the future if we want to generalize (e.g., a batch of distributions/multioutput/etc.) then could do `(n,b)` (aligning with `_ensure_batch_vector()`).
- [ ] Shorten `Multivariate.dimension` to `Multivariate.dim`?
- [ ] Define alias for float type / move other type definitions to `custom_types.py`
- [ ] Add `dtype` argument to `_ensure` functions, providing a central location for array dtype conversion. 

Questions:
- [ ] Does the `Multivariate` class essentially mean "a sample from this distribution has shape (d,)".
- [ ] Naming convention for distributions? e.g., GaussianDistribution vs. Gaussian vs. GaussianDist

Closes #25

### DIFF
--- a/tests/test_distributions.py
+++ b/tests/test_distributions.py
@@ -2,6 +2,7 @@
 import numpy as np
 import pytest
 
+from probpipe.distributions import Distribution
 from probpipe import EmpiricalDistribution
 
 # Conditional imports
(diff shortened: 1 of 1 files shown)


examples/Untitled.ipynb
jhuggins: Seems like this should be deleted?

PR #34, opened 2025-11-18
New architecture: Node, workflow, module

erozercan:
As a start three new files are added: node.py, module_node.py, and workflow_node.py

node.py is basically an abstract base class

workflow_node.py is the WorkflowFunction subclass. It provides automated distribution conversion for a single function, 
defaults to being a Prefect node (as task). It doesn't include broadcasting support yet. The type conversion logic is the same that of Module's. 

module_node.py: is the Module subclass. It is a container for multiple workflow functions and shared state between them, a module's dependencies is the union of all the dependencies of workflow functions. It supports Approach 1-  module always passes itself as first argument. It exposes .state as an unrestricted dictionary. Also it allows each workflow function to access module state via the explicit first argument module.


jhuggins (changes requested):
I don't think the way the module node is written will quite work. We want the workflow functions to created within subclasses of `ModuleNode`. Something like:

```python
class MyModule(ModuleNode):
    @wf
    def fit(self, node_dep : Node, prior : Distribution):
        # does stuff

    @wf
    def predict(self, dist : Distribution):
        # does other stuff
```

Here, `@wf` would effectively turn `fit` and `predict` into `Workflow` nodes. The `MyModule` class would automatically get its dependencies determined by the collection of node arguments to `fit` and `predict` (in this case just node_dep). When a `MyModule` object is created, the dependencies required by each workflow function would be passed on to instantiate that workflow function. If multiple `@wf` functions in a module have the same name, we can enforce that they are the same type too and get reused across workflow functions.

erozercan:
Workflow nodes are now defined as workflow functions (@wf) inside module classes, rather than as standalone nodes. I introduced AbstractModule and @abstractwf (@wf + @abstractmodule) to allow modules to declare interfaces (e.g. data sources, inference methods, predictive steps) without committing to a specific implementation. Concrete modules implement these interfaces and can be swapped without changing the surrounding pipeline. Overall, this clarifies the separation between module roles and execution logic and lays the groundwork for cleaner APIs and future signature improvements...

You can check out how they are used by looking at our time series example.

PR #40, opened 2026-03-12
Add broadcasting support for Workflow nodes

### LINKED ISSUE #13: Add support for broadcasting
We would like to be able to call a module's workflow functions by passing in a `Distribution` object over an input or multiple inputs and then have the function return a distribution over the outputs. Since a single `Distribution` object will define a joint distribution over multiple parameters that are given as separate inputs. Therefore, we need to be able to propagate uncertainty in such cases. This requires
1. Named variables/components within joint distributions 
2. A standard for how to jointly propagate.

Here's an illustration of broadcasting in general and how the distribution could be specified as an input multiple times to handle joint broadcasting. Then, on the backend, the distribution class can take care of joint sampling by checking whether multiple arguments got the same distribution assigned to them. 

```python
from probpipe import Workflow, ConditionalDistribution
from probpipe.modeling import MissingDataImputer, PredictiveModel
from probpipe.typing import ArrayLike


@Workflow
def example_workflow(covariates: ArrayLike, responses: ArrayLike, 
					 target_model_data: ArrayLike
					 imputer: MissingDataImputer, 
					 reg_model: PredictiveModel,
					 target_model: ConditionalDistribution) -> Distribution:
	"""The purpose of this workflow is to fit a "target model" that
	relies on regression parameter estimates from an auxiliary 
	model. Due to model misspecification, a cut posterior approach
	is used, with further UQ needed because some covariate values
	are missing. The workflow returns the posterior mean of the 
	parameters of interest, which have remaining uncertainty 
	due to Monte Carlo error when fitting the second model. 
	"""
	# first, impute distribution for missing covariates
	full_covariate_dist = imputer.impute(covariates=covariates)
	# next, fit regression model, propagating covariate uncertainty
	# here, the `covariates` argument is an array, but we 
	# pass in the `Distribution` object `full_covariate_dist`
	reg_posterior = reg_model.fit(covariates=full_covariate_dist, 
								  responses=responses)
	# The regression model is auxiliary; need to use its posterior as 
	# input to fit target model, where parameters of the regression
	# posterior are split into two arguments (beta and sigma2).
	# Since covariates get used again, also need to propagate
	# the covariate uncertainty here. 
	target_posterior = target_model.conditional(
		beta=reg_posterior, sigma2=reg_posterior, 
		covariates=full_covariate_dist, data=target_model_data)
		
	# return posterior mean of parameter of interest with, uncertainty
	return target_posterior.mean('alpha')
```

jhuggins:
## Summary
- When a `Distribution` is passed where a concrete type is expected, `Workflow` automatically samples and calls the wrapped function per-sample, returning an `EmpiricalDistribution` (or a plain list for non-numeric results)
- `EmpiricalDistribution` arguments with small support are enumerated (cartesian product) with weight propagation instead of resampled, using a greedy budget-aware cutoff when the product of support sizes would exceed `n_broadcast_samples`
- Parallel execution via `ThreadPoolExecutor` (plain Python) or Prefect `task.map()` (when `workflow_kind` is set)
- `n_broadcast_samples` configurable at construction and overridable at call time

Closes #13

### DIFF
--- a/tests/test_broadcasting.py
+++ b/tests/test_broadcasting.py
@@ -30,13 +30,16 @@ class TestBroadcastingSingleArg:
         def add_one(x: NDArray) -> NDArray:
             return x + 1.0
 
-        w = Workflow(func=add_one, n_broadcast_samples=30)
+        w = Workflow(func=add_one, n_broadcast_samples=500)
         g = Gaussian(mean=np.array([0.0]), cov=np.eye(1), rng=rng)
         result = w(x=g)
 
-        # Mean of result should be approximately 1.0
-        assert result.n == 30
+        assert result.n == 500
         assert result.dim == 1
+        # add_one shifts the mean by 1: Gaussian(0,1) + 1 → mean ≈ 1.0
+        assert abs(float(result.mean()) - 1.0) < 0.3, (
+            f"Expected mean ≈ 1.0, got {float(result.mean()):.3f}"
+        )
 
     def test_scalar_return(self, rng):
         def norm(x: NDArray) -> float:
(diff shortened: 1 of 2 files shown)


tests/test_broadcasting.py
```diff
@@ -0,0 +1,255 @@
...
+            return np.tile(x, (n_samples, 1))
+
+        w = Workflow(func=sample_wrapper, n_broadcast_samples=5)
+        g = Gaussian(mean=np.array([1.0]), cov=np.eye(1), rng=rng)
+        # n_samples=3 should go to the function, not be intercepted
+        result = w(x=g, n_samples=3)
+
+        assert isinstance(result, EmpiricalDistribution)
+        # Broadcasting uses the construction default (5), n_samples=3 goes to function
+        assert result.n == 5
```
arob5: Maybe worth making `MIN_BROADCAST_SAMPLES` a global or a class variable? Then this test wouldn't require hard-coding the default value.

Issue #29, opened 2025-11-05
Automatic Distribution/Type Conversion System


jhuggins:
The goals of the auto-type conversion system are to 
1. make is easy to integrate diverse probabilistic and quasi-probabilistic ML/statistics packages into ProbPipe and 
2. enforce/encourage a distribution-centric approach.

Here's a first draft at an approach:
- Define a *converter registry*, which tracks a set of available *type converters*. 
- A type converter operates in terms of *promises*, which are a list of types that the converter promises it can change any object of that type into an equivalent ProbPipe `Distribution` subtype. It can also provide a promise to convert certain ProbPipe `Distribution` types to some other set of types. For example, a scipy type converter could promise to convert the standard list of `scipy.stats` distributions to/from equivalent PropPipe types. 
- Type converters also provide a `check` function, which returns information about whether a specific conversion could be performed, how the conversion  *would* be done (e.g., exact vs. approximate), and whether additional options are available (e.g., setting the approximation error).

PR #56, opened 2026-03-22
Add converter registry for automatic distribution type conversion

jhuggins:
## Summary

Implements issue #29: automatic distribution/type conversion system.

- New `probpipe.converters` package with a global `converter_registry` singleton
- **`Converter` ABC** with `check()` (cheap feasibility probe with `ConversionInfo` including method, error estimate, and cost) and `convert()` (actual conversion)
- **Three built-in converters:**
  - `ProbPipeConverter` (priority 100) — handles all ProbPipe↔ProbPipe conversions via moment-matching
  - `TFPConverter` (priority 50) — bidirectional raw `tfd.*` ↔ ProbPipe with exact parameter extraction for known types
  - `ScipyConverter` (priority 25) — bidirectional `scipy.stats.*` ↔ ProbPipe (optional dependency)
- `Distribution.from_distribution()` delegates to the registry (backward compatible)
- `WorkflowFunction` uses `converter_registry.is_distribution_type()` instead of hard-coded `DISTRIBUTION_TYPES`, enabling external distribution types to participate in broadcasting
- Users can register custom converters for their own distribution types

## Test plan
- [x] All 714 existing tests pass (no regressions)
- [x] 34 new converter tests covering: registry basics, ProbPipe↔ProbPipe, TFP↔ProbPipe, scipy↔ProbPipe, custom converter registration, from_distribution() backward compatibility

### DIFF
--- a/probpipe/converters/_probpipe.py
+++ b/probpipe/converters/_probpipe.py
@@ -1,10 +1,13 @@
 """ProbPipe-to-ProbPipe converter.
 
-Converts between ProbPipe distribution types using exact parameter
-extraction (same-class) or moment-matching via ``expectation()``
-(cross-family).  When moment-matching, the BootstrapDistribution
-objects for the estimated mean and variance are stored in provenance
-metadata so users can inspect conversion error.
+Handles conversions between ProbPipe distribution types:
+
+- **Same-class**: returns the source unchanged (no copy, no provenance).
+- **Cross-family**: moment-matches using the source's ``mean()`` and
+  ``variance()`` methods, which delegate to TFP for analytical moments
+  or fall back to Monte Carlo via the ``@monte_carlo`` decorator.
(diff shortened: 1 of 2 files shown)


probpipe/converters/_probpipe.py
```diff
@@ -0,0 +1,503 @@
...
+#
+# Each function has signature:
+#   (source, key, **kwargs) -> Distribution
+# ---------------------------------------------------------------------------
+
+def _convert_to_normal(source, key, **kw):
+    from ..distributions.continuous import Normal
+    if isinstance(source, Normal):
+        return source
+    _, m, v = _sample_moments(source, key, kw.pop("num_samples", 1024))
```
arob5: Understandably may want different default num samples for different distributions, but defaults are all 1024 so seems like this should be defined as a constant.

PR #74, opened 2026-03-29
Add BroadcastDistribution for joint input–output broadcasting

jhuggins:
## Summary

- **`BroadcastDistribution`** replaces `EmpiricalDistribution` as the return type of `WorkflowFunction` broadcasting, preserving the joint distribution over broadcast inputs and function outputs
- **`marginalize()`** returns the output-only marginal: `_ArrayMarginal` (stackable arrays), `_MixtureMarginal` (distribution outputs as a mixture), or `_ListMarginal` (non-stackable)
- `_MixtureMarginal` dynamically supports protocols (`SupportsSampling`, `SupportsMean`, `SupportsVariance`, `SupportsLogProb`) based on what the component distributions support
- `marginalize=True` parameter on `WorkflowFunction` for memory opt-out (skips storing input samples)
- `_ArrayMarginal` supports `SupportsExpectation` for MC expectation computation
- Updated all tests, docs (README, getting-started, index, API), and example notebooks (03, 04, 06, 08) to use the new API

## Test plan

- [x] All 1296 tests pass (`pytest -x -q`)
- [x] 38 new unit tests in `test_broadcast_distribution.py`
- [x] 34 rewritten tests in `test_broadcasting.py`
- [x] Example notebooks 03, 04, 06, 08 re-executed successfully
- [ ] CI passes on Python 3.12 and 3.13

### DIFF
--- a/probpipe/core/distribution.py
+++ b/probpipe/core/distribution.py
@@ -1763,7 +1763,7 @@ def _make_marginal(
     *,
     output_distributions: list | None = None,
     name: str | None = None,
-) -> _ArrayMarginal | _MixtureMarginal | _ListMarginal:
+) -> MarginalizedBroadcastDistribution:
     """Factory to construct the appropriate marginal subtype."""
     if output_distributions is not None:
         return _make_mixture_marginal(output_distributions, weights, name=name)
@@ -1805,6 +1805,20 @@ class BroadcastDistribution(Distribution[dict], SupportsSampling, SupportsNamedC
     Call :meth:`marginalize` to obtain the output-only marginal, which
     supports moment protocols (mean, variance, etc.) when the output
     data permits.
+
+    .. note::
+
+       ``BroadcastDistribution`` does **not** inherit from
+       :class:`~probpipe.distributions.joint.JointDistribution`.
+       ``JointDistribution`` requires all leaves to be
+       ``ArrayDistribution`` instances with TFP shape semantics
+       (``batch_shape``, ``event_shape``), but a broadcast output can be
+       any type — arrays, distributions, strings, etc. — and input
+       samples are plain arrays without distribution metadata.  The two
+       hierarchies serve different roles: ``JointDistribution`` models
+       structured probabilistic variables; ``BroadcastDistribution``
+       records the empirical input–output mapping of a function
+       evaluation.
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


probpipe/core/distribution.py
```diff
@@ -1429,3 +1432,514 @@ def __repr__(self) -> str:
...
+"""
+
+
+def _make_marginal(
+    output_samples: Any,
+    weights: Array | None,
+    *,
+    output_distributions: list | None = None,
+    name: str | None = None,
+) -> _ArrayMarginal | _MixtureMarginal | _ListMarginal:
```
arob5: Reason not to type hint the output using `MarginalizedBroadcastDistribution` here?

probpipe/core/distribution.py
```diff
@@ -1429,3 +1432,514 @@ def __repr__(self) -> str:
...
+    # Single array result (e.g., from vmap)
+    return _ArrayMarginal(jnp.asarray(output_samples, dtype=jnp.float32), weights, name=name)
+
+
+# ---------------------------------------------------------------------------
+# BroadcastDistribution — joint over broadcast inputs and function output
+# ---------------------------------------------------------------------------
+
+
+class BroadcastDistribution(Distribution[dict], SupportsSampling, SupportsNamedComponents):
```
arob5: Should this inherit from `JointDistribution`? Still need to look into how the joint distribution abstraction has changed in this recent round of updates, so could be an irrelevant question

probpipe/core/distribution.py
```diff
@@ -1429,3 +1432,514 @@ def __repr__(self) -> str:
...
+        self._weights = weights
+        self._broadcast_args = list(broadcast_args)
+        self._name = name
+        self._approximate = True
+        self._marginal_cache: _ArrayMarginal | _MixtureMarginal | _ListMarginal | None = None
+
+    # -- basic properties ---------------------------------------------------
+
+    @property
+    def n(self) -> int:
```
arob5: The `n` property seems to be used in a variety of places throughout the code. It would seem to be a good idea to standardize its use and document this in the style guide.

jhuggins: Done -- documented the .n property convention in STYLE_GUIDE.md (section 1.8). Lists all classes that define .n with their meanings and provides guidance for new classes.

PR #94, opened 2026-04-01
Make Likelihood generic, replace IterativeForecaster with IncrementalConditioner

### LINKED ISSUE #75: Likelihood, GenerativeLikelihood, and IterativeForecaster should all be generic rather than specific to arrays
Furthermore, the IterativeForecaster implementation should be improved and aligned with current implementation, including using `SimpleModel` to represent the model, then defaulting to condition_on, but supporting some other condition_on workflow function. The name is also not aligned with its functionality. Call it IncrementalConditioner or something like that

jhuggins:
## Summary

Closes #75.

- `Likelihood[P, D]` and `GenerativeLikelihood[P, D]` now use PEP 695 generics with `P` (parameter type) and `D` (data type) instead of `ArrayLike`
- `IterativeForecaster` replaced by `IncrementalConditioner[P, D]`:
  - Takes a prior `Distribution[P]` and `Likelihood[P, D]` directly
  - Builds `SimpleModel` internally on each `update()` (current posterior becomes the prior)
  - Defaults to `condition_on` for posterior updates; accepts a custom `condition_fn` workflow
  - Simpler API: `update(data)` instead of `update(approx_post, likelihood, data)`
- `GenerativeLikelihood` no longer required by `IncrementalConditioner` (was stored but never called)

## Test plan

- [x] Full test suite passes (1369 passed, 5 skipped)
- [ ] CI passes on Python 3.12 and 3.13


probpipe/modeling/_simple.py
```diff
@@ -62,6 +62,19 @@ def __init__(
...
+        # are truthful: the methods only exist when the prior supports them.
+        if isinstance(prior, SupportsLogProb):
+            self._log_prob = self._prior_log_prob
+            self._unnormalized_log_prob = self._prior_log_prob
+            self._prob = lambda value: jnp.exp(self._log_prob(value))
+            self._unnormalized_prob = lambda value: jnp.exp(self._unnormalized_log_prob(value))
+        if isinstance(prior, SupportsSampling):
+            self._sample = self._prior_sample
+            self._sampling_cost = prior._sampling_cost
+            self._preferred_orchestration = prior._preferred_orchestration
```
arob5: Confused why these are set to the prior density/sampler. If a `SimpleModel` is a distribution, then presumably it encodes the joint distribution.

arob5 (approved):

PR #96, opened 2026-04-01
Fix potential bugs in main

erozercan:
When trying to understand what's new in Probpipe, I looked closely how they were implemented and check for possbile bugs. I am following the new features from the notebooks under the `/examples` directory. I started from `02_transformations.ipynb` that's why I named the branch as bug_fix_transformations. But, I think I don't think I need to create a new branch and PR that corresponds to each notebook. So, I will continue from this branch and PR to fix the possible bugs across the all new implementations.


jhuggins (commented):
All looks good. We just need to resolve whether we should adjust the protocols.

Issue #98, opened 2026-04-01
Improve handing of weights


arob5:
Weighted samples show up in many different places across the ProbPipe codebase. Two examples are `EmpiricalDistribution` and `BroadcastReplicateDistribution`. There seems to be a good amount duplicated logic at various points. This will likely influence #97.

We should standardize this to avoid duplicating logic and standardize the way we use weights. 

## Considerations

1. Consistent API for user-facing functions that accept weights. This should include option to pass log unnormalized weights. 
2. Centralized function to handle validation and numerically-stable normalization.
3. It perhaps makes sense to have some lower-level functions accept already normalized weights to avoid repeating weight validation (assuming that this has already been handled), but need to think about the right way to do this.
4. Need to define weights consistently so that they work both for TFP-style shape semantics, as well as non-array generic settings (e.g., `EmpiricalDistribution[T]` works with numpy object arrays.
5. As part of this update, we may also want to write standardized functions for computing weighted means, variances, and covariances over arrays (or pytrees of arrays). This logic is also duplicated throughout the codebase; one example is in the KDE code.

Need to determine best way to do this, but some potential ideas:

1. Move common logic to `_utils.py`
2. Define a `Weights` class that handles all this internally but acts like a normal array

PR #99, opened 2026-04-01
Add protocol-based distribution conversion and KDEDistribution

### LINKED ISSUE #95: Distribution conversion automation
Currently there is no automatic conversion if a distribution doesn't support a needed functionality (that is, a specific Protocol). A canonical example is a discrete approximation (such as from MCMC) is available but SupportLogProb is needed. The distribution conversion registry should provide a way to convert a distribution without SupportsLogProb into one that satisfies SupportsLogProb. This should be doable with minimal changes. The crux is to include canonical / default approaches for converting between SupportsSampling and SupportsLogProb, perhaps with more specialized methods in certain cases, such as for EmpiricalDistribution or when there is support for computing expectations or means/variances. 

Example:
```python
# MCMC approximation 
posterior = condition_on(my_model, data)
# automatic conversion because prior must inherit SupportsLogProb
second_model = SimpleLogProbModel(likelihood=likelihood, prior=posterior) 
# uses NUTS be default 
second_posterior = condition_on(second_model, other_data)

# no conversion if only sampling needed 
alt_second_model = SimpleSamplingModel(likelihood=likelihood, prior=posterior) 
# uses SBI (e.g., ABC)
alt_second_posterior = SimpleSamplingModel(likelihood=likelihood, prior=posterior) 
```

jhuggins:
## Summary

- **Protocol-based conversion**: `converter_registry.convert(dist, SupportsLogProb)` now accepts `@runtime_checkable` protocols as targets, resolving to concrete types via registered resolver functions. This enables automatic conversion of sampling-only distributions (e.g., MCMC posteriors) into distributions that support density evaluation.
- **KDEDistribution**: New Gaussian KDE distribution backed by TFP `MixtureSameFamily`, satisfying `SupportsLogProb` + `SupportsSampling` + moment protocols. Provides a nonparametric alternative to moment-matching for protocol conversion.
- **WorkflowFunction integration**: `_convert_distributions()` now handles protocol type hints, auto-converting distributions that don't satisfy the required protocol.

## Architecture

Protocol resolution follows the existing `Converter` pattern rather than special-casing the registry. A new `ProtocolConverter` (priority 200) intercepts protocol targets, resolves them to concrete types via registered resolvers, and delegates back to the registry for the actual conversion. `ConverterRegistry` remains a simple dispatcher.

## Changes

- `probpipe/converters/_protocol_converter.py` – New `ProtocolConverter` class with `_is_protocol()`, resolvers, delegation to registry
- `probpipe/converters/__init__.py` – Register `ProtocolConverter` with `SupportsLogProb` and `SupportsSampling` resolvers
- `probpipe/converters/_probpipe.py` – `_convert_to_kde()` + dispatch table entry
- `probpipe/distributions/kde.py` – New `KDEDistribution` class
- `probpipe/core/node.py` – Protocol conversion in `_convert_distributions()`
- `probpipe/distributions/__init__.py` – Export `KDEDistribution`
- `probpipe/__init__.py` – Top-level export
- `tests/test_converters.py` – 25 new tests

## Test plan

- [x] `TestProtocolConversion` (9 tests): already-satisfied returns same, scalar/MV empirical conversion, check protocol, unregistered raises, from_distribution API, provenance
- [x] `TestKDEDistribution` (16 tests): construction, log_prob, sampling, moments, weights, bandwidth, protocols, conversions, repr
- [x] Full test suite: 1393 passed, 5 skipped, 0 failures

Closes #95

### DIFF
--- a/probpipe/distributions/kde.py
+++ b/probpipe/distributions/kde.py
@@ -9,41 +9,23 @@ that supports density evaluation.
 
 from __future__ import annotations
 
-from collections.abc import Callable
-from typing import Any
-
-import jax
 import jax.numpy as jnp
 import tensorflow_probability.substrates.jax.distributions as tfd
 
-from ..core.distribution import ArrayDistribution, _mc_expectation
+from ._tfp_base import TFPDistribution
 from ..core.constraints import Constraint, real
-from ..core.protocols import (
-    SupportsCovariance,
-    SupportsLogProb,
-    SupportsMean,
-    SupportsSampling,
-    SupportsVariance,
-)
-from ..custom_types import Array, ArrayLike, PRNGKey
+from ..custom_types import ArrayLike
 
 __all__ = ["KDEDistribution"]
 
 
-class KDEDistribution(
-    ArrayDistribution,
-    SupportsSampling,
-    SupportsLogProb,
-    SupportsMean,
-    SupportsVariance,
-    SupportsCovariance,
-):
+class KDEDistribution(TFPDistribution):
     """Gaussian kernel density estimate as a ProbPipe distribution.
 
     Wraps a TFP ``MixtureSameFamily(Categorical, MultivariateNormalDiag)``
     to provide a smooth density approximation from a set of weighted
-    samples.  Satisfies :class:`SupportsLogProb` and
-    :class:`SupportsSampling`.
+    samples.  Inherits all protocol implementations from
+    :class:`TFPDistribution`.
 
     Parameters
     ----------
@@ -59,9 +41,6 @@ class KDEDistribution(
     name : str or None
         Distribution name for provenance.
     """
-
-    _sampling_cost: str = "low"
-    _preferred_orchestration: str | None = None
 
     def __init__(
         self,
@@ -119,7 +98,7 @@ class KDEDistribution(
             components_distribution=components,
         )
 
-    # -- Distribution interface ------------------------------------------------
+    # -- KDE-specific properties -----------------------------------------------
 
     @property
     def name(self) -> str | None:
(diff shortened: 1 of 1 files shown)


probpipe/converters/_registry.py
```diff
@@ -38,16 +45,66 @@ class ConversionInfo:
...
+    """Choose Normal or MultivariateNormal based on event dimensionality."""
+    from ..distributions.continuous import Normal
+    from ..distributions.multivariate import MultivariateNormal
+
+    try:
+        event_shape = dist.event_shape
+    except AttributeError:
+        # If no event_shape, try dim
+        try:
+            if dist.dim <= 1:
```
arob5: When do we use `dim`? `PyTreeArrayDistribution` defines `event_size`, which seems like what we'd want here.

probpipe/distributions/kde.py
```diff
@@ -0,0 +1,216 @@
...
+        sample_var = jnp.einsum("n,nd->d", self._weights, diff ** 2)
+        return sample_var + self._bandwidth ** 2
+
+    def _cov(self) -> Array:
+        m = self._mean()
+        if self._scalar:
+            return self._variance()
+        flat = self._samples
+        diff = flat - m
+        sample_cov = jnp.einsum("ni,nj,n->ij", diff, diff, self._weights)
```
arob5: Weighed sample cov and sample var logic is likely duplicated in multiple points throughout the codebase. My thought is that this should be standardized in a utils file or something like this, perhaps in conjunction with standardizing the weight code: #98. This would avoid redundancy and clearly define the shape requirements/assumptions.

PR #120, opened 2026-04-08
Scalability validation via prefect

### LINKED ISSUE #90: validate scalability when using prefect
Apply to an example where we need to distribute tasks across machines

erozercan:
Goal is to close #90 

---

For more comprehensive planing read `global-prefect-config_plan.md`

### Summary:

- Adds a plan document (`global-prefect-config_plan.md`) for introducing a global Prefect configuration system to ProbPipe
- Currently, `WorkflowFunction` orchestration (`workflow_kind`, task runner) must be set per-instance, which means high-level ops like `condition_on` cannot use Prefect at all
- This plan proposes a `prefect_config` singleton that lets users set orchestration defaults once and have them apply everywhere- including the bagged posterior broadcasting path

### Key design decisions:

- `_UNSET` sentinel distinguishes "not specified" (use global) from explicit `None` (no Prefect)
- Lazy resolution via `effective_workflow_kind` property so config changes after construction take effect
- Graceful fallback: global config silently degrades if Prefect isn't installed; explicit per-instance settings raise `ImportError`
- Task runner support: `prefect_config.task_runner` is threaded into `@flow` decorators for Ray/Dask distribution

### Test plan:
- 12 new tests covering defaults, overrides, global off switch, graceful fallback, validation
- All existing `test_prefect_orchestration.py` tests pass unchanged


jhuggins (commented):
1. We have a builtin GLM likelihood that you should use rather than rolling your own
2. It would be nice to return the full posterior, rather than just the mean. This is a more realistic scenario 
3. We should set the task runner globally rather than passing it to each workflow function on creation. We may need to make other adjustments to how workflow functions are assigned to tasks or flows so things work smoothly with prefect without the user having to make lots of manual choices

probpipe/core/node.py
```diff
@@ -334,7 +337,7 @@ def __init__(
         self,
         *,
         func: Callable,
-        workflow_kind: str | None = None,   # "task" or "flow" or None
+        workflow_kind: WorkflowKind | str | None = WorkflowKind.DEFAULT,
```
jhuggins: add a follow-up issue to remove str and None as options later

probpipe/core/node.py
```diff
@@ -348,7 +351,13 @@ def __init__(
         self._func = func
         self._sig = inspect.signature(func)
         self._hints = get_type_hints(func)
-        self._workflow_kind = workflow_kind
+        # Convert legacy string / None values to WorkflowKind enum
+        if workflow_kind is None:
```
jhuggins, line 379: remove this legacy conversion code as part of the follow-up issue too

probpipe/core/node.py
```diff
@@ -391,6 +400,46 @@ def __init__(
...
+        ``TASK`` / ``FLOW`` with Prefect missing raises ``ImportError``
+        for explicit per-instance settings, but falls back to ``OFF``
+        for values inherited from global config (graceful degradation).
+        """
+        raw = self._workflow_kind_raw
+
+        # 1. Per-instance explicit (non-DEFAULT) override
+        if raw is not WorkflowKind.DEFAULT:
+            if raw in (WorkflowKind.TASK, WorkflowKind.FLOW) and task is None:
+                raise ImportError(
```
jhuggins: I'm not sure we want this to be an error – maybe just a one-off warning?

global-prefect-config_plan.md
jhuggins, line 1: remove before merging

jhuggins (changes requested):
Looking good! Just a few small items to address. I think we want TASK/FLOW to mean, "if available, run this as a prefect task/flow" – having it fail when prefect isn't installed is very aggressive. But some kind of warning message might make sense to give the user a heads up

probpipe/core/node.py
erozercan: Added # TODO: "remove str | None in follow-up issue" comment to the type hint

probpipe/core/node.py
erozercan: Added # TODO: "remove this legacy conversion in follow-up issue" comment to the conversion block

probpipe/core/node.py
erozercan: Changed `raise ImportError(...)` to `warnings.warn(...)` + `return WorkflowKind.OFF`. Now when someone requests `TASK`/`FLOW` but Prefect isn't installed, they get a friendly warning instead of a crash, and execution continues normally without Prefect.

jhuggins (approved):

Issue #93, opened 2026-03-31
BootstrapReplicateDistribution should support an arbitrary input distribution as long as it supports sampling


jhuggins:

jhuggins:
Resolved by PR-B (#155). The generic `BootstrapReplicateDistribution[T]` accepts a `SupportsSampling` source directly — when the source is sampleable, each replicate is `n` i.i.d. draws from `source._sample`. See the regression test in `tests/test_bootstrap_replicate.py::test_generic_sampling_source`.

(GitHub didn't auto-close on PR-B's merge because the PR body listed multiple issues in a comma-separated `Closes …, …, …` line, which the auto-close keyword only recognises as a single match. Closing manually now.)

PR #123, opened 2026-04-10
feat: Record/RecordArray hierarchy + bracket-only field access + NumericRecordDistribution

### LINKED ISSUE #92: make JointDistribution generic
The base implementation shouldn't depend on JAX. Similar to other recent changes, `JointArrayDistribution` will serve the same purpose as the current `JointDistribution`. Other distributions that inherit from `JointDistribution` need to be adjusted to inherit from either `JointDistribution` or `JointArrayDistribution`, as appropriate. For example, `ProductDistribution` should be generic. And `JointEmpirical` may need to be also divided to into `JointEmpirical` and `JointArrayEmpirical`

### LINKED ISSUE #113: Design: data ingestion from xarray and future data backends
### Background

ProbPipe currently has no data ingestion abstraction. Users with xarray data must call `.values` before passing arrays to any ProbPipe function. As xarray datasets become common inputs — and as future backends (dask, Polars, Arrow, database-backed arrays) emerge — the ingestion boundary needs to be formalized.

---

### Current data flow

All inference methods immediately materialize `observed` into backend-native arrays:

- **JAX methods** (TFP, RWMH): `jnp.asarray(observed)` at the top of `_build_target_log_prob`
- **PyMC/nutpie**: pass `observed` directly to the backend (which handles its own conversion)
- **CmdStan**: expects `dict`, merges with model data

`condition_on` itself is typed `observed: Any` and never inspects the type — it passes through to the inference method. The `@workflow_function` broadcasting layer handles `Distribution` objects transparently (e.g., `BootstrapReplicateDistribution` triggers automatic sampling).

This means any object that satisfies `jnp.asarray()` or `__array__` already works. The gap is xarray `Dataset`/`DataArray` objects (which carry coordinate metadata that should be preserved) and future lazy/remote backends (which require explicit materialization before entering JAX).

---

### Proposed approach: `materialize()` utility

A lightweight conversion function that handles the data ingestion boundary:

```python
def materialize(data, *, variables=None):
    """Convert structured data to backend-native arrays.

    Parameters
    ----------
    data : xr.Dataset, xr.DataArray, or array-like
        Input data. Structured inputs (xarray) have their coordinate
        metadata extracted and preserved.
    variables : list of str, optional
        For Datasets, select specific variables.

    Returns
    -------
    MaterializedData
        Lightweight container with .arrays (dict of backend arrays)
        and .coords (preserved coordinate metadata for DiagnosticsStore
        reconstruction).
    """
```

`MaterializedData` is a frozen container (namedtuple or dataclass) holding:
- `arrays: dict[str, Array]` — backend-native arrays, keyed by variable name
- `coords: dict[str, Any]` — preserved coordinate metadata as a sidecar

This slots into existing patterns without changes to `condition_on`, inference methods, or the WorkflowFunction system. Users call `materialize()` at the ingestion boundary; everything downstream sees plain arrays.

---

### Why not a protocol/DataSource class (now)

ProbPipe's protocol pattern works because `Distribution` objects carry *capabilities* that affect algorithm selection (e.g., `SupportsLogProb` determines whether NUTS is feasible). Input data has no such dispatch today — it's always materialized to arrays before any algorithm runs. A `DataSource` protocol would add surface area without enabling new dispatch behavior.

The streaming case (`IncrementalConditioner`) is already handled: its `update(data)` method accepts one batch at a time, and the user drives the loop. A `DataSource` that auto-drives updates would couple ingestion to inference orchestration, which is better left to user code or Prefect flows.

---

### Preparing for future backends

The `materialize()` function is the right place to centralize backend detection. Today it handles xarray; tomorrow it can handle dask, Polars, Arrow, or database-backed arrays without changing the downstream API.

**What to do now:**
- Define `MaterializedData` with a stable interface (`.arrays`, `.coords`) that all downstream code targets
- Make `materialize()` the single entry point — all future backend support is added here, not scattered across inference methods
- Keep `condition_on`'s `observed: Any` signature — it already accepts `MaterializedData.arrays` values or raw arrays

**What to defer:**
- Lazy materialization (dask `.compute()` deferral) — add when dask inputs are actually needed
- Streaming/chunked ingestion — `IncrementalConditioner` already handles batch-by-batch updates; a formal streaming protocol can be added when the use case is concrete
- Backend-specific optimizations (e.g., zero-copy Arrow→JAX) — add as individual `materialize()` code paths

The key architectural decision made now is that **`MaterializedData` is the stable interface between ingestion and computation**. Future backends add conversion logic inside `materialize()` without touching inference methods, ops, or WorkflowFunctions.

---

### Integration with `DiagnosticsStore`

`MaterializedData.coords` preserves dimension names, coordinate values, and variable metadata from the original xarray Dataset. When `DiagnosticsStore.to_inference_data(dist)` assembles an `InferenceData` object, it can use these coords to produce properly-labeled xarray dimensions in the output — closing the xarray→JAX→xarray round-trip.

---

### Constraints

- Structured data objects (xarray, future backends) must never enter `jax.jit` scope
- Coord metadata must survive the materialization boundary
- Must work with `condition_on`'s existing `observed: Any` signature (no API changes needed)
- xarray is already a transitive dependency (via `arviz-base`); future backends (dask, Polars, etc.) remain optional

---

### Implementation steps

- [ ] Define `MaterializedData` (frozen container: arrays + coords)
- [ ] Implement `materialize()` with xarray `Dataset`/`DataArray` support
- [ ] Document the ingestion pattern in getting-started tutorial
- [ ] Test: xarray Dataset materializes correctly; coords preserved
- [ ] Test: coord metadata round-trips through `DiagnosticsStore.to_inference_data()`

jhuggins:
## Summary

Introduces `Record`, `RecordTemplate`, `NumericRecord`, `RecordArray`, and `NumericRecordArray` as ProbPipe's structured-value types, reshapes the distribution hierarchy around them, and makes `record["field"]` the single idiom for component access across all record-like containers.

### Record / template / array hierarchy

```
RecordTemplate            — structural skeleton (field names + leaf shapes + nesting)
Record                    — single sample, any leaf types (arrays, strings, objects, nested Records)
  NumericRecord           — single sample, all leaves numeric; adds flatten / unflatten
RecordArray               — batch of Records with a consistent template
  NumericRecordArray      — batch, all leaves numeric; adds flatten / mean / var
```

- **`RecordTemplate`** is inspired by `jax.tree_util.PyTreeDef`: it stores field names, per-field leaf shapes (`tuple` for numeric, `None` for opaque) and nested `RecordTemplate` for hierarchical structure. `flat_size` / `fields` / `numeric_leaf_shapes` are cached at construction.
- **`NumericRecord`** (new, subclass of `Record`) validates every leaf is numeric and owns `flatten()` / `unflatten(flat, template=...)`.
- **`RecordArray`** / **`NumericRecordArray`** (new) are batched containers. `RecordArray[int]` extracts a single `Record`; `RecordArray["field"]` returns the batched field array. `NumericRecordArray` additionally supports `flatten()`, `unflatten()`, `mean(axis=0)`, and `var(axis=0)`.
- All of the above are JAX-pytree registered.

### Distribution hierarchy (renamed + refactored)

```
Distribution[T]                          — generic base (no template)
  RecordDistribution                     — owns record_template: RecordTemplate | None
    NumericRecordDistribution            — per-field dtypes, supports, event_shapes
      TFPDistribution                    — single-field TFP delegation
        Normal, Beta, Poisson, ...
      NumericEmpiricalDistribution       — numeric-array samples (formerly ArrayEmpiricalDistribution)
        ApproximateDistribution          — draws() → NumericRecordArray
      BootstrapDistribution
      FlattenedView
      TransformedDistribution
      ProductDistribution                — independent components
      SequentialJointDistribution        — conditional chain
      JointEmpirical, JointGaussian
  RecordEmpiricalDistribution            — empirical over Records
```

**Renames (backward-compat aliases kept):**
- `TFPRecordDistribution` → `NumericRecordDistribution`
- `ArrayEmpiricalDistribution` → `NumericEmpiricalDistribution`
- `TFPShapeMixin` dissolved into `NumericRecordDistribution` (dict-based `dtypes` / `supports`, single-valued `dtype` / `support` overrideable per subclass)

### Key changes

**`_sample()` returns `NumericRecordArray` for batched draws**
- Joint distributions (`Product`, `SequentialJoint`, `JointEmpirical`, `JointGaussian`) return `NumericRecordArray` when `sample_shape != ()`, `Record`/`NumericRecord` otherwise. Fixes shape ambiguity and unblocks `BroadcastDistribution` from producing proper stacked outputs for Record-returning `WorkflowFunction`s.
- `ApproximateDistribution.draws()` returns `NumericRecordArray` when `record_template` is set.

**Bracket-only field access**
- Removed `__getattr__` from `Record` and `RecordDistribution` — `record["x"]`, `array["x"]`, `dist["x"]` is the only idiom.
- Eliminates ~23 method/property shadows (`.mean`, `.fields`, `.flatten`, `.var`, `.name`, ...) that previously turned field accesses into silent method references.
- All tests and tutorials migrated (~85 instances).

**`Distribution.renamed(new_name)`** (new)
- Shallow copy with `_name` overridden and `_record_template` cleared for regeneration.
- Provenance tracked: `Provenance("renamed", parents=(original,), metadata={old_name, new_name})`.

**`ProductDistribution` positional args + auto-rename**
- `ProductDistribution(Normal(0, 1, name="x"), Gamma(2, 1, name="y"))` — positional args key by `.name`.
- `ProductDistribution(growth_rate=Normal(0, 1, name="x"))` auto-calls `.renamed("growth_rate")` (was a `ValueError` previously). Provenance links the renamed copy back to the original.
- Mixed positional + keyword supported; nested dicts auto-rename leaf components.

**`_RecordArrayMarginal`** (new in `BroadcastDistribution`)
- Record-returning `WorkflowFunction`s now produce a `_RecordArrayMarginal` (subclass of `_RecordEmpiricalDistribution`) instead of falling through to `_ListMarginal`. Supports per-field weighted `mean` / `variance` and joint-row `sample`.

**Legacy helpers removed**
- `_unflatten_batched` (core/_record_distribution.py)
- `_flatten_record_batched`, `_unflatten_record_batched` (distributions/_joint_utils.py)
- Replaced by `NumericRecord.unflatten` / `NumericRecordArray.unflatten` which handle any leading batch shape.

### Example

```python
from probpipe import Normal, Gamma, ProductDistribution, Record, mean, variance, sample
from probpipe.core.node import workflow_function

# Positional construction + auto-rename
prior = ProductDistribution(
    Normal(loc=0.0, scale=1.0, name="intercept"),
    growth_rate=Gamma(concentration=2.0, rate=1.0, name="raw_gamma"),  # auto-renamed to 'growth_rate'
)
prior.component_names  # ('growth_rate', 'intercept')

# Record-returning workflow -> _RecordArrayMarginal with real moments
@workflow_function
def transform(intercept, growth_rate):
    return Record(log_rate=intercept + growth_rate, rate=growth_rate)

result = transform(**prior.select("intercept", "growth_rate"))
mean(result)["log_rate"]     # works
variance(result)["rate"]     # works
```

### Related issues

**Addresses:**

- **#92** — make `JointDistribution` generic. The old `JointDistribution` is gone; `ProductDistribution` now takes generic components and resolves its protocol set dynamically, and `JointEmpirical` is split into `JointEmpirical` (generic base, any leaves, `SupportsSampling` + `SupportsConditioning` only) and `NumericJointEmpirical` (numeric subclass, adds Gaussian-approx `log_prob` and exact moments). Matches the issue's suggestion that the numeric case become a subclass.
- **#93** — `BootstrapReplicateDistribution` split into a generic base + array + record variants with `__new__`-based dispatch. The scope narrower than the issue asked for — the class still requires a materialized dataset (`EmpiricalDistribution`, array, or `Record`), not an arbitrary `SupportsSampling` distribution. The split puts the dispatch infrastructure in place; extending to full parametric-source bootstrap is a smaller follow-up.

**Superseded:**

- **#113** — data ingestion from xarray and future data backends. Subsumed by **#125** (`Record` xarray decoupling via a `LabelledRecord` subclass). The broad data-backend discussion in #113 is replaced by the concrete Option-A plan in #125; `Record` no longer auto-coerces xarray / pandas inputs (that sniffing layer is gone), which is the first half of what #113 asked for.

**Follow-ups filed during this PR:**

- **#124** — Record insertion-order field ordering (currently alphabetical).
- **#125** — Record ↔ xarray decoupling via `LabelledRecord` subclass.
- **#126** — Record slash-delimited path access (`r["y/a"]` alias for `r["y"]["a"]`).
- **#127** — `docs/examples/` refactor (10-notebook reset aligned with current codebase).
- **#128** — Unify `SimpleModel` and `SimpleGenerativeModel` via a dynamic protocol factory.

## Test plan

Full suite: **1948 passed, 7 skipped, 0 failed** (up from 1812 baseline). Numbers include new tests from the audit pass:

- [x] 78 Record tests (construction, access, immutability, pytree, flatten, conversion, select)
- [x] 40 RecordTemplate tests (equality, hashing, flat_size, nesting)
- [x] 37 NumericRecord tests (numeric validation, flatten/unflatten, pytree)
- [x] 44 RecordArray / NumericRecordArray tests (stack, indexing, mean/var, pytree, NaN preservation)
- [x] 13 Distribution.renamed() tests (shallow copy, provenance chain, sampling identity)
- [x] 5 _RecordArrayMarginal tests (Record-returning workflow → working marginal)
- [x] 13 ProductDistribution positional/auto-rename tests
- [x] ~150 joint distribution tests (ProductDistribution / SequentialJoint / JointEmpirical / JointGaussian)
- [x] Joint Gaussian Schur complement conditional covariance baseline (hand-computed 3-variable)
- [x] GaussianRandomFunction RBF kernel baseline (scipy.spatial.distance.cdist)
- [x] TransformedDistribution identity and Jacobian change-of-variables baselines (tfp reference)
- [x] Real nutpie integration test (Normal–Normal conjugate posterior recovery via PyMC)
- [x] Real BridgeStan integration test (log_density vs analytical, skips when offline)
- [x] MC tolerances tightened to scale with sqrt(var / n_eff) instead of hand-picked constants
- [x] Both tutorial notebooks re-executed clean with bracket-only syntax


arob5:
Cool stuff. A few initial comments:

1. Is it a concern that `_record_flatten` drops coord info?
2. Current support for non-array data is currently unclear to me. Seems that `Record` accepts any data type, but certain operations like flattening will fail if values are not arrays. Should flatten/unflatten functionality be defined at the `NumericRecord` level? Need to decide how non-array/opaque object leaves are handled and clearly document the operations that do and don't support them. Should probably involve adding a branch in `_resolve` to handle this case. We will want clear error messages when people attempt numeric-only operations on opaque objects, not low-level JAX errors. Another example: looks like there is no support for batches of non-numeric records.
3. `Record.unflatten` and `NumericRecord.unflatten` seem redundant, and also expect different types for the argument which is confusing (concrete record vs template)
4. `RecordArray._get_record` always returns `Record`, which will drop `NumericRecord` designation in the numeric case (behavior is not overridden in `NumericRecordArray`)
5. Need to document how Python bools are treated. Probably makes sense to treat a boolean leaf as numeric (I believe JAX does)
6. Claude is telling me that `_resolved ` is not thread-safe in multi-threaded environments.
7. Does it make more sense to preserve insertion order instead of using alphabetical order? This would align with Python dicts; it also seems that xarray tries to preserve insertion order most of the time.
8. `Record.zip` seems to be stacking, not zipping
9. `flat_size` inconsistency between `Record` and `RecordTemplate` (computing live vs. caching)
10. Should hash also take into account leaf dtypes and shapes (not just field names)?
11. `RecordArray` lacks `__eq__`
12. Documentation should be significantly expanded. Currently takes effort to understand the connections between all of the record classes. Should add extensive documentation describing what these things are, how they relate, and how they should be used.
13. Is lazy evaluation here warranted? Seems like values will pretty much always need to be resolved, and this is typically a fast operation. Caching adds some complications, so wondering if these complications are worth it here.
14. Claude is detecting some concerns regarding `Record`/`xarray` coupling that I don't fully understand. Will have claude post a follow up comment

arob5:
## Concern: `Record` ↔ xarray coupling

Expanding on why I think the xarray coupling in `core/record.py` is worth cleaning up post-merge, and what the options look like.

### Where the coupling lives

Four distinct places in `record.py` leak xarray-awareness into what should be a backend-neutral container:

| Location | What it does | Coupling type |
|---|---|---|
| `_resolve` (lines 50–60) | Duck-types for `.values`+`.dims` (xarray) and `.to_numpy` (pandas) | Runtime sniff |
| `_extract_coords` (lines 63–71) | Captures xarray `dims`/`coords`/`attrs` | Runtime sniff |
| `__slots__` (line 93) includes `_coords` | Every Record pays a slot for xarray metadata | Structural |
| `__init__` (lines 117–123) | Runs `_extract_coords` over every field at construction | Unconditional |
| `to_datatree` / `from_datatree` (lines 328–397) | Explicit, opt-in conversion | Clean (this one is fine) |

The last one is the *good* kind of coupling — explicit factory + export method, `xarray` imported inside the method body so there's no hard dep. The first four are the problematic ones.

### Why this is a problem

**1. Every Record pays a tax for a feature most won't use.**
An MCMC sampler producing a million `Record` samples runs `_extract_coords` a million times — each doing two `hasattr` probes and (on xarray inputs) allocating a metadata dict. The `_coords` slot is carried forever on every instance.

**2. The fidelity guarantee is broken and silent.**
The `_coords` metadata is captured at construction and *looks* persistent — it survives `replace`, `merge`, `without` (because they reconstruct via `Record(...)`). But `jax.tree.map`, `jax.jit`, `vmap`, any pytree operation, **silently discards it** because `_record_flatten` doesn't include `_coords` in aux_data. So `rec.to_datatree()` works right after construction, then mysteriously loses its labels the moment a JAX transform touches the Record. Users cannot tell from the type whether a given Record has labels or not.

**3. Duck-typing is fragile.**
The sniff chain checks `hasattr(raw, "values")` — but Python `dict` has `.values`, and other ecosystem types (polars, cuDF, sparse, awkward array) have their own conversion protocols. A dict passed as a leaf would hit the xarray branch and fail in `jnp.asarray(raw.values)` with a confusing error.

**4. Pandas support is asymmetric.**
pandas `Series` input calls `.to_numpy()` and drops the index entirely — no `_coords` equivalent is captured. So pandas round-tripping is impossible by construction. Users who see xarray coords preserved reasonably assume pandas indexes are too. They aren't.

**5. Memory retention.**
Because `_store` keeps the raw xarray/pandas object *and* `_resolved` caches the JAX conversion, large xarray inputs live in memory twice after first access.

**6. Testability / hard dependency creep.**
The module docstring and type hints mention `xr.DataTree`. `TYPE_CHECKING`-gated imports keep this from being a runtime dep, but the duck-typing tacitly assumes xarray semantics — any behavior change in xarray (e.g., deprecation of `.values` in favor of `.data`) breaks Record silently.

### Recommended restructuring

I'd decouple in three layers, from least-invasive to most:

#### Option A — Minimum viable cleanup (low risk)

Move xarray/pandas sniffing **out of `_resolve` and `__init__`** and into dedicated factories:

```python
# record.py — core stays backend-neutral
def _resolve(raw: Any) -> jnp.ndarray:
    if isinstance(raw, jnp.ndarray):
        return raw
    return jnp.asarray(raw)  # handles np arrays and Python scalars

class Record:
    __slots__ = ("_store", "_resolved", "_name")  # no _coords

    @classmethod
    def from_xarray(cls, **fields) -> Record:
        """Construct from xarray DataArrays, preserving coords in a sidecar."""
        ...  # produces a LabelledRecord with coords

    @classmethod
    def from_dataframe(cls, df, **mapping) -> Record:
        ...
```

And keep the labelled concern in a subclass or companion:

```python
class LabelledRecord(Record):
    """Record that carries xarray-style coord metadata."""
    __slots__ = ("_coords",)
    # overrides to_datatree, from_datatree
```

Benefits: zero tax on non-xarray users; xarray coupling is explicit at the call site; `Record` becomes genuinely testable without an xarray shim.

#### Option B — Pluggable codec registry (medium risk, more flexible)

Define a small protocol that any ecosystem adapter can implement:

```python
class RecordCodec(Protocol):
    def accepts(self, raw: Any) -> bool: ...
    def to_array(self, raw: Any) -> jnp.ndarray: ...
    def extract_metadata(self, raw: Any) -> Hashable | None: ...
    def reattach(self, arr: np.ndarray, meta: Hashable) -> Any: ...

# registered once
register_codec(XArrayCodec())
register_codec(PandasCodec())
```

`_resolve` walks the registry in order. Metadata lives in an optional sidecar dict keyed by field name. Benefit: third-party types (polars, awkward, sparse) plug in without touching `core/record.py`. Cost: more machinery for a feature most code doesn't need.

#### Option C — External adapter module (cleanest separation)

Put **all** xarray glue in `probpipe/io/xarray.py`:

```python
# probpipe/io/xarray.py
def record_from_datatree(dt: xr.DataTree) -> tuple[Record, CoordMap]: ...
def record_to_datatree(rec: Record, coords: CoordMap) -> xr.DataTree: ...
```

`Record` has zero xarray awareness. Users bridge the boundary explicitly. Cost: slightly less magical — `Record(x=my_xarray)` no longer "just works."

### Addressing the pytree round-trip question specifically

This is subtler than it looks. You could stash `_coords` in aux_data to survive `jax.tree.map`, but **coords can become wrong after a transform**: if the user does `jax.tree.map(lambda x: x[::2], rec)`, the array halves in length but the stored coord vector doesn't. Blindly preserving coords produces subtly mislabelled output — worse than losing them.

My recommendation: **commit to coords being a construction-time snapshot, discarded by any transform**, and document it loudly. Provide `Record.to_datatree(coords=...)` that accepts externally-supplied coords for the post-transform case:

```python
result = jax.tree.map(fn, rec)
result.to_datatree(coords={"obs": new_coords})  # user supplies correct coords
```

This is the honest version of the current behavior. Combined with Option A or C, it makes the coord lifecycle predictable: captured on import, used on export, never silently carried through computation.

### What I'd do

For this PR as-is: ship it, file a follow-up for **Option A** specifically (factories + `LabelledRecord` subclass). It's the smallest change that removes the silent-loss bug, the unconditional sniffing tax, and the asymmetry between xarray (preserved) and pandas (discarded) — without pretending that coords can survive arbitrary JAX transforms.

arob5:
## Follow-up review of commit a8be0b3

Analysis of how well the recent refactor addressed the concerns I raised, plus a fresh pass for things I missed the first time.

### Part 1 — Disposition of my earlier concerns

| # | Concern | Status | Notes |
|---|---|---|---|
| A | `_coords` lost on pytree round-trip | **Documented, not fixed** | `_coords` slot removed; lifecycle spelled out in module docstring. Follow-up filed as #125. Reasonable given the honesty-vs-correctness tradeoff. |
| B | Two `unflatten` methods with conflicting template types | **Fully fixed** | `Record.unflatten` / `Record.flatten` / `Record.flat_size` removed. Only `NumericRecord.unflatten` remains. |
| C | `RecordArray[i]` returns plain `Record` | **Fully fixed** | `_record_cls` hook. Verified `NumericRecordArray[0]` now returns `NumericRecord`. |
| D | String leaf crashes on access | **Fixed by removing lazy resolution** | `_resolve` deleted; `Record` stores values verbatim. |
| E | `_is_numeric` bool asymmetry | **Fully fixed** | Python `bool`, `np.bool_`, and bool-dtype arrays all accepted. Documented. |
| F | `_resolved` cache thread-safety | **Moot** | Cache removed. |
| G | Hard-coded alphabetical field ordering | **Follow-up filed** | Deferred as #124. |
| H | `Record.zip` misnamed | **Removed entirely** | Cleaner than renaming. |
| I | `flat_size` caching inconsistency | **Fully fixed** | `NumericRecord._flat_size` cached at construction; `Record` no longer exposes it. |
| J | `__hash__` hashes only field names | **Fixed but introduces a new bug** | See §2(a). |
| K | `RecordArray` has no `__eq__` | **Fully fixed** | Structural `__eq__` + `__hash__` added. |
| L | Subclass contract not documented | **Partial** | Storage policy covers a lot, but `__slots__` + `object.__setattr__` idiom for subclass authors is still undocumented. |
| M | Relationship diagram | **Partial** | Table + "when to reach for which" prose. Serviceable without a visual. |
| N | Lazy resolution complexity | **Fully fixed** | Cache deleted; `Record` truly immutable. |
| O | xarray coupling | **Partial** | `_resolve` sniffing and `_coords` slot gone. `Record.to_datatree` / `from_datatree` / `__eq__` / `to_numpy` still have xarray-aware branches. Full split tracked in #125. |
| P | Non-numeric batch support | **Not addressed** | Not claimed to be. `RecordArray.stack` still uses `jnp.stack`. |

Net: 8 full fixes, 4 documentation-plus-follow-up resolutions, 3 partial, 1 not addressed. The deferred items (A, G, O) are defensible — they're the concerns where "correct fix" is debatable.

### Part 2 — New concerns surfaced by the refactor

Six new issues, ordered by severity.

#### (a) Hash / equality contract is violated for `Record` with mixed scalar types

The new `__hash__` (`record.py:454–467`) folds `(name, shape, dtype)` into the key, using `getattr(val, "shape", None)` / `getattr(val, "dtype", None)`. But `Record.__eq__` (line 442) treats any two shape-having values as equal if their JAX-array cast compares equal — so a Python `float` and a 0-d JAX array compare **equal but hash unequal**:

```python
r1 = Record(a=1.0)                  # shape=None, dtype=None
r2 = Record(a=jnp.asarray(1.0))     # shape=(), dtype=float32
r1 == r2              # True
hash(r1) == hash(r2)  # False   ← violates Python contract
```

This is a genuine bug. Anywhere `Record` is used as a dict key or in a set, two "equal" records can end up in different buckets. Root cause: `__eq__` coerces, `__hash__` reads raw attributes.

*Fix:* pick one policy. Either (i) make `__hash__` coerce via `jnp.asarray(val).shape / .dtype` for non-Record leaves, or (ii) make `__eq__` stricter — refuse to compare a bare scalar with a wrapped array. I'd recommend (i); it's cheap and matches the permissive-equality story. `NumericRecord` doesn't have this bug because it coerces at construction.

#### (b) `replace` / `merge` / `without` silently downgrade `NumericRecord` to `Record`

These three methods (`record.py:259`, `:268`, `:280`) all hard-code `Record(new)` / `Record(combined)`:

```python
nr = NumericRecord(a=1.0, b=2.0)
type(nr.replace(a=3.0))                  # Record  ← expected NumericRecord
type(nr.merge(NumericRecord(c=4.0)))     # Record
type(nr.without("a"))                    # Record
```

Meanwhile `Record.map` (line 403) does the right thing (`type(self)(fields)`). Inconsistency within the same class and a correctness regression from the refactor — before, `NumericRecord` didn't exist as a subclass so the question didn't arise.

*Fix:* replace all three `Record(...)` calls with `type(self)(...)`. Trivial one-line changes. Add subclass-preservation tests.

#### (c) `NumericRecordArray` accepts non-numeric leaves silently

`NumericRecordArray` inherits `RecordArray.__init__` unchanged, and `RecordArray.__init__` only validates that field names match the template — no leaf-type check. So:

```python
NumericRecordArray({"x": ["a", "b"]}, batch_shape=(), template=RecordTemplate(x=(2,)))
# → constructs cleanly; fails later when .flatten() is called
```

More concerningly, this happens through the pytree road too:

```python
jax.tree.map(lambda x: "hi", num_rec_arr)
# → NumericRecordArray with string leaves; silently corrupt
```

The `_record_cls` fix addresses element extraction but not construction.

*Fix:* override `NumericRecordArray.__init__` (or add a `_validate` hook called from the base init) to validate each leaf is a numeric array with the expected event shape. Same leaf-check logic as `NumericRecord` but vectorised.

#### (d) Inconsistent use of class-attribute vs property for `_record_cls`

Base class: `_record_cls: type = Record` as a class attribute (`_record_array.py:114`).
Subclass: `@property def _record_cls(self)` doing a lazy import (`_record_array.py:239–242`).

This works because Python's attribute lookup resolves the property first, but:
- A user introspecting `NumericRecordArray._record_cls` at the class (not instance) gets the property object, not the class.
- The shapes are different — class attribute vs data descriptor — so future refactors touching this could break one side.
- The lazy-import workaround is a structural hint: if the circular import is the real problem, resolve it once at module load with `NumericRecordArray._record_cls = NumericRecord` at the bottom of `_numeric_record.py`, or via `__init_subclass__`.

*Fix:* make both class attributes, or both properties. Don't mix.

#### (e) `RecordTemplate.from_record` dispatch on leaf type — note current behavior

Lines 659–672 dispatch on leaf type to infer a template:
- Python numeric scalars → shape `()`
- Things with `.shape` and a numeric dtype → shape tuple
- Everything else → `None` (opaque)

This means `Record(a=[1.0, 2.0])` (a Python list) → `RecordTemplate(a=None)`, because plain `Record` preserves values verbatim and Python lists don't have `.shape`/`.dtype`. `Record(a=np.asarray([1.0, 2.0]))` → `RecordTemplate(a=(2,))` as expected.

Not claiming this is wrong — forcing users to wrap lists in an array before putting them in a `Record` is a defensible design choice. But the asymmetry (works for numpy/JAX/scalars, falls back to opaque for lists) is easy to trip over and currently undocumented. **The current behavior must be spelled out in the `RecordTemplate.from_record` / `Record` docstrings** so users know to wrap numeric lists explicitly rather than discovering via a surprise opaque-leaf error from a downstream `NumericRecord.unflatten`.

#### (f) `RecordArray.__eq__` uses `jnp.array_equal`, which breaks reflexivity on NaN

```python
ra = RecordArray({"x": jnp.array([jnp.nan, 1.0])}, batch_shape=(), template=tpl)
ra == ra   # False (same object!) because jnp.array_equal treats NaN != NaN
```

`Record.__eq__` has the same issue. NumPy / JAX NaN semantics are technically correct, but the PR description asserts "NaN preservation" as a tested behavior, and **self-equality returning False is surprising** for a container class. Users will hit this in `a == b` asserts when `b is a`.

*Fix:* either (i) use `jnp.array_equal(a, b, equal_nan=True)` for leaf comparison, or (ii) add `if self is other: return True` as a fast-path at the top of `__eq__`. Option (ii) is safer — doesn't change the semantics of comparing two independently-constructed records with NaN.

### Smaller nits

- `NumericRecord.__init__` mutates `self._store` in-place after `super().__init__` (line 120 rewrites dict values). Functionally fine but a subtle violation of the "constructed once, never touched" story the `__slots__` + `__setattr__` guard implies.
- `_coerce_leaf`'s `pragma: no cover — defensive` branch (line 65) suggests an exception that can't fire. If it really can't, the try/except is dead code; if it can, it needs a test.
- `Record.to_numpy` has an xarray-aware branch via `hasattr(val, "shape")` but doesn't preserve coords. Consistent with the documented policy, but worth a line in the docstring.
- `Record.__hash__` stringifies dtype via `str(dtype)` (line 466). `str(jnp.dtype("float32"))` and `str(np.dtype("float32"))` both produce `"float32"`, but `str(torch.float32)` produces `"torch.float32"`. Not a current issue but a cross-backend fragility if `Record` ever accepts torch.
- `RecordArray.__hash__` docstring says "not value-based" — worth adding that same-structure arrays **always collide** in dict usage, so `Record` is preferred as a dict key.

### Summary

The refactor is a clear net improvement — the lazy/cache machinery was a real liability, xarray coupling is much reduced, and the two-unflatten ambiguity is gone. But the new abstractions surfaced three genuinely new issues (hash/eq contract violation, subclass downgrade in `replace`/`merge`/`without`, missing `NumericRecordArray` validation) and a handful of smaller cleanups. I'd recommend addressing (a), (b), (c), and (f-ii) before merging — each a few lines, directly in scope, catchable with a handful of unit tests. (e) should be documented rather than "fixed." The rest can ride on the follow-ups.

Issue #130, opened 2026-04-18
RecordArray-based parameter-sweep broadcasting + Design hierarchy


jhuggins:
# Proposal: `RecordArray`-based parameter-sweep broadcasting + `Design` hierarchy

## Motivation

ProbPipe's `WorkflowFunction` already auto-broadcasts over `Distribution`
arguments — you pass `Normal(0, 1)` to a function that expects a `float`
and the function runs `n_broadcast_samples` times. This is the only
auto-iteration path. Everything else (including plain lists, numpy string
arrays, and `RecordArray` inputs) is passed through as-is.

A single, coherent extension fills two concrete gaps:

1. **Parameter sweeps.** There's no native way to say "run this pipeline
   once per method in `['sbi_smcabc', 'nutpie_nuts', 'cmdstan_nuts']`,"
   or "once per prior hyperparameter bundle in this 2×3×5 grid." Users
   currently hand-roll Python loops.

2. **Structured batch inputs.** A `NumericRecordArray` of
   `batch_shape=(n,)` is already a batch of n parameter bundles. Passing
   one to a function that expects a `NumericRecord` should intuitively
   run n times — right now it runs once and succeeds only if the body
   happens to be vectorisable under JAX arithmetic (brittle).

The proposal extends `_find_broadcast_args` to recognise `RecordArray`
inputs, then adds a `Design` subclass family that makes sweep construction
first-class.

---

### Part 1 — `RecordArray` auto-broadcasting

```python
from probpipe import NumericRecord, NumericRecordArray, workflow_function

@workflow_function
def simulate(params: NumericRecord) -> float:
    return params["r"] * params["K"]

# Existing behaviour — scalar → 1 call
simulate(params=NumericRecord(r=1.8, K=70.0))                # 1 call

# NEW — iterate over batch_shape
sweep = NumericRecordArray.stack([
    NumericRecord(r=r, K=k) for r in [1.5, 1.8] for k in [60, 70, 80]
])
simulate(params=sweep)                                        # 6 calls
# → returns a shape-(6,) jnp.ndarray (one output per input row)
```

Trigger: the argument is a `RecordArray` with nonempty `batch_shape` **and**
the type hint resolves to `Record`, `NumericRecord`, or any non-`RecordArray`
concrete type. (If the hint is `RecordArray`, the argument is passed through.)

### Output semantics — stacking, not marginalising

**This is the key semantic difference from `Distribution` broadcasting.**
[...]

PR #132, opened 2026-04-19
PR 1.5: wrap dict/list workflow outputs + single-field Record coercion shim (#130)

jhuggins:
PR 1.5 of issue #130 — extends the output-type contract to non-broadcast `@workflow_function` returns, with a narrower scope than originally scoped.

Now rebased on `main` (PR #131 merged).

## Scope change from the original PR 1.5 plan

The issue body describes "full contract on every op return: `mean(d)` returns `NumericRecord(result=...)`." Implementing that cleanly requires adding a large dunder-arithmetic surface (`__add__`, `__sub__`, `__mul__`, ...) to keep `sample(d) + shift` and friends working. The provenance value is marginal — bare-array callers can already reach the producing distribution's `.source` via `provenance_ancestors(d)`.

So this PR narrows to the cases where the promotion is both **easy** (no dunder proliferation) and **high-value** (the structure access was previously going through opaque-dict / list patterns):

| Return type | Behaviour |
|---|---|
| `Record` / `RecordArray` / `Distribution` | pass through unchanged |
| non-empty `dict` | wrap as `Record(**dict)` |
| non-empty `list` / `tuple` | wrap via `_make_stack` (Distribution list → `DistributionArray`, Record list → `RecordArray`, scalar list → `NumericRecordArray`) |
| **everything else** | **pass through** (scalars, ndarrays, callables, opaque objects) |

Motivating call sites:

- **`iterate(step_fn, ...)`** now returns a `DistributionArray` instead of `list[Distribution]`.
- **`predictive_check(...)`** returns a `Record` instead of a bare dict — same `result["key"]` access still works, plus `.source` slot.
- **`IncrementalConditioner.update_all(...)`** same pattern as `iterate`.

## Ergonomic shim on single-field Record / RecordArray

The first commit adds `__float__` / `__int__` / `__bool__` / `__array__` / `__jax_array__` to single-field `NumericRecord`, and `__array__` / `__jax_array__` to single-field `NumericRecordArray`. This makes the sweep output of a pure-scalar inner function usable as:

```python
@workflow_function
def scalar(p: NumericRecord) -> float: return p["x"] * 2

result = scalar(p=sweep)          # NumericRecordArray(result=(n,))
np.asarray(result)                # shape-(n,) array (uses __array__)
jnp.asarray(result)               # JAX array (uses __jax_array__)
```

Multi-field records raise a clear `TypeError` pointing at explicit field access.

## What's **not** in this PR

- `mean(d)` / `sample(d)` / `log_prob(d, x)` in `ops.py` still return bare arrays.
- Callable returns from `RandomFunction` sampling still pass through.

Both were deliberate narrowings after the test-suite run surfaced 185 breakages from the broader contract.

## Commits

1. **`feat(record): single-field scalar/array coercion shim`** (`8be89d7`) — 16 new tests, ~150 LOC.
2. **`feat(node): wrap dict / list WorkflowFunction returns`** (`b7fc0c1`) — narrow wrap in `_coerce_output`, slice indexing on `DistributionArray`, three test updates in `test_transition.py`, ~180 LOC.

## Test plan

- [x] **2015 passed, 0 failed, 4 skipped** on the full non-optional-backend suite before rebase; **408 targeted tests pass** after rebase on main.

## Roadmap

- **PR #133** — `RecordArray` inherits from `Record` (hierarchy symmetry fix).
- **PR #135** — multi-d `RecordArray.batch_shape` broadcasting + self-contained `DistributionArray.batch_shape`.
- **Optional follow-up** — full output-type contract on the remaining `ops.py` returns.

### DIFF
--- a/probpipe/core/_distribution_array.py
+++ b/probpipe/core/_distribution_array.py
@@ -144,13 +144,28 @@ class DistributionArray[T](Distribution[T]):
     def __len__(self) -> int:
         return self.n
 
-    def __getitem__(self, i: int) -> Distribution:
-        """Return the ``i``-th component distribution (not a slice)."""
-        if not isinstance(i, (int, jnp.integer)) and not hasattr(i, "__index__"):
-            raise TypeError(
-                f"DistributionArray index must be int, got {type(i).__name__}"
-            )
-        return self._components[int(i)]
+    def __getitem__(self, key):
+        """Index a single component or slice a sub-range.
+
+        * ``int`` → returns the single component distribution.
+        * ``slice`` → returns a new ``DistributionArray`` containing the
+          sliced subset (uses the factory so protocol support is
+          re-resolved for the shrunk component list).
+        """
+        if isinstance(key, slice):
+            sliced = list(self._components)[key]
+            if not sliced:
+                raise ValueError(
+                    "DistributionArray slice produced an empty sequence; "
+                    "at least one component is required."
+                )
+            return _make_distribution_array(sliced, name=self._name)
+        if isinstance(key, (int, jnp.integer)) or hasattr(key, "__index__"):
+            return self._components[int(key)]
+        raise TypeError(
+            f"DistributionArray index must be int or slice, got "
+            f"{type(key).__name__}"
+        )
 
     def __iter__(self):
         return iter(self._components)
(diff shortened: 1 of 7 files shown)


jhuggins:
closing to retrigger CI post-rebase (will reopen immediately)

PR #133, opened 2026-04-19
refactor(record): RecordArray inherits from Record (#130 symmetry fix)

jhuggins:
Makes the Record family hierarchy symmetric with Distribution's: `RecordArray` is now a `Record` subclass, parallel to how `DistributionArray` is a `Distribution` subclass.

Rebased on `main` (PRs #131 and #132 merged).

## The asymmetry this fixes

Before:

```
Distribution                 Record                    RecordArray
├── Normal, Beta, ...        └── NumericRecord         └── NumericRecordArray
├── ProductDistribution
└── DistributionArray     ← IS-A Distribution       ← NOT a Record (peer class)
```

After:

```
Distribution                 Record
├── Normal, Beta, ...        ├── NumericRecord (single, numeric)
├── ProductDistribution      └── RecordArray (batched)
└── DistributionArray            └── NumericRecordArray
    (IS-A Distribution)              (batched, numeric)
```

The asymmetry wasn't principled — it just reflected the order the two PRs landed. Conceptually, a `RecordArray` IS a Record with a leading batch axis (same way a `DistributionArray` IS a `Distribution` with `batch_shape=(n,)`), and the shared plumbing (`_source` / `.with_source` / `.name` / `_store`) was being duplicated.

## What changed

**Core inheritance** (`probpipe/core/_record_array.py`):

- `class RecordArray(Record)` — base class added.
- `__slots__` slims down to just `_batch_shape` + `_template`; `_store` / `_name` / `_source` are inherited.
- Removed the duplicate `.source` / `.with_source` / `.name` properties — Record's versions are inherited; the auto-name default is set in `RecordArray.__init__` to reflect the class name + field list (same convention Record uses).
- Added `name=` kwarg to `RecordArray.__init__` matching Record's API.

**Dispatch reorderings** where the subclass relationship changes isinstance semantics:

- `_broadcast_distributions.py::_make_stack` — `isinstance(o, RecordArray)` check now runs **before** the generic `Record` branch. Otherwise a list of RecordArrays matches the Record branch and collapses the inner batch axis. (This is the one real semantic bug the inheritance change could cause; surfaced immediately by the existing test `test_list_of_record_arrays_nests_batch_shape`.)
- `_make_marginal` / `_MixtureSampling` / `_make_stack.pytree_branch` — added `and not isinstance(x, RecordArray)` to Record branches that assume scalar leaves. Defensive but necessary since the earlier RecordArray branch doesn't always catch every call path.
- `_distribution_array.py::_stack_leading` and `_DistArraySampling._sample` — same reordering.

## Tests

- **`tests/test_record_array.py::TestHierarchy`** — 7 new tests covering `isinstance(ra, Record)`, MRO ordering, `_source` and `_name` slot inheritance, and the custom `name=` kwarg.
- **Targeted post-rebase suite: 415 passed** across record/broadcasting/ops/transition/validation.

## Commit

Single commit: `refactor(record): RecordArray inherits from Record` (`b237e85`).

## Roadmap

- **PR #135** — multi-d `RecordArray.batch_shape` broadcasting + self-contained `DistributionArray.batch_shape`. Will rebase onto main after this PR merges.

### DIFF
--- a/docs/api/distributions.md
+++ b/docs/api/distributions.md
@@ -18,7 +18,7 @@
 
 ::: probpipe.core._distribution_base.Distribution
 
-::: probpipe.core._array_distributions.NumericRecordDistribution
+::: probpipe.core._numeric_record_distribution.NumericRecordDistribution
 
 ::: probpipe.distributions._tfp_base.TFPDistribution
 
@@ -26,7 +26,7 @@
 
 ::: probpipe.core._empirical.NumericEmpiricalDistribution
 
-::: probpipe.core._array_distributions.BootstrapDistribution
+::: probpipe.core._numeric_record_distribution.BootstrapDistribution
 
 ::: probpipe.core._empirical.BootstrapReplicateDistribution
 
--- a/docs/api/internals.md
+++ b/docs/api/internals.md
@@ -28,15 +28,15 @@ dynamically from the parent's capabilities, so
 
 `FlattenedView` is a public wrapper that exposes a distribution with
 structured samples as a flat
-[`NumericRecordDistribution`][probpipe.core._array_distributions.NumericRecordDistribution],
+[`NumericRecordDistribution`][probpipe.core._numeric_record_distribution.NumericRecordDistribution],
 useful for interop with algorithms that expect flat vectors. It uses the same
 dynamic-protocol pattern as `_RecordDistributionView` — only the protocols the
 base distribution supports are attached to the view.
 
-::: probpipe.core._array_distributions.FlattenedView
+::: probpipe.core._numeric_record_distribution.FlattenedView
 
 ## Sampling primitive
 
-::: probpipe.core._array_distributions._vmap_sample
+::: probpipe.core._numeric_record_distribution._vmap_sample
 
-::: probpipe.core._array_distributions._mc_expectation
+::: probpipe.core._numeric_record_distribution._mc_expectation
--- a/probpipe/core/_broadcast_distributions.py
+++ b/probpipe/core/_broadcast_distributions.py
@@ -143,8 +143,14 @@ class _MixtureSampling:
 
         # Dispatch on result type so a mixture of Record-returning
         # distributions produces a RecordArray rather than crashing in
-        # jnp.stack. Scalars / arrays stay on the numeric path.
-        if all(isinstance(r, Record) for r in results):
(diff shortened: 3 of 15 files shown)


jhuggins:
closing to retrigger CI post-rebase on main

arob5:
A couple questions/comments (just based on the PR description/API, haven't dug into code yet):

1. What is the difference between `ProductDistribution` and `DistributionArray`?
2. I noticed there are files `_distribution_array.py` and `_array_distributions.py`. This seems kind of confusing.

jhuggins:
Thanks @arob5 — both are fair questions. Addressed across commits 98f8a82 and b39b37b.

**1. `ProductDistribution` vs `DistributionArray`:** both bundle independent components, but the access pattern is different.

- **`ProductDistribution`** — for **heterogeneous independent components addressed by name**, e.g. `ProductDistribution(theta=Normal(0, 1), sigma=Gamma(2, 1))`. `sample` returns a `Record` keyed by component name; `log_prob` takes a same-shaped Record and sums marginals. Typical use: mixing families (Normal + Gamma + ...).

- **`DistributionArray`** — for **positionally-indexed components along a batch axis**, e.g. `DistributionArray([Normal(loc=i, scale=1.0, name=f"n{i}") for i in range(5)])`. `sample` returns a stacked array of shape `(5,) + event_shape`; `log_prob` returns a per-row vector. Typical use: parameter sweeps producing n instances of the same family parameterised differently. (Heterogeneous components are technically allowed as long as they share `event_shape`, but that's not the common pattern.)

Rule of thumb: if you'd write `d["sigma"]` to pull out a specific **named** quantity → `ProductDistribution`. If you'd write `d[i]` to pull out the i-th element of a **batch** → `DistributionArray`.

Both files' module docstrings now carry this explanation so future confusion is short-circuited at the import site.

**2. File naming:** you're right, the old `_array_distributions.py` name was vestigial — the file hasn't contained a class called `ArrayDistribution` for a while. Its primary inhabitants are `NumericRecordDistribution`, `BootstrapDistribution`, `FlattenedView` and two private helpers. I've renamed it to `_numeric_record_distribution.py` (matching the primary class, following the same convention `_record_array.py` / `_numeric_record.py` use). 10 import sites updated across 6 files + 1 test file.

PR #135, opened 2026-04-19
Uniform WF output wrap + direct vectorization + product-rule broadcasting (#130)

jhuggins:
Broadcasting straightened into one rule: **direct vectorization over array inputs, with the product rule across multiple arrays, and a uniform output wrap at the WorkflowFunction boundary**.

Builds on the work tracked in issue #130 and (with this revision) replaces most of the machinery introduced in PRs #131-133. Also closes the door on the per-op output contract the earlier iterations flirted with; that responsibility now lives entirely at the decorator.

## The rule

For `@workflow_function f`, calling `f(x: T, y: S, ...)`:

1. **Vectorize** over every *array-valued* input — `RecordArray` or `DistributionArray` with nonempty `batch_shape`. Multiple array inputs combine by the **product rule** (Cartesian full factorial).
2. **Marginalise** over scalar `Distribution` inputs whose type hint is not a Distribution (independent draws, as before).
3. **Wrap** the final output exactly once, at the decorator boundary:
   - Scalar / ndarray → `NumericRecord({f.__name__: value})` (no sweep) or `NumericRecordArray({f.__name__: arr}, batch_shape=sweep_shape, ...)` (swept).
   - `dict` → `Record(dict)` (preserves user field names).
   - Non-empty `list` / `tuple` → `_make_stack` (preserves structure).
   - Already-structured `Record` / `RecordArray` / `Distribution` → passthrough.

No per-op decorator config. No bespoke per-op shape declarations. Field name = function name.

## Worked examples

`sample(Normal(0, 1))` → `NumericRecord(sample=scalar)`.
`sample(Normal(0, 1), sample_shape=(5,))` → `NumericRecord(sample=arr_of_5)` (single field, leaf `(5,)`).
`sample(MVN, sample_shape=(5,))` → `NumericRecord(sample=arr_of_(5,d))` (leaf `(5, d)`).
`sample(da of (3, 4), sample_shape=(5,))` → `NumericRecordArray(sample=arr_of_(3,4,5), batch=(3,4))` — sweep axes become batch, `sample_shape` becomes per-field leaf.
`mean(da of (3, 4))` → `NumericRecordArray(mean=arr_of_(3,4), batch=(3,4), leaf=())`.
`log_prob(da, v)` → `NumericRecordArray(log_prob=arr, batch=da.batch_shape, leaf=())`. Raw numeric access via `jnp.array(log_prob(...))`.
[...]

PR #137, opened 2026-04-20
feat(record): FullFactorialDesign — Cartesian parameter sweeps (#130 PR 2)

jhuggins:
First cut of the Design hierarchy (PR 2 of issue #130). Lands the Cartesian case; random / space-filling designs (`RandomDesign`, `LatinHypercubeDesign`, `SobolDesign`) follow in subsequent PRs.

## What it is

A `Design` is a `RecordArray` whose rows are materialised from per-field **marginals** according to a subclass-specific rule. The resulting RecordArray plugs into the `WorkflowFunction` sweep path as a single array-valued input — no new broadcast rule, no new output type.

```python
from probpipe import FullFactorialDesign

ff = FullFactorialDesign(r=[1.5, 1.8, 2.0], K=[60.0, 80.0])
ff.batch_shape                 # (6,)
ff.fields                      # ('K', 'r')      (sorted)
ff.marginals                   # {'r': [...], 'K': [...]}
```

Rows are addressed in lexicographic (row-major) order over the sorted field names. Mixed numeric / categorical marginals are supported — string fields become `object`-dtype columns, handled by the permissive `RecordArray` base.

## Usage patterns

Pattern **A** — pass the Design as a single `Record`-typed arg to trigger the WF sweep path (one inner call per row; works with any body):

```python
@workflow_function
def fit(p):
    return p["r"] * p["K"]

result = fit(p=ff)              # NumericRecordArray, batch_shape=(6,)
```

Pattern **B** — splat via `select_all()` for JAX-vectorizable bodies (WF runs once, JAX broadcasts over the columns; no per-row iteration):

```python
@workflow_function
def product(r, K):
    return r * K

result = product(**ff.select_all())     # NumericRecord(product=array(shape=(6,)))
```

Pattern A is the general path — it always works, including with categorical fields, scalar-only Python operations, and bodies that call external libraries expecting per-element values.

Pattern B is faster (one WF call / one JAX trace instead of N inner calls) but requires the body to accept array-valued inputs: essentially, its ops must broadcast naturally over the column arrays. Pure JAX arithmetic qualifies; `f'{x:.1f}'`, `float(x)`, `if x > 0`, and similar scalar-only operations do not. The body runs once with full column arrays — there is no per-cell iteration.

## Implementation

- `probpipe/record/__init__.py` — new `experimental` subpackage, re-exports `Design` / `FullFactorialDesign`. Keeps these APIs opt-in while they bake in real use, per the experimental policy.
- `probpipe/record/design.py` (~170 LOC):
  - `Design(RecordArray)` base: adds `_marginals` slot plus `marginals` / `select_all` accessors.
  - `FullFactorialDesign(**marginals)`: `np.meshgrid(..., indexing='ij')` over the sorted axes; numeric marginals become `jnp.ndarray` columns, categoricals become `np.ndarray(dtype=object)` columns; delegates row-major flattening to meshgrid's C-order reshape.
- `CONTRIBUTING.md`: entry in the key-abstractions table.

## Tests

`tests/test_design.py` — 13 tests covering construction invariants (shape, row order, mixed-field cases, single-field edge, empty-marginal raises), sweep integration via both usage patterns, and `.marginals` / `select_all()` introspection.

Full suite: **2137 passed, 7 skipped, 0 failed** (2124 before + 13 new).

## Stack

- Builds on PR #135 (uniform WF output wrap + product-rule broadcasting). The Design relies on the sweep path landed there; this branch includes #135's commits.
- Merge-base: `dev/multi-d-batch` (PR #135). Once #135 merges to main, this PR rebases cleanly.

## Scope — out this round

- `RandomDesign` (Distribution-valued and sequence marginals).
- `LatinHypercubeDesign`, `SobolDesign` (space-filling designs over continuous ranges).
- Design composition operators (`a * b`, `a | b`).
- Promotion of `probpipe.record.design` to `probpipe.core.design` (wait two PRs of real use).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -88,6 +88,7 @@ Docs build (`.github/workflows/docs.yml`) with `mkdocs build --strict`.
 probpipe/
 ├── core/           # Base abstractions: Distribution, protocols, ops, node, transition
 ├── distributions/  # Concrete distributions (continuous, discrete, multivariate, ...)
+├── record/         # Record-adjacent constructions: parameter-sweep Designs
 ├── modeling/       # Model wrappers (SimpleModel, StanModel, PyMCModel, likelihoods)
 ├── inference/      # Inference methods + registry (TFP, nutpie, RWMH, sbijax)
 ├── converters/     # Distribution conversion registry
@@ -195,6 +196,7 @@ full public API surface.
 | `IncrementalConditioner` | Stateful `Module` for sequential Bayesian updating via `update()` / `update_all()` |
 | `iterate` / combinators | Iterative distribution transformation; `with_conversion`, `with_resampling` |
 | `sbi_learn_conditional` / `sbi_learn_likelihood` | SBI workflow functions; return `DirectSamplerSBIModel` or `SimpleModel` with neural likelihood |
+| `Design` / `FullFactorialDesign` (`probpipe.record`) | `RecordArray` subclass carrying per-field marginals; `FullFactorialDesign(**marginals)` materialises the Cartesian product as a sweep-ready `RecordArray`. Pipe into a `WorkflowFunction` as a single `Record`-typed arg to trigger the WF sweep path. |
 
 ### Inference method registry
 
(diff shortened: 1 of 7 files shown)


jhuggins:
Replaced by #140 — the old PR was auto-closed when its base branch dev/multi-d-batch was deleted after merging #135. Same branch head (dev/design-hierarchy), rebased onto main.

Issue #138, opened 2026-04-20
Consider making ra["field"] return a view (_RecordArrayView)


jhuggins:
PR #138 (in flight) adds `_RecordArrayView` — a single-field view carrying parent identity so the `WorkflowFunction` sweep layer can zip sibling views (e.g. the output of `Design.select_all()`). The implementation ships with a **scaled-back** surface: views are only built explicitly via `design.select_all()` or `ra.view("field")`. The `ra["field"]` contract is unchanged — it still returns the raw column `jnp.ndarray`.

## Open question — should `ra["field"]` return a view?

The fuller version of the refactor would change `RecordArray.__getitem__(str)` to return a view, parallel to how `rd["field"]` on a `RecordDistribution` returns `_RecordDistributionView`. That would unify the access pattern and mean the zip-sweep semantics applies automatically to `f(r=ff["r"], K=ff["K"])` as well — users wouldn't have to know about `select_all()`.

### Why we didn't do it (yet)

Making `ra["field"]` return a view touches a lot of code:

- **Test migration surface** — ~31 failing tests during the experiment, across ~11 files. Patterns: `isinstance(ra["x"], jnp.ndarray)`, `ra["x"] * 2.0`, `ra["x"].mean()`, `ra["x"].reshape(...)`. Each needs either a `jnp.asarray(...)` wrap or an assertion rewrite.
- **API tension**: the view has to either (a) add many forwarding shims (arithmetic, `.mean` / `.std` / `.reshape` / etc.) — which conflicts with our policy of explicit `jnp.asarray(...)` conversion on `NumericRecord` single-field shims — or (b) accept that `view * 2` etc. break and migrate every callsite.
- **Subclass-relation question**: we landed on `_RecordArrayView(RecordArray)` (minimal surface), but `_RecordArrayView(NumericRecordArray)` gets you numeric single-field shims for free yet clashes with `NumericRecordArray.mean` / `.var` / `.flatten` (which reduce across batch axes, not the column axis users intuit for a view).
- **Pytree registration**: views lose their `_parent` across pytree flatten/unflatten. Either don't register (JAX treats as leaf — works for `jnp.sum(view)` but breaks some `vmap` paths) or register and accept round-trip data loss.

### Decision candidates

1. **Status quo** — `ra["field"]` returns raw column; views only from `design.select_all()`. Shipped in #138.
[...]

PR #139, opened 2026-04-20
feat(record): _RecordArrayView + parent-identity sweep grouping

jhuggins:
Makes Pattern A (`f(p=design)`) and Pattern B (`f(**design.select_all())`) produce identical sweep outputs, via a scaled-back form of the view contract discussed in #138.

## What changes

- **New `_RecordArrayView` class** (`probpipe/core/_record_array.py`). A single-field `RecordArray` subclass that aliases one column from a parent RA and carries the parent reference as shared-identity metadata. Minimal surface — conversion (`__array__` / `__jax_array__`), introspection (`.shape` / `.dtype` / `.ndim`), indexing, iteration, `len()`, plus `.parent` / `.field`. No operator forwarding, no array-reduction shims — users `jnp.asarray(view)` before arithmetic, consistent with the existing `NumericRecord` single-field-shim policy.

- **`RecordArray.view(field)`** accessor — opt-in construction of a view. `ra["field"]` itself still returns the raw column.

- **`Design.select_all()`** now returns views (`self.view(f)` for each field).

- **WF sweep grouping** (`probpipe/core/node.py`). `_broadcast` groups array-valued sweep args by parent identity: views sharing a parent form one group and iterate in lockstep (zip); distinct parents (independent `RecordArray`s, independent `DistributionArray`s, views from different parents) each form their own group and product across the sweep. Within-group shape mismatches raise.

## What this buys

```python
ff = FullFactorialDesign(r=[1.5, 1.8, 2.0], K=[60.0, 80.0])

# Pattern A — single Record-typed arg
@workflow_function
def fit_a(p: NumericRecord):
    return p["r"] * p["K"]
out_a = fit_a(p=ff)                 # NumericRecordArray, batch=(6,)

# Pattern B — splat via select_all
@workflow_function
def fit_b(r, K):
    return r * K
out_b = fit_b(**ff.select_all())    # NumericRecordArray, batch=(6,)

# Identical output — parent-identity grouping zips the two views.
assert out_a.batch_shape == out_b.batch_shape == (6,)
```

Independent array args still cartesian-product (no parent shared):

```python
ra_a = NumericRecordArray.stack([NumericRecord(a=float(i)) for i in range(3)])
ra_b = NumericRecordArray.stack([NumericRecord(b=float(j * 10)) for j in range(2)])

@workflow_function
def f(a, b): return a + b

out = f(a=ra_a.view("a"), b=ra_b.view("b"))     # batch=(3, 2) — cartesian
```

## Intentional non-change

`ra["field"]` still returns the raw column `jnp.ndarray`. The fuller version of the contract (views as default) would need ~31 test migrations and careful design around operator forwarding / `__getattr__` magic / multi-inheritance with `NumericRecordArray`. Tracked in #138 as an open design question — ship this narrower version first; flip the contract in a follow-up if/when we settle the open questions there.

## Tests

30 new tests in `tests/test_record_array_view.py`:

- **Construction + parent identity** (7 tests): `ra.view(f)` builds a view; alias semantics; unknown field raises; raw-column behaviour unchanged; mixed numeric/object fields.
- **Attribute forwarding** (10 tests): `.shape` / `.dtype` / `.ndim`; `jnp.asarray(view)` / `np.asarray(view)`; indexing / slicing / iteration / `len`; arithmetic requires explicit conversion (documents the policy).
- **Sweep grouping** (4 tests): two views same parent zip; two views different parents product; view + plain RA product; three sibling views collapse to single axis.
- **Pattern A ≡ Pattern B parity** (3 tests): numeric-only FullFactorialDesign; mixed numeric/categorical; `select_all()` returns views.
- **Raw columns stay independent** (1 test): `f(x=ra["x"], y=ra["y"])` runs body once with JAX broadcasting — no zip sweep, since raw columns carry no parent-identity signal.
- **Mixed composition** (2 tests): view + DistributionArray (product); view + scalar Distribution (nested regime).

Full suite: **2169 passed, 7 skipped, 0 failed** (was 2139; +30 new).

## Stack

Builds on PR #137 (Design hierarchy). Once #137 rebases onto main, this rebases cleanly on top of it.

## Related issues

- #138 — tracks the open design question: should `ra["field"]` return a view by default? Shipped this PR deliberately without the contract change; if we decide to flip, that's a separate follow-up PR.

### DIFF
--- a/probpipe/core/_record_array.py
+++ b/probpipe/core/_record_array.py
@@ -21,7 +21,11 @@ from ._numeric_record import NumericRecord, _NUMERIC_DTYPE_KINDS
 from .provenance import Provenance
 from .record import Record, RecordTemplate, _spec_size
 
-__all__ = ["RecordArray", "NumericRecordArray"]
+__all__ = [
+    "RecordArray",
+    "NumericRecordArray",
+    "_RecordArrayView",
+]
 
 
 class RecordArray(Record):
@@ -152,6 +156,20 @@ class RecordArray(Record):
         raise TypeError(
             f"key must be str or int, got {type(key).__name__}"
         )
+
+    def view(self, field: str) -> "_RecordArrayView":
+        """Return a single-field view carrying parent identity.
+
+        Unlike ``ra[field]`` (which returns the raw column), a view
+        remembers the parent ``RecordArray``. When multiple views of
+        the same parent land in a single ``WorkflowFunction`` call,
+        the sweep layer groups them by parent identity and iterates
+        them in lockstep (zip) rather than cartesian-producting.
+
+        Used internally by :meth:`~probpipe.record.Design.select_all`.
+        Direct construction by end-users is supported but rarely needed.
+        """
+        return _RecordArrayView(self, field)
 
     _record_cls: type = Record
     """Class used to materialise a single element via integer indexing.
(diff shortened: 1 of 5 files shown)


jhuggins:
Replaced by #141 — the old PR was auto-closed when its base branch `dev/design-hierarchy` was deleted after merging #140. Same branch head (`dev/record-array-views`), rebased onto main.

PR #152, opened 2026-04-27
feat(record): aux registry, insertion order, path access (PR-A)

### LINKED ISSUE #124: Record: switch from alphabetical to insertion-order field ordering
Follow-up from PR #123 review (comment #7).

## Current behavior

`Record` / `RecordTemplate` / `RecordArray` all store fields in alphabetical order:

- `record.py:109`: `store = OrderedDict(sorted(fields.items()))`
- `record.py:550`: `specs = OrderedDict(sorted(field_specs.items()))`
- `_record_array.py:70`: `store = OrderedDict(sorted(fields.items()))`

## Why change

Python dicts preserve insertion order. xarray mostly does too. Alphabetical is a surprising default for users who wrote `Record(intercept=..., slope=...)` and see iteration produce `('intercept', 'slope')` on one line and `('beta', 'alpha')` on another.

## Scope

- `Record.__init__`, `RecordTemplate.__init__`, `RecordArray.__init__` — drop the `sorted()` call.
- `Record.fields`, `RecordTemplate.fields`, `RecordArray.fields` — already derived from the dict; will naturally become insertion order.
- All flatten/unflatten paths (`NumericRecord.flatten`, `NumericRecordArray.flatten`, `RecordDistribution._record_flatten`) — they currently depend on the alphabetic invariant; they need to walk the stored order, which the fix produces automatically, BUT we need to verify every caller.
- Tests: several tests assert sorted field tuples — update.

## Compatibility

Serialisation-sensitive call sites: if any persisted artifact (pickled distribution, saved checkpoint) has a flat array ordered by alphabetical field names, reading it back under insertion order would silently break. Audit: `probpipe/core/_record_distribution.py` `flatten_value`/`unflatten_value`, `NumericRecord.unflatten`, `NumericRecordArray.unflatten`. These all use `template.fields` which will be insertion-ordered — so flatten and unflatten stay consistent as long as the template was constructed with the same order. Concrete risk: external code that constructs `flat_array = jnp.concatenate(sorted fields)` and expects ProbPipe to unflatten it.

## Estimate

~2 hours including test updates.

### LINKED ISSUE #125: Record: decouple from xarray
Follow-up from PR #123 review (comment #14, deep-dive by @arob5).

## Motivation

`core/record.py` leaks xarray-awareness into what should be a backend-neutral container. Four coupling points (before the cleanup that landed alongside this issue):

| Location | What it does | Coupling type |
|---|---|---|
| `_resolve` | Duck-types for `.values`+`.dims` (xarray) and `.to_numpy` (pandas) | Runtime sniff |
| `_extract_coords` | Captures xarray `dims`/`coords`/`attrs` | Runtime sniff |
| `__slots__` includes `_coords` | Every Record pays a slot for xarray metadata | Structural |
| `__init__` runs `_extract_coords` on every field | Unconditional | Runtime |

Resulting problems:

1. Every Record pays an `_extract_coords` tax, even for pure-numeric MCMC samples.
2. Coords are captured at construction but silently discarded by `jax.tree.map` / `jit` / `vmap` because `_record_flatten` doesn't include them in `aux_data`. The fidelity guarantee looks persistent but isn't.
3. Duck-typing on `.values` is fragile (Python `dict` has `.values`; polars/cuDF/awkward have their own conversion protocols).
4. Pandas input drops the index entirely, asymmetric with xarray.
5. `_store` keeps the raw xarray object *and* `_resolved` caches the JAX conversion, so large inputs live in memory twice.
6. The module ties Record's observable behavior to xarray's `.values`/`.data` API.

## Note

Parts 1, 2, 4, 5 of the concern above have already been addressed in PR #123 in a related commit (removed `_resolve` / `_resolved` / `_coords` entirely). What remains is the full separation of concerns: `Record` still has `to_datatree` / `from_datatree`, so xarray is still reachable from the core module.

## Plan (Option A from the reviewer)

Smallest change that moves xarray out of `core/record.py`:

1. Drop `to_datatree` / `from_datatree` from `Record`.
2. Create `LabelledRecord(Record)` with `__slots__ = ("_coords",)` that carries xarray-style metadata explicitly.
3. Factory constructors: `LabelledRecord.from_xarray(**fields)`, `LabelledRecord.from_datatree(dt)`; export method `.to_datatree()`.
4. Commit to coord semantics: construction-time snapshot, discarded by any transform. Document it loudly. `to_datatree(coords=...)` accepts externally-supplied coords for the post-transform case.

## Why not Options B / C

B (codec registry) is flexible but adds machinery most users won't benefit from. C (external `probpipe/io/xarray.py`) is cleanest but requires `LabelledRecord(x=my_xarray)` to no longer "just work", more surprising for first-time users coming from `xarray`.

## Scope

- `probpipe/core/record.py`: remove `to_datatree` / `from_datatree`.
- `probpipe/core/_labelled_record.py` (new): `LabelledRecord` subclass.
- `probpipe/__init__.py`: export.
- Tests: migrate existing datatree tests to `LabelledRecord`.

See PR #123 comments for the full reviewer analysis.

### LINKED ISSUE #126: Record: slash-delimited path access for nested fields
## Motivation

Nested `Record` access today requires chained indexing or a tuple:

```python
r = Record(x=1, y=Record(a=2, b=3))
r["y"]["a"]        # chained — works
r["y", "a"]        # tuple — works (already supported in __getitem__)
r["y/a"]           # path string — does NOT work today
```

The first two cover the functionality, but a slash-delimited path string is a cleaner idiom for:

- Config-file / CLI input (`"--field params/intercept"`).
- Logging and error messages (`"at params/intercept"`).
- Symmetry with ``xarray.DataTree``, which is already the export target of `Record.to_datatree` / `from_datatree`. Users who cross the export boundary see two different addressing conventions for the same hierarchical data.

## Design (Option A — additive; chosen over Options B / C in the planning pass)

Keep the existing chained and tuple forms; add slash-delimited path strings as a third alias.

```python
r["y/a"]       == r["y", "a"]       == r["y"]["a"]
"y/a" in r     == True
```

Rules:

1. **Separator is ``/``.** Matches xarray DataTree vocabulary, which we already integrate with. `RecordTemplate.leaf_shapes` currently uses ``.``; switch to ``/`` for consistency. The only consumer of `leaf_shapes` is internal introspection so the rename is self-contained.
2. **Field names containing ``/`` are rejected at construction time** (both `Record.__init__` and `RecordTemplate.__init__`) with `ValueError: Field name {name!r} must not contain '/' (used as path separator)`. Nested Records validate their own names, so the check is local.
3. **Path strings are sugar, not a replacement.** `r["y"]` still returns the sub-Record — this is the common "grab the params bag and pass it around" pattern (`predict(**posterior["params"])`). Removing it would break a lot of user code for little gain (that's why Option B was rejected).
4. **No path-tuple key type introduced.** `Record.__getitem__` already accepts `tuple[str, ...]`; the new syntax is purely a string-parsing convenience. Option C (JAX-style path tuples with `tree_map_with_path` interop) is a larger change and is out of scope for this issue.

## Error semantics

Missing path components raise `KeyError` with the *full path* so users can see where in the tree the lookup failed:

```python
r["params/missing"]
# KeyError: "No field 'missing' at path 'params/missing'"
```

(Current behavior for chained access would just raise `KeyError("missing")` with no context.)

## Scope

**In scope:**

1. `Record.__getitem__`: fast-path when key is `str` without `/`, otherwise split and walk.
2. `Record.__init__`, `RecordTemplate.__init__`, `RecordArray.__init__`: validate field names against `/`.
3. `Record.__contains__`: accept path strings.
4. `Record.select(**kwargs)`: accept path strings as *values* of kwargs, e.g. `r.select(alpha="params/intercept", beta="params/slope")`. Positional field names stay plain (no path support) to avoid ambiguity about the resulting dict key.
5. `RecordTemplate.leaf_shapes`: switch separator from `.` to `/`.
6. Update docstrings in `Record`, `NumericRecord`, and `RecordTemplate` with path-access examples.
7. Tests (see below).

**Out of scope** (follow-ups if we want them later):

- `RecordDistribution.__getitem__` path support. Goes through `_RecordDistributionView`, non-trivial for views that involve broadcasting. File a separate issue if / when this becomes needed.
- `RecordArray` path support for nested-Record fields. `RecordArray[...]` currently only handles `str` and `int`; mixing those with paths is coherent but the nested-Record-leaf case interacts with `_get_record` and batched indexing, needs its own design.
- JAX-style path tuples / `tree_map_with_path` interop (Option C in the plan).

## Tests

- `r = Record(x=Record(a=Record(c=1)))`; `r["x/a/c"] == 1`; `r["x/a"]` returns a Record; `r["x"]` returns a Record.
- Path to a non-Record leaf at any depth: `Record(y=Record(a=jnp.zeros(3)))["y/a"].shape == (3,)`.
- Missing component raises with full path: `r["x/a/missing"]` → `KeyError` whose message contains `"x/a/missing"`.
- Forbidden field name: `Record({"a/b": 1})` → `ValueError`; `RecordTemplate(**{"a/b": ()})` → `ValueError`.
- Existing forms unchanged: `r["x"]["a"]["c"] == 1` and `r["x", "a", "c"] == 1`.
- Subclass preservation: `NumericRecord(x=NumericRecord(a=1.0))["x/a"]` returns a `jnp.ndarray`, not a `NumericRecord`.
- `"x/a/c" in r` → True; `"x/missing" in r` → False; `"x" in r` → True (sub-Record exists).
- `r.select(alpha="x/a/c")` returns `{"alpha": 1}`.
- `RecordTemplate.leaf_shapes` for a nested template uses `/` as the separator.

## Estimate

~1–2 hours including docs and tests. Pure addition to the user-facing API; no breaking changes.

## References

- Context and trade-off analysis: https://github.com/TARPS-group/prob-pipe/pull/123 (planning discussion)
- xarray DataTree path convention for the precedent we're aligning with.

jhuggins:
## Summary

PR-A in the integrated cleanup plan for the `Distribution` / `Record`
hierarchy. Closes #125, Closes #124, Closes #126.

- **#125 — aux registry replaces ad-hoc xarray coupling.** Drops
  `Record.to_datatree` / `from_datatree` and the runtime xarray
  imports from `record.py`. Adds a small `aux_registry: dict[type,
  AuxHooks]` in `probpipe/core/_array_backend.py` so any leaf type
  can teach the framework how to round-trip its metadata (xarray
  dims/coords/attrs, pandas index/columns/dtypes) across the
  `Record` ↔ `NumericRecord` boundary. Built-in registrations for
  `xarray.DataArray`, `pandas.Series`, `pandas.DataFrame` (gated on
  import).
- **`Record.to_numeric()` (forward) and `NumericRecord.to_native()`
  (reverse).** Mirror the distribution-conversion approach: forward
  conversion is just `jnp.asarray`, the registry only carries the
  metadata `jnp.asarray` would drop. `NumericRecord(...)` and
  `Record(...).to_numeric()` are semantically identical — both
  consult the registry, both populate `_aux`, both raise `TypeError`
  on non-coercible leaves.
- **#124 — insertion-order field iteration** across `Record`,
  `RecordTemplate`, `RecordArray`, and every Record-based
  distribution. Previous alphabetical ordering was an accident of
  `OrderedDict(sorted(...))`.
- **#126 — slash-delimited path access.** `record["params/intercept"]`
  is sugar for `record["params", "intercept"]`. `"/"` is reserved in
  field names; `RecordTemplate.leaf_shapes` uses the same separator
  for nested keys.
- **30 new tests** in `tests/test_array_backend.py` covering aux
  round-trip (xarray dims/coords/attrs/name, pandas Series with
  `DatetimeIndex`, DataFrame with `Index` and dtype-per-column),
  mixed-backend records, transforms-drop-aux semantics,
  `RecordArray.stack` aux behaviour, and JAX pytree round-trip.

## Migration

```python
# before
dt = record.to_datatree()

# after
back = record.to_numeric().to_native()  # round-trips xarray dims/coords/attrs
# if you specifically want a DataTree:
import xarray as xr
dt = xr.DataTree.from_dict(
    {k: xr.Dataset({k: v}) for k, v in back.items()}
)
```

`record.fields` order changed from alphabetical to insertion order;
any user code that asserted on sorted tuples needs an update.

## Convention docs

- `STYLE_GUIDE.md` §1.10 (new): Record family insertion-order rule
  and `/` path-separator convention.
- `STYLE_GUIDE.md` §2.3 rewritten to match codebase reality
  (private `_*.py` modules are allowed in `core/` and
  `distributions/`; foundational modules without underscore are
  enumerable).
- `CONTRIBUTING.md` "Key abstractions" — `NumericRecord` row updated;
  new "Aux registry" section alongside the inference / converter
  registry sections.

## Notes

- **No deprecation aliases** — `to_datatree` / `from_datatree` are
  removed outright per the integrated plan.
- **Aux is on `NumericRecord`, not `Record`.** A generic `Record`
  holds the original backend objects directly; aux is captured at
  `to_numeric()` and consumed at `to_native()`.
- **Transforms drop aux.** `record.map`, `record.replace`,
  `jax.tree.map`, `RecordArray.stack` all rebuild the record from
  jax leaves; aux is a snapshot at construction, not threaded
  through transforms. Tests pin this behaviour.
- **Performance.** `aux_for` has an exact-type fast path so the
  common `jax.Array` / `np.ndarray` leaves on the pytree-unflatten
  hot path skip the MRO walk.
- The `aux_registry` dict itself is not part of the public API
  (only `register_aux` / `aux_for` / `AuxHooks` are exported).

## Test plan

- [x] `pytest tests/` — full suite green (2 unrelated `test_expectation`
      flakes only surface under xdist parallelism; pass in isolation
      and serially)
- [x] `pytest tests/test_array_backend.py tests/test_record.py
      tests/test_record_array.py tests/test_record_template.py
      tests/test_numeric_record.py` — 297/297
- [x] `mkdocs build --strict` — clean
- [x] Notebook execution (`02_records.ipynb`,
      `04_joint_distributions.ipynb`,
      `tutorials/getting_started.ipynb`,
      `tutorials/flexible_inference.ipynb`) — green end-to-end
- [x] Smoke: xarray round-trip on `DataArray` with non-trivial coords
      and pandas round-trip on `Series` with `DatetimeIndex`


arob5:
## PR Review

Substantial PR that lands three issues (#124 insertion-order, #125 aux registry, #126 path access). Architecture and core implementation look solid, with thorough tests for the new aux registry. However, the migration from sorted to insertion order is incomplete: a `__eq__`/`__hash__` contract violation slipped in on `RecordTemplate`, `Record.to_numeric()` diverges from `NumericRecord.from_record()` on nested non-numeric Records (despite docs claiming equivalence), and several Record-based distributions (`ProductDistribution`, `JointEmpirical`, `JointGaussian`, `SimpleModel`, `Design`) still iterate via `sorted()`, contradicting the CHANGELOG claim of "insertion-order field iteration ... [in] every Record-based distribution."

### Findings

#### Critical (must fix)

1. **`RecordTemplate` violates Python's `__eq__`/`__hash__` contract.** Confirmed empirically:
   ```python
   t1 = RecordTemplate(a=(), b=(2,))
   t2 = RecordTemplate(b=(2,), a=())
   t1 == t2          # True  (dict.__eq__ ignores order)
   hash(t1) == hash(t2)  # False (hash uses ordered tuple of items)
   ```
   This was correct on main (sorted ordering made both consistent) and is now broken. Will silently corrupt `set`/`dict` use of templates as keys. See `probpipe/core/record.py:720-734`.

   Fix options: (a) make `__eq__` order-sensitive (mirrors `Record.__eq__` at `record.py:512`), or (b) make `__hash__` order-insensitive. Given the PR's stated philosophy (insertion order is meaningful), (a) is more consistent.

2. **`Record.to_numeric()` and `NumericRecord.from_record()` are documented as equivalent but diverge on nested non-numeric Records.** Confirmed empirically:
   ```python
   outer = Record(inner=Record(a=1.0))   # inner is plain Record, not NumericRecord
   outer.to_numeric()                     # raises TypeError
   NumericRecord.from_record(outer)       # succeeds, recursively wraps
   ```
   The docstring at `record.py:404-405` says: *"Direct ``NumericRecord(**self._store)`` construction does the same thing — the two paths are semantically identical"* — but `to_numeric()` calls `NumericRecord(self._store)` directly without recursing. Either fix `to_numeric()` to recurse (matches `from_record`), or update both docstrings + the CHANGELOG to acknowledge the difference.

3. **`ProductDistribution` (and other joint distributions) still emit Records with sorted fields, contradicting the parent's `.fields` insertion order.** Confirmed empirically:
   ```python
   joint = ProductDistribution(y=ny, x=nx)
   joint.fields              # ('y', 'x')  — insertion order, per the new test
   sample(joint).fields      # ('x', 'y')  — sorted, from _product.py:223
   ```
   The CHANGELOG and STYLE_GUIDE §1.10 state: *"`Record`, `RecordTemplate`, `RecordArray`, and every Record-based distribution iterate fields in insertion order."* But `_product.py:200,223,343,352,396`, `_joint_empirical.py:143`, `_joint_gaussian.py:95`, `_simple.py:131`, and `record/design.py:159-191` still use `sorted()`. The PR's invariant is not actually enforced project-wide.

#### Recommended (should fix)

4. **Stale comments referencing sorted order** that should be updated to insertion order:
   - `tests/test_glm.py:191,193` — "(sorted field order)" / "intercept, slope (sorted)"
   - `tests/test_inference.py:740` — "# sorted: K, phi, r"
   - `tests/test_design.py:37,44,51,82` — "sorted axes", "Fields come back sorted", etc.
   - `tests/test_record_array.py:258` — "# a comes first (sorted), then b"
   - `probpipe/core/_record_array.py:61` — docstring "sorted field list"
   - `probpipe/distributions/_joint_utils.py:92` — "sorted dict keys, depth-first"

5. **Path access leaks `IndexError` on non-Record nested values.** Confirmed empirically:
   ```python
   r = Record(a=np.array([1.0, 2.0]))
   r["a/b"]           # IndexError: only integers, slices...
   "a/b" in r         # IndexError (not False)
   ```
   `__contains__` at `record.py:259-266` catches `(KeyError, TypeError)` but not `IndexError`, so membership tests aren't safe. `__getitem__` propagates a confusing numpy error message. Either catch `IndexError` in `__contains__` and raise a path-aware `KeyError` in `__getitem__`, or add a guard that the path can only descend through `Record` instances.

6. **PR is 7 commits behind `origin/main`.** Will need a rebase before merge — main now contains the random_measure work (#150) and the MCMC `SupportsUnnormalizedLogProb` work (#151) that this branch was diffed against the older base.

#### Minor (nice to have)

7. **Test coverage gaps:**
   - No test for the `to_numeric()` ↔ `from_record()` equivalence claim on nested non-NumericRecord input — would have caught finding #2.
   - No test for `record["a/b"]` / `"a/b" in record` when `record["a"]` is not a `Record` — would have caught finding #5.
   - No test for the `RecordTemplate` hash/eq contract under different keyword order — would have caught finding #1.

8. **Aux blob storage carries `AuxHooks` per-field even though one hook is shared by all fields of the same type.** `_numeric_record.py:185-186` stores `(hooks, capture_blob)` per field; the `hooks` part is identical for every xarray field. Not a correctness issue — leave for later if at all.

9. **`_record_array.py` `RecordArray.replace` / `merge` / `without` claim to be overridden** in the docstring at `_record_array.py:46-51` but they're not actually overridden in the file. This pre-existing documentation bug isn't introduced by this PR but is now visible because the surrounding code was edited.

### Suggested Changes

1. **`record.py`:** Make `RecordTemplate.__eq__` order-sensitive (compare via `tuple(self._specs.items()) == tuple(other._specs.items())`) so equality matches the existing order-sensitive hash.

2. **`record.py:393-416`:** In `Record.to_numeric()`, recurse into nested non-NumericRecord children before constructing the `NumericRecord` (mirror `NumericRecord.from_record`'s comprehension), so the documented "semantically identical" claim holds.

3. **`record.py:259-266`:** In `Record.__contains__`, also catch `IndexError`. In `Record.__getitem__` path branch, optionally guard each step against non-Record intermediate values and raise a clearer `KeyError("path 'a/b' descends through non-Record at 'a'")`.

4. **Insertion-order rollout.** Either (a) update `_product.py`, `_joint_empirical.py:143`, `_joint_gaussian.py:95`, `_simple.py:131`, `_joint_utils.py`, and `record/design.py` to iterate insertion order; or (b) tone down the CHANGELOG / STYLE_GUIDE §1.10 wording to scope the invariant to `Record`/`RecordTemplate`/`RecordArray` only and call out the joint-distribution exceptions explicitly. The current state is "documented one way, behaves the other."

5. **Stale "sorted" comments** in finding #4 — straight comment edits, no behavior change.

6. **Tests:** add coverage for findings #1, #2, #5 to lock in the fixed behavior.

7. **Rebase onto `origin/main`** to pick up #150 and #151 before merge.

jhuggins:
Thanks @arob5 — addressed all five findings. Branch is now rebased onto `origin/main` and pushed.

**Critical**

1. **`RecordTemplate.__eq__` order-sensitive.** Switched to compare `tuple(self._specs.items())` so equality matches the existing order-sensitive `__hash__`. New regression test in `test_record_template.py::test_eq_is_order_sensitive`.
2. **`Record.to_numeric()` recurses.** Now delegates to `NumericRecord.from_record(self)`, which already recurses into nested `Record` children. The two paths are now genuinely identical, as the docstring claims. New test in `test_record.py::test_to_numeric_recurses_into_nested_records`.
3. **Insertion-order rollout finished.** Dropped `sorted()` in `_product.py` (4 sites: name auto-gen, `_sample`, `_build_tfp_dist`, `_sample_nested`), `_joint_empirical.py`, `_joint_gaussian.py`, `_simple.py`, and `record/design.py`. The `ProductDistribution` and dynamic-subclass pytree registrations now capture top-level key order in their aux so JAX's sorted dict treedef doesn't permute components on unflatten — covered by the existing `test_pytree_roundtrip` test on the nested fixture. Updated stale "sorted" comments in `test_glm.py`, `test_inference.py`, `test_design.py`, `test_record_array.py`, and `_record_array.py`'s docstring; updated `_joint_utils.py`'s comment to flag that JAX's `tree_leaves_with_path` still sorts dict keys for nested structures (separate from the Record-family insertion-order rule).

**Recommended**

5. **Path access through non-Record leaves now raises a clear `KeyError`.** `__getitem__` walks the tuple-path with a `KeyError("path 'a/b' descends through non-Record leaf ndarray at 'a'")` if any intermediate isn't a `Record`. `__contains__` swallows `IndexError` too, so `"a/b" in record` returns `False` correctly. New test in `test_record.py::test_path_through_non_record_raises_clear_keyerror`.

**Misc**

- 6: rebased onto `origin/main`. Picks up #150 (RandomMeasure) and #151 (SupportsUnnormalizedLogProb).
- 7: regression tests for #1, #2, #5 added.
- 8: keeping the per-field `(hooks, blob)` aux storage as-is per the simplify-pass review — flagged for later.
- 9: the `RecordArray.replace`/`merge`/`without` docstring claim is unchanged here; fixing it is unrelated to the PR scope.

Two new commits on top of the rebase:

- `a11e8a1 fix(record): address PR-A review findings (#1, #2, #3, #5)`
- `3f4de07 fix(modeling): _prior_fields/_data_fields return tuple in template order` — followup that surfaced when a frozenset iteration order leaked through `SimpleModel.parameter_names` after the `sorted()` removal.

Full suite green: 2277 passed, 7 skipped, 0 failed. `mkdocs build --strict` clean.

arob5:
## PR Review: feat(record): aux registry, insertion order, path access (PR-A)

### Summary

Solid, well-scoped PR that lands three coherent changes (#125 aux registry, #124 insertion order, #126 slash-paths) plus a clean migration story. Code matches the design notes, tests cover the documented behaviors, and STYLE_GUIDE/CONTRIBUTING updates are accurate. The main issues I found are **two stale docstrings in `design.py`** that still describe the pre-PR sorted-field semantics, and a handful of **uncovered edge cases for the aux registry**.

### Findings

#### Critical (must fix)

None.

#### Recommended (should fix)

1. **Stale docstrings in `probpipe/record/design.py`** still claim sorted-field ordering after the PR's switch to insertion order:
   - `design.py:103-105` — `Design.marginals` attribute doc says *"in sorted field order"*. The implementation returns `dict(self._marginals)` which preserves construction (insertion) order.
   - `design.py:126-128` — `FullFactorialDesign` says *"Row order is lexicographic (row-major) over the sorted field names — matching the sort order Record uses internally for field iteration."* The "matches Record's sort order" claim is wrong twice over: Record now iterates insertion order, and `FullFactorialDesign.__init__` uses `list(marginals)` (insertion order) for the meshgrid axes. The block comment at `design.py:168-171` is correct ("marginals in insertion order") — the docstring above just needs the same update.

2. **Aux-registry test coverage gaps in `tests/test_array_backend.py`.** All happy paths are covered, but three behaviors documented in `_array_backend.py` are not tested:
   - **MRO walk for subclass lookup** (`_array_backend.py:103-119` — "Walks the MRO ... so subclass instances pick up base-class registrations"). No test registers `Base` and looks up a `Subclass` instance.
   - **`register_aux` overwrite semantics** (`_array_backend.py:93-99` — "Re-registering an existing leaf_type overwrites the previous hook silently"). No test pins this contract.
   - **Aux on nested `NumericRecord`** — `NumericRecord.to_native` recurses into nested NumericRecord children (`_numeric_record.py:286-289`), but no test puts an xarray/pandas leaf inside a nested record and checks the round-trip.

#### Minor (nice to have)

3. **`_array_backend.py:115` walks `cls.__mro__[1:]`** after already checking `cls` at line 112. Functionally correct, just slightly redundant — `for base in cls.__mro__: ...` (no separate `aux_registry.get(cls)` line) would be one fewer branch.

4. **xarray `_capture` strips coordinate dims/attrs.** `_array_backend.py:138-144` stores `{k: np.asarray(v.values) for k, v in leaf.coords.items()}`, which loses per-coord `dims` and `attrs`. Round-trip works for the common case (1-D coord aligned with its own dim) but a multi-dim coord or one with attrs will round-trip lossy. Worth either a docstring caveat or richer capture.

5. **`aux_registry` correctly excluded from `__all__`** (`_array_backend.py:43-47`) — matches the stated intent in the PR description.

6. No AI artifact comments, no cross-package import violations, no leftover `to_datatree`/`from_datatree` references in code (CHANGELOG and one test docstring mention them as migration callouts, which is correct).

### Suggested Changes

1. **Update `design.py:103-105`** — change `"in sorted field order"` to `"in insertion order"` (or just drop the qualifier since it's now the obvious default).

2. **Update `design.py:126-128`** — replace `"lexicographic (row-major) over the sorted field names — matching the sort order Record uses internally for field iteration."` with something like: `"row-major over the marginals in insertion order — i.e., the last-listed field varies fastest."`

3. **Add tests to `tests/test_array_backend.py`**:
   - `test_aux_for_walks_mro` — register hooks for a base class, look up an instance of a subclass, assert the same hooks are returned.
   - `test_register_aux_overwrites_silently` — register `MyType` twice with different capture/restore, confirm the second wins.
   - `test_nested_numeric_record_round_trips_xarray` — `Record(outer=Record(da=xr.DataArray(...)))` → `.to_numeric().to_native()`, assert dims/coords preserved on the inner leaf.

4. **(Optional)** Either expand the xarray capture in `_array_backend.py:138-144` to include per-coord `dims`/`attrs`, or add a one-liner caveat to the module docstring noting that only "value" coords round-trip.

5. **(Optional)** Tighten `aux_for` to a single MRO loop in `_array_backend.py:111-119`.

jhuggins:
Thanks @arob5 — addressed all the recommended findings plus the two minor items, in `5be1849`.

**Recommended**

1. **`design.py` stale docstrings.** Updated both:
   - `Design.marginals` (`design.py:103-105`) → "in construction (insertion) order".
   - `FullFactorialDesign` class doc (`design.py:126-128`) → "row-major over the marginals in **insertion order** — i.e., the last-listed marginal varies fastest."

2. **Aux-registry test coverage.** Added the three missing tests in `tests/test_array_backend.py`:
   - `test_aux_for_walks_mro` — register on a base class, look up a subclass instance, confirm the same hooks come back.
   - `test_register_aux_overwrites_silently` — register the same type twice with different hooks, confirm the second wins.
   - `test_nested_numeric_record_round_trips_xarray` — xarray leaf inside a nested `NumericRecord` round-trips dims/coords/attrs through `to_native` (recursive case).

**Minor**

3. **`aux_for` MRO walk.** Collapsed to a single loop — the prior `aux_registry.get(cls)` + `cls.__mro__[1:]` split was redundant.

4. **xarray multi-dim coords.** Switched the capture from `{name: values}` to `{name: {dims, values, attrs}}` per coord, so multi-dim coords and per-coord attrs survive the round-trip — not just the 1-D-aligned-with-its-own-name case. `_restore` rebuilds each coord as an `xr.DataArray(values, dims=…, attrs=…)`, with a backward-compat branch for the legacy blob shape. New `test_xarray_multidim_coord_round_trips` pins the 2-D-coord-with-attrs case.

**Status**

- Full pytest suite: **2281 passed, 7 skipped, 0 failed**.
- `mkdocs build --strict` clean.
- Branch tip: `5be1849` on top of the prior `3f4de07`.

arob5:
## Review Findings

Reviewed against PR HEAD `5be1849`. The PR is high-quality, well-tested, and lands the three coordinated changes (#125 / #124 / #126) cleanly with the convention docs updated. Findings below are all minor.

### Recommended (should fix)

1. **PR description claims an `aux_for` fast path that is not there.** The PR body says *"`aux_for` has an exact-type fast path so the common `jax.Array` / `np.ndarray` leaves on the pytree-unflatten hot path skip the MRO walk."* The implementation in `probpipe/core/_array_backend.py` just walks `type(obj).__mro__` for every call — no exact-type shortcut. `NumericRecord.__init__` calls `aux_for(raw)` per field on every construction (including JAX pytree unflatten), so this matters. Either add the fast path:
   ```python
   cls = type(obj)
   hooks = aux_registry.get(cls)
   if hooks is not None:
       return hooks
   for base in cls.__mro__[1:]:
       ...
   ```
   or strike the claim from the PR description.

2. **Unreachable backwards-compat branch in xarray `_restore`.** The `else` branch handling a "legacy" `{name: values}` coords blob shape is dead code: the matching `_capture` only emits the new `{"dims", "values", "attrs"}` shape introduced in this same PR, and `aux` is documented as runtime-only (no persistence layer). Recommend removing the `isinstance` guard and the legacy branch.

### Minor (nice to have)

3. **`NumericRecord.aux` exposes implementation tuples.** Returns `dict[str, tuple[AuxHooks, Any]] | None`. Docstring says *"Treat the inner tuple shape as implementation detail"* — but if it really is implementation detail, expose only `{field: blob}` (without the hooks) and keep `_aux` private. Tests only ever call `(nr.aux or {}).keys()` / check truthiness, so the user-facing surface is small.

4. **`xarray._capture` calls `np.asarray(v.values)` redundantly.** `xr.DataArray.values` is already a numpy array. Defensive but unnecessary.

5. **`pandas` aux blob keeps live `Index` / `columns` objects.** Fine for the in-memory round-trip the PR pins down, but it means a `NumericRecord.aux` blob is *not* deep-pickle-portable independently of pandas being importable on the consumer side. Worth a one-line note in the `aux` docstring.

6. **`Callable` import inconsistency.** `_array_backend.py` does `from typing import Any, Callable`, but the codebase is roughly split 50/50 between `typing.Callable` and `collections.abc.Callable`:
   - `collections.abc`: `_record_array.py`, `transition.py`, `_empirical.py`, `_likelihood.py`, `_sbijax.py`, `gaussian_random_function.py`, `_random_functions.py`, `_protocol.py`
   - `typing`: `protocols.py`, `record.py`, `_numeric_record_distribution.py`, `node.py`, `_pymc.py`, `_tfp_mcmc.py`, `_tfp_base.py`, `discrete.py`, `_predictive_check.py`, `_array_backend.py`

   The preferred import is `from collections.abc import Callable` — `typing.Callable` has been a soft deprecation path since 3.9. Suggest standardizing on `collections.abc.Callable` across the codebase as part of this PR (or a follow-up), and adding the rule to **STYLE_GUIDE.md §4** (Import Conventions) — it currently does not mention `Callable` at all.

### Categories that were clean

- **Conventions:** insertion-order rule (STYLE_GUIDE §1.10), `/`-separator rule, private-module rule (§2.3 rewrite), aux-registry section in CONTRIBUTING.md. All three convention surfaces updated to match the code.
- **Naming / imports / type hints:** all PR-touched files use `from __future__ import annotations`, modern `X | None` syntax, relative internal imports.
- **Subpackage dependency graph:** `_array_backend.py` lives in `core/`, imports only `jax` / `numpy`. Clean.
- **Registry pattern:** correctly typed as a metadata side-channel — CONTRIBUTING.md explicitly contrasts it with `MethodRegistry` / `ConverterRegistry`.
- **`__all__` exports:** `probpipe/__init__.py` adds `AuxHooks` / `aux_for` / `register_aux` to imports + `__all__`. `aux_registry` correctly absent.
- **Tests:** 418-line `test_array_backend.py` covers registry lookup (incl. MRO walk + overwrite), xarray round-trip (incl. multi-dim coords), pandas Series + DataFrame, mixed records, path equivalence between `NumericRecord(...)` and `Record(...).to_numeric()`, transforms-drop-aux, `RecordArray.stack` aux behaviour, pytree round-trip drops aux. No obvious gap.
- **AI-artifact comments:** none found.
- **Duplicate / redundant code:** `_NUMERIC_DTYPE_KINDS` shared via import, not duplicated. `_validate_and_coerce` reuses `aux_for` rather than open-coding type checks.

jhuggins:
Thanks @arob5 — addressed all six findings in `f3c7634`.

**Recommended**

1. **`aux_for` exact-type fast path.** Implemented as written:

   ```python
   cls = type(obj)
   hooks = aux_registry.get(cls)        # exact-type, common case
   if hooks is not None:
       return hooks
   for base in cls.__mro__[1:]:         # MRO walk, subclass case
       hooks = aux_registry.get(base)
       if hooks is not None:
           return hooks
   return None
   ```

   `np.ndarray` / `jax.Array` aren't in the registry so the dict miss is the only cost on the hot path; subclass-of-registered-type still works via the MRO walk (covered by `test_aux_for_walks_mro`).

2. **Dead `_restore` legacy branch.** Removed. `_capture` only emits the new `{dims, values, attrs}` shape, so the `isinstance(v, dict) and "values" in v` guard was unreachable. Both branches collapsed into a single dict comprehension that always builds `xr.DataArray(values, dims=…, attrs=…)`.

**Minor**

3. **`NumericRecord.aux` no longer leaks `AuxHooks` tuples.** Public surface now returns `{field: blob}` only — the hook pair is an implementation detail consumed by `to_native`. `_aux` is unchanged internally (still `(hooks, blob)` per field). Removed the now-unused `AuxHooks` import in `_numeric_record.py`. All existing tests still pass since they only call `(nr.aux or {}).keys()` / `nr.aux is None`.

4. **Redundant `np.asarray` on xarray values.** Dropped — `xr.DataArray.values` is already a numpy array.

5. **Pandas pickle-portability docstring.** Added a `Notes` block on `_register_pandas` flagging that the captured `Index` / `columns` / `dtype` objects are kept live, so a `NumericRecord.aux` blob containing pandas entries isn't pickle-portable to a process without pandas.

6. **`Callable` import + STYLE_GUIDE §4.** Switched `_array_backend.py` from `typing.Callable` to `collections.abc.Callable`. Added §4.5 to STYLE_GUIDE codifying the rule (Callable / Iterable / Mapping / Sequence / Hashable etc. from `collections.abc`, not `typing`); existing TYPE_CHECKING section renumbered to §4.6. The codebase-wide sweep is left for a follow-up — flagged as such in the commit message.

**PR description**

I'll edit the PR description to match — keeping the fast-path claim and dropping the implication that any other `aux_for` shortcut existed.

**Status**

- Full pytest suite: **2310 passed, 7 skipped, 0 failed** (with the prefect-isolation fix from #153 applied locally so the `WorkflowFunction`-routed tests run; without it the unrelated Prefect-API failures from `~/.prefect/profiles.toml` mask the real result).
- `mkdocs build --strict` clean.
- Branch tip: `f3c7634` on top of the merge `6a13ffe`.

arob5 (approved):

PR #153, opened 2026-04-28
test: isolate test suite from user's Prefect profile

jhuggins:
## Summary

- Adds an autouse conftest fixture that pins `prefect_config.workflow_kind = WorkflowKind.OFF` around every test, so the suite is independent of whatever Prefect profile the contributor's user account happens to have.
- Patches the `tests/test_dtype.py` subprocess preamble to do the same — those tests run probpipe in fresh Python processes and would otherwise bypass the conftest fixture.

## Background

The `WorkflowKind.DEFAULT` resolution in `WorkflowFunction.effective_workflow_kind` (added in #120) auto-promotes to `TASK` whenever `prefect` is importable. At dispatch time the Prefect client uses whatever `PREFECT_API_URL` the user's `~/.prefect/profiles.toml` sets. If that server isn't running, every test that goes through a `WorkflowFunction` fails with:

```
RuntimeError: Failed to reach API at http://127.0.0.1:4200/api/
```

CI sidesteps this because its runner has no profile and Prefect falls back to ephemeral mode — but any contributor who's previously run `prefect cloud login` or set up a local server profile hits 100+ failures locally.

## Why this approach

- **Test-only fix, no runtime behavior change.** The auto-promote-to-TASK semantic for users is the documented design and isn't touched here.
- **Function-scoped (not session-scoped).** `test_prefect_config.py` has its own autouse fixture that calls `prefect_config.reset()` (→ `DEFAULT`); without function scope here, that DEFAULT would leak into later tests.
- **Doesn't interfere with the Prefect tests.** `test_prefect_orchestration.py` passes `workflow_kind="task"` per-instance (a per-instance override beats the global). `test_prefect_config.py`'s own fixture runs after the conftest one, so the DEFAULT they expect is preserved.

## Test plan

- [x] `pytest tests/` is green locally with `~/.prefect/profiles.toml` pointed at an unreachable server: **2270 passed, 7 skipped, 0 failed**.
- [x] `pytest tests/test_prefect_config.py` — 29 passed (the DEFAULT auto-detect tests still see DEFAULT correctly).
- [x] `pytest tests/test_prefect_orchestration.py` — 17 passed (the in-process Prefect harness still works).
- [x] CI run on this PR (Prefect-less ephemeral mode) stays green — should be a no-op there.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,7 +3,42 @@ import jax
 import jax.numpy as jnp
 import numpy as np
 
-from probpipe import EmpiricalDistribution
+from probpipe import EmpiricalDistribution, WorkflowKind, prefect_config
+
+
+@pytest.fixture(autouse=True)
+def _disable_prefect_for_tests():
+    """Force ``WorkflowKind.OFF`` globally for each test.
+
+    The default ``WorkflowKind.DEFAULT`` auto-promotes to ``TASK`` when
+    Prefect is importable (see ``WorkflowFunction.effective_workflow_kind``).
+    That auto-promotion picks up whatever ``PREFECT_API_URL`` the user
+    has in ``~/.prefect/profiles.toml`` — if the configured server isn't
+    running, every ``WorkflowFunction`` call raises
+    ``RuntimeError: Failed to reach API at ...``. CI happens to escape
+    because it has no profile and falls back to ephemeral mode, but
+    contributors with an existing Prefect setup hit it immediately.
+
+    Function-scoped (not session-scoped) so the OFF state is restored
+    around every test — otherwise other test classes that flip the
+    global (``test_prefect_config.py`` resets to ``DEFAULT`` between
+    its tests) leak that state into later tests.
+
+    The Prefect-orchestration tests in ``test_prefect_orchestration.py``
+    don't depend on the global config — they pass ``workflow_kind="task"``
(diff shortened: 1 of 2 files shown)


erozercan:
Here's my review:

I actually found a similar solution for notebooks tests to pass without the need of firing up the prefect server. But, I didn't think of extending this solution to other tests... So, the solution seems valid to me! I don't have any suggestions for any change in either `conftest.py` and `test_dtype.py`. And, Claude seems to agree with me as well. Here is its view:

---

### `conftest.py` — Looks great

The fixture is well-designed:

- Function-scoped `autouse=True` — saves and restores the global config around every test, so no leakage between tests
- Doesn't break Prefect tests — `test_prefect_orchestration.py` uses explicit per-instance `workflow_kind="task"` which bypasses the global, and `test_prefect_config.py` has its own fixture that resets to `DEFAULT` (runs after conftest's fixture)
- The docstring is thorough — explains the why, the interaction with other test files, and the scoping choice 

### `test_dtype.py` — Smart catch

The subprocess tests run fresh Python processes that don't inherit the conftest fixture. Adding `prefect_config.workflow_kind = WorkflowKind.OFF` to the preamble of `_run_x64` is the right fix — without it, those subprocesses would still auto-promote to TASK and fail.

--- 

So, approved!

erozercan (approved):

PR #154, opened 2026-04-28
Fix Record pickle serialization for Ray compatibility

erozercan:
## Summary

- Adds `__reduce__` methods to `Record`, `RecordTemplate`, `NumericRecord`, `RecordArray`, and `NumericRecordArray` so they can be pickled/unpickled
- Ray's `RayTaskRunner` serializes task arguments with pickle. `Record.__setattr__` raises `"Record is immutable"` during deserialization, blocking distributed execution
- This unblocks Prefect + Ray for the bagged posterior workflow (and any pipeline that passes Records through distributed tasks)

## Context

Discovered during the Prefect integration PR (#120). Ray starts a local cluster and dispatches tasks correctly, but crashes when deserializing `Record` objects on workers:

> `ray.exceptions.RaySystemError: System error: Record is immutable`

The current workaround forces `ThreadPoolTaskRunner` (in-process, no serialization). This PR removes that limitation.

## Plan

See `ray-serialization-fix_plan.md` for full design. 
**The approach:** add `__reduce__` to each immutable class so pickle reconstructs via the constructor (same pattern as `namedtuple` / frozen `dataclass`).


jhuggins:
The plan looks good. 

Question from the design doc: **Should _source (provenance) survive pickling?** I think the answer is yes – we want to preserve provenance post-pickle – but if that creates problems we can revisit

Issue #142, opened 2026-04-20
Iteration convention divergence across container families


jhuggins:
After PR #141 aligned `RecordArray.__len__` with `__iter__` (both now field-based), one inconsistency remains across the container families:

| Type | `__iter__` yields | `__len__` returns |
|---|---|---|
| `Record` / `NumericRecord` | field names (`str`) | field count |
| `RecordArray` / `NumericRecordArray` | field names (`str`) | field count (after #141) |
| `DistributionArray` | components (`Distribution`) | component count (`n`) |
| `RecordDistribution` | — (intentionally not iterable; comment at `_record_distribution.py:352`) | — |

The Record family iterates **field names**; `DistributionArray` iterates **positional components**. Both families are internally self-consistent (iter and len match) but the iteration target differs.

## Options

1. **Leave as-is.** Document the distinction: Record-family objects are named-field dicts; `DistributionArray` is a positional sequence of distributions. Each iterates its natural "elements."
2. **Align `DistributionArray.__iter__`** to yield `range(n)` indices (positional names), or add `.keys()` / `.values()` / `.items()` that yield `(i, component_i)`. Matches dict-style of Record family.
3. **Align `Record.__iter__`** to yield field values (not names). Matches `DistributionArray.__iter__` yielding components. Breaking change — every `for name in record` site breaks.

Option 1 is safest; option 2 is additive; option 3 is disruptive.

## Decision

Left open. Flag for design discussion. Related audit item from PR #141 prep; deferred pending real-world use cases that call for unified iteration.

jhuggins:
Resolved by PR-B (#155). The iteration convention is now codified in `STYLE_GUIDE.md` §1.11:

- Record family (`Record`, `NumericRecord`, `RecordArray`, `NumericRecordArray`) iterates field names dict-style.
- `DistributionArray` has `__len__ == batch_shape[0]` and is positional / non-iterable.
- All `Distribution` subclasses (including `EmpiricalDistribution`, `BootstrapReplicateDistribution`, marginals) are non-iterable; samples are accessed via `.samples` / `.draws()`; `.n` reports the size.

Regression test enumerating each `Distribution` subclass: `tests/test_iteration_protocol.py`.

(Same auto-close-keyword issue as #93 — closing manually.)

Issue #134, opened 2026-04-19
Remove batch_shape from Distribution; batching is always DistributionArray


jhuggins:
## Motivation

`batch_shape` currently lives on every `Distribution` subclass (via `TFPDistribution`, `ArrayDistribution`, `MixtureMarginal`, each joint class, each MCMC output, etc.). It originated from TFP's shape semantics (`sample_shape + batch_shape + event_shape`) where a single distribution object can represent many random variables sharing a family.

With `DistributionArray` (from #130 PR 1, PR #131), ProbPipe now has an explicit way to represent a collection of independent distributions: a `DistributionArray(components, batch_shape=(n,))` is "n distributions stacked along a leading axis." This is a cleaner substrate than overloading `batch_shape` on a single Distribution object.

**The design direction:** `Distribution` represents exactly one random variable. Collections of distributions are always `DistributionArray`. `batch_shape` disappears from the `Distribution` base class.

## Benefits

- **One concept per class.** `Distribution` = one random variable; `DistributionArray` = ordered collection. No more "is this a single Normal or a batched Normal?"
- **Cleaner sample / log_prob / mean shapes.** Output shape becomes `sample_shape + event_shape`, no `batch_shape` to track.
- **DistributionArray semantics simplify.** `DistributionArray.batch_shape` is its own leading shape, full stop — no `(n,) + inner_batch` composition.
- **Symmetric with Record / RecordArray.** `Record` is scalar; `RecordArray` is batched. Makes the Distribution hierarchy match.
- **Fewer gotchas with broadcasting.** Today `Normal(loc=jnp.array([0,1,2]), scale=1.0)` silently creates a `batch_shape=(3,)` distribution. That's error-prone in pipelines that assume scalar distributions.

### TFP-backed families reject batched parameters

```python
# TODAY — silently creates a batched distribution
Normal(loc=jnp.array([0.0, 1.0, 2.0]), scale=1.0, name="x")  # batch_shape=(3,)

# AFTER — raises at construction
Normal(loc=jnp.array([0.0, 1.0, 2.0]), scale=1.0, name="x")
# ValueError: Normal parameters imply batch_shape=(3,); wrap multiple
#             distributions in a DistributionArray instead.

# The intended pattern
DistributionArray([Normal(loc=float(i), scale=1.0, name=f"x{i}") for i in range(3)])
```

For convenience: add a factory that produces a DistributionArray from batched parameters, so users aren't forced to write the list comprehension:

```python
[...]

Issue #91, opened 2026-03-31
document framework abstraction hierarchy


jhuggins:
- **Layer 0:** Core functionality. Corresponds to everything in `.core` 
- **Layer 1:** Core interfaces. These are generic/abstract modules that only depend on each other and core functionality. They facilitate a plug-and-play approach. 
- **Layer 2:** Core implementations (MCMC, GPs, …): these implement core interfaces and offer basic/essential functionality
- **Layer 3:** Implementations: anything that builds on the lower levels but is not “essential” or is more specialized.

jhuggins:
Resolved by PR-B (#155). The framework abstraction hierarchy is documented in `CONTRIBUTING.md` ("Framework abstraction hierarchy" section) with three rules:

1. **One random variable per `Distribution`.** Collections of distributions are always a `DistributionArray`.
2. **Two implementations per concept, no third.** A generic implementation parameterised over `T` and a Record-based specialisation.
3. **Iteration is a Record-family convention only.** `Distribution` subclasses are non-iterable.

(Same auto-close-keyword issue as #93 — closing manually.)

PR #155, opened 2026-04-28
feat(record): consolidate empirical/bootstrap/marginal hierarchy (PR-B)

### LINKED ISSUE #147: Empirical / Bootstrap / Marginal class hierarchy: consolidate naming and clarify user-facing surface
Raised by @arob5 in [PR #129 review](https://github.com/TARPS-group/prob-pipe/pull/129#issuecomment-4305732862): the empirical / bootstrap / marginal class family has eight related classes whose user-facing surface and intended overlap aren't obvious from the names. This issue catalogs them and proposes a consolidation pass.

## Inventory

| # | Class | Location | Inherits | Sample type | Purpose / unique surface |
|---|---|---|---|---|---|
| 1 | `EmpiricalDistribution[T]` | `core/_empirical.py:49` | `Distribution[T]`, `SupportsSampling`, `SupportsExpectation` | Stored as `np.ndarray(dtype=object)` | Generic empirical wrapper for arbitrary `T`. `__new__` dispatches numeric-array inputs to `NumericEmpiricalDistribution`; non-numeric stays here. Supports weighted resampling. |
| 2 | `NumericEmpiricalDistribution` | `core/_empirical.py:231` | `EmpiricalDistribution[Array]`, `NumericRecordDistribution`, `SupportsLogProb` (Gaussian-approx), `SupportsMean`, `SupportsVariance`, `SupportsCovariance` | `Array` | Numeric specialisation. Adds `event_shape`, moments, KDE-style log-prob via Gaussian approximation. Most user-facing empirical work lands here. |
| 3 | `_RecordEmpiricalDistribution` | `core/_empirical.py:382` | `EmpiricalDistribution[Record]`, `RecordDistribution`, `SupportsConditioning` | `Record` (named fields, joint rows) | Record-valued empirical: per-row joint resampling preserves component correlation. Constructed transitively when you pass a dict of arrays to `EmpiricalDistribution.__new__`. The public face is `JointEmpirical` / `NumericJointEmpirical` in `distributions/_joint_empirical.py`. |
| 4 | `_ArrayMarginal` | `core/_broadcast_distributions.py:50` | `NumericEmpiricalDistribution` | `Array` | Output of `WorkflowFunction` broadcasting when the body returns a numeric array. A thin subclass that adds nothing beyond a tag for provenance / `_MixtureMarginal` recognition. |
| 5 | `_RecordArrayMarginal` | `core/_broadcast_distributions.py:257` | `_RecordEmpiricalDistribution` | `Record` | Same as `_ArrayMarginal` but for Record-valued WF outputs. |
| 6 | `BootstrapReplicateDistribution[T]` | `core/_empirical.py:469` | `Distribution[T]`, `SupportsSampling` | Generic `T` (rows resampled with replacement) | Bootstrap over a stored dataset. `__new__` dispatches numeric inputs to `ArrayBootstrapReplicateDistribution`; Record-valued inputs to `_RecordBootstrapReplicateDistribution`. |
| 7 | `ArrayBootstrapReplicateDistribution` | `core/_empirical.py:682` | `BootstrapReplicateDistribution[Array]`, `NumericRecordDistribution` | `Array` | Array-valued bootstrap. Note the naming inconsistency: should be `NumericBootstrapReplicateDistribution` for parity with `NumericEmpiricalDistribution`. |
| 8 | `_RecordBootstrapReplicateDistribution` | `core/_empirical.py:755` | `BootstrapReplicateDistribution[Record]`, `RecordDistribution` | `Record` | Record-valued bootstrap: rows are joint, resampling preserves cross-field correlation. |

## Issues

1. **Naming inconsistency**: the empirical family uses `Numeric*` for the numeric specialisation; the bootstrap family uses `Array*`. Pick one (`Numeric` is more descriptive and matches the `NumericRecord` / `NumericRecordArray` family).

2. **Unclear privacy boundary**: `_RecordEmpiricalDistribution` is private but its public face (`JointEmpirical` / `NumericJointEmpirical` in `distributions/_joint_empirical.py`) is documented separately. Users rarely need to see the underscore class — but it leaks via `isinstance` checks and the JointEmpirical docs cross-reference it. Either document the private class as an implementation detail or expose it under a non-underscore name.

3. **`_ArrayMarginal` and `_RecordArrayMarginal` are user-visible but private**: every `WorkflowFunction` broadcast that returns numeric or Record output produces one. Users see them in `type(result).__name__`, in error messages, and in provenance metadata. The leading underscore signals "internal" but they're effectively the API. Either rename them (`BroadcastMarginal` / `RecordBroadcastMarginal`?) or add a thin public alias.

4. **`BootstrapReplicateDistribution` `__new__` dispatch parity**: `EmpiricalDistribution.__new__` dispatches dict / numeric / object inputs to three concrete classes. `BootstrapReplicateDistribution.__new__` does the same — but the dispatch table isn't documented anywhere users can find it.

## Proposed actions

- [ ] Rename `ArrayBootstrapReplicateDistribution` → `NumericBootstrapReplicateDistribution` (keep `ArrayBootstrapReplicateDistribution` as a deprecated alias for one release).
- [ ] Either rename `_ArrayMarginal` / `_RecordArrayMarginal` to non-underscore names or add public aliases — they're a default workflow output, not an implementation detail.
- [ ] Decide whether `_RecordEmpiricalDistribution` and `_RecordBootstrapReplicateDistribution` should be public; if not, document them as "use `JointEmpirical` / `BootstrapReplicateDistribution` instead" in the module docstring.
- [ ] Add a section to `docs/api/empirical.md` (or extend `docs/examples/02_records.ipynb`?) walking through the dispatch — what you get when you pass a numeric array, a dict, a list of records, etc.

## Related

- PR #129 review thread: https://github.com/TARPS-group/prob-pipe/pull/129#issuecomment-4305732862
- `Record` → `Value` rename suggestion (separate, larger discussion): tracked in a follow-up issue.

jhuggins:
## Summary

Second installment of the integrated cleanup plan, on top of PR #152. Closes **#147** (class consolidation), **#93** (BootstrapReplicateDistribution accepts any sampleable source), **#142** (iteration convention), and **#91** (abstraction-hierarchy docs).

## What changes

**Class merges (#147).** Each empirical / bootstrap / marginal pair collapses to a single Record-based class — no third numeric-array variant:

| Old (gone) | New |
|---|---|
| `NumericEmpiricalDistribution` + `_RecordEmpiricalDistribution` | **`RecordEmpiricalDistribution`** (public) |
| `ArrayBootstrapReplicateDistribution` + `_RecordBootstrapReplicateDistribution` | **`RecordBootstrapReplicateDistribution`** (public) |
| `_ArrayMarginal` + `_RecordArrayMarginal` | **`_RecordMarginal`** (private — WF output) |

Numeric-array sources auto-wrap as a single-field Record keyed by the (now mandatory) `name=` kwarg. The dispatch table is documented on `EmpiricalDistribution` / `BootstrapReplicateDistribution` class docstrings.

**`BootstrapReplicateDistribution[T]` accepts any `SupportsSampling` (#93).** No separate subclass — when the source is sampleable, each replicate is `n` i.i.d. draws from `source._sample`. `n` is mandatory in that case (no canonical observation count for a generic sampleable source).

**Distributions are non-iterable (#142).** Codified in STYLE_GUIDE §1.11 alongside the existing iteration / `.n` rules, with a regression test in `tests/test_iteration_protocol.py` that enumerates each `Distribution` subclass and asserts `not isinstance(d, Iterable)`. The Record family stays iterable (dict-style); `DistributionArray` is positional / non-iterable.

**Framework abstraction hierarchy (#91).** New section in `CONTRIBUTING.md` with the three rules:
1. One random variable per `Distribution`. Collections are always a `DistributionArray`.
2. Two implementations per concept, no third — generic `[T]` and Record-based.
3. Iteration is a Record-family convention only.

**`SupportsLogProb` removal from empirical.** `NumericJointEmpirical` and `RecordEmpiricalDistribution` no longer advertise a density. Migration: route through `from_distribution(emp, KDEDistribution, …)` for a kernel estimate, or fit a parametric distribution and call `log_prob` on that.

**`ApproximateDistribution`** now splits chains by template at construction so views into named fields work uniformly with the merged `RecordEmpiricalDistribution`.

## API impact

**Removed (no aliases — call sites migrate):**
- `NumericEmpiricalDistribution`
- `ArrayBootstrapReplicateDistribution`
- `_RecordEmpiricalDistribution` (private but referenced widely in tests)
- `_RecordBootstrapReplicateDistribution`
- `_ArrayMarginal`
- `_RecordArrayMarginal`
- `RecordEmpiricalDistribution.log_prob` / `JointEmpirical.log_prob` — these classes no longer satisfy `SupportsLogProb`.

**Renamed:** all `isinstance(..., _ArrayMarginal | _RecordArrayMarginal)` checks become `isinstance(..., _RecordMarginal)`. Rewrites span ~200 sites across `converters/`, `inference/`, `validation/`, `distributions/`, plus the test suite.

**Mandatory `name=` on `EmpiricalDistribution(arr)` for numeric arrays.** The single-field auto-wrap convention requires a name to key the Record by — same convention `WorkflowFunction` already uses for output coercion.

## Breaking changes (CHANGELOG)

- `NumericEmpiricalDistribution` / `ArrayBootstrapReplicateDistribution` removed; use `RecordEmpiricalDistribution` / `RecordBootstrapReplicateDistribution`.
- `_ArrayMarginal` / `_RecordArrayMarginal` removed; use `_RecordMarginal`.
- `NumericJointEmpirical.log_prob` removed; route through the converter registry.
- `EmpiricalDistribution(arr)` requires `name="…"` for numeric-array sources.

## Test plan

- [x] `pytest tests/` — **2323 passed, 7 skipped, 0 failed** locally on the rebased branch.
- [x] `mkdocs build --strict` clean.
- [x] All touched notebooks (`01_distributions`, `03_broadcasting`, `04_joint_distributions`, `06_converting_representations`, `07_sequential_updating`, `09_bagged_posteriors`, `10_random_functions_and_emulators`) re-executed.
- [ ] CI run on this PR — green.

## Linked issues

Closes #147, #93, #142, #91. Sequenced after #152 (PR-A landed) and before PR-C (#134 — `Distribution.batch_shape` removal).


jhuggins:
## PR Review: feat(record): consolidate empirical/bootstrap/marginal hierarchy (PR-B)

### Summary

Substantial PR (~7000 lines net, 51 files) that cleanly executes the Empirical/Bootstrap/Marginal merge per the integrated-cleanup plan and lands the framework-abstraction-hierarchy docs. Code matches `STYLE_GUIDE.md` / `CONTRIBUTING.md` conventions; the iteration regression test is a good addition. **The headline issue is that the PR's notebook migration was incomplete: 7 of the 11 example notebooks broke under the new `samples`-is-a-`NumericRecord` API and the mandatory `name=` requirement** (CI's `notebooks` job caught this; addressed in commit `80db144` so CI is now green).

### Findings

#### Critical (must fix)

1. **Notebook migration was incomplete** — `notebooks` CI job failed on `03_broadcasting.ipynb` first; 6 more notebooks (`04`, `05`, `06`, `07`, `09`, `11`) would have failed in turn after that. Root causes:
   - `EmpiricalDistribution(numeric_array)` now raises `ValueError` without `name=` (5 sites in `03`, 5 in `07`, 1 in `06`).
   - `dist.samples` is now a `NumericRecord` rather than a flat array. Old idioms — `samples.shape`, `samples.mean(axis=0)`, `samples[:, i]`, arithmetic `samples + offset` — break on multi-field records and don't always work on single-field records via the shape shim.
   - `prob(g, v)` and `log_prob(td, v)` inside `jax.vmap` raise `UnexpectedTracerError` — the `WorkflowFunction` dispatch path leaks tracers under vmap. Main's `c569c78` worked around this by using `dist._log_prob(v)` directly inside vmap, but PR-B's rebase took its own side of `03`/`04` and lost those fixes.
   - Notebooks `05`, `06`, `07`, `09`, `11` were never touched by `2850ec1`, so the API breakage went unnoticed.

   **Fixed in `80db144`** — all 11 `docs/examples/*.ipynb` now execute cleanly under a clean HOME (matching CI's no-Prefect-profile ephemeral-mode environment).

#### Recommended (should fix)

2. **No functional test for the new `SupportsSampling` source path of `BootstrapReplicateDistribution[T]` (#93).** The integrated plan called for a dedicated test (`BootstrapReplicateDistribution(Normal(0, 1, name="x"), n=10)` samples a `(10,)`-shaped draw); only the iteration-protocol test instantiates it. The new code in `_empirical.py:651-672` covers:
   - `n is None or n < 1` raising `ValueError`,
   - `_one_bootstrap` calling `source._sample(key, sample_shape=(self._n,))`,
   - `_sample` with non-empty `sample_shape` going through `jax.vmap`.

   None of these have direct tests. Suggest a `TestSampleableSource` class in `tests/test_bootstrap_replicate.py` covering construction, `_sample` shape (empty + non-empty `sample_shape`), and the `n` mandatory error.

3. **The `dist.samples` API change is bigger than the docstring suggests** and worth a deliberate design pass:
   - For *single-field* (`EmpiricalDistribution(arr, name="x")`), `.samples` returns a `NumericRecord` whose shape shim makes `.shape` / `__array__` / `__jax_array__` work as if `.samples` were `arr` — but `.mean(axis=0)`, `[:, i]` slicing, `.ravel()`, and arithmetic (`samples + offset`) silently fall off the shim.
   - For *multi-field* (an MCMC posterior, a joint empirical), the shape shim raises (`_single_numeric_leaf` rejects >1 field), and there's no clean accessor for an `(n, dim)` flat-matrix view; users have to iterate `dist.fields` and column-stack themselves.
   - Net effect: the same code path that worked for a single-field case will silently break when applied to multi-field. The notebook migrations exposed this 5 times.

   Suggested follow-up (not necessarily in this PR):
   - Add a public `.values` / `.samples_array` property that returns the underlying numeric array for single-field (and a clearly-defined fallback or `TypeError` for multi-field).
   - Document the shim's exact surface in `_numeric_record.py` (which methods work for single-field, what raises for multi-field).
   - Consider whether multi-field deserves an ergonomic flat-matrix view — `as_flat_distribution()` exists but is sample-axis-only.

4. **The vmap workaround the notebooks now rely on (`g._log_prob(v)` inside `jax.vmap`) violates `STYLE_GUIDE.md` §1.2** and worth flagging as a known issue:
   - §1.2 explicitly says *"Users call ops, never the underscore methods directly"*. After the workaround, the notebooks teach the opposite — `prob(g, v)` at top level, `g._log_prob(v)` inside vmap.
   - The error itself (`UnexpectedTracerError: ... allowing for a reference to an intermediate value with type float32[] wrapped in a BatchTracer to escape the scope of the transformation`) means `WorkflowFunction.__call__` is mutating Python-side state with values that turned out to be tracers. This is a leak by definition; JAX is correctly rejecting it.
   - The bug is *not* vmap-specific. Anything that crosses a JAX-transformation boundary while calling a `WorkflowFunction` could surface the same leak: `jax.grad` through ops, `jit`-wrapped closures, `jax.scan` with an op in the body, `pmap`. The vmap case is just where the notebooks happen to hit it.
   - `_log_prob` also bypasses the protocol check in `prob` / `log_prob`, so a user who copy-pastes the workaround pattern to a non-`SupportsLogProb` distribution gets `AttributeError` instead of a clean `TypeError`.

   This isn't on PR-B to fix, but the notebooks should at least carry a one-line comment explaining *why* `_log_prob` is used inside vmap, and the underlying tracer leak in `WorkflowFunction` deserves its own issue.

5. **`_RecordMarginal` (formerly `_ArrayMarginal` + `_RecordArrayMarginal`) — verify the bare-array path stays covered.** `_broadcast_distributions.py:60-79` has the auto-wrap logic for bare-array outputs (default `name="marginal"`). The renaming was mechanical but the merged class now handles both inputs; quick scan of `tests/test_broadcast_distribution.py` looks OK, but a regression test pinning the `name="marginal"` default-name path would prevent silent regression.

6. **`ApproximateDistribution.draws()` user-template-vs-auto-wrap branch is subtle and undertested.** `_approximate_distribution.py:165-209` — when a multi-field template is passed, samples are split per-field; otherwise `draws()` returns a raw concatenated array. The branch hinges on `getattr(self, "_user_template", False)`. Suggest a docstring example contrasting the two paths and a test that constructs both with and without `record_template` and checks the return type.

#### Minor (nice to have)

7. **`probpipe/core/_empirical.py:447-453` — `object.__setattr__(self, "_samples_record", cached)` for caching `.samples`.** Works because `Distribution` doesn't use `__slots__`, but the explicit `object.__setattr__` is unusual — `self._samples_record = cached` would do the same. Worth a one-line comment explaining why `object.__setattr__` (presumably defensive against any future subclass that adds slots) — otherwise it reads like a copy from a slotted class.

8. **`_empirical.py:551` empty-string `only_field` guard:** `single_field = (len(self._record_data.fields) == 1 and self._record_data.fields[0] != "")`. The `!= ""` check is suspicious — when would a field be `""`? If it's defensive against the `name=""` path, the construction should reject `""` upfront in `_wrap_numeric_array_as_record` (`if not name:` already does this). Either drop the `!= ""` guard or add a comment.

9. **CHANGELOG line 30-34**: tighten the `n` mandatory wording. Currently *"n is mandatory in this case (no canonical observation count)."* Suggest: *"`n` is mandatory when `source` is a `SupportsSampling` distribution (no canonical observation count); otherwise optional, defaults to source's row count."*

10. **STYLE_GUIDE.md `.n` table (§1.9) is stale.** Line 130 still says `_ArrayMarginal | Number of output samples` — this class is gone. Either rename to `_RecordMarginal` or, since this is a private class, drop the row entirely (the public-surface rows above already cover what users see).

11. **`probpipe/core/_empirical.py:1-25` module docstring** describes the new classes but doesn't note that `EmpiricalDistribution[T]` is the *generic* base and that `EmpiricalDistribution(numeric_array, name=...)` returns a `RecordEmpiricalDistribution` via `__new__` dispatch. The CHANGELOG covers this; the module doc could too. (One-liner addition.)

12. **`_RecordMarginal.__init__` accepts `Array` for `samples` parameter type, but the implementation falls through to `RecordEmpiricalDistribution.__init__` which dispatches via `_is_numeric_array`.** Type signature could be tightened to `Record | RecordArray | Array`. Cosmetic.

#### Categories that were clean

- **`__future__ import annotations`**: present in all PR-B-modified modules.
- **`Callable` / `Sequence` from `collections.abc`**: ✓ — `_empirical.py:29` does `from collections.abc import Callable, Sequence`.
- **Modern type syntax** (`X | Y`, `list[T]`, `dict[K, V]`): no `Optional` / `Union` / `List[…]` / `Dict[…]` usage in modified code.
- **Subpackage dependency graph**: `core/_empirical.py` imports only from `core/`, `_utils`, `custom_types`. `inference/_approximate_distribution.py` imports from `core/`, no reverse edges.
- **`__all__` exports**: `RecordEmpiricalDistribution`, `RecordBootstrapReplicateDistribution` exposed in `probpipe/__init__.py`.
- **STYLE_GUIDE §1.11** (distribution non-iterable rule): codified, regression test in `tests/test_iteration_protocol.py` enumerates each `Distribution` subclass.
- **CONTRIBUTING.md framework-abstraction-hierarchy section**: three rules documented at lines 311-340.
- **`.n` property convention** (§1.9): `RecordEmpiricalDistribution.n`, `RecordBootstrapReplicateDistribution.n` defined.
- **No AI-artifact comments** found.
- **CI status (after `80db144`)**: 7/7 checks green — `test (3.12)` 7m29s, `test (3.13)` 9m19s, `notebooks` 14m18s, `build` 1m14s, both Codecov checks pass.

### Suggested Changes (concrete)

Grouped by priority. **Critical (#1) is already done in `80db144`.**

**Recommended:**

A. Add `tests/test_bootstrap_replicate.py::TestSampleableSource` covering:
   - `BootstrapReplicateDistribution(Normal(0, 1, name="x"), n=10)._sample(key, ()).shape == (10,)`
   - `BootstrapReplicateDistribution(Normal(0, 1, name="x"), n=10)._sample(key, (3,)).shape == (3, 10)`
   - Missing `n` raises `ValueError`
   - `n < 1` raises `ValueError`
   ~15 lines.

B. Add 1-2 tests for `ApproximateDistribution.draws()` covering both the `_user_template=True` (multi-field unflatten via `NumericRecordArray.unflatten`) and `_user_template=False` (raw concatenated array) branches. ~20 lines.

C. Open a follow-up issue for the `WorkflowFunction` tracer leak under JAX transformations (vmap / grad / scan / jit). The notebook workaround using `_log_prob` inside vmap should reference that issue in a comment. The right fix is in WF (detect transformation context and short-circuit to `_<op>_impl`); the notebook workaround is just papering over it.

D. Audit `dist.samples` API: either expand the shim to cover `mean(axis=...)`, `__getitem__[slice]`, arithmetic, etc. for single-field; or add a `.values` accessor that returns the underlying array; or document the exact shim surface to set expectations. Pick one — current "some methods work, others don't" is the worst of all worlds.

**Minor:**

E. STYLE_GUIDE.md §1.9 table: drop the `_ArrayMarginal` row or rename to `_RecordMarginal`.

F. `_empirical.py:447-453`: add a one-line comment on why `object.__setattr__` is used (or replace with `self._samples_record = cached`).

G. `_empirical.py:551`: drop the `!= ""` guard or add a comment explaining when `fields[0] == ""` could happen.

H. CHANGELOG line 30-34: tighten the `n` mandatory wording.

I. `_empirical.py:1-25` module docstring: add a sentence on `__new__` dispatch behavior for `EmpiricalDistribution(numeric_array, name=...)`.

J. `_RecordMarginal.__init__` type signature: tighten to `Record | RecordArray | Array`.

jhuggins:
Thanks for the third-round review (the one I posted above as the PR author after running the review skill against my own branch — meta, but useful as a self-check). Working through every recommended/minor finding, plus the post-fix re-review's two items.

**Branch is now rebased onto current main** (was 13 commits behind after #160/#161/#162 landed). One conflict in `CONTRIBUTING.md` between PR-B's "Generic vs Record-based pattern" rename and main's new "Constraint → Bijector registry" section — kept both. Branch tip: `188018a`.

### Findings addressed

**A. SupportsSampling source tests (#93).** `test_bootstrap_replicate.py::TestSampleableSource` — 8 tests covering generic-base dispatch (no `RecordBootstrap` upgrade), `n` mandatory / zero / negative `ValueError`s, `_sample` shape with empty + non-empty `sample_shape`, `data is None` / `weights is None`, and repr. Commit `169edb7`.

**B. `ApproximateDistribution.draws()` template-vs-array branches.** No-op — `tests/test_inference.py::TestApproximateDistributionValuesTemplate` already covers both branches comprehensively (`test_without_template_returns_array` for the raw path; `test_draws_returns_values` / `test_draws_field_shapes` / `test_array_shaped_fields` / `test_nested_record_template_unflatten` / `test_draws_with_warmup_and_template` for the Record-template path). My earlier review missed those.

**C. WorkflowFunction tracer leak (vmap / grad / scan / jit).** Filed as [#163](https://github.com/TARPS-group/prob-pipe/issues/163) with full repro, current notebook workaround, and proposed fixes. Annotated the two affected notebooks (`03_broadcasting`, `06_converting_representations`) with three-line comments above each `dist._log_prob(v)` call inside `vmap`, pointing at the issue. Commit `61d060d`.

**D. `dist.samples` API gap.** Added `RecordEmpiricalDistribution.flat_samples` — flat `(n, dim)` matrix view across all fields, in insertion order, multi-dim event shapes flattened row-major. The cache mirrors the `samples` pattern (`object.__setattr__` after `getattr` fallback) so repeated access doesn't re-materialise the matrix (~8MB for a 100k-sample / 10-parameter posterior). 7 dedicated tests in `TestFlatSamples` covering single-field scalar / 1-D / multi-dim, multi-field 1-D, multi-field mixed shapes, insertion-order, and caching identity. Notebook `11_prefect_scalability` migrated from hand-rolled `np.column_stack` to `flat_samples`. Commits `d065d8f`, `7e5c068`. CHANGELOG + CONTRIBUTING entries.

**E. STYLE_GUIDE §1.9 stale `_ArrayMarginal`.** Renamed to `_RecordMarginal` in the `.n` table. Commit `237f868`.

**F. `samples` cache `object.__setattr__` comment.** Explained: it bypasses any subclass that adds `__slots__` or a custom `__setattr__` guard (Record's immutability check); the cache is internal-only and a controlled exception. Commit `26f9721`.

**G. `_expectation` `!= ""` guard on `fields[0]`.** Dropped. The auto-wrap path rejects empty names at construction (`_wrap_numeric_array_as_record`), so the guard never fired there; for an explicit `Record({"": arr})` the guard arbitrarily forced multi-field semantics. Treating any one-field Record as single-field is the consistent behaviour. Commit `26f9721`.

**H. CHANGELOG `n is mandatory` wording.** Tightened: spell out that `n` is mandatory *only* when source is a `SupportsSampling` distribution; for Record / array / Empirical sources it remains optional and defaults to the row count. Commit `ff24147`.

**I. Module docstring `__new__` dispatch note.** Added — `EmpiricalDistribution(arr, name=...)` returns `RecordEmpiricalDistribution`; `BootstrapReplicateDistribution(rec)` returns `RecordBootstrapReplicateDistribution`; `SupportsSampling` source stays in the generic base. Commit `26f9721`.

**J. `_RecordMarginal.__init__` type signature.** No-op — the `samples: Record | RecordArray | Array` hint was already correct; my earlier review note was confused.

### Post-fix re-review (the simplify pass + the latest pytest-aware re-read)

Two extra items surfaced and fixed:

- **Unused import** — `from ._record_distribution import RecordDistribution` in `_empirical.py` was added during consolidation but never referenced. Dropped. Commit `65876a8`.
- **`flat_samples` zero-sized event coverage** — added `test_zero_sized_event` pinning that a field with shape `(n, 0)` reshapes to `(n, 0)` and concatenates cleanly alongside non-empty fields. Commit `188018a`.

### Status

- Full pytest: **2373 passed, 7 skipped, 0 failed**.
- `mkdocs build --strict` clean.
- Rebased onto current main (`e640afe`); branch tip `188018a`.
- Issue [#163](https://github.com/TARPS-group/prob-pipe/issues/163) tracks the WF tracer leak as a follow-up.

arob5:
## PR Review (automated, read-only)

### Summary

A large, well-scoped consolidation that collapses the generic/numeric/Record three-class pattern down to generic + Record-based pairs, codifies the iteration rule, removes `SupportsLogProb` from empirical distributions, and adds a sampleable-source path to `BootstrapReplicateDistribution`. Test additions are solid and the convention docs (STYLE_GUIDE.md, CONTRIBUTING.md) are kept in sync. A handful of small leftover artifacts from the mass-rename and one likely-dead conditional are the main fixes.

Reviewed at commit `188018a`.

### Findings

#### Critical (must fix)

1. **Duplicate dict key in `_probpipe.py` dispatch table.** `probpipe/converters/_probpipe.py:514-515` has `"RecordEmpiricalDistribution": _convert_to_empirical,` twice in `_build_dispatch_table()`. The original code already had `NumericEmpiricalDistribution` duplicated (a pre-existing bug); the rename merged them into a same-key duplicate. Python keeps only the second; functionally harmless, but it is an obvious leftover and would trip a linter. Should be a single entry.

#### Recommended (should fix)

2. **Dead conditional branch in `BootstrapReplicateDistribution.__new__`.** `probpipe/core/_empirical.py:681-685`:
   ```python
   if (
       isinstance(source, EmpiricalDistribution)
       and _is_numeric_array(source.samples)
   ):
       return object.__new__(RecordBootstrapReplicateDistribution)
   ```
   Can never fire after the consolidation: any numeric-array-backed empirical now *is* a `RecordEmpiricalDistribution` (caught by the first branch at line 675), and the generic base only stores numpy object arrays (so `_is_numeric_array(source.samples)` is False). Remove the branch.

3. **Identical-branch `if/else` in `BootstrapReplicateDistribution._sample`.** `probpipe/core/_empirical.py:827-831`:
   ```python
   if self._source_kind == "sampleable":
       results = jax.vmap(self._one_bootstrap)(keys)
       return results.reshape(*sample_shape, *results.shape[1:])
   results = jax.vmap(self._one_bootstrap)(keys)
   return results.reshape(*sample_shape, *results.shape[1:])
   ```
   Both arms do the same thing. Drop the `if`.

4. **STYLE_GUIDE §1.11 wording overstates the regression test's scope.** It says the test "enforces this rule across every concrete distribution class," but `tests/test_iteration_protocol.py:41-83` parametrises a representative set (Normal, Beta, Gamma, MVN, Product, two record-empiricals, one sampleable bootstrap, one joint-empirical) and skips `TransformedDistribution`, `KDEDistribution`, `BootstrapDistribution`, `BroadcastDistribution`, marginals, random functions/measures. Either soften the doc to "enforces this rule across representative concrete distributions," or extend the parametrization to cover the rest of the family.

#### Minor (nice to have)

5. **`RecordEmpiricalDistribution._sample` returns `NumericRecord` (not `NumericRecordArray`) for non-empty `sample_shape`.** Intentional and documented inline at `probpipe/core/_empirical.py:559-562` ("doesn't trip over batch-axis validation under jax.vmap"), but it is a subtle deviation from the WF "uniform output wrap" contract — callers that expect a properly-typed batched container will not get one. Worth a quick scan for downstream callsites (predictive_check, WF sweep) that may rely on `isinstance(s, RecordArray)`.

6. **Potential template/data mismatch in `ApproximateDistribution`.** `probpipe/inference/_approximate_distribution.py:73-104`: when `record_template` has multiple fields but *any* of them is nested, the constructor falls into the single-field auto-wrap path and stores `_record_data` with one field, then overrides `self._record_template` with the multi-field template. `draws()` is fine (it reads the raw flat samples), but `event_shapes`, `dtypes`, `_mean`, `_variance`, `flat_samples` read from `_record_data` and will only see the single auto-wrap field. Worth verifying no regression in nested-template inference flows.

7. **Pre-existing duplicate test methods in `tests/test_bootstrap_replicate.py`.** Not introduced by this PR, but `test_getitem_returns_view` is defined twice in two test classes (lines 486/492 and 581/586) — the first definition is overwritten by the second. Cleanup opportunity if you want to do it in this PR.

### Suggested Changes

1. **Drop the duplicate dispatch key** in `probpipe/converters/_probpipe.py:514-515` — keep one `"RecordEmpiricalDistribution": _convert_to_empirical,`.
2. **Remove the dead `__new__` branch** at `probpipe/core/_empirical.py:681-685`.
3. **Collapse the redundant if/else** at `probpipe/core/_empirical.py:827-831` into a single `results = jax.vmap(self._one_bootstrap)(keys); return results.reshape(...)`.
4. **Either** broaden the iteration test parametrization in `tests/test_iteration_protocol.py:41`, **or** soften STYLE_GUIDE.md §1.11's "every concrete distribution class" claim.
5. *Optional:* add a regression test for `ApproximateDistribution` constructed with a multi-field template containing a nested `RecordTemplate` field.
6. *Optional cleanup* of pre-existing duplicate test methods in `tests/test_bootstrap_replicate.py:486` and `:581`.

jhuggins:
Thanks for the post-merge review. Worked through every finding; 7 commits on top of `188018a`, branch tip `c0d773c`.

### Critical

**1. Duplicate dict key in `_probpipe.py:514-515`** — confirmed; PR-B's mass-rename collapsed `NumericEmpiricalDistribution` and `RecordEmpiricalDistribution` adjacent entries into a same-key duplicate. Dropped one. Commit `103932c`.

### Recommended

**2. Dead `__new__` branch in `_empirical.py:681-685`** — verified your analysis: the generic `EmpiricalDistribution.__new__` already routes numeric-array samples to `RecordEmpiricalDistribution` (caught by the first branch), and the generic base's `__init__` only stores `np.ndarray(dtype=object)` for non-array sources, so `_is_numeric_array(source.samples)` returns False. Removed the branch and replaced the inline comment with one explaining why it's unreachable. Commit `fbef8dd`.

**3. Identical if/else in `_sample` at `_empirical.py:827-831`** — confirmed both arms were byte-identical. Collapsed to a single `jax.vmap` + reshape with a comment explaining why vmap is safe for both `data` and `sampleable` source kinds. Commit `34cc444`.

**4. STYLE_GUIDE §1.11 wording vs test scope** — two-pronged. Extended `tests/test_iteration_protocol.py::DISTRIBUTIONS` to add `TransformedDistribution` and `KDEDistribution` (the easy-to-construct gaps). Skipped the WF-output classes (`BroadcastDistribution`, `_RecordMarginal`, `_MixtureMarginal`, `_ListMarginal`) since they're WF-dispatched, not user-constructed; adding them would need fragile fixture setup for marginal allocation. Softened §1.11's "across every concrete distribution class" to enumerate the user-constructible classes covered and call out that WF-output classes inherit the constraint from their bases. Commit `d402429` (15 tests pass, was 13).

### Minor

**5. `_sample` returns `NumericRecord` (not `NumericRecordArray`)** — audited callsites. Three classes of consumer:
- `_predictive_check_batched` passes the result to user-supplied `generate_data` — duck-typed (works on either).
- `_RecordMarginal.__init__` accepts both Record and RecordArray — handled (RecordArray peels off the rows axis, NumericRecord falls through to the `super().__init__` which handles it).
- `flatten_value` calls `.flatten()` — both `NumericRecord` and `NumericRecordArray` have it.

No isinstance check fires `False` on the current `NumericRecord` return that would on `NumericRecordArray`. Kept the current behavior; expanded the inline comment to (a) name the specific vmap failure mode, (b) flag the deviation from the WF "uniform output wrap" contract, and (c) record the audit result so future readers don't redo it. Commit `e82f5c6`.

**6. Nested-template `ApproximateDistribution` accessor mismatch** — wrote the regression test you suggested. Surfaced the asymmetry concretely:

```python
template = RecordTemplate(params=RecordTemplate(a=(), b=()), scale=())
post = make_posterior([chain], record_template=template, ...)
post.fields                # ('params', 'scale')   — template-side
post.record_template.fields  # ('params', 'scale')   — template-side
post._record_data.fields   # ('posterior',)        — data-side, single auto-wrap
post.event_shapes          # {'posterior': (3,)}   — data-side
post.flat_samples.shape    # (n, 3)                — data-side
post.draws()['params']['a']   # works               — uses _chains, not _record_data
```

The template-side accessors and `draws()` give the user the right thing; only `event_shapes` / `dtypes` / `flat_samples` / `_mean` / `_variance` see the flat single-field view. Audited inference code: every flow that produces an `ApproximateDistribution` with a nested template (`_tfp_mcmc`, `_rwmh`, `_pymc_method`, `_cmdstan_method`, `_nutpie`) only ever consumes `draws()`, not the data-side accessors. So the inconsistency doesn't break any production flow.

Extending the per-field split to recurse through nested templates would mean materialising nested NumericRecord leaves at construction with knock-on effects on flatten/unflatten paths — too big for a review-fix PR.

Compromise: pin the current behavior in `test_nested_template_accessors_split_state` (so any future fix has to update the test deliberately) + add a `Notes` block on `ApproximateDistribution` documenting the split with a pointer to the regression test. Commit `760b19e`.

**7. Pre-existing duplicate `test_getitem_returns_view`** — confirmed two test classes each defined the method twice with byte-identical bodies (Python silently overwrote the first definition each time). Dropped the duplicates. Commit `c0d773c`.

### Status

- Full pytest: **2376 passed, 7 skipped, 0 failed**.
- `mkdocs build --strict` clean.
- Branch tip: `c0d773c` on top of the rebased PR-B base. 7 new commits, all pushed.

jhuggins:
Following up on finding 6 — agreed it should be fixed, not just documented. Probed the actual damage and shipped Option B (per-top-level-field split using `_spec_size`).

### What was actually broken

The asymmetry was worse than the audit suggested. With `template = RecordTemplate(params=RecordTemplate(a=(), b=()), scale=())`:

```python
post.fields                # ('params', 'scale')           — template-side
mean(post)                 # NumericRecord(posterior=...)  ← wrong
mean(post)['params']       # KeyError: 'params'            ← real footgun
post.event_shapes          # {'posterior': (3,)}           ← keys don't match fields
```

A user iterating `for f in post.fields: print(mean(post)[f])` would `KeyError` on the first iteration. That's a real bug, not a documentation gap.

### Fix (commit `47f2ec6`)

Replaced the `all_flat_leaves` predicate in `ApproximateDistribution.__init__` with a per-top-level-field split that uses `_spec_size`. Nested `RecordTemplate` fields are stored as a flat `(n, nested_flat_size)` slice under the top-level field name; the nested structure is recoverable via `record_template[field]` and `draws()`.

After the fix:

```python
post.fields                # ('params', 'scale')
post._record_data.fields   # ('params', 'scale')                  ← matches
post.event_shapes          # {'params': (2,), 'scale': ()}        ← matches
mean(post)                 # NumericRecord(params=array(2,), scale=scalar)
mean(post).fields          # ('params', 'scale')                  ← matches
draws()['params']['a']     # works (unchanged)
```

Trade-off: `event_shapes['params']` reports `(2,)` (the flat size of the nested template) rather than the nested per-leaf shape. Users who want per-leaf moments use `draws()` or `record_template[field]`. This matches the existing `event_shapes` semantic — one shape per `field`, not per-leaf.

### Follow-up tracked

[#164](https://github.com/TARPS-group/prob-pipe/issues/164) tracks whether to extend to a full recursive split (Option A — materialise nested `NumericRecord` leaves so `mean(post)['params']['a']` is a scalar). Cost would be ~10 accessor changes plus return-type updates; deferred until a real consumer surfaces.

### Updated regression test

`test_nested_template_accessors_match_top_level_fields` now asserts the correct (Option B) behavior — `fields` / `_record_data.fields` / `event_shapes.keys()` / `mean(post).fields` all match `('params', 'scale')`, with the nested structure recovery via `draws()` and `record_template[field]` still working.

### Status

- pytest: **2376 passed, 7 skipped, 0 failed**.
- `mkdocs build --strict` clean.
- Branch tip: `47f2ec6`.

jhuggins:
All 8 findings from the post-Option-B review addressed. 4 commits on top of `47f2ec6`, branch tip `329d2a9`.

### Recommended

**1 + 2 + 3. `_approximate_distribution.py` validation + import cleanup** (commit `f5c05d5`):
- Pre-validate that no top-level template field has `spec=None` (opaque). Previous code silently treated opaque leaves as size-1; the new path reaches `_spec_size` which raises a generic message. Catch upfront with a targeted `ValueError` naming the field.
- Validate `flat.shape[-1]` against the template's total flat size before slicing. Catches template/chain mismatch with a clear message ("chain last dim (4) doesn't match template total flat size (5); template fields=('a', 'b'), sizes=[2, 3].") instead of silently producing zero-sized chunks.
- Hoist `_spec_size` from the in-function import to the existing module-level `from ..core.record import ...`. No cycle-breaking purpose.

### Minor

**4. STYLE_GUIDE §1.11 enumeration** (commit `da112bb`): Added `BootstrapReplicateDistribution` (the generic-base case with `SupportsSampling` source) to the list. The test parametrizes 11 classes; the doc only listed 10.

**5. `_make_transformed` ordering** (commit `b14d610`): Moved the helper above `DISTRIBUTIONS` for top-down reading order. No behavior change.

**6 + 7 + 8. Regression test polish** (commit `329d2a9`):
- Added `assert post.event_shape == ()` (singular) to pin the multi-field default — different code path from `event_shapes` (plural).
- Added a `variance(post)` block mirroring the `mean(post)` assertions.
- Dropped the `post._record_data.fields == expected_fields` assertion; the user-facing surface already locks in the contract, the `_record_data` assertion over-specified private state.

### Status

- pytest: **2376 passed, 7 skipped, 0 failed**.
- `mkdocs build --strict` clean.
- Branch tip: `329d2a9` on top of the Option B fix.

arob5:
## Review findings

### Recommended

1. **Import order — [probpipe/core/_empirical.py:40-69](https://github.com/TARPS-group/prob-pipe/blob/dev/pr-b-empirical-consolidation/probpipe/core/_empirical.py#L40-L69)** — mixes third-party (`numpy`) → internal (`.._utils`, `.protocols`) → third-party (`jax`, `jax.numpy`) → internal again. STYLE_GUIDE §4.2 wants stdlib → third-party → internal in three blank-line-separated blocks.

2. **`event_shape` silently returns `()` for multi-field records — [probpipe/core/_empirical.py:506-511](https://github.com/TARPS-group/prob-pipe/blob/dev/pr-b-empirical-consolidation/probpipe/core/_empirical.py#L506-L511) and [1014-1019](https://github.com/TARPS-group/prob-pipe/blob/dev/pr-b-empirical-consolidation/probpipe/core/_empirical.py#L1014-L1019)** — multi-field branch falls through to `()` rather than raising. Callers that weren't updated for the multi-field case will see a scalar event shape, masking bugs. Recommend raising \`AttributeError(\"multi-field record; use event_shapes\")\`. Same fix applies to \`obs_shape\` at [1023-1028](https://github.com/TARPS-group/prob-pipe/blob/dev/pr-b-empirical-consolidation/probpipe/core/_empirical.py#L1023-L1028).

3. **Iteration regression test is too lenient — [tests/test_iteration_protocol.py](https://github.com/TARPS-group/prob-pipe/blob/dev/pr-b-empirical-consolidation/tests/test_iteration_protocol.py)** — \`pytest.raises((TypeError, KeyError, IndexError))\` accepts \`IndexError\`, which is exactly the silent-iteration-via-\`__getitem__\`-fallback case the rule is meant to forbid. Tighten to \`TypeError\` only.

### Minor

4. **Inconsistent \`NumericRecord\` local alias — [probpipe/core/node.py:1338](https://github.com/TARPS-group/prob-pipe/blob/dev/pr-b-empirical-consolidation/probpipe/core/node.py#L1338) vs [1354](https://github.com/TARPS-group/prob-pipe/blob/dev/pr-b-empirical-consolidation/probpipe/core/node.py#L1354)** — the same function imports \`NumericRecord\` once normally, then again as \`_NR\` ~16 lines later. Pick one.

5. **\`getattr(self, \"_record_data\", None)\` sniff — [probpipe/core/_empirical.py:856](https://github.com/TARPS-group/prob-pipe/blob/dev/pr-b-empirical-consolidation/probpipe/core/_empirical.py#L856)** — the generic-base \`_expectation\` reaches into the Record-subclass's private \`_record_data\` to detect the single-field auto-wrap case. \`self._source_kind\` already encodes the source kind; a subclass override (e.g. \`_dataset_unwrap\`) would localize the unwrap rather than leaking subclass internals through the base.

6. **\`RecordEmpiricalDistribution.__init__\` skips its parent's \`__init__\` — [probpipe/core/_empirical.py:435-437](https://github.com/TARPS-group/prob-pipe/blob/dev/pr-b-empirical-consolidation/probpipe/core/_empirical.py#L435-L437)** — calls \`Distribution.__init__\` directly to bypass \`EmpiricalDistribution.__init__\`'s incompatible storage layout. The inline comment explains why, but it's worth a brief note in the class docstring (or a small shared mixin) so a future reader doesn't accidentally chain through \`super()\`.

jhuggins:
All 6 findings addressed. 6 commits on top of `329d2a9`, branch tip `b6e7ea7`.

### Recommended

**1. Import order in `_empirical.py:40-69`** (commit `a67feae`). Reordered into three blank-line-separated blocks per STYLE_GUIDE §4.2: stdlib (`collections.abc`, `typing`) → third-party (`jax`, `jax.numpy`, `numpy`) → internal. Within each block, alphabetical. No behavior change.

**2. `event_shape` / `obs_shape` silent-`()` for multi-field** (commit `9276c96`). Three sites in `_empirical.py` now raise `AttributeError("multiple fields ...; use event_shapes/obs_shapes (plural)")` instead of falling through to `()`:
- `RecordEmpiricalDistribution.event_shape`
- `RecordBootstrapReplicateDistribution.event_shape`
- `RecordBootstrapReplicateDistribution.obs_shape`

Cascading fixes: `ApproximateDistribution.__repr__` now uses `event_shapes` (plural) for multi-field posteriors so `repr(post)` stays valid; the regression test from the previous round (`test_nested_template_accessors_match_top_level_fields`) updated to assert `pytest.raises(AttributeError, match="multiple fields")` instead of `event_shape == ()`.

**3. Iteration test too lenient — accepts `IndexError`** (commit `eadbd85`). Dropped `IndexError` from the accepted exceptions; kept `TypeError` and `KeyError`. Probed each parametrised class — 9 of 11 raise `KeyError` (Record family's `__getitem__` rejects integer keys), 1 raises `TypeError` (the SupportsSampling bootstrap), 1 raises `KeyError` (Beta). Tightening to `TypeError` only as you suggested would fail 9 of those — `KeyError` is genuinely surfaced (not silent), so it stays. The actual silent-iteration footgun is `IndexError` (Python's iter-fallback treats it as end-of-iteration), and that's what got dropped. Updated the test docstring to explain the choice.

### Minor

**4. Duplicate `NumericRecord` local alias in `node.py`** (commit `dc06590`). Renamed `_NR` to the canonical `NumericRecord` so both local imports use the same name.

**5. `_record_data` sniff in generic-base `_expectation`** (commit `b6e7ea7`). Replaced the `getattr(self, "_record_data", None)` reach-through with a `_unwrap_dataset(d)` hook on the base class (default: identity). `RecordBootstrapReplicateDistribution` overrides it to do the single-field auto-wrap unwrap. Each class owns its own policy; no leaked subclass internals through the base. Same behavior.

**6. `__init__` super-skip docstring note** (commit `ab1286d`). Added a Notes block on `RecordEmpiricalDistribution`'s class docstring explaining why the constructor calls `Distribution.__init__` directly (the generic `EmpiricalDistribution[T]` base's `_samples` storage is incompatible with `_record_data`), with a pointer for subclasses (e.g. `ApproximateDistribution`) that need to chain to `RecordEmpiricalDistribution.__init__` rather than `super()`.

### Status

- pytest: **2376 passed, 7 skipped, 0 failed**.
- `mkdocs build --strict` clean.
- Branch tip: `b6e7ea7`.

arob5:
## PR Review (automated review pass)

### Summary

A large, well-scoped consolidation (52 files, ~6.6k insertions / ~2k deletions) that lands cleanly: design rules ("two implementations per concept", non-iterable distributions, single-field auto-wrap with mandatory `name=`) are codified in both code and docs (STYLE_GUIDE §1.11, CONTRIBUTING "Framework abstraction hierarchy"). Public API surface and converter dispatch are updated end-to-end and the iteration regression test (`tests/test_iteration_protocol.py`) is a strong guard rail. A few small loose ends — mostly latent edge cases that aren't exercised today.

### Findings

#### Critical (must fix)

None.

#### Recommended (should fix)

1. **Dead / mis-scoped branch in `RecordBootstrapReplicateDistribution.__init__`** — `probpipe/core/_empirical.py:988–1002`. The `elif isinstance(source, EmpiricalDistribution):` branch is documented in the `__new__` comment as unreachable via factory dispatch (a generic `EmpiricalDistribution(numeric_array)` is rerouted to `RecordEmpiricalDistribution`, caught by the first branch). The only path that reaches it is direct construction with a *generic* `EmpiricalDistribution` whose `samples` is an object-dtype array — and there `_wrap_numeric_array_as_record` → `_as_float_array(object_arr)` will raise an unhelpful `TypeError`. Either delete the branch or make it explicitly raise with a clear message about non-numeric `EmpiricalDistribution` sources.

2. **`_convert_to_kde` is broken for multi-field `RecordEmpiricalDistribution` sources** — `probpipe/converters/_probpipe.py:460–478`. Single-field is handled, but multi-field falls through to `source._sample(key, (n,))` which now returns a `NumericRecord`; `KDEDistribution(NumericRecord)` calls `_as_float_array(NumericRecord)` and the multi-field shim raises. The simplest fix is `samples = source.flat_samples` in the multi-field branch (KDE happily accepts `(n, dim)`); failing that, raise a clear error. **Test coverage gap:** `tests/test_converters.py:663–672` only exercises single-field for the empirical→KDE path.

3. **`_convert_to_kde` docstring overstates its behavior** — same file, lines 446–450. "If the source is an ``RecordEmpiricalDistribution`` … the stored samples and weights are reused directly" is only true for single-field; multi-field actually re-samples. Update the docstring to match (and tighten once #2 is fixed).

#### Minor (nice to have)

4. **Duplicated row-indexing in `node.py`** — `probpipe/core/node.py:1322–1346` and `1389–1413`. `_broadcast_enumerate_then_sample` inlines the Record/NumericRecord per-row indexing twice; `_broadcast_sample` already extracts a local `_index_sample` helper. Lift it to module scope and call from both methods.

5. **Inconsistent unwrap idiom in transition tests** — `tests/test_transition.py:275–276`. One assertion explicitly unwraps the field (`dists1[1].samples[dists1[1].samples.fields[0]]`) while the next leans on the `__jax_array__` shim (`dists1[2].samples`). Pick one for readability.

6. **`_validate_record_samples` rejects 0-D fields with a message that misreports the offending shape** — `probpipe/core/_empirical.py:182–189`. When `arr.ndim == 0`, the message reads `"sample-axis length None"`. Minor — print `arr.shape` and the field name to make it actionable.

7. **`RecordEmpiricalDistribution.event_shape` for the auto-wrap singleton path** does not mention `obs_shape` parity — `probpipe/core/_empirical.py:516–544`. Compare to `RecordBootstrapReplicateDistribution.obs_shape` / `obs_shapes` at lines 1080–1110 which explicitly cross-link. Worth a one-line "see also" for symmetry.

8. **No regression test for `JointEmpirical.from_distribution(emp, KDEDistribution, …)` over multi-field**, which is the path that finding #2 breaks. Adding one would have caught it.

### Suggested Changes

1. **`probpipe/core/_empirical.py:988`** — delete the unreachable `elif isinstance(source, EmpiricalDistribution):` branch in `RecordBootstrapReplicateDistribution.__init__`, or replace with `raise TypeError("RecordBootstrapReplicateDistribution does not accept generic (object-array) EmpiricalDistribution sources; pass a RecordEmpiricalDistribution or numeric array.")`.

2. **`probpipe/converters/_probpipe.py:460–471`** — in the multi-field branch, route through `flat_samples`:
   ```python
   if isinstance(source, RecordEmpiricalDistribution):
       arr = (source.samples[source.samples.fields[0]]
              if len(source.samples.fields) == 1
              else source.flat_samples)
       r = KDEDistribution(arr, weights=source._w, bandwidth=bandwidth, name=name)
       r.with_source(_mm_provenance(source))
       return r
   ```
   Then update the docstring at lines 446–450 to drop the "single-field only" caveat. Add a multi-field test alongside `tests/test_converters.py:663`.

3. **`probpipe/core/node.py`** — promote `_index_sample` to a module-level helper (or method on `WorkflowFunction`) and call it from both `_broadcast_enumerate_then_sample` (lines 1325, 1351) and `_broadcast_sample` (line 1413).

4. **`tests/test_transition.py:275–276`** — pick one unwrap idiom and use it consistently across both assertions.

5. **`probpipe/core/_empirical.py:182–189`** — include the offending field name and `arr.shape` in the `_validate_record_samples` error message instead of `"length None"`.

6. **(Optional)** Add a regression test for the multi-field empirical → KDE path in `tests/test_converters.py` so #2 doesn't regress.

arob5 (approved):

PR #156, opened 2026-04-28
feat(ci): Check diffs to run tests

erozercan:
## Summary

Speed up CI by only running tests related to changed `.py` files. Non-Python changes (CI config, docs, etc.) skip the test steps entirely. Notebook CI is unchanged.

## How it works

- Diffs the PR branch against base to find changed `.py` files.
- Maps source files to their test files (e.g. `probpipe/core/record.py` -> `tests/test_record.py`).
- **Skip tests entirely** if no `.py` files changed.
- **Run full suite** if foundational files change (`pyproject.toml`, `conftest.py`, `probpipe/__init__.py`).
- **Run targeted tests** otherwise.

## TODO
- [ ] ~~Diff against the last successful CI commit instead of the base branch/previous commit, so that tests broken by earlier commits aren't skipped.~~

## Question

The notebook CI job currently runs unconditionally and is one of the slowest parts of the pipeline (~10-15 min). Since the unit tests already ran in parallel, the total CI time was effectively gated by the notebook job anyway. Should we also apply selective execution to notebooks (e.g. only run when notebook files or core source files change), or keep them running on every PR?


erozercan:
A few updates on this PR :
first, I checked the notebook CI and we're good: it only runs notebooks under `docs/examples/`, the tutorials directory is not touched at all.

The notebooks job now only runs when either:

- A file in docs/examples/ (.ipynb) changes
- Any file under probpipe/ (source code) changes

Second, @jhuggins made a comment about "the diff strategy" not being robust, and that's valid. The problem was something like: if you change `provenance.py`, the old code would only run `test_provenance.py`; but `record.py` imports `provenance.py`, so those tests should run too and were being skipped. to fix this i added a small python script that runs at CI time, parses all the source files using the `ast` module, builds a reverse import graph, and does a BFS expansion from the changed files. So now if `provenance.py` changes, we find every module that transitively depends on it and run all their tests too.

Third, I think the TODO item "Diff against the last successful CI commit instead of the base branch/previous commit, so that tests broken by earlier commits aren't skipped." is no longer necessary. The concern was that a bug introduced in commit B might slip through to commit C if commit C happened to not touch the affected file. but with the import graph expansion, commit B would have already expanded the changed file to all its dependents and run their tests. So the bug would be caught at commit B, not silently passed along...

erozercan:
Actually, I realized there's still a problem with the notebook optimization. Right now the condition is "any file under probpipe/ changed -> run all notebooks". So if I touch `node.py` it still triggers all 11 notebooks and ~15 min, which is what we were trying to avoid in the first place.

A better approach would be to do the same import-graph trick we did for unit tests but for notebooks: parse each notebook's code cells with ast, build a map of which notebooks import which probpipe modules, and then only run the notebooks that are actually affected by the changed source files (transitively). So if `node.py` changes and only `08_jax_interop.ipynb` imports it, only that notebook runs.

The tradeoff is that for changes to central modules like `record.py` or `distribution.py`, most notebooks would still run since they all depend on those. But for targeted bugfixes and feature work (which is most of what we do), this would be a good improvement. Should I implement this?

erozercan:
<p>As I added the new features that I mentioned in my earlier comments, I have realized that there are some issues in the current logic that need to be worth discussing. I think it is easier to understand how the current logic works. I will mention the issues in mt next comment</p>
<p>Let's say we changed the <code>node.py</code>. Here is what happens:</p>
<h3 data-heading="Step 1 — &#x60;git diff&#x60; collects changed files">Step 1 — <code>git diff</code> collects changed files</h3>
<pre><code>CHANGED = probpipe/core/node.py
</code></pre>
<h3 data-heading="Step 2 — Build the reverse import graph">Step 2 — Build the reverse import graph</h3>
<p>The script reads every <code>.py</code> file under <code>probpipe/</code> and builds:</p>
<pre><code>rdeps[X] = {all modules that import X}
</code></pre>
<p>For <code>node.py</code>, the direct importers (depth 1) are:</p>
<pre><code>probpipe.core.node  &#x3C;──  probpipe.__init__
                    &#x3C;──  probpipe.core.ops
                    &#x3C;──  probpipe.core.transition
                    &#x3C;──  probpipe.inference._nutpie
                    &#x3C;──  probpipe.inference._rwmh
                    &#x3C;──  probpipe.inference._sbijax
                    &#x3C;──  probpipe.modeling._likelihood
                    &#x3C;──  probpipe.validation._predictive_check
</code></pre>
<h3 data-heading="Step 3 — BFS expansion">Step 3 — BFS expansion</h3>
<p>Starting from <code>probpipe.core.node</code>, BFS discovers all modules that <strong>transitively</strong> depend on it:</p>
<pre><code>Seed:    probpipe.core.node
</code></pre>
<pre><code>Depth 1: probpipe.__init__               (via node)
         probpipe.core.ops               (via node)
         probpipe.core.transition        (via node)
         probpipe.inference._nutpie      (via node)
         probpipe.inference._rwmh        (via node)
         probpipe.inference._sbijax      (via node)
         probpipe.modeling._likelihood   (via node)
         probpipe.validation._predictive_check  (via node)


Depth 2: probpipe.inference.__init__     (via _nutpie)
         probpipe.modeling.__init__      (via _likelihood)
         probpipe.modeling._simple       (via _likelihood)
         probpipe.modeling._simple_generative  (via _likelihood)
         probpipe.validation.__init__    (via _predictive_check)

Depth 3: probpipe.inference._tfp_mcmc   (via _simple)
</code></pre>
<pre><code>Total affected: 15 modules (including node.py itself)
</code></pre>
<h3 data-heading="Step 4 (test job) — Map affected source files to test files">Step 4 (test job) — Map affected source files to test files</h3>
<p>For each affected module, the script checks if <code>tests/test_&#x3C;name>.py</code> exists:</p>

Affected file | Test found?
-- | --
probpipe/core/node.py | no test_node.py exists
probpipe/core/ops.py | tests/test_ops.py
probpipe/core/transition.py | tests/test_transition.py
probpipe/__init__.py | no test___init__.py
probpipe/inference/_nutpie.py | test_nutpie.py
... | ...


<p><strong>Tests that run:</strong> <code>tests/test_ops.py tests/test_transition.py tests/test_nutpie.py</code></p>
<p>Instead of the full ~60-test suite, only 3 targeted files run.</p>
<p>note: <code>test___init__.py</code> check is trivial.</p>
<h3 data-heading="Step 5 (notebooks job) — Check which notebooks import affected modules">Step 5 (notebooks job) — Check which notebooks import affected modules</h3>
<p>Each notebook's code cells are parsed. The notebooks all do <code>import probpipe</code> at minimum. The match condition is:</p>
<pre><code>af_mod == imp  OR  af_mod.startswith(imp + '.')
</code></pre>
<p>Because <code>probpipe.__init__</code> is in the affected set at depth 1, and every notebook imports <code>probpipe</code>, the check <code>probpipe.modeling._simple_generative.startswith('probpipe.')</code> is <code>True</code> — so <strong>all 11 notebooks run</strong>.</p>

erozercan:
### Issues

#### Issue 1: Step 4 (test job) — Map affected source files to test files

The current CI logic maps each changed source file to a test file by matching names (e.g., `node.py` → `test_node.py`). This works in the simple case but breaks in several real situations — something like `config.py` has no corresponding `test_config.py` at all, even though `test_prefect_config.py` exists and tests it.

**Initial idea:** enforce a naming convention — every test file must be named after the source file it tests (stripping the leading underscore), and add a step in `ci.yml` that throws a warning or error if a test file doesn't follow the convention. The problem with this approach is that one-to-one mapping isn't realistic: some test files cover multiple source files (e.g., `test_distributions.py`, `test_inference.py`), and enforcing strict naming on everyone is hard to maintain in practice.

**Better idea: restructure the `tests/` directory to mirror the `probpipe/` package layout.** This is already partially done — `tests/linalg/` exists and mirrors `probpipe/linalg/`. We'd just make it consistent everywhere:

tests/
core/           ← tests for probpipe/core/
distributions/  ← tests for probpipe/distributions/
inference/      ← tests for probpipe/inference/
modeling/       ← tests for probpipe/modeling/
converters/     ← tests for probpipe/converters/
linalg/         ← already done ✓
validation/     ← tests for probpipe/validation/
record/         ← tests for probpipe/record/
(root)          ← integration/cross-cutting tests



Instead of one-to-one file matching, CI maps a changed source file to its subfolder: `probpipe/core/node.py` → run all of `tests/core/`. This is robust, simple to reason about, and scales naturally. Cross-cutting tests (integration tests that span multiple modules) stay in the `tests/` root and only run on foundational changes.

The main downside is that it requires reorganizing 50+ existing test files — but this can be done quickly with Claude Code rather than by hand.

---

#### Issue 2: Step 5 (notebooks job) — Check which notebooks import affected modules

The notebook optimization (only run notebooks affected by the changed module) breaks down because of how `probpipe/__init__.py` participates in the BFS. When `node.py` changes, `probpipe/__init__.py` gets pulled into the affected module set (since it re-exports from `node.py`). Because all example notebooks have imports like `from probpipe import (...)`, they all match against `probpipe.__init__` being affected — so all 11 notebooks end up running, defeating the purpose of the optimization.

I added a fix that excludes `__init__.py` from the transitive BFS propagation in the notebook job (so `probpipe/__init__.py` is never added to the affected set unless it was directly modified). However, this still doesn't fully solve the problem — because even without `__init__.py` in the affected set, the BFS seed `probpipe.core.node` itself satisfies `af_mod.startswith('probpipe.')`, so any notebook doing `from probpipe import (...)` still matches.

The real fix is to change the example notebooks to use explicit submodule imports:

```python
# instead of this
from probpipe import WorkflowFunction, Normal

# do this
from probpipe.core.node import WorkflowFunction
from probpipe.distributions.continuous import Normal
```

With explicit imports, the matching logic works exactly as intended — a notebook that imports probpipe.core.node runs when node.py changes, and a notebook that only imports probpipe.distributions.continuous doesn't.

Note: both of these improvements rely on conventions — structured test directories and explicit submodule imports in notebooks — that are easy to drift from as the codebase grows. One way to enforce this without manual review overhead is to add a Claude Code PR reviewer to the repo: a reviewer that automatically runs on every PR and checks that new test files are placed in the right subfolder and that new notebook imports use the explicit submodule format. Independent of this CI logic, a Claude Code reviewer making automatic reviews on every commit would be broadly useful for the repo — catching things like missing tests, inconsistent API patterns, or documentation gaps before @jhuggins sees them in review.

jhuggins (approved):

PR #172, opened 2026-05-01
PR-B.1: Address remaining review findings on PR-B (#155)

jhuggins:
## Summary

Six small follow-up commits addressing arob5's post-merge review of PR-B (#155). Each commit lands the fix plus its targeted regression test where applicable. Independently revertable.

## Commits

1. **cleanup(empirical):** replace unreachable `RecordBootstrap` branch with explicit raise (finding 0.1) — `RecordBootstrapReplicateDistribution` now raises a clear `TypeError` for generic (object-array) `EmpiricalDistribution` sources instead of falling through to a confusing `_as_float_array` error.
2. **fix(converters):** `_convert_to_kde` handles multi-field `RecordEmpirical` via `flat_samples` (findings 0.2, 0.3, gap A, gap B) — multi-field `RecordEmpiricalDistribution` → `KDEDistribution` now routes through `flat_samples` (was broken). Adds tests for single-field unchanged path, multi-field new path, weight preservation, and explicit object-array rejection.
3. **refactor(node):** extract `_index_sample` helper for broadcast row-indexing (finding 0.4, gap C) — promotes inline Record/NumericRecord per-row indexing duplicated across `_broadcast_enumerate_then_sample` / `_broadcast_sample` to a shared module-level helper. Direct unit test plus integration coverage.
4. **test(transition):** unify field-access idiom in adjacent assertions (finding 0.5).
5. **fix(empirical):** include field name + shape in `_validate_record_samples` error messages (finding 0.6, gap E) — replaces the unhelpful "sample-axis length None" message with one naming the field and the actual shape. Audits sibling validators in `_wrap_numeric_array_as_record` for the same gap.
6. **docs(empirical):** cross-link `event_shape` / `obs_shape` between empirical and bootstrap classes (finding 0.7) — symmetric "See Also" sections so the single-field-only / multi-field-raises accessors point at each other.

## Test plan

- [x] Targeted suites (`test_bootstrap_replicate.py`, `test_distributions.py`, `test_converters.py`, `test_broadcast_distribution.py`, `test_transition.py`, `test_protocols.py`) all green
- [x] Full pytest: 2390 passed, 7 skipped, 0 failed
- [x] mkdocs strict (no docstring formatting changes that should affect builds; verify in CI)

### DIFF
--- a/probpipe/converters/_probpipe.py
+++ b/probpipe/converters/_probpipe.py
@@ -56,8 +56,9 @@ def _mm_provenance(source, mean_result=None, var_result=None):
 
 def _point_estimate(x):
     """Extract a plain array from a value that may be a BootstrapDistribution
-    or a single-field NumericRecord (the auto-wrap form returned by the
-    merged ``RecordEmpiricalDistribution._mean``)."""
+    or a single-field NumericRecord (the auto-wrap form returned by
+    ``RecordEmpiricalDistribution._mean`` — numeric arrays auto-wrap
+    as a single-field Record)."""
     from ..core.distribution import BootstrapDistribution
     from ..core._numeric_record import NumericRecord
     if isinstance(x, BootstrapDistribution):
@@ -453,9 +454,13 @@ def _convert_to_kde(source, key, **kw):
     view) so KDE sees one row per sample with all parameters
     concatenated. Other sources fall back to drawing fresh samples.
 
-    A generic ``EmpiricalDistribution`` (object-array storage) is
-    rejected with a clear ``TypeError`` rather than silently
-    producing a degenerate KDE.
+    Raises
+    ------
+    TypeError
+        If *source* is a generic ``EmpiricalDistribution``
+        (non-numeric / object-array storage). KDE requires numeric
+        samples; route the source through
+        ``RecordEmpiricalDistribution`` or supply a numeric source.
     """
     from ..distributions.kde import KDEDistribution
 
--- a/probpipe/converters/_probpipe.py
+++ b/probpipe/converters/_probpipe.py
@@ -23,8 +23,9 @@ from ..custom_types import PRNGKey
 from .._utils import _auto_key
 from ..core.constraints import _supports_compatible
 from ..core.distribution import (
-    RecordEmpiricalDistribution,
     Distribution,
+    EmpiricalDistribution,
+    RecordEmpiricalDistribution,
 )
 from ..core.provenance import Provenance
 from ._registry import ConversionInfo, ConversionMethod, Converter
@@ -445,9 +446,16 @@ def _convert_to_empirical(source, key, **kw):
 def _convert_to_kde(source, key, **kw):
     """Convert any distribution to a KDEDistribution.
 
-    If the source is an ``RecordEmpiricalDistribution`` (or subclass),
-    the stored samples and weights are reused directly.  Otherwise,
-    samples are drawn from the source.
+    For a ``RecordEmpiricalDistribution`` source the stored samples
+    and weights are reused directly: single-field empiricals pass the
+    field's stacked array straight to KDE; multi-field empiricals
+    pass ``source.flat_samples`` (the ``(n, total_dim)`` flat-matrix
+    view) so KDE sees one row per sample with all parameters
+    concatenated. Other sources fall back to drawing fresh samples.
+
+    A generic ``EmpiricalDistribution`` (object-array storage) is
+    rejected with a clear ``TypeError`` rather than silently
+    producing a degenerate KDE.
     """
     from ..distributions.kde import KDEDistribution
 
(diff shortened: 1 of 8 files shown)


arob5:
## PR Review: PR-B.1: Address remaining review findings on PR-B (#155)

### Summary

A clean follow-up PR addressing seven targeted reviewer findings (0.1–0.7) plus two doc gaps. Each commit is small, scoped, and independently revertable, with regression tests where applicable. Adheres well to ProbPipe conventions; only minor stylistic suggestions.

### Findings

#### Critical (must fix)

None.

#### Recommended (should fix)

1. **`_index_sample` docstring isn't NumPy-style.** `probpipe/core/node.py:295-326` — has prose + a bullet list but lacks explicit `Parameters` / `Returns` sections. The Style Guide §3.3 requires NumPy-style for function docstrings; even with `Any` types, the parameter `i` and the dispatch contract belong in named sections.

2. **`_convert_to_kde` no longer documents the object-array rejection in its main signature contract.** `probpipe/converters/_probpipe.py:445-462` — the docstring describes the new behavior in the body, but a `Raises` section listing `TypeError` for generic `EmpiricalDistribution` would make this discoverable in API docs (mkdocstrings will surface it).

#### Minor (nice to have)

3. **`_index_sample` does local imports of `Record` / `NumericRecord` per call.** `probpipe/core/node.py:321-323` — the previous `_broadcast_sample` hoisted these to function scope (called once per dispatch), and `_broadcast_enumerate_then_sample` also had `from .record import Record` outside the inner loop. The new helper recomputes them on every row. After the first call they're cached in `sys.modules`, so the cost is one dict lookup per row — negligible at 128 broadcast samples but visible for 100k. If module-load circularity is the only blocker, consider a one-shot `_record_cls = None` module-level cache that lazily populates on first call. Low priority.

4. **The "merged `EmpiricalDistribution`" phrasing in two docstrings is ambiguous.** `probpipe/core/node.py:301-302` and the comment at `probpipe/core/_empirical.py:991-998` refer to "the merged `EmpiricalDistribution`" without context. A future reader (post-PR-B's history fading) won't know what was merged. Replace with something like "the unified `EmpiricalDistribution` (post-PR-B)" or just describe the behavior directly without referencing the merge.

5. **The test `test_object_array_empirical_to_kde_rejected` doesn't import `EmpiricalDistribution` at module top.** `tests/test_converters.py:741-743` — uses `from probpipe import EmpiricalDistribution` inside the test function. Other tests in the same module import it at the top. Inconsistent but harmless.

6. **`RecordBootstrapReplicateDistribution.__init__` error message wording is slightly off.** `probpipe/core/_empirical.py:1004-1009` — the message says "got `{type(source).__name__}` with object samples" but `source.samples` could be any non-numeric type, not strictly `object`. Saying "with non-numeric samples" or referencing `source.samples.dtype` (as `_convert_to_kde` does) would be more accurate.

7. **No explicit `Raises` documentation for the new `TypeError` branch in `RecordBootstrapReplicateDistribution.__init__`.** `probpipe/core/_empirical.py:972-977` — class docstring describes `source` as "Record | RecordEmpiricalDistribution | EmpiricalDistribution | array-like" at `probpipe/core/_empirical.py:959`. With the new behavior, generic `EmpiricalDistribution` is no longer accepted. Either tighten the type hint description (drop `EmpiricalDistribution` and clarify "RecordEmpiricalDistribution only") or add a `Raises` section calling out the rejection.

### Observations on what's done well

- **Dead-code analysis is correct.** The removed `EmpiricalDistribution` branch in `RecordBootstrapReplicateDistribution.__init__` was indeed unreachable for numeric-array-backed empiricals (those route to `RecordEmpiricalDistribution` via `__new__`), and the new explicit raise improves the error path for the only remaining caller (direct construction with a generic-base instance).
- **Cross-links between `event_shape` and `obs_shape` symmetric accessors** are exactly the right scope and don't bloat the docstrings.
- **`_index_sample` extraction** correctly identifies the three drift-prone callsites and consolidates them. Test coverage is thorough (bare array, 1-D bare array, single-field unwrap, multi-field with scalar event, multi-field with non-trivial event shapes).
- **Validation message tests** pin both field name and shape, which is the right level of specificity for actionable error messages.
- **Multi-field KDE conversion via `flat_samples`** correctly reuses the existing `flat_samples` abstraction rather than re-flattening inline.
- **Tests for the object-array rejection paths** explicitly assert the routing premise (`assert not isinstance(emp, RecordEmpiricalDistribution)`) before exercising the raise — defensive against future `__new__` changes.

### Suggested Changes

Group A — Docstring polish (recommended):

1. Restructure `_index_sample` docstring at `probpipe/core/node.py:295-326` to use NumPy-style `Parameters` and `Returns` sections; keep the dispatch bullets as a `Notes` section.
2. Add a `Raises` section to `_convert_to_kde` at `probpipe/converters/_probpipe.py:446-460` documenting `TypeError` for generic `EmpiricalDistribution` sources.
3. Add a `Raises` section to `RecordBootstrapReplicateDistribution` class docstring (or `__init__` if you prefer) at `probpipe/core/_empirical.py:945-967` documenting the `TypeError` for generic `EmpiricalDistribution` sources, and tighten the `source` parameter description to drop the misleading bare `EmpiricalDistribution` mention.

Group B — Minor wording/style (nice to have):

4. Replace "the merged `EmpiricalDistribution`" with self-contained phrasing in `probpipe/core/node.py:301` and `probpipe/core/_empirical.py:991`.
5. Tighten the error message at `probpipe/core/_empirical.py:1006` to reference the actual non-numeric character of the samples rather than asserting "with object samples".
6. Move `from probpipe import EmpiricalDistribution` to the top of `tests/test_converters.py` to match adjacent style.

Group C — Performance (low priority, only if you care):

7. Consider hoisting the `Record`/`NumericRecord` imports in `_index_sample` (`probpipe/core/node.py:321-323`) to a module-level lazy cache pattern if circularity rules them out at top level. Negligible impact at typical broadcast sizes; skip unless a perf trace flags it.

PR #173, opened 2026-05-01
PR-C.1: SupportsArrayBackend protocol + from_batched_params factory

jhuggins:
## Summary

PR-C.1 (Phase 1 of the four-PR `Distribution.batch_shape` cleanup, plan: `.claude/plans/pr-c-distribution-batch-shape.md`). **Additive only — no breaking change.** Introduces the migration path that PR-C.2 will rely on when TFP-backed constructors begin rejecting batched parameters.

Key piece: distributions own their own array logic via the new `SupportsArrayBackend` capability protocol. `DistributionArray` becomes a thin consumer that detects the protocol and delegates to whatever backend the class returns. Same pattern as the rest of the protocol family (`SupportsSampling`, `SupportsLogProb`, …) — capability declared on the distribution side, dispatched by the consumer.

This is cleaner than putting backend-specific knowledge in `DistributionArray` subclasses: each distribution family owns its batching story; new families (user distributions, joints, future `RandomMeasure[T]`) integrate by implementing one classmethod; no `DistributionArray` subclass proliferation; no special WF-sweep hook needed.

## Stack

This PR is stacked on top of **#172 (PR-B.1)**. Base is `dev/pr-b-loose-ends`; once #172 merges, GitHub will auto-rebase this onto `main`.

## Commits

1. **feat(protocols):** `SupportsArrayBackend` + `_DistributionArrayBackend` protocols — capability + the produced-backend interface. No implementations yet.
2. **feat(tfp):** `_TFPArrayBackend` + `TFPDistribution._make_array_backend` — every concrete TFP-backed distribution (`Normal`, `Beta`, `Gamma`, `MultivariateNormal`, `Dirichlet`, …) inherits the protocol.
3. **refactor(distribution-array):** `_backend` slot + `_from_backend` private classmethod + lazy cell materialisation. Literal-array path unchanged.
4. **feat(distribution-array):** `from_batched_params` classmethod with protocol dispatch — backend path for `SupportsArrayBackend` classes; literal-array fallback otherwise. Per-cell name auto-suffix; `batch_shape` inferred from broadcast of array params; `MultivariateNormal`-style heterogeneous event ranks require explicit `batch_shape`.
5. **feat(distribution):** `Distribution.from_batched_params` class-method alias — `Normal.from_batched_params(loc=..., scale=..., name="x")` for free.
6. **docs(changelog):** documents the new surface; previews the upcoming PR-C.2 breaking change.

## Migration path (preview of PR-C.2)

```python
# Today (still works)
n = Normal(loc=jnp.zeros(5), scale=1.0, name="x")
n.batch_shape  # (5,)

# After PR-C.2 lands, the above raises with a hint pointing at:
da = DistributionArray.from_batched_params(
    Normal, loc=jnp.zeros(5), scale=1.0, name="x",
)
# Or the ergonomic per-class alias:
da = Normal.from_batched_params(loc=jnp.zeros(5), scale=1.0, name="x")

da.batch_shape  # (5,)
da[2].name      # "x_2"
da[2].loc       # 2.0
da._backend     # _TFPArrayBackend(Normal, batch_shape=(5,), name='x')
```

The factory is performance-equivalent to the legacy form because `_TFPArrayBackend` wraps the same TFP-batched distribution under the hood.

## Test plan

- [x] New tests pass: `tests/test_protocols.py::TestSupportsArrayBackendProtocolSurface` (4), `tests/test_tfp_array_backend.py` (24), `tests/test_distribution_array.py::TestBackendDelegatedStorage` (10) + `TestFromBatchedParams` (12) + `TestDistributionFromBatchedParamsAlias` (5)
- [x] Full pytest: 2443 passed, 7 skipped, 0 failed
- [ ] mkdocs strict (verify in CI)

## What's deferred

- **Vectorised dispatch via the WF sweep.** Backed `DistributionArray`s expose the backend's vectorised `_sample` / `_log_prob` / `_mean` / `_variance` / `_cov`, but the WF sweep still routes through `_flat_component(i)` for cell-by-cell dispatch. The lazy-materialisation benefit on construction is real today; the vectorised-ops benefit is a clean follow-up commit (small, can land in PR-C.1.1 or PR-C.2). This is documented as a known optimisation, not a regression.
- **`_RecordArrayBackend`** for `RecordEmpiricalDistribution` / `RecordBootstrapReplicateDistribution` — the perf-preservation argument is strongest for TFP, so the Record backend is deferred to a follow-up issue. The `from_batched_params(RecordEmpiricalDistribution, ...)` path goes through the literal-array fallback today.
- **Per-family `Normal.batched(...)` shortcuts** — the `Distribution.from_batched_params` alias already gives `Normal.from_batched_params(...)` for free. Defer if users ask.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,47 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **`SupportsArrayBackend` capability protocol** (`probpipe.SupportsArrayBackend`)
+  declares that a `Distribution` subclass can produce a fused storage
+  backend for `DistributionArray`. Implemented by every TFP-backed
+  concrete class (`Normal`, `Beta`, `Gamma`, `MultivariateNormal`,
+  `Dirichlet`, …) via inheritance from `TFPDistribution`. Distribution
+  classes that don't implement the protocol still work in a
+  `DistributionArray` via the literal-array fallback path.
+
+- **`DistributionArray.from_batched_params(dist_cls, *, name, batch_shape=None, **batched_params)`**
+  factory + ergonomic per-class alias **`Distribution.from_batched_params(*, name, batch_shape=None, **batched_params)`**.
+  Constructs a `DistributionArray` of homogeneous components,
+  dispatching on `SupportsArrayBackend`: TFP-backed classes get a
+  fused TFP-batched backend; other classes fall back to one
+  `dist_cls` instance per cell. Per-cell names auto-suffix
+  `f"{name}_{flat_index}"`. `batch_shape` is inferred from broadcast
+  of array-valued params; classes with heterogeneous per-param event
+  ranks (`MultivariateNormal`, `Dirichlet`) require explicit
+  `batch_shape=...`.
+
+  ```python
+  # Recommended ergonomic form
+  da = Normal.from_batched_params(loc=jnp.zeros(5), scale=1.0, name="x")
+  da.batch_shape       # (5,)
+  da[2].name           # "x_2"
+
+  # Equivalent universal form
+  da = DistributionArray.from_batched_params(
+      Normal, loc=jnp.zeros(5), scale=1.0, name="x",
+  )
+  ```
+
+  **Preview of upcoming PR-C.2 breaking change:** in the next release,
+  `Normal(loc=jnp.zeros(5), scale=1.0, name="x")` (and similar
+  TFP-backed classes constructed with batched parameters) will raise
+  `ValueError` with a hint pointing at this factory. Migrate now to
+  avoid the breaking change. The factory is performance-equivalent
+  to the legacy form because it shares the TFP-batched backend under
+  the hood.
+
 ### Changed (breaking)
 
 - **Empirical / Bootstrap / Marginal class consolidation.** The
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -137,6 +137,17 @@ full public API surface.
    dispatch time.  Protocols are dynamically included on composite
    distributions (`ProductDistribution`, `TransformedDistribution`)
    based on component capabilities.
+
+   Most protocols are *instance-level*: the contract is "this
+   `Distribution` instance can do X" and the runtime check is
+   `isinstance(my_dist, SupportsX)`. A small number are *class-level*:
+   `SupportsArrayBackend` declares `_make_array_backend` as a
+   `@classmethod`, so the contract is "this class can produce a
+   batched form" and the runtime check is on the class itself
+   (`isinstance(MyDistribution, SupportsArrayBackend)`). New
+   protocols default to instance-level; reach for class-level only
+   when the capability is genuinely a class concern (e.g., a fused-
+   storage factory that doesn't depend on per-instance state).
 4. **Private method convention** — protocols define `_method()` (e.g.,
    `_sample`, `_log_prob`, `_mean`). The public API is via ops:
    `sample(dist)`, not `dist.sample()`.
--- a/STYLE_GUIDE.md
+++ b/STYLE_GUIDE.md
@@ -30,11 +30,18 @@ _make_array_backend
 
 Most `Supports*` protocols are instance-level capabilities (does
 `isinstance(my_dist, SupportsSampling)` hold?). `SupportsArrayBackend`
-is the exception: its method is a `@classmethod`, so the runtime check
-is on the **class** (`isinstance(MyDistribution, SupportsArrayBackend)`),
-not on instances. The corresponding *backend* interface that
-`_make_array_backend` returns (`_DistributionArrayBackend`) is private
-to the library — user code never imports or constructs it.
+is the exception: its method is a `@classmethod`, so the contract
+lives at class scope. Always check
+`isinstance(MyDistribution, SupportsArrayBackend)` (i.e. on the
+class object). `isinstance(an_instance, SupportsArrayBackend)`
+returns `True` too — instances inherit class attributes, and
+`runtime_checkable` just looks for the named attribute — but the
+result is misleading because the protocol's contract is the class
+declaring `_make_array_backend`, not the instance.
+
+The corresponding *backend* interface that `_make_array_backend`
+returns (`_DistributionArrayBackend`) is private to the library —
+user code never imports or constructs it.
 
 ### 1.2 Ops (public API)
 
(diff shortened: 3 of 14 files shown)


arob5:
## PR Review: PR-C.1: SupportsArrayBackend protocol + from_batched_params factory

### Summary

Solid additive PR introducing the `SupportsArrayBackend` capability protocol and `DistributionArray.from_batched_params` factory as the migration path for PR-C.2. Architecture is clean — capability declared on the distribution side, dispatched by the consumer, mirroring the rest of the protocol family. Tests are thorough (80 new tests pass). Findings are minor and mostly cosmetic.

### Findings

#### Critical (must fix)
None.

#### Recommended (should fix)

1. **Unused imports in `probpipe/distributions/_tfp_base.py`**:
   - `prod` from `.._utils` (line 18) is imported but never used in the module.
   - `_DistributionArrayBackend` from `..core.protocols` (line 30) is imported but only referenced in a docstring (line 201). `_TFPArrayBackend` doesn't inherit from it or use it at runtime, so this import is dead. If you want to keep the structural-conformance assertion, consider `assert isinstance(_TFPArrayBackend, type) and issubclass(_TFPArrayBackend, _DistributionArrayBackend)` — but that won't work because Protocol with non-method members behaves unevenly with `issubclass`. Cleanest: drop the import.

2. **`_from_backend` hardcodes `_approximate = False`** in `probpipe/core/_distribution_array.py:360`. This works for `_TFPArrayBackend` (TFP distributions are exact) but bakes in a current-implementation assumption. When `_RecordArrayBackend` lands later (the deferred follow-up the PR description mentions), `is_approximate` will need to be derived from the backend's underlying samples. Options: (a) ask the backend (`getattr(backend, "is_approximate", False)`), (b) add a `TODO` flagging this for the Record-backend follow-up, or (c) leave a docstring note that any approximate backend is responsible for setting `instance._approximate = True` post-construction.

3. **`Distribution.from_batched_params` return type is `Any`** (`probpipe/core/_distribution_base.py:178`). Could be narrowed to a forward-ref `"DistributionArray"` for better IDE / typechecker support. The reason for the local import is the cycle, but the annotation can still be a string.

#### Minor (nice to have)

4. **Edge case: scalar-only params + explicit `batch_shape`**. `DistributionArray.from_batched_params(Normal, loc=0.0, scale=1.0, batch_shape=(5,), name="x")` will reach `_TFPArrayBackend.__init__`, which constructs `Normal(loc=0.0, scale=1.0)` with TFP `batch_shape=()`, then the sanity check raises `ValueError`. Either document that "explicit batch_shape requires at least one batched array param" or have the backend broadcast the scalars to the declared batch shape. Test `test_no_array_params_requires_explicit_batch_shape` only covers the inference path; the explicit-shape-with-scalars path would similarly fail but with a cryptic shape-mismatch error.

5. **`_flat_component(-1)` behavior diverges from `da[-1]`**. `__getitem__` wraps negatives via `% dim` (`_distribution_array.py:453`), but `_TFPArrayBackend._normalize_index` rejects negatives (`_tfp_base.py:302`). Since `_flat_component` is documented as "row-major flat index over `batch_shape`", its semantics for negative indices are unspecified. The literal-array path returns `self._components[-1]` (Python tuple semantics, which works); the backend path raises `IndexError`. Tighten: either reject negatives in both paths or accept them in both.

6. **`DistributionArray.__init__` field annotation**. `self._components: tuple[Distribution, ...] | None = components` (`_distribution_array.py:166`) widens the visible type — `components` is always a non-`None` tuple in this code path; only `_from_backend` sets it to `None`. Consider declaring the field at class scope (so its widened type is documented once) and assigning `self._components = components` here without the `| None` annotation noise.

7. **Tests do all imports inside test methods**. `tests/test_tfp_array_backend.py` does this at module top, but `tests/test_distribution_array.py::TestBackendDelegatedStorage` and `TestFromBatchedParams` import inside each test method (e.g. `from probpipe import Normal, DistributionArray` repeated 12+ times). STYLE_GUIDE §8.3 prefers module-scope fixtures / imports. Minor; refactor only if you're touching these tests for another reason.

8. **`test_cell_returns_fresh_scalar_normal` pins "no caching"** via `assert backend.cell(0) is not backend.cell(0)` (`test_tfp_array_backend.py:1605`). This locks in an implementation detail; a future caching optimisation would force you to update the test. Consider testing the observable consequence (e.g., that mutating the returned cell doesn't affect the next call) instead of identity.

9. **STYLE_GUIDE addendum is good but slightly off**. The new note says "calling `isinstance(an_instance, SupportsArrayBackend)` is meaningless because the protocol method is a classmethod." Strictly, `isinstance(an_instance, SupportsArrayBackend)` would actually return `True` because `runtime_checkable` uses `hasattr`, and instances inherit class attributes. Rewording: "...is misleading — the runtime check returns True because instances inherit the classmethod, but the contract is at class scope. Always check `isinstance(MyDistribution, SupportsArrayBackend)`."

10. **CONTRIBUTING.md not touched**. The PR adds a new "Generic vs class-level" protocol category (every other `Supports*` is instance-level). The "Protocol Design" section in CONTRIBUTING.md / `STYLE_GUIDE.md §7` could mention that classmethod-level protocols exist and are distinguished by where the runtime check lives. The `STYLE_GUIDE.md §1.1` addendum partially covers this — extending §7 to call out the pattern explicitly would help future readers.

### Suggested Changes

1. **Drop unused imports** in `probpipe/distributions/_tfp_base.py`:
   - Remove `from .._utils import prod` (line 18).
   - Remove `_DistributionArrayBackend` from the `..core.protocols` import (line 30).

2. **Make `_approximate` propagate from the backend** in `probpipe/core/_distribution_array.py:360`:
   ```python
   instance._approximate = getattr(backend, "is_approximate", False)
   ```
   Or, if you'd rather defer to PR-C.2 / the Record-backend follow-up, leave a `# TODO(pr-c.X)` comment explaining that the hardcoded `False` will need revisiting.

3. **Narrow the `from_batched_params` return type** in `probpipe/core/_distribution_base.py:178`:
   ```python
   ) -> "DistributionArray":
   ```

4. **Document or fix the scalar-only + explicit-batch-shape edge case** in `_TFPArrayBackend.__init__` (`probpipe/distributions/_tfp_base.py:240`). Either broadcast the scalars internally or extend the existing `ValueError` message to suggest "pass at least one parameter with leading shape matching `batch_shape`".

5. **Align negative-index handling** between `__getitem__` and `_TFPArrayBackend._normalize_index`. Either accept negatives in both (modulo) or reject in both. If you keep the current split, document it on `_flat_component`.

6. **Clarify the STYLE_GUIDE note** (`STYLE_GUIDE.md:78-82`) about `isinstance` on instances — it does succeed via attribute inheritance, but it's misleading rather than meaningless.

7. **Replace identity-based no-caching test** in `tests/test_tfp_array_backend.py:1605` with a behavioural assertion (or delete it — the test already covers the value-correctness of `cell(0)`).

PR #188, opened 2026-05-17
feat(inference): MinibatchedDistribution + BlackJAX SGMCMC

jhuggins:
## Summary

Adds the `MinibatchedDistribution` random-measure machinery **and** the two BlackJAX-backed SGMCMC kernels that consume it, in a single PR so the new abstraction ships with a real consumer rather than waiting for a follow-up. Net: ~700 LOC source / ~1000 LOC tests / +59 tests / full suite 2557 → 2616.

Stacked on #187; retarget to `main` once that merges.

```python
from probpipe import condition_on

# Old (full-batch NUTS): the default.
posterior = condition_on(model, data)

# New (SGMCMC; opt-in via method=):
posterior = condition_on(
    model, data,
    method="blackjax_sgld",
    batch_size=64, num_steps=2000, num_warmup=500, step_size=1e-3,
)
```

Or, for direct random-measure consumers (custom kernels, diagnostics, …):

```python
from probpipe import MinibatchedDistribution, Record, random_unnormalized_log_prob

m = MinibatchedDistribution(model, Record(X=X, y=y), batch_size=64)
rf = random_unnormalized_log_prob(m)              # RandomFunction[Record, Array]
target = rf._sample(k)                            # θ → log~D_B(θ) for one minibatch
grad = jax.grad(target)(theta)                    # unbiased gradient estimate
```

### What's in (the abstraction)

- **`ConditionallyIndependentLikelihood[P, D]`** Protocol extending `Likelihood[P, D]` with `per_datum_log_likelihood(params, datum)`. Required by `MinibatchedDistribution` so non-i.i.d. likelihoods can't silently produce wrong gradient estimates.
- **`_default_per_datum_log_likelihood`** helper — length-1-batch fallback for likelihoods that want a default rather than an efficient override.
- **`MinibatchedDistribution`** (public) + three private helpers: `_FixedMinibatchDistribution` (one realisation), `_RandomMinibatchLogProb` (the random-function form), `_MinibatchLogProbAtPoint` (the `Distribution[Array]` at fixed θ).
- Existing concrete likelihoods extended with `per_datum_log_likelihood`: `GLMLikelihood` (efficient direct override), `_NLELikelihood` / `_NRELikelihood` (default fallback).
- `SimpleModel.prior` / `SimpleModel.likelihood` and `SimpleGenerativeModel.prior` / `SimpleGenerativeModel.likelihood` as public read-only properties — symmetric across both model wrappers.

### What's in (the consumer)

- **`BlackJAXSGLDMethod`** (priority 30, `method="blackjax_sgld"`).
- **`BlackJAXSGHMCMethod`** (priority 25, `method="blackjax_sghmc"`).
- Both consume any `SimpleModel` with a `ConditionallyIndependentLikelihood` plus a `batch_size=` kwarg. Both wrap the model + data in a `MinibatchedDistribution` internally and feed BlackJAX via the per-step random-measure draw — the kernel stays oblivious to the minibatching convention, so future `RandomMeasure[Record]` subclasses (tempered targets, synthetic likelihoods, …) plug in unchanged.
- Priorities sit below the full-batch gradient methods, so SGMCMC is opt-in via `method=...`; it doesn't auto-win on a routine `condition_on(model, observed)` call.

### Deliberately out

- **No public `split_log_density(model, observed)` utility.** The `SimpleModel` decomposition is just `model.prior._log_prob` and `model.likelihood.per_datum_log_likelihood` — two attribute reads with no shared machinery to extract. `MinibatchedDistribution` reads them directly. When the BlackJAX migration of full-batch MCMC and VI lands later, that PR will decide whether the existing dispatch logic in `probpipe/inference/_tfp_mcmc.py:_build_target_log_prob` is worth factoring into a *private* shared helper.

### Class summary

```
RandomMeasure[Record]
└── MinibatchedDistribution               (public; SupportsSampling +
                                           SupportsRandomUnnormalizedLogProb)

Distribution[Record]
└── _FixedMinibatchDistribution           (private; one minibatch's surrogate)

RandomFunction[Record, Array]
└── _RandomMinibatchLogProb               (private; SupportsSampling — returns
                                           a callable θ → log~D_B(θ))

InferenceMethod
├── BlackJAXSGLDMethod                    (public; method="blackjax_sgld")
└── BlackJAXSGHMCMethod                   (public; method="blackjax_sghmc")
```

### Tests

`tests/test_minibatch.py` — 29:
- **Construction (7):** SimpleModel path, explicit-callable path, rejects bare `Likelihood` / non-SimpleModel / out-of-range `batch_size` / nested Records.
- **Accessors (2):** all convenience properties + default-name format.
- **Protocols (4):** `RandomMeasure` / `SupportsSampling` / `SupportsRandomUnnormalizedLogProb` / non-iterable.
- **Sampling (8):** inner type, op-layer wrap, batched `DistributionArray` with component-type check, factorisation identity, `with_replacement` repeats, `rescale=False` raw-sum, `batch_size == N` degenerate full-batch, `batch_size == 1` vmap-over-length-1-axis edge.
- **Mathematical correctness (2):** MC mean ≈ full log-density at `atol=0.5` over 2000 vmapped minibatches; MC gradient ≈ full gradient at `atol=0.75` per coordinate (tightened from initial draft's `1.0` / `2.0`).
- **Random-function op (4):** zero-arg / one-arg forms, callable return, batched-sample-not-supported.
- **JIT traceability (1):** `jax.grad` of the log-density callable compiles under `jax.jit`.

`tests/test_blackjax_sgmcmc.py` — 17:
- **Registry (3):** SGLD/SGHMC registered, priorities below full-batch.
- **Check (4):** rejects bare `SupportsLogProb`, non-factorisable likelihood, missing `batch_size`; feasible for well-formed input.
- **Gradient-estimator correctness (1):** direct deterministic check that `_build_grad_estimator(measure)(theta, key)` matches `jax.grad(prior + N/b·Σ per_datum)(theta)` on the same minibatch (no MC noise; catches sign/scale bugs at `rtol=1e-5`).
- **Convergence (2):** both kernels recover the true coefficients of a 200-row Bayesian logistic regression to `atol=0.3` over 5000 retained samples, with chain finiteness and a per-coordinate std lower bound to catch a stuck chain.
- **Reproducibility (2):** same-seed → identical chain; different seeds → different chains.
- **Dispatch (5):** `condition_on(method="blackjax_sgld", …)` works end-to-end, chain shape, warmup discarded, user-supplied `init=`, `with_replacement=` kwarg threads through.

`tests/test_conditionally_independent_likelihood.py` — 8:
- Protocol structure, `GLMLikelihood` (Bernoulli + Normal families) per-datum sums to `log_likelihood`, default-fallback correctness for both Array and Record data, bare `Likelihood` rejected, NLE / NRE membership.

Plus one accessor test added to each of `tests/test_simple_model.py` and `tests/test_simple_generative_model.py`, and `tests/test_iteration_protocol.py` extended to enumerate `MinibatchedDistribution`.

## Test plan

- [x] `pytest tests/` — 2616 passed, 7 skipped (full suite, ~6m).
- [x] `pytest tests/test_blackjax_sgmcmc.py tests/test_minibatch.py tests/test_conditionally_independent_likelihood.py` — 56 passed.
- [x] `mkdocs build --strict` — clean.
- [x] Manual smoke: SGLD recovers logistic-regression coefficients within MC tolerance; MC log-density / gradient over 2000 minibatches match full-data values within SE.


probpipe/inference/_blackjax_sgmcmc.py
```diff
@@ -0,0 +1,264 @@
...
+Methods are registered at low priority (30 / 25) so they only fire when
+explicitly requested via ``condition_on(model, observed,
+method="blackjax_sgld", batch_size=…, …)``.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+import blackjax
```
SoymilkWinsAgain, line 29: `inference/__init__.py` registers the `BlackJAX` methods unconditionally, which imports `inference/_blackjax_sgmcmc.py` and its import `blackjax`, but `pyproject.toml` does not declare `blackjax` as a dependency. In a normal dev environment, import `probpipe` fails immediately with `ModuleNotFoundError`: No module named 'blackjax'. This breaks the whole package and test collection, not just the new SGMCMC feature.

probpipe/inference/_minibatch.py
```diff
@@ -0,0 +1,538 @@
...
+
+        # Resolve the (log_prior, per_datum_log_likelihood) callables.
+        if per_datum_log_likelihood is not None:
+            if not isinstance(model, SupportsLogProb):
+                raise TypeError(
+                    "When per_datum_log_likelihood is supplied, model must "
+                    "be SupportsLogProb (it provides the log-prior)."
+                )
+            self._log_prior_fn = model._log_prob
+            self._per_datum_log_lkl_fn = per_datum_log_likelihood
```
SoymilkWinsAgain: The explicit `per_datum_log_likelihood=` branch in is broken for `SimpleModel`. The test in `tests/test_minibatch.py` constructs exactly this path with `SimpleModel + explicit callable`, but only asserts that construction succeeds.

At actual log-density evaluation time, this branch sets `self._log_prior_fn = model._log_prob`. For `SimpleModel`, `_log_prob` expects a `(params, data)` tuple, not a single `theta`. Later, `_FixedMinibatchDistribution._unnormalized_log_prob(theta)` passes `theta` directly into that callable, so this path will fail during unpacking/call.

SoymilkWinsAgain (changes requested):

probpipe/inference/_blackjax_sgmcmc.py
jhuggins: Fixed in 7e5952f: added `blackjax>=1.3` to `[project] dependencies` in `pyproject.toml`, and dropped the misleading inline comment in `inference/__init__.py` that claimed blackjax was already a core dep. `pip install -e .` now produces a wheel that `import probpipe` resolves cleanly.

probpipe/inference/_minibatch.py
jhuggins: Fixed twice over:

1. Initial fix (1e27f14): the branch now uses `model.prior._log_prob` when the model is a `SimpleModel`, so the log-prior callable has the right unary signature at evaluation time. The existing construction-only test was replaced by `test_explicit_callable_path_evaluates`, which evaluates the unnormalized log-density and cross-checks it against the auto-detected path over the same minibatch draw.

2. Subsequent refactor (80b8dfe): the entire explicit-callable branch is gone. `MinibatchedDistribution` now takes `(prior, likelihood, data, batch_size)` directly — no `model` parameter, no `per_datum_log_likelihood=` escape hatch, no `rescale=` switch either. The bug surface is removed by simplification rather than guarded against.

probpipe/inference/_blackjax_sgmcmc.py
```diff
@@ -0,0 +1,267 @@
...
+                ),
+            )
+        return MethodInfo(feasible=True, method_name=self.name)
+
+    # -- execution -----------------------------------------------------------
+
+    def execute(self, dist: Any, observed: Any, **kwargs: Any) -> ApproximateDistribution:
+        """Run the SGMCMC kernel; return an :class:`ApproximateDistribution`."""
+        batch_size: int = kwargs["batch_size"]
+        num_steps: int = kwargs.get("num_steps", 1000)
```
SoymilkWinsAgain, line 142: I believe this only read `num_steps`, not `num_results` in inference API. Therefore, if the user use `condition_on` to input `num_results=200`, it will still silently treat the parameter as 1000. It is recommended to at least support the `num_results` alias, or explicitly prohibit the mixing of the two.

CHANGELOG.md
```diff
@@ -7,8 +7,138 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+      batch_size=64, num_steps=2000, num_warmup=500, step_size=1e-3,
+  )
+  ```
+
+  Priorities sit below the full-batch gradient methods
+  (`tfp_nuts=100`, `tfp_hmc=90`, …, `tfp_rwmh=50`), so SGMCMC is
+  opt-in via `method=...` — it does not auto-win on a routine
+  `condition_on(model, observed)` call.
+
+- **`MinibatchedDistribution`** (`probpipe.MinibatchedDistribution`)
```
SoymilkWinsAgain, line 83: The `MinibatchedDistribution` documentation is inconsistent with the code: the documentation states it supports `SupportsSampling`, and the examples use `MinibatchedDistribution(model, Record(...), batch_size=64)` and `sample(m)`, but the current code signature is `MinibatchedDistribution(prior, likelihood, data, batch_size)`, and tests explicitly state it does not support sampling. This may mislead users in the examples in the release notes and requires an update.

SoymilkWinsAgain (changes requested):

probpipe/inference/_blackjax_sgmcmc.py
jhuggins: This is correct behavior. num_results is also affects by burn-in and number of chains used, so the two aren't equivalent

CHANGELOG.md
jhuggins: Fixed in acf1d81. The CHANGELOG entry now reflects the post-refactor API: constructor takes `(prior, likelihood, data, batch_size)` directly (no `model` parameter, no `per_datum_log_likelihood=` escape hatch), and explicitly notes that `MinibatchedDistribution` does *not* satisfy `SupportsSampling` — the public consumption path is `random_unnormalized_log_prob(m)._sample(key)` (the `SupportsRandomUnnormalizedLogProb` surface). Example code rewritten to match. Also dropped the stale `(Phase 3)` / `(Phase 5)` phase references in the same pass.

Swept the rest of the repo for other stale references: source docstrings + tests are current as of 5196d90; STYLE_GUIDE, CONTRIBUTING, and the tutorial notebooks don't mention `MinibatchedDistribution`. `docs/api/inference.md` auto-renders from the (now-correct) class docstring.

PR #192, opened 2026-05-19
Bug fixed: the original code in ``node.py`` did not correctly handle `self._parallel`

SoymilkWinsAgain:
Fixes a bug in `WorkflowFunction._execute_many_threaded` where `parallel=True` was incorrectly interpreted as an explicit worker count. In Python, `bool` is a subclass of `int`, so the previous check:

```python
max_workers = self._parallel if isinstance(self._parallel, int) else None
```
treated `True` as `1`. As a result, `parallel=True` effectively created a `ThreadPoolExecutor` with a single worker, even though the documented behavior is to use the executor's default worker count.

The fix explicitly handles `parallel=True` before checking for integer values.

### DIFF
--- a/probpipe/core/node.py
+++ b/probpipe/core/node.py
@@ -1478,25 +1478,24 @@ class WorkflowFunction(Node):
         if kind is WorkflowKind.FLOW:
             return self._execute_many_prefect_flow(call_value_list)
 
-        if self._parallel is not False:
+        if self._parallel is not False and self._parallel is not None:
             return self._execute_many_threaded(call_value_list)
 
         return [self._func(**v) for v in call_value_list]
 
     def _execute_many_threaded(self, call_value_list: list[dict[str, Any]]) -> list:
-        if self._parallel is True:
-            max_workers = None  # ThreadPoolExecutor default
+        max_workers = None  # ThreadPoolExecutor default for parallel=True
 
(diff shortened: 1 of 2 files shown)


probpipe/core/node.py
```diff
@@ -1468,22 +1468,51 @@ def _execute_many(self, call_value_list: list[dict[str, Any]]) -> list:
...
-        if self._parallel:
+
+        if self._parallel is not False:
             return self._execute_many_threaded(call_value_list)
+
         return [self._func(**v) for v in call_value_list]
 
     def _execute_many_threaded(self, call_value_list: list[dict[str, Any]]) -> list:
-        max_workers = self._parallel if isinstance(self._parallel, int) else None
+        if self._parallel is True:
```
jhuggins: no need for an if statement here; this only gets run if parallel = true, so just set max_works = None, then have the `if isinstance(self._parallel, int) and not isinstance(self._parallel, bool)` check.

jhuggins (changes requested):

probpipe/core/node.py
SoymilkWinsAgain: The if statement has been modified. Now it follows this logic.

jhuggins (approved):

Issue #196, opened 2026-05-20
WorkflowFunction Class Refactor


SoymilkWinsAgain:
## Summary

`WorkflowFunction` is the central execution facade for `ProbPipe` operations and user-defined workflow functions. It should remain the public facade, but it should no longer be the implementation hub for call resolution, distribution normalization, broadcast planning, sampling, sweep execution, JAX vectorization, execution dispatch, output coercion, and provenance assembly. In a word, this class is turning into a big ball of mud.

This plan proposes a staged internal refactor that keeps the public API stable while moving distinct responsibilities into private core modules. The immediate goal is maintainability and testability. The longer-term goal is to expose a **backend-neutral** internal execution interface: execution adapters should receive plain functions, plain call dictionaries, and resolved execution configuration rather than a `WorkflowFunction` instance.

## Status Quo of `WorkflowFunction`

`WorkflowFunction` currently does much more than wrap and call a user function. It is the public facade for user-defined workflow functions, but it also owns call resolution, type normalization, broadcast planning, sampling, sweep execution, local vectorization, backend dispatch, output coercion, and provenance assembly. The result is a class where unrelated changes require a maintainer to reason across several implementation domains at once.

### Current Attributes

| Category | Attribute | Source | Current responsibility |
|---|---|---|---|
| Distribution broadcast | `DEFAULT_N_BROADCAST_SAMPLES` | Class attribute | Default distribution broadcast sample count, currently `128`. |
| Public facade and call resolution | `_func` | `__init__` argument `func` | Wrapped user function. |
| Public facade and call resolution | `_sig` | `inspect.signature(func)` | Function signature used for positional binding, defaults, and validation. |
| Public facade and call resolution | `_hints` | `get_type_hints(func)` | Type hints used to identify dependencies, distributions, protocols, and array behavior. |
| Execution dispatch | `_workflow_kind_raw` | `__init__` argument `workflow_kind` | Per-instance Prefect orchestration setting before global fallback resolution. |
| Public facade and call resolution | `_name` | `__init__` argument `name` or `func.__name__` | Display and operation name used by results, provenance, and orchestration labels. |
| Public facade and call resolution | `__doc__` | Copied from `func.__doc__` | Introspection metadata for docs and help output. |
| Public facade and call resolution | `__name__` | Set from `_name` | Callable name exposed to introspection. |
[...]

PR #201, opened 2026-05-21
refactor(core) [OUTDATE]: new characterization tests for `WorkflowFunction` call

SoymilkWinsAgain:
This is the PR 0 stage of #196 , following the plan in #194.

Adding 2 new characterization tests in `tests`. No functional changes.

- `tests/test_workflow_call_resolution.py`: Covers positional binding, `**kwargs`, bind/default/missing input, module dependency/input resolution, dependency type validation, reserved call-time overrides, and seed override.
- `tests/test_workflow_distribution_normalization.py`: Covers TFP external distribution transformation, protocol hint transformation, 0-d `DistributionArray` unwrap, size-1 sweep, and unhinted external distribution broadcast.

Result: all passed.

### DIFF
--- a/tests/test_workflow_call_resolution.py
+++ b/tests/test_workflow_call_resolution.py
@@ -4,6 +4,8 @@ These tests lock down the public-call boundary before WorkflowFunction
 internals are split into smaller private modules.
 """
 
+from __future__ import annotations
+
 import jax.numpy as jnp
 import pytest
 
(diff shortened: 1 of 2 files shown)


jhuggins:
Nice tight characterisation-test PR. Findings below — all non-blocking (the PR adds no source code, so nothing here prevents landing), but a few are worth tightening before these tests become the contract for the upcoming refactor.

### Recommended

1. **Test-class grouping (STYLE_GUIDE §8.2).** Both files use flat module-level test functions; the cases cluster cleanly into `Test*` classes:
   - `test_workflow_call_resolution.py` → `TestArgumentBinding`, `TestModuleResolution`, `TestReservedOverrides`.
   - `test_workflow_distribution_normalization.py` → `TestHintedDistributionConversion`, `TestDistributionArrayHandling`, `TestUnhintedExternalDistribution`.

2. **Direct underscore-method calls in `test_workflow_distribution_normalization.py`** (lines 22, 42, 58, 77 — `dist._mean()`, `dist._log_prob(...)`). STYLE_GUIDE §1.2 says users call ops. Either swap to the public op (`mean(dist)`, `log_prob(dist, value)`) or add a one-line comment that the bypass is deliberate (the tests are probing the *input* distribution passed into the wrapped function, so unwrapping is the point).

3. **`test_size_one_distribution_array_remains_a_sweep` couples to the function's `__name__`** (line 92): `result["mean_of_normal"]` will `KeyError` confusingly if the helper gets renamed. Suggest `result[result.fields[0]]` or pinning the field name via a local constant.

### Minor

4. **Missing `from __future__ import annotations`** at the top of both files (STYLE_GUIDE §4.1 says "Every module starts with…").

5. **Duplicate inline setup** that fixtures (§8.3) would DRY: `def add(x, y)` × 3 and `def identity(x)` × 2 in the call-resolution file, `def mean_of_normal(...)` × 3 in the distribution-normalization file, plus `Normal(0, 1, name="x")` × 2.

6. **`test_protocol_hint_converts_distribution_that_lacks_protocol`** asserts only `isinstance(seen[0], SupportsLogProb)` (line 49). Pinning the concrete converted type (`KDEDistribution` or whatever the registry resolves to) would catch silent rerouting by a future converter change.

7. **`test_unhinted_external_distribution_converts_then_broadcasts` `atol=0.25`** (line 114) on an 8-sample mean is generous. With `seed=0` it's deterministic, so a tighter tolerance — or a per-sample structural check via `seen_shapes` + a recorded-samples comparison — would catch numerical drift earlier.

8. **`test_module_inputs_dependencies_and_defaults_resolve_for_workflow_methods`** — the `assert dep is self.child_nodes["dep"]` invariant (line 94) fires inside the workflow execution; if the wiring is broken the assertion runs deep in the call stack. Pulling it out to a top-level pre-invocation assert (`assert module.step.child_nodes["dep"] is dep`) makes the failure mode crisper.

Happy to expand on any of these if useful.

jhuggins (changes requested):
See my previous comment for suggested changes.

PR #184, opened 2026-05-13
Implementing multiple dispatch

tcikezu:
Plan document: see https://github.com/TARPS-group/prob-pipe/blob/7753788b0be61ea93277bf00958f52d102be7c5c/multipledispatch_plan.md

PR #204, opened 2026-05-22
feat(core): extract BaseDispatchRegistry; add BinaryDispatchRegistry (Stage 1)

tcikezu:
## Summary

Stage 1 of the multiple-dispatch plan from PR #184 / @arob5's consolidated design.

- Extracts the arity-independent logic from the previous `MethodRegistry` into a new `BaseDispatchRegistry` abstract base class (`_registry.py`). Hoists registration, priority management (incl. the `OPT_IN_ONLY_PRIORITY` sentinel + override warnings inherited from PR #190), and the `check`/`execute` loop onto the base; arity-specific subclasses override `_cache_key`, `_find_methods`, and `_format_key`.
- Adds `UnaryDispatchMethod` / `UnaryDispatchRegistry` as the concrete single-argument subclasses. **The original `Method` / `MethodRegistry` aliases are removed** (per @jhuggins' direction during review): inference-method classes now subclass `UnaryDispatchMethod` directly (or `InferenceMethod`, the unchanged domain re-export from `probpipe.inference`).
- Adds `BinaryDispatchMethod` / `BinaryDispatchRegistry` for two-argument dispatch, dispatching on the joint type of the first two positional arguments via paired `((left_types,), (right_types,))` pre-filters.

## ⚠️ Breaking change

`Method` and `MethodRegistry` are removed from `probpipe.core._registry` and from the `probpipe.inference` re-export surface. External code that imported either name will need to migrate to `UnaryDispatchMethod` / `UnaryDispatchRegistry` (or `InferenceMethod` for inference-method subclasses). The CHANGELOG entry documents the migration.

## Design notes

- `supported_types()` is **not** on `BaseDispatchMethod` — its return shape is arity-dependent. `UnaryDispatchMethod.supported_types() -> tuple[type, ...]`; `BinaryDispatchMethod.supported_types() -> tuple[tuple[type, ...], tuple[type, ...]]`.
- Opt-in filtering uses Approach A: a shared `_is_auto_dispatchable(m)` helper on the base, called by each subclass's `_find_methods`. Single point of truth for the sentinel check.
- `_format_key` hook on the base produces `"TypeName"` (unary) or `"(TypeA, TypeB)"` (binary) in error messages.
- `BinaryDispatchRegistry._cache_key` raises `TypeError` (with a clear "at least two positional arguments" message) when called with fewer than 2 args, so a 1-arg call gives a useful error rather than a bare `IndexError`. Under-arity behaviour is documented in the base class docstring.
- `BaseDispatchRegistry.register` rejects empty / falsy `method.name` with `ValueError`, in addition to the existing duplicate-name check.
- `BaseDispatchRegistry.__init__` takes no `name`/`description` yet — catalog wiring is Stage 2.

## Files changed

- `probpipe/core/_registry.py` — registry hierarchy extracted; aliases retired; arity-guard, empty-name guard, and docstring augmentations applied per review feedback
- `probpipe/inference/_registry.py` — `InferenceMethod` aliases now point at `UnaryDispatchMethod`; singleton typed as `UnaryDispatchRegistry`
- `probpipe/inference/__init__.py` — re-exports the new hierarchy symbols (`BaseDispatch*`, `UnaryDispatch*`, `BinaryDispatch*`, `MethodInfo`)
- `probpipe/inference/_blackjax_sgmcmc.py` — module-docstring `:class:` cross-ref updated to point at `UnaryDispatchMethod`
- `tests/test_binary_registry.py` — new; covers the binary dispatch surface (registration, priority ordering, type pre-filter, subclass matching, opt-in-only, `set_priorities`, cache invalidation) plus `TestArgumentCountGuards` (0-arg / 1-arg behaviour) and `TestRegistrationGuards` (empty/`None`-name rejection)
- `tests/inference/test_inference_registry.py` — renamed `MethodRegistry` references to `UnaryDispatchRegistry`; `FakeMethod` now subclasses `UnaryDispatchMethod`
- Doc updates: `CHANGELOG.md` (new entry), `CONTRIBUTING.md` (Key abstractions table expanded to document the hierarchy), `STYLE_GUIDE.md` (registry naming-rule lead-in updated), `docs/api/extending.md` (mkdocstrings auto-render block + prose updated to `InferenceMethod`)

## Test plan

- [x] `tests/test_binary_registry.py` — full binary dispatch coverage (incl. argument-count guards + registration guards)
- [x] `tests/inference/test_inference_registry.py` — all existing tests still pass after alias retirement; new `Bare` class test re-targeted at `UnaryDispatchMethod`
- [x] `mkdocs build --strict` — clean (the `:class:` cross-ref fix is the doc-build prerequisite)
- [x] `ruff check` — neutral against baseline on touched files


probpipe/core/_registry.py
arob5: Artifact from previous PR, but let's get rid of the reference to the issue here

probpipe/core/_registry.py
```diff
@@ -12,12 +12,21 @@
...
 See ``docs/api/extending.md`` for the tier criteria a contributor
 should use when choosing a priority for a new method.
+
+**Arity-typed subclasses.** :class:`UnaryDispatchRegistry` and
+:class:`BinaryDispatchRegistry` differ only in ``supported_types()``
+signature, cache-key construction, and the pre-filter predicate.  Arity
+is a type-level property: ``BinaryDispatchRegistry[KLMethod].execute(p)``
+is a static type error rather than a runtime one.
+
+``Method`` and ``MethodRegistry`` are aliases for the unary subclasses
```
arob5: @jhuggins do we want these aliases or can we just retire the old names?

jhuggins: Retire the old names

probpipe/core/_registry.py
```diff
@@ -92,23 +113,73 @@ def priority(self) -> int:
...
+# ---------------------------------------------------------------------------
+
+
+class BaseDispatchRegistry[M: BaseDispatchMethod](ABC):
+    """Priority-based dispatch registry (arity-independent base).
+
+    Holds all arity-independent logic: registration, priority
+    management, opt-in filtering, and the ``check``/``execute`` loop.
+    Arity-specific subclasses override :meth:`_cache_key`,
+    :meth:`_find_methods`, and :meth:`_format_key`.
```
arob5, line 169: This class represents a registry with particular priority-based dispatch mechanics (which, for example, operate differently than the converter priority-based dispatch). I think it'd be good to augment the docstring here so that it gives a concrete description of what this class really is; i.e., how the dispatch works. Also a good place to explain the motivation for priority overrides and how they work.

probpipe/core/_registry.py
tcikezu, line 246: I was wondering, would we like to assert that `method.name` is not the empty string?

probpipe/core/_registry.py
```diff
@@ -232,38 +317,119 @@ def execute(
...
+        Applies an ``issubclass`` pre-filter on ``supported_types()``
+        and excludes opt-in-only methods via
+        :meth:`~BaseDispatchRegistry._is_auto_dispatchable`.
 
         Uses ``issubclass`` rather than ``isinstance`` because we are
         comparing *types*, not instances.  ``supported_types()`` must
         return concrete classes (not protocols with non-method members),
         since ``issubclass`` does not work reliably with such protocols.
         """
-        if key_type not in self._type_cache:
```
tcikezu, line 263: I don't understand this variable name change

arob5:
## PR Review: feat(core): extract BaseDispatchRegistry; add BinaryDispatchRegistry (Stage 1)

### Summary

Clean, well-structured refactor. The extraction of `BaseDispatchRegistry` and addition of `BinaryDispatchRegistry` is architecturally sound, and the 38-test suite covers the binary dispatch surface thoroughly. There are one correctness gap, a few documentation hygiene items, and a handful of minor test coverage gaps.

---

### Findings

#### Critical (must fix)

**1. `BinaryDispatchRegistry._cache_key` crashes with `IndexError` on a 1-arg call**

`_registry.py`, `BinaryDispatchRegistry._cache_key`:

```python
def _cache_key(self, args: tuple) -> tuple[type, type]:
    return (type(args[0]), type(args[1]))   # IndexError if len(args) == 1
```

`BaseDispatchRegistry.check`/`execute` guard against 0 args (`if not args:`), but `_cache_key` is called for any `len(args) >= 1`. A caller that passes a single object to a `BinaryDispatchRegistry` gets a bare `IndexError: tuple index out of range` with no indication that two arguments are required. Should raise `TypeError`:

```python
def _cache_key(self, args: tuple) -> tuple[type, type]:
    if len(args) < 2:
        raise TypeError(
            "BinaryDispatchRegistry requires at least two positional arguments; "
            f"got {len(args)}"
        )
    return (type(args[0]), type(args[1]))
```

---

#### Recommended (should fix)

**2. `CONTRIBUTING.md` "Key abstractions" table is stale**

```
| `MethodRegistry` | Generic priority-based dispatch; used by the inference method registry |
```

This PR introduces a three-class hierarchy (`BaseDispatchRegistry` → `UnaryDispatchRegistry`, `BinaryDispatchRegistry`) and two new abstract method bases. The table entry should be expanded to document the hierarchy — as written, `CONTRIBUTING.md` describes the old, flat design.

**3. New public symbols not re-exported from any `__init__.py`**

`__all__` in `_registry.py` lists `BaseDispatchMethod`, `UnaryDispatchMethod`, `BinaryDispatchMethod`, `BaseDispatchRegistry`, `UnaryDispatchRegistry`, `BinaryDispatchRegistry` as public symbols, but none is re-exported from any package `__init__.py`. Any implementor of a custom `BinaryDispatchMethod` subclass must write:

```python
from probpipe.core._registry import BinaryDispatchMethod   # underscore module
```

At minimum `BinaryDispatchMethod` and `BinaryDispatchRegistry` should be accessible from `probpipe.inference` (or `probpipe.core`) without reaching into an underscore module.

**4. Module docstring first line is stale**

`_registry.py`, line 1:

```
"""Generic priority-based method registry.
```

The module now defines a dispatch registry *hierarchy*. The one-line summary no longer fits and should be updated (e.g., `"Priority-based dispatch registry hierarchy: unary and binary arity-typed subclasses of a shared abstract base."`).

---

#### Minor (nice to have)

**5. `BaseDispatchMethod.priority` docstring doesn't mention the sentinel meaning**

```python
@property
def priority(self) -> int:
    """Higher priority methods are tried first.  Default 0."""
    return OPT_IN_ONLY_PRIORITY
```

"Default 0" loses the semantic: the default is the *opt-in-only sentinel*, not just an arbitrary zero. Suggest: `"Higher priority methods are tried first. Defaults to ``OPT_IN_ONLY_PRIORITY`` (opt-in only; not auto-dispatched)."`

**6. No test coverage for the 0-arg early-exit paths**

The new `if not args:` branches in `BaseDispatchRegistry.check` and `execute` (returning `MethodInfo(feasible=False, ...)` and raising `TypeError` respectively) are untested. These represent a behavioral change from the previous "default to `object`" behavior and are exactly the two lines Codecov flagged as missing.

**7. `BinaryDispatchRegistry._find_methods` docstring is slightly imprecise**

> "Both left and right type slots must pass an ``issubclass`` check against ``supported_types()[0]`` and ``supported_types()[1]`` respectively."

The `any(...)` in the implementation means a method matches when the left arg type is a subclass of *at least one* type in `supported_types()[0]` **and** the right arg type is a subclass of at least one in `supported_types()[1]`. The current phrasing doesn't convey the `any()` semantics.

---

### Suggested Changes

**Change 1 — Fix `BinaryDispatchRegistry._cache_key` to give a useful error (Critical)**

```python
# before
def _cache_key(self, args: tuple) -> tuple[type, type]:
    return (type(args[0]), type(args[1]))

# after
def _cache_key(self, args: tuple) -> tuple[type, type]:
    if len(args) < 2:
        raise TypeError(
            "BinaryDispatchRegistry requires at least two positional arguments; "
            f"got {len(args)}"
        )
    return (type(args[0]), type(args[1]))
```

Add a corresponding test in `test_binary_registry.py`:

```python
def test_execute_single_arg_raises_type_error(self):
    reg = BinaryDispatchRegistry()
    with pytest.raises(TypeError, match="at least two positional arguments"):
        reg.execute(Left())
```

**Change 2 — Update `CONTRIBUTING.md` key abstractions table (Recommended)**

Replace the single `MethodRegistry` row with something like:

```
| `BaseDispatchRegistry` | Abstract base for priority-based registries; holds registration, priority management, opt-in filtering, `check`/`execute` loop. |
| `BinaryDispatchRegistry` | Two-argument dispatch registry; dispatches on the joint type of the first two positional args. |
```

**Change 3 — Re-export new public symbols (Recommended)**

In `probpipe/inference/__init__.py`, add the new symbols so they are importable without reaching into an underscore module:

```python
from ..core._registry import (
    BaseDispatchMethod, UnaryDispatchMethod, BinaryDispatchMethod,
    BaseDispatchRegistry, UnaryDispatchRegistry, BinaryDispatchRegistry,
    MethodInfo,
)
```

**Change 4 — Update module docstring first line (Recommended)**

```python
# before
"""Generic priority-based method registry.

# after
"""Priority-based dispatch registry hierarchy.
```

**Change 5 — Improve `BaseDispatchMethod.priority` docstring (Minor)**

```python
# before
"""Higher priority methods are tried first.  Default 0."""

# after
"""Higher priority methods are tried first.

Defaults to :data:`OPT_IN_ONLY_PRIORITY` — the method is excluded
from auto-dispatch and only reachable by explicit ``method="..."``
until a contributor assigns a positive priority.
"""
```

**Change 6 — Add tests for 0-arg early exits (Minor)**

```python
def test_check_no_args_returns_infeasible(self):
    reg = BinaryDispatchRegistry()
    info = reg.check()
    assert not info.feasible

def test_execute_no_args_raises(self):
    reg = BinaryDispatchRegistry()
    with pytest.raises(TypeError, match="No arguments provided"):
        reg.execute()
```

**Change 7 — Tighten `_find_methods` docstring in `BinaryDispatchRegistry` (Minor)**

Replace "Both left and right type slots must pass an ``issubclass`` check..." with: "A method is included when the left arg type is a subclass of at least one type in ``supported_types()[0]`` **and** the right arg type is a subclass of at least one type in ``supported_types()[1]``."

arob5:
@tcikezu once you address the items above (both the comment immediately above as well as my older ones), should be good to merge

arob5:
## PR Review: extract `BaseDispatchRegistry`; add `BinaryDispatchRegistry` (Stage 1)

### Summary
A clean, well-structured refactor that hoists arity-independent dispatch logic into an abstract `BaseDispatchRegistry`/`BaseDispatchMethod` and adds unary + binary concrete subclasses. The code adheres closely to ProbPipe conventions, binary-path test coverage is thorough (38 tests), and the convention docs (STYLE_GUIDE, CONTRIBUTING, extending.md, CHANGELOG) were all updated consistently. Two real issues: stale doc cross-references that will trip `mkdocs --strict`, and a PR description that contradicts the actual code.

### Recommended (should fix)

1. **Stale `:class:` cross-references to the removed `Method` symbol.** Two module docstrings still reference `:class:\`~probpipe.core._registry.Method\``, which no longer exists after this PR:
   - `probpipe/inference/_blackjax_mcmc.py:3`
   - `probpipe/inference/_blackjax_sgmcmc.py:3`

   These files weren't touched by the PR, but the PR removed the symbol they point to. mkdocstrings/griffe will fail to resolve the reference, and CI runs `mkdocs build --strict` (per CONTRIBUTING.md), which fails on warnings. Update both to `UnaryDispatchMethod` (or `InferenceMethod`).

2. **PR description contradicts the code.** The body says *"`Method` and `MethodRegistry` remain as aliases — no call-site changes anywhere."* In fact the aliases are removed (no `Method =`/`MethodRegistry =` defined anywhere at head; dropped from `__all__` in both `core/_registry.py` and `inference/__init__.py`), and call sites *were* updated (inference `__init__`, `_registry`, `test_inference_registry`). The CHANGELOG correctly says they're removed. This matters because removing `Method`/`MethodRegistry` from the public `probpipe.inference` namespace is a **breaking change** for external importers — please reconcile the description so the compat impact is assessed accurately. (The change itself is fine and documented in the CHANGELOG.)

### Minor (nice to have)

3. **`supported_types()` called twice per candidate** in `BinaryDispatchRegistry._find_methods` — `m.supported_types()[0]` and `...[1]` each invoke the method. Only on cache-miss, but binding to a local (`st = m.supported_types()`) is cleaner and cheaper.

4. **Asymmetric `check()` under-arity behavior on the binary registry.** With zero args, `check()` returns an infeasible `MethodInfo` ("No arguments provided"); with exactly one arg it *raises* `TypeError` from `_cache_key`. Both are tested and arguably intentional, but `check()` is documented as a cheap *probe* — having it raise in one under-arity case and return infeasible in another is a subtle inconsistency worth a deliberate decision.

5. **`register()` docstring is now incomplete.** It says only "invalidates the lookup cache" but the method now also raises `ValueError` for an empty/falsy name (new guard) and for a duplicate name. Add a `Raises` section.

6. **Bare `tuple` type hint** on `_cache_key(self, args: tuple)` in the base and subclasses. STYLE_GUIDE §5 favors precise hints; `tuple[Any, ...]` is more informative.

7. **Cryptic placeholder in the module docstring** — `BinaryDispatchRegistry[KLMethod].execute(p)` references an undefined `KLMethod`. Consider a named-but-explained example or drop it.

### Checked, no issues
- **Naming / imports / types / dependency graph**: PEP 695 generics, single-underscore private helpers, `from __future__ import annotations`, and `__all__` updates all conform; no illegal cross-imports.
- **AI-artifact comments**: none. The `# Approach A: single point of truth` comment ties to the design notes and is intentional.
- **Duplicate code**: the unary vs. binary `_find_methods` overlap is the intended cost of the abstract split; shared opt-in filtering is correctly centralized in `_is_auto_dispatchable`.
- **Test coverage**: thorough for the binary path (registration, priority ordering, both-slot/subclass pre-filter, opt-in-only, `set_priorities` warnings, cache invalidation, arity guards, name guards). The new empty-name guard is covered via the binary suite.
- **Inference methods**: all subclass `InferenceMethod` (= `UnaryDispatchMethod`) with flat `supported_types()` tuples — consistent with the unary contract. No dangling imports of the removed names anywhere in `probpipe/` or `tests/`.

### Suggested changes
1. `_blackjax_mcmc.py:3` & `_blackjax_sgmcmc.py:3` — `Method` → `UnaryDispatchMethod` in the `:class:` refs (doc-build blocker).
2. PR description — correct the "remain as aliases / no call-site changes" sentence.
3. `BinaryDispatchRegistry._find_methods` — bind `st = m.supported_types()` once.
4. `register` — add a `Raises` section (`ValueError` on empty/duplicate name).
5. Optional: `tuple` → `tuple[Any, ...]`; clarify/remove the `KLMethod` example; decide whether binary `check()` with one arg should return infeasible rather than raise.

🤖 Review generated with Claude Code

arob5 (approved):

PR #206, opened 2026-05-22
refactor(core) [#196 PR-1]: Result-contract extraction

SoymilkWinsAgain:
## Summary

This is the stage 1 of #196, i.e. result-contract extraction. Only small changes implemented, no functional changes. Fixed a small lint dept in `probpipe/core/node.py` (see below for detail). Also added my contact information in `pyproject.toml` file. No new tests introduced.

## What changed

1. Added `probpipe/core/_workflow_result.py` with `BroadcastMode`, `BROADCAST_*`, `_wrap_as_record`, and `_coerce_output`.
2. Removed the moved helper implementation from `probpipe/core/node.py`.
3. Cleaned imports made unused by the move: `Literal`, `np, _make_marginal`, `NumericRecord`, `_is_numeric_leaf, and top-level Record`.
4. Changed the code format in `probpipe/core/node.py` to conform to modern typing rules does not introduce any functional changes. Including: `from typing import Callable, Mapping` -> `from collections.abc import Callable, Mapping`. And `Mapping[str, "Node"]` -> `Mapping[str, Node]`
5. Added my contact information in `pyproject.toml`.

### DIFF
--- a/probpipe/core/node.py
+++ b/probpipe/core/node.py
@@ -1,20 +1,20 @@
 from __future__ import annotations
 
-from abc import ABC, abstractmethod
-from typing import Any, get_type_hints
-from collections.abc import Callable, Mapping
 import inspect
 import logging
+import warnings
+from abc import ABC, abstractmethod
+from collections.abc import Callable, Mapping
 from concurrent.futures import ThreadPoolExecutor
 from itertools import product as cartesian_product
 from types import MappingProxyType
-import warnings
+from typing import Any, get_type_hints
 
 import jax
 import jax.numpy as jnp
 
 try:
-    from prefect import task, flow
+    from prefect import flow, task
 except ImportError:
     task = flow = None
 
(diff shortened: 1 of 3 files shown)


probpipe/core/node.py
```diff
@@ -32,15 +32,13 @@
...
-    _make_marginal,
 )
 from ._broadcast_distributions import _make_stack
 from ._distribution_array import DistributionArray, _make_distribution_array
-from ._numeric_record import NumericRecord, _is_numeric_leaf
 from ._record_array import RecordArray, _RecordArrayView
 from ._record_distribution import _RecordDistributionView
 from .provenance import Provenance
-from .record import Record
+from . import _workflow_result as _workflow_result
```
jhuggins: I don't think there's a need for `as _workflow_result`

probpipe/core/node.py
```diff
@@ -77,111 +75,14 @@
...
-    if provenance is not None and hasattr(value, "with_source"):
-        try:
-            value.with_source(provenance)
-        except RuntimeError:
-            # Value already has a source (e.g., an inner marginal that
-            # the broadcasting layer built with its own provenance).
-            # Leave the existing source in place.
-            pass
-    return value
+# Compatibility aliases for tests or downstream code that imported these
```
jhuggins, line 78: I would prefer not to retain aliases unless absolutely necessary

SoymilkWinsAgain:
One more thing:

Rewrote

```python
try:
    value.with_source(provenance)
except RuntimeError:
    pass
```

into

```python
with suppress(RuntimeError):
    value.with_source(provenance)
```

to shorten the code.

probpipe/core/node.py
jhuggins: also, not in correct (alphabetical) order

probpipe/core/node.py
SoymilkWinsAgain: This is something I intentionally kept. It's a deliberate compatibility layer to ensure that older tests and potential downstream code can still import `_coerce_output` from `probpipe.core.node`. It's reasonable for this PR. I'll delete it if it's no longer needed.

jhuggins (approved):

SoymilkWinsAgain:
Retaining compatibility aliases are necessary because there are some existing tests that do use something like `from probpipe.core.node import _coerce_output`. Removing these aliases would require much larger code changes, which is far beyond the scope of this PR. Plan to address that in a future clean-up-dedicated PR.

PR #211, opened 2026-05-26
refactor(core) [#196 PR-2]: Execution extraction

SoymilkWinsAgain:
This is the stage 2 of #196. 

Changes:

1. The following files have been added: `probpipe/core/_workflow_execution.py`, containing `WorkflowExecutionConfig`, `WorkflowExecutionRequest`, `execute_many(request)`, and the `sequential/thread/Prefect` executor.

2. `probpipe/core/node.py` retains `effective_workflow_kind` and adds `_make_execution_config()` and `_make_execution_request()`. The original `_execute_many*` and `_map_task` have become compatibility wrappers. (P.S. This was intentionally designed to pass pytests. Will become clear during the future clean-up PR.)

3. `tests/test_workflow_parallel.py` has been rewritten in groups of `Test*`, directly overriding the execution module and retaining compatibility tests for the `WorkflowFunction` private wrapper.

### DIFF
--- a/probpipe/core/_workflow_execution.py
+++ b/probpipe/core/_workflow_execution.py
@@ -10,7 +10,7 @@ from __future__ import annotations
 from collections.abc import Callable
 from concurrent.futures import ThreadPoolExecutor
 from dataclasses import dataclass
-from typing import Any, Literal, Protocol
+from typing import Any, Literal
 
 try:
     from prefect import flow, task
@@ -44,37 +44,20 @@ class WorkflowExecutionRequest:
     execution: WorkflowExecutionConfig
 
 
-class _WorkflowExecutor(Protocol):
-    def execute_many(self, request: WorkflowExecutionRequest) -> list[Any]:
-        """Execute the request and return results in call order."""
-        ...
-
-
-class _SequentialExecutor:
-    def execute_many(self, request: WorkflowExecutionRequest) -> list[Any]:
-        return [request.func(**values) for values in request.call_value_list]
-
-
-class _ThreadExecutor:
-    def execute_many(self, request: WorkflowExecutionRequest) -> list[Any]:
-        return execute_many_threaded(request)
-
-
-class _PrefectTaskExecutor:
-    def execute_many(self, request: WorkflowExecutionRequest) -> list[Any]:
-        return execute_many_prefect_task(request)
-
-
-class _PrefectFlowExecutor:
-    def execute_many(self, request: WorkflowExecutionRequest) -> list[Any]:
-        return execute_many_prefect_flow(request)
-
-
 def execute_many(request: WorkflowExecutionRequest) -> list[Any]:
     """Execute all call dictionaries using the configured dispatch mode."""
     if not request.call_value_list:
         return []
-    return _resolve_executor(request.execution).execute_many(request)
+
+    if request.execution.mode == "sequential":
+        return [request.func(**values) for values in request.call_value_list]
+    if request.execution.mode == "thread":
+        return execute_many_threaded(request)
+    if request.execution.mode == "prefect_task":
+        return execute_many_prefect_task(request)
+    if request.execution.mode == "prefect_flow":
+        return execute_many_prefect_flow(request)
+    raise ValueError(f"Unknown workflow execution mode: {request.execution.mode!r}")
 
 
 def execute_many_threaded(request: WorkflowExecutionRequest) -> list[Any]:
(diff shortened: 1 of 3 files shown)


jhuggins (changes requested):
Overall looks great. Just a few few minor things before merging: 

1. **Collapse the executor layer.** The `_WorkflowExecutor` Protocol + four
   one-line `_*Executor` classes + `_resolve_executor`
   (`_workflow_execution.py:47-70`, `154-163`) are indirection without payoff:
   three classes just forward to the same-named module functions and the
   sequential one inlines its body. `execute_many` can switch directly on
   `request.execution.mode`. This also matches the PR-1 sibling
   `_workflow_result.py`, which uses plain functions. ~25 fewer lines.

2. **Dead wrappers / cleanup tracking.** Only `_execute_many` has production
   callers now; `_execute_many_threaded`, `_map_task`, and
   `_execute_many_prefect_task/flow` are kept solely to satisfy the old tests.
   Fine as a staging step — could we either (a) point those unit tests at the
   module functions and drop the wrappers here, or (b) add a checkbox to #196 so
   they don't linger?

3. **Stale error text** (`_workflow_execution.py:170`): the `ValueError` reads
   `"self._parallel must be…"` inside a standalone function; the sibling
   `TypeError` already says `"parallel must be…"`. Drop the `self.`.

4. **Flow-guard wording** (`_workflow_execution.py:190-195`): the
   flow-availability guard raises `"Prefect task execution was requested…"` —
   should say "task or flow".

5. **Coverage gap.** The `parallel`-truthy → `"thread"` branch in
   `_make_execution_config` (`node.py:432-433`) is no longer exercised through
   the live path — `test_threaded_wrapper_delegates_to_execution_module` forces
   `mode="thread"`. A one-liner asserting `_make_execution_config()` returns
   `"thread"` for truthy `parallel` with `kind=OFF` would restore what the old
   `parallel=True` test covered.

6. `Attributes` sections on `WorkflowExecutionConfig` / `WorkflowExecutionRequest`
   — `parallel: bool | int` and `prefect_task_runner` semantics aren't obvious
   from the hints alone.

7. The empty-list guard is now triple-checked on the live path (wrapper →
   `execute_many` → per-mode fn); could thin to just the dispatcher plus the
   direct-call entry points.

SoymilkWinsAgain:
NOTE: The issue of duplicate guards mentioned in review point 7 does exist. However, these guards were introduced through all pytests to preserve the old code execution logic. Therefore, I only plan to remove one guard from `map_task()` in `node.py`.

Removing more guards is actually feasible, but this would require modifying pytests, which I don't plan to do for now, as it might introduce unknown changes. Furthermore, in practice, these guards have almost no impact on performance, making the changes largely meaningless.

SoymilkWinsAgain:
For point 2: A clearer note has been added because the current plan includes a clean-up phase, so it doesn't need to be done here. See the final stage in #196.

Point 6 is related to #213 which involves a wider range of modifications. Therefore, it will not be processed in this PR.

SoymilkWinsAgain (commented):
Point 1, 3, 4, 5, 7 has fixed.

For point 2: A clearer note has been added because the current plan includes a clean-up phase, so it doesn't need to be done here. See the final stage in (#196).

Point 6 is related to #213 which involves a wider range of modifications. Therefore, it will not be processed in this PR.

probpipe/core/_workflow_execution.py
```diff
@@ -0,0 +1,168 @@
...
+
+def _ensure_prefect_task_available() -> None:
+    if task is None:
+        raise RuntimeError(
+            "Prefect task or flow execution was requested, but Prefect is not installed. "
+            "Install with: pip install probpipe[prefect]"
+        )
+
+
+def _ensure_prefect_flow_available() -> None:
```
jhuggins: Since both prefect tasks and flows will be available or not, reduce to a single _ensure_prefect_available() check.

jhuggins (changes requested):

probpipe/core/_workflow_execution.py
SoymilkWinsAgain: There is now only one `_ensure_perfect_available()`.

SoymilkWinsAgain:
Initially, `_ensure_perfect_task_available()` and `_ensure_perfect_flow_available()` were different, but after working on point 4, they are actually the same. There is now only one `_ensure_perfect_available()`.

jhuggins (approved):

PR #214, opened 2026-05-29
refactor(core): Redesigning `@workflow_function` API

### LINKED ISSUE #213: Redesigning `@workflow_function` API
## Summary and Motivation

Currently, the private `WorkflowExecutionConfig` API, as well as the public API of the `WorkflowFunction` class, i.e. the `@workflow_function` decorator, uses a single `parallel` argument to represent two different concepts:

- whether local threaded execution is enabled;
- how many local workers should be used.

At the moment, `parallel` can be either a `bool` or an `int`:

- `parallel=False`: disable local threaded execution;
- `parallel=True`: enable local threaded execution and automatically choose `max_workers`;
- `parallel=<int>`: enable local threaded execution and use the given value as `max_workers`.

This is logically valid, but it has some practical issues.

First, `bool` is a subclass of `int` in Python. As a result, checks based on `isinstance(parallel, int)` can easily behave incorrectly unless `bool` is handled explicitly first. I previously submitted a related debug PR, see #192. There may be other similar edge cases, especially after the config is passed to the execution backend, where such behavior can be harder to test and debug.

Second, the current API mixes several different concepts:

- JAX vectorization vs row-wise/function-call dispatch;
- local sequential vs local threaded execution;
- orchestration through Prefect task/flow.

The existing `vectorize` field is also becoming less accurate as a name, because `"loop"` is not vectorization. The more general concept is dispatch strategy.

## Proposed Change

I propose replacing the public `parallel` control with an explicit `dispatch` strategy and keeping `max_workers` only as a configuration option for local threaded dispatch.

```python
# Public API
WorkflowFunction(
    ...,
    dispatch="auto",  # "auto" | "jax" | "sequential" | "thread"
    workflow_kind=WorkflowKind.DEFAULT,  # orchestration strategy
    max_workers=None,  # local ThreadPoolExecutor worker count
)

# Private API
WorkflowExecutionConfig(
    mode="sequential" | "thread" | "prefect_task" | "prefect_flow",
    max_workers=None,
    name="...",
    prefect_task_runner=...,
)
```

Where:
- `dispatch="jax"`: must use JAX vmap.
- `dispatch="sequential"`: Python sequential row-wise/function-call dispatch.
- `dispatch="thread"`: Python `ThreadPoolExecutor` dispatch, `max_workers` applies here.
- `dispatch="auto"`: Prioritize JAX, then fallback to sequential, i.e. `dispatch="sequential"`.

This keeps the execution layers separate:
1. `dispatch`: JAX vmap vs Python sequential vs Python threaded dispatch.
2. `workflow_kind`: off vs Prefect task/flow orchestration.
3. `max_workers`: worker-count configuration only for local threaded dispatch.

**Capability warning: This is a breaking API change.**
- `parallel=<bool | int>` is removed.
- `vectorize` is replaced or renamed to `dispatch`.
- users who want local threaded execution should write `dispatch="thread", max_workers=<int | None>`.
- users who want the old non-JAX loop behavior should write `dispatch="sequential"`.

## Impact Assessment

The implementation change should be manageable, but the API impact is non-trivial because this affects the parameters of the core `@workflow_function` decorator.

This change may require updates to:

- internal imports and call sites;
- demo scripts;
- documentation or example notebooks;
- any downstream code using parallel or vectorize.

### `Prefect` backend

The `dispatch` field controls local/JAX function dispatch. `workflow_kind` controls orchestration.

The current priority remains:

```text
Prefect TASK/FLOW > local thread parallel (if parallel == True) > sequential
```

If `workflow_kind` resolves to `TASK` or `FLOW`, Prefect wraps the selected computation for orchestration. Local thread settings should not control Prefect scheduling.

### `dispatch` field

The `dispatch` field controls the function-call dispatch strategy:

- `"jax"` uses JAX vmap;
- `"sequential"` uses Python sequential row-wise/function-call dispatch;
- `"thread"` uses Python ThreadPoolExecutor dispatch;
- `"auto"` chooses JAX if possible, otherwise falls back to Python sequential dispatch.

From a functional and design perspective, `max_workers` should not affect JAX vmap, Prefect, Ray, or Dask concurrency. It only applies to local `ThreadPoolExecutor` dispatch.

### Control logic branches

- `dispatch="thread", max_workers=None` -> local threaded dispatch with automatic worker selection.
- `dispatch="thread", max_workers=<positive int>` -> local threaded dispatch with an explicit worker count.
- `dispatch!="thread"` and `max_workers is not None` -> warning, because `max_workers` has no effect.
- `workflow_kind in {TASK, FLOW}` and `dispatch="thread"` -> warning, because `Prefect` orchestration takes priority over local thread dispatch.
- `workflow_kind in {TASK, FLOW}` and `max_workers is not None` -> warning, because `max_workers` does not control Prefect scheduling.
- `dispatch="jax"` and `max_workers is not None` -> warning, because `max_workers` does not control JAX vmap.

SoymilkWinsAgain:
## Summary

Refs #213.

This PR replaces the old `WorkflowFunction` `vectorize` / `parallel` controls with a single explicit dispatch API:

```python
WorkflowFunction(
    ...,
    dispatch="auto",  # "auto" | "jax" | "sequential" | "thread"
    max_workers=None,
)
```
Previously, the public API split behavior across `vectorize` and `parallel`, and `parallel` also overloaded two meanings: whether local threaded execution was enabled, and how many workers should be used. This was easy to misuse because bool is a subclass of int, and it also blurred the boundary between vectorization, local threaded execution, and orchestration.

This PR makes the public API explicit:

- `dispatch="auto"` probes JAX traceability and uses JAX when possible, otherwise falls back to local sequential dispatch.
- `dispatch="jax"` requires JAX dispatch and fails clearly if the function/path is not JAX-traceable.
- `dispatch="sequential"` uses local row-wise/function-call dispatch without threads.
- `dispatch="thread"` uses local row-wise/function-call dispatch through `ThreadPoolExecutor`. `max_workers` only configures dispatch="thread".

This is an intentional breaking API change. `vectorize=` and `parallel=` are no longer accepted.

## Behavior

Dispatch and orchestration are now treated as separate layers:

- `dispatch` controls how broadcast samples are dispatched: JAX vmap, local sequential calls, or local threaded calls.
- `workflow_kind` controls whether the dispatch is wrapped in Prefect task/flow orchestration.

The new behavior is:

- `dispatch="auto"`: use JAX if the wrapped function and broadcast path are JAX-traceable; otherwise use sequential dispatch.
- `dispatch="jax"`: use JAX vmap; raise a clear `ValueError` if tracing fails or the broadcast path is unsupported.
- `dispatch="sequential"`: use local row-wise dispatch.
- `dispatch="thread", max_workers=None`: use `ThreadPoolExecutor` with default worker selection.
- `dispatch="thread", max_workers=<positive int>`: use `ThreadPoolExecutor(max_workers=...)`.
- `max_workers` with any non-thread dispatch: warn and ignore it.
- `Prefect TASK / FLOW` orchestration remains separate; local thread settings do not control Prefect scheduling and warn when ignored.

## Implementation Notes

- Private `WorkflowExecutionConfig` remains a resolved execution contract. It carries execution facts such as `mode`, `max_workers`, `name`, and `prefect_task_runner`.
- `WorkflowExecutionConfig` does not carry public intent fields like `dispatch`, `vectorize`, or `parallel`.
- `_workflow_execution.execute_many(...)` dispatches directly on resolved `mode`.
- The explicit JAX path now reuses the same trace probe as auto-dispatch, but `dispatch="auto"` still probes only once.
- Docs and notebooks have been updated to use `dispatch=` terminology.

## Tests
```bash
ruff check --select I,F probpipe/core/node.py probpipe/core/_workflow_execution.py tests/core/test_workflow_parallel.py tests/core/test_broadcasting.py tests/core/test_prefect_config.py tests/core/test_prefect_orchestration.py tests/core/test_provenance.py tests/core/test_record_array_view.py tests/core/test_recordarray_broadcasting.py tests/core/test_workflow_call_resolution.py tests/core/test_workflow_distribution_normalization.py tests/distributions/test_joint_empirical.py tests/distributions/test_joint_gaussian.py tests/distributions/test_product.py tests/distributions/test_sequential_joint.py tests/inference/test_inference.py
ruff check probpipe/core/_workflow_execution.py tests/core/test_workflow_parallel.py tests/core/test_broadcasting.py tests/core/test_recordarray_broadcasting.py tests/core/test_prefect_orchestration.py
pytest -p no:xdist -o addopts= tests/core/test_workflow_parallel.py tests/core/test_broadcasting.py tests/core/test_prefect_config.py tests/core/test_prefect_orchestration.py tests/core/test_provenance.py tests/core/test_recordarray_broadcasting.py tests/core/test_workflow_call_resolution.py tests/core/test_workflow_distribution_normalization.py -q
pytest -p no:xdist -o addopts= tests/distributions/test_product.py tests/distributions/test_joint_empirical.py tests/distributions/test_joint_gaussian.py tests/distributions/test_sequential_joint.py tests/inference/test_inference.py -q
mkdocs build --strict
git diff --check
```

### DIFF
--- a/probpipe/core/_workflow_execution.py
+++ b/probpipe/core/_workflow_execution.py
@@ -67,7 +67,7 @@ def execute_many_threaded(request: WorkflowExecutionRequest) -> list[Any]:
     if not request.call_value_list:
         return []
 
-    max_workers = _resolve_max_workers(request.execution.max_workers)
+    max_workers = _validate_max_workers(request.execution.max_workers)
     with ThreadPoolExecutor(max_workers=max_workers) as pool:
         return list(pool.map(lambda kwargs: request.func(**kwargs), request.call_value_list))
 
@@ -136,11 +136,11 @@ def execute_many_prefect_flow(request: WorkflowExecutionRequest) -> list[Any]:
     return mapped_flow()
 
 
-def _resolve_max_workers(max_workers: int | None) -> int | None:
+def _validate_max_workers(max_workers: int | None) -> int | None:
     if max_workers is None:
         return None
 
-    if isinstance(max_workers, bool) or not isinstance(max_workers, int):
+    if not isinstance(max_workers, int):
         raise TypeError(
             f"max_workers must be None or a positive int; got {max_workers!r}"
         )
--- a/probpipe/core/_workflow_execution.py
+++ b/probpipe/core/_workflow_execution.py
@@ -30,7 +30,8 @@ class WorkflowExecutionConfig:
     """Resolved execution settings for ordered workflow calls."""
 
     mode: WorkflowExecutionMode
-    parallel: bool | int = False
+    parallel: bool = False
+    max_workers: int | None = None
     name: str = "workflow"
     prefect_task_runner: Any | None = None
 
@@ -67,7 +68,7 @@ def execute_many_threaded(request: WorkflowExecutionRequest) -> list[Any]:
     if not request.call_value_list:
         return []
 
-    max_workers = _resolve_max_workers(request.execution.parallel)
+    max_workers = _resolve_max_workers(request.execution.max_workers)
     with ThreadPoolExecutor(max_workers=max_workers) as pool:
         return list(pool.map(lambda kwargs: request.func(**kwargs), request.call_value_list))
 
@@ -136,20 +137,19 @@ def execute_many_prefect_flow(request: WorkflowExecutionRequest) -> list[Any]:
     return mapped_flow()
 
 
-def _resolve_max_workers(parallel: bool | int) -> int | None:
-    if isinstance(parallel, int) and not isinstance(parallel, bool):
-        if parallel < 1:
-            raise ValueError(
-                f"parallel must be True, False, or a positive int; got {parallel!r}"
-            )
-        return parallel
-
-    if parallel is True:
+def _resolve_max_workers(max_workers: int | None) -> int | None:
+    if max_workers is None:
         return None
 
-    raise TypeError(
-        f"parallel must be True, False, or a positive int; got {parallel!r}"
-    )
+    if isinstance(max_workers, bool) or not isinstance(max_workers, int):
+        raise TypeError(
+            f"max_workers must be None or a positive int; got {max_workers!r}"
+        )
+    if max_workers < 1:
+        raise ValueError(
+            f"max_workers must be None or a positive int; got {max_workers!r}"
+        )
+    return max_workers
 
 
 def _ensure_prefect_available() -> None:
--- a/probpipe/core/_workflow_execution.py
+++ b/probpipe/core/_workflow_execution.py
@@ -30,7 +30,6 @@ class WorkflowExecutionConfig:
     """Resolved execution settings for ordered workflow calls."""
 
     mode: WorkflowExecutionMode
-    parallel: bool = False
     max_workers: int | None = None
     name: str = "workflow"
     prefect_task_runner: Any | None = None
--- a/probpipe/core/node.py
+++ b/probpipe/core/node.py
@@ -7,7 +7,7 @@ from abc import ABC, abstractmethod
 from collections.abc import Callable, Mapping
 from itertools import product as cartesian_product
 from types import MappingProxyType
-from typing import Any, Literal, TypeAlias, get_args, get_type_hints
+from typing import Any, ClassVar, Literal, TypeAlias, get_args, get_type_hints
 
 import jax
 import jax.numpy as jnp
@@ -95,9 +95,10 @@ __all__ = [
     "AbstractModule",
 ]
 
-WorkflowFunctionDispatch: TypeAlias = Literal[
+_WorkflowFunctionDispatch: TypeAlias = Literal[
     "auto", "jax", "sequential", "thread"
 ]
+
 
 class InputFrozenError(Exception):
     pass
@@ -303,7 +304,7 @@ class WorkflowFunction(Node):
         - ``"thread"``: local row-wise/function-call dispatch through
           ``ThreadPoolExecutor``.
     max_workers : int or None
-        Worker count for threaded loop execution. ``None`` lets
+        Worker count for threaded sequential execution. ``None`` lets
         ``ThreadPoolExecutor`` choose automatically; a positive integer
         sets the worker count explicitly. Only applies when ``dispatch`` is
         ``"thread"`` and execution resolves to local thread dispatch. JAX
@@ -314,7 +315,9 @@ class WorkflowFunction(Node):
     """
 
     DEFAULT_N_BROADCAST_SAMPLES: int = 128
-    _VALID_DISPATCH_STRATEGIES = get_args(WorkflowFunctionDispatch)
+    _VALID_DISPATCH_STRATEGIES: ClassVar[tuple[str, ...]] = get_args(
+        _WorkflowFunctionDispatch
+    )
 
     def __init__(
         self,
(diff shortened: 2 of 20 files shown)


SoymilkWinsAgain:
Implement the new design

probpipe/core/_workflow_execution.py
```diff
@@ -136,20 +136,19 @@ def mapped_flow():
...
-
-    if parallel is True:
+def _resolve_max_workers(max_workers: int | None) -> int | None:
+    if max_workers is None:
         return None
 
-    raise TypeError(
-        f"parallel must be True, False, or a positive int; got {parallel!r}"
-    )
+    if isinstance(max_workers, bool) or not isinstance(max_workers, int):
```
jhuggins: no need to check for bool type

probpipe/core/node.py
```diff
@@ -286,19 +287,24 @@ class WorkflowFunction(Node):
         Default number of samples drawn when broadcasting.  Can be overridden
         at call time by passing ``n_broadcast_samples=…`` (provided the
         wrapped function does not itself declare a parameter with that name).
-    vectorize : str
-        Vectorization strategy for broadcasting:
+    dispatch : str
+        Function-call dispatch strategy for broadcasting:
 
         - ``"auto"`` (default): probe with ``jax.make_jaxpr``; on success
```
jhuggins, line 298: I think auto should fall back to threaded, not sequential. only use sequential if max_workers = 1.

probpipe/core/node.py
```diff
@@ -314,12 +320,40 @@ def __init__(
         bind: dict[str, Any] | None = None,         # construction-time bindings (defaults/config)
         module: Any | None = None,                  # typically a Module; kept as Any to avoid import cycles
         n_broadcast_samples: int | None = None,      # default number of samples for broadcasting
-        vectorize: str = "auto",                     # "auto" | "jax" | "loop"
-        parallel: bool | int = False,               # True/int for ThreadPoolExecutor, or Prefect .map()
+        dispatch: str = "auto",                     # "auto" | "jax" | "sequential" | "thread"
```
jhuggins: define a Literal for stronger typing hint here

probpipe/core/node.py
```diff
@@ -314,12 +320,40 @@ def __init__(
...
+            raise TypeError(
+                f"WorkflowFunction no longer accepts {names}; use dispatch= instead."
+            )
+
+        if dispatch not in ("auto", "jax", "sequential", "thread"):
+            raise ValueError(
+                "dispatch must be one of 'auto', 'jax', 'sequential', or "
+                f"'thread'; got {dispatch!r}"
+            )
+        if max_workers is not None:
```
jhuggins: seems redundant with `_resolve_max_workers`

jhuggins (changes requested):

probpipe/core/node.py
SoymilkWinsAgain: Confirmed not to change this time. Leaving the fallback behavior robust and conservative this time

probpipe/core/_workflow_execution.py
SoymilkWinsAgain: I intentionally left this because the design plan explicitly rejected this parameter for bool types (or it will be treated as True=1, False=0). However, I've removed it. As long as it doesn't cause confusion, it's fine.

probpipe/core/node.py
SoymilkWinsAgain: These two are logically the same, but their engineering purposes differ: one is to raise errors to the public API asap, while the other is to prevent internal execution errors. I extracted a helper function for code reuse but still retained this design.

jhuggins (approved):

PR #221, opened 2026-06-03
refactor(core) [#196 PR-5]: Sweep execution extraction & clean-up

SoymilkWinsAgain:
This is the stage 5 of #196, also the last stage. Distribution broadcast extraction is considered a legacy issue and will be addressed later.

## Changes
1. Added `probpipe/core/_workflow_sweep.py`.
2. None JAX-vmap row-wise sweep now explicitly handled by `_workflow_execution.execute_many(WorkflowExecutionRequest(...))`.
3. Change `execution config` to a lazy factory that passes to the `sweep` module, which only parses the row-wise path, avoiding warnings/side effects from changing JAX-vmap or early failure paths.
4. Deleted all wrappers and aliases left before. Now say goodbye to old names.
5. Fixed a small bug in `WorkflowFunction._broadcast_distributions_only`.  (Not directly related to the current topic, but there's no harm in making the change. I just happened to see it, or I might forget about it next time.)

### DIFF
--- a/probpipe/core/node.py
+++ b/probpipe/core/node.py
@@ -7,7 +7,7 @@ from abc import ABC, abstractmethod
 from collections.abc import Callable, Mapping
 from itertools import product as cartesian_product
 from types import MappingProxyType
-from typing import Any, ClassVar, Literal, TypeAlias, get_args
+from typing import Any, ClassVar, Literal
 
 import jax
 import jax.numpy as jnp
@@ -31,10 +31,9 @@ from . import (
     _workflow_execution,
     _workflow_plan,
     _workflow_result,
+    _workflow_sweep,
 )
-from ._broadcast_distributions import _make_stack
-from ._distribution_array import DistributionArray, _make_distribution_array
-from ._record_array import RecordArray, _RecordArrayView
+from ._record_array import RecordArray
 from ._record_distribution import _RecordDistributionView
 from .distribution import (
     BroadcastDistribution,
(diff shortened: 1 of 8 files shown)


probpipe/core/node.py
```diff
@@ -273,8 +258,11 @@ class WorkflowFunction(Node):
     """
 
     DEFAULT_N_BROADCAST_SAMPLES: int = 128
-    _VALID_DISPATCH_STRATEGIES: ClassVar[tuple[str, ...]] = get_args(
-        _WorkflowFunctionDispatch
+    _VALID_DISPATCH_STRATEGIES: ClassVar[tuple[str, ...]] = (
```
jhuggins: Shouldn't be needed anymore – baked into _WorkflowFunctionDispatch already, so having it here too is worth for maintainability.

probpipe/core/node.py
```diff
@@ -908,9 +640,15 @@ def _broadcast_distributions_only(
           tracing.
         """
         MIN_BROADCAST_SAMPLES = 5  # Recommended minimum samples
+        
+        if not isinstance(n_broadcast_samples, int):
```
jhuggins, line 638: Add pytest.raises coverage for n_broadcast_samples=2.5 and 0/-1.

jhuggins (changes requested):

probpipe/core/node.py
SoymilkWinsAgain: A silly over-design from AI: it silently deleted my `get_args()` and then added these with no reason... Rolled back.

jhuggins (approved):

PR #230, opened 2026-06-04
nutpie+PyMC: drop auto-transformed value vars at chain extraction

### LINKED ISSUE #225: PyMCModel: chain extraction double-counts auto-transformed RVs (HalfNormal, LogNormal, ...)
## Summary

When a PyMC model includes a free RV with an automatic unconstrained transform — `pm.HalfNormal`, `pm.LogNormal`, `pm.Exponential`, `pm.Gamma`, any positive-support prior — the inference chain carries both the constrained value and the transformed (unconstrained) value, but `PyMCModel.record_template` counts the named RV once. The mismatch surfaces as a clear shape-comparison error at posterior assembly time:

```
ValueError: chain last dim (N+1) doesn't match template total flat size (N);
template fields=(..., 'sigma', ...), sizes=[..., 1, ...].
```

This affects any user who writes the most natural PyMC idioms for variance / scale parameters.

## Reproducer

```python
import numpy as np
import pymc as pm
from probpipe import condition_on
from probpipe.modeling import PyMCModel

def model_fn(y=None):
    with pm.Model() as m:
        mu    = pm.Normal('mu', 0, 1)
        sigma = pm.HalfNormal('sigma', 1)      # auto log-transform
        pm.Normal('obs', mu, sigma, observed=y)
    return m

obs = np.random.randn(50).astype(np.float32)
post = condition_on(PyMCModel(model_fn), obs)   # raises
```

## Proposed fix

The template and the extracted chain should agree on RV count, with `posterior['sigma']` being the *constrained* (positive-support) samples — that's what users expect.

Two implementation options:

1. **Filter the chain to constrained-only.** In the inference dispatch path (`probpipe/inference/_nutpie.py` and any sibling that runs PyMC models), exclude auto-introduced unconstrained value-vars when extracting chain samples. Use `model.value_vars` vs `model.free_RVs` (or check the `transform` attribute of each RV) to identify them, then drop the transformed columns. The recommended path.
2. **Have the template include transformed slots.** Less natural — leaks an implementation detail into the user-facing structure, and `posterior['sigma_log__']` is rarely what anyone wants.

A regression test using a `HalfNormal` (or any positive-support prior) covers this.

## Workaround for users today

Replace the auto-transformed prior with an unconstrained Normal + explicit exponentiation:

```python
log_sigma = pm.Normal('log_sigma', 0.0, 0.5)
sigma     = pm.math.exp(log_sigma)   # tensor, not a free RV
```

Then recover `sigma` as a derived posterior, e.g., via a `@workflow_function` that exponentiates `posterior['log_sigma']`. Workable but awkward — and it rules out the most natural PyMC idioms (HalfNormal, HalfCauchy, Exponential, LogNormal, Gamma, ...).

jhuggins:
Closes #225.

## Diagnosis

`PyMCNutsMethod.execute` already filtered correctly via `dist._param_names` (in `_pymc_method.py`), so the bug only fired on the nutpie path. `_nutpie.py::_extract_chains` iterated every `posterior.data_vars`, which for a PyMC model includes both the user-named constrained RV (`sigma`) and its auto-introduced unconstrained value var (`sigma_log__`). That doubled the column count for any RV with an automatic transform — `HalfNormal`, `Exponential`, `Gamma`, `LogNormal`, `HalfCauchy`, ... — and shape-mismatched against `record_template`.

## Change

- `_extract_chains` gains an optional `keep_names: list[str] | tuple[str, ...] | None` argument. When supplied, only those data-vars are extracted, in the given order. Default `None` preserves the existing "extract everything" behaviour for the Stan path.
- `condition_on_nutpie` passes `getattr(model, "_param_names", None)`. PyMCModel exposes that attribute (constrained free RV names, in template order); StanModel does not, so the fallback path is unchanged.

This mirrors how `_pymc_method.py::_extract_pymc_chains` has always worked — the nutpie path was just inconsistent.

## Tests

- **Unit test** (`TestExtractChains.test_keep_names_filters_out_transformed_value_vars`): mock posterior with `mu`, `sigma`, `sigma_log__`; asserts that `keep_names=["mu", "sigma"]` drops the transformed column and preserves order.
- **Integration test** (`TestNutpieIntegration.test_auto_transformed_rv_not_double_counted`): the canonical reproducer from the issue — `pm.HalfNormal('sigma', 1)` with 50 observations. Asserts `draws.fields == ("mu", "sigma")` (no `sigma_log__`) and `min(sigma_draws) > 0` (constrained values, not log-transformed).

## Test plan

- [ ] CI: `pytest tests/inference/test_nutpie.py` (nutpie + pymc installed)
- [x] AST-validates locally; logic mirrors the already-working `_pymc_method.py` path.

### DIFF
--- a/probpipe/inference/_nutpie.py
+++ b/probpipe/inference/_nutpie.py
@@ -53,7 +53,15 @@ def condition_on_nutpie(
         chains=num_chains, seed=random_seed, **kwargs,
     )
 
-    chains, param_names = _extract_chains(trace, num_chains)
+    # For PyMC models, restrict extraction to the user-named (constrained)
+    # free RVs. Nutpie's posterior also exposes auto-introduced
+    # unconstrained value vars (e.g. ``sigma_log__`` for a
+    # ``HalfNormal('sigma', ...)``); pulling them in would double-count
+    # those RVs and shape-mismatch against ``record_template``.
+    keep_names = getattr(model, "_param_names", None)
+    chains, param_names = _extract_chains(
+        trace, num_chains, keep_names=keep_names,
+    )
 
     return make_posterior(
         chains, parents=(model,), algorithm="nutpie_nuts",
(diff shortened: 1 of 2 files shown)


jhuggins:
Closing — could not reproduce the underlying bug with the project's pinned nutpie/pymc range. Detailed probe evidence on #225.

PR #231, opened 2026-06-04
PyMCModel: data-dependent template shapes via stateless _record_template_for

### LINKED ISSUE #224: PyMCModel: record_template uses unconditioned-build shapes; breaks data-size-dependent RVs
## Summary

`PyMCModel` builds the wrapped PyMC model twice: once during `__init__` with no data (the user's `X=None` sentinel branch), and again at inference time via `_pymc_model(data)`. `record_template` reads RV shapes from the *unconditioned* build (`probpipe/modeling/_pymc.py` around L148–164), but the inference path actually samples from the *conditioned* build. For any RV whose shape depends on data size — per-observation random effects, hierarchical group-level effects where the group count varies, etc. — the two builds produce different shapes and the resulting `NumericRecordTemplate` doesn't match what the chain returns.

The undocumented assumption is "RV shapes are data-independent." It's a real assumption — currently nothing enforces it and nothing warns when it's violated.

## Reproducer

```python
import numpy as np
import pymc as pm
from probpipe import NumericRecordArray, NumericRecordTemplate, condition_on
from probpipe.modeling import PyMCModel

N = 50
data = NumericRecordArray(
    {'X': np.zeros(N, dtype=np.float32), 'y': np.zeros(N, dtype=np.float32)},
    batch_shape=(N,),
    template=NumericRecordTemplate(X=(), y=()),
)

def model_fn(X=None, y=None):
    if X is None:
        X = np.ones(1, dtype=np.float32)        # sentinel — shape (1,)
    with pm.Model() as m:
        intercept = pm.Normal('intercept', 0, 1)
        alpha     = pm.Normal('alpha', 0, 1, shape=X.shape[0])   # data-dependent shape
        pm.Poisson('y', mu=pm.math.exp(intercept + alpha), observed=y)
    return m

# Unconditioned build sets alpha.shape = (1,); inference uses (N,).
# Template thinks alpha is 1-dim; chain returns N — shape mismatch on posterior assembly.
post = condition_on(PyMCModel(model_fn), data)
```

## Proposed fix

Have `record_template` introspect from the *conditioned* model rather than from `self._unconditioned_model`. Two implementation options:

1. **Cache the most recently conditioned model.** Have `_pymc_model(data)` store the resulting `pm.Model` on `self._last_conditioned_model`, and have `record_template` prefer it (falling back to `_unconditioned_model` only when conditioning hasn't happened yet). Smallest diff.
2. **Compute the template at inference time.** Make `record_template` a method that takes the model directly (or compute it inline in the inference path after the rebuild). Cleaner semantically since the property naturally implies "shape of the chain you'll get back," but a bigger API change.

Option (1) is the recommended path.

Either way, document that PyMC RVs whose shape depends on data are now supported. A regression test should cover a per-observation random-effects model.

## Workaround for users today

Capture the real data size via closure so the sentinel build matches the conditioned size:

```python
N = int(data.batch_shape[0])

def model_fn(X=None, y=None):
    if X is None:
        X = np.zeros(N, dtype=np.float32)   # sentinel matches inference size
    ...
```

Workable but brittle — the model function silently depends on a module-level `N` pulled from the dataset.

jhuggins:
Closes #224.

## Diagnosis

`PyMCModel.record_template` / `event_shape` read RV shapes from `_unconditioned_model` (the no-data build done at `__init__`), but inference samples from the data-conditioned rebuild in `_pymc_model(data)`. For any RV whose shape depends on data size (per-observation random effects, hierarchical group-level effects with varying group counts), the two builds disagreed and posterior assembly failed with the issue's exact error:

```
ValueError: chain last dim (13) doesn't match template total flat size (2);
template fields=('intercept', 'alpha'), sizes=[1, 1].
```

(Verified end-to-end: the integration regression test below reproduces it on `main`.)

## Change — stateless `_record_template_for(model, names)`

The template is now a **pure function of a model build**, not of cached instance state:

- `PyMCModel._record_template_for(model, names)` builds the template from a passed PyMC build over an explicit name list. A shared `_param_rvs(model, names)` iterator backs it.
- The inference paths already hold the conditioned `pm.Model` locally, so they build the template from that exact build — template and chain provably come from the same model. `_compile_for_nutpie` now returns `(compiled, pymc_build)` so `condition_on_nutpie` can do the same (Stan targets return `None` and keep the old `getattr` path).
- The public `record_template` property reports the **declared** (no-data) shapes — unchanged from `main` — and documents that data-dependent shapes resolve at inference.

**No instance cache.** An earlier revision cached the conditioned model on `self._last_conditioned_model` and had the property read it. That introduced a concurrency race: under threaded broadcast (`ThreadPoolTaskRunner` / `dispatch="thread"`), two inference calls sharing one `PyMCModel` could interleave their cache writes, so one call's template would be built from another's data. The stateless design removes that shared mutable state entirely.

## Per-build parameter set: dynamic RVs and partial conditioning

The set of parameters to infer is resolved per build by `_conditioned_param_names(model)`:

- **Dynamic RVs are rejected** with a clear `ValueError` (tracked in #232): a canonical parameter the build dropped (subtractive), or a new **non-observed** free RV the build introduced (additive). Per-variable data-dependent *shapes* remain supported (the point of this PR).
- **Partial conditioning is supported**: an observed variable the caller leaves unsupplied becomes a free RV in the conditioned build and is *inferred* — it is included in the template and the chain rather than silently dropped. Supplied observed variables are observed (not free) and excluded.

Chain extraction and the template share this one ordered name set, so columns align by construction (PR #234 generalizes that into name-keyed assembly).

## Tests (`tests/modeling/test_pymc_model.py`)

- `test_data_dependent_shape_reflects_conditioned_build` — the bare property reports the declared `(1,)`; the conditioned build reports `(N,)`; no cached state.
- `test_data_dependent_shape_inference_recovers_correct_layout` — end-to-end `condition_on(method="pymc_nuts")`; asserts `draws["alpha"].shape == (num_results, N)`. Fails on `main`.
- `test_dynamic_rv_set_rejected` / `test_additive_dynamic_rv_set_rejected` — both directions of a changed RV *set* raise the clear `ValueError`.
- `test_partial_conditioning_includes_unsupplied_observed` / `test_partial_conditioning_via_inference` — an unsupplied observed variable is inferred and present in the posterior.
- nutpie helper tests updated for the `(compiled, pymc_build)` return.

## Test plan

- [x] `pytest tests/modeling/test_pymc_model.py tests/inference/test_nutpie.py` — 38 passed
- [x] `pytest tests/inference/ tests/modeling/` — 395 passed, 4 skipped
- [ ] CI green

### DIFF
--- a/probpipe/inference/_nutpie.py
+++ b/probpipe/inference/_nutpie.py
@@ -47,7 +47,7 @@ def condition_on_nutpie(
             "Install it with: pip install nutpie"
         ) from e
 
-    compiled = _compile_for_nutpie(model, data)
+    compiled, pymc_build = _compile_for_nutpie(model, data)
     trace = nutpie.sample(
         compiled, draws=num_results, tune=num_warmup,
         chains=num_chains, seed=random_seed, **kwargs,
@@ -55,9 +55,16 @@ def condition_on_nutpie(
 
     chains, param_names = _extract_chains(trace, num_chains)
 
+    # Build the template from the same conditioned model nutpie sampled
+    # (PyMC only); Stan models carry their own record_template, if any.
+    if pymc_build is not None:
+        record_template = model.record_template_for(pymc_build)
+    else:
+        record_template = getattr(model, "record_template", None)
+
     return make_posterior(
         chains, parents=(model,), algorithm="nutpie_nuts",
-        auxiliary=trace, record_template=getattr(model, "record_template", None),
+        auxiliary=trace, record_template=record_template,
         num_results=num_results, num_warmup=num_warmup, num_chains=num_chains,
     )
 
@@ -67,15 +74,22 @@ def condition_on_nutpie(
 # ---------------------------------------------------------------------------
 
 
-def _compile_for_nutpie(model: Any, data: Any) -> Any:
-    """Compile a model for nutpie sampling."""
+def _compile_for_nutpie(model: Any, data: Any) -> tuple[Any, Any]:
+    """Compile a model for nutpie sampling.
+
+    Returns ``(compiled, pymc_build)``. ``pymc_build`` is the
+    data-conditioned ``pm.Model`` for PyMCModel targets (so the caller
+    can derive a matching ``record_template``), and ``None`` for Stan
+    targets.
+    """
     if hasattr(model, "_bridgestan_model"):
         import nutpie
-        return nutpie.compile_stan_model(model._bridgestan_model(data=data))
+        return nutpie.compile_stan_model(model._bridgestan_model(data=data)), None
 
     if hasattr(model, "_pymc_model"):
         import nutpie
-        return nutpie.compile_pymc_model(model._pymc_model(data=data))
+        pymc_build = model._pymc_model(data=data)
+        return nutpie.compile_pymc_model(pymc_build), pymc_build
 
     raise TypeError(
         f"condition_on_nutpie does not support {type(model).__name__}. "
--- a/probpipe/inference/_nutpie.py
+++ b/probpipe/inference/_nutpie.py
@@ -48,14 +48,17 @@ def condition_on_nutpie(
 
     compiled, pymc_build = _compile_for_nutpie(model, data)
 
-    # Build the template from the same conditioned model before sampling,
-    # so a non-concrete or dynamic-RV model fails fast with a clear error
-    # rather than an opaque KeyError during chain extraction. PyMC derives
-    # it from the conditioned build; Stan models carry their own
+    # Resolve the PyMC param names + template from the same conditioned
+    # build before sampling, so a dynamic-RV or non-concrete model fails
+    # fast with a clear error rather than an opaque KeyError during chain
+    # extraction. The names cover canonical params plus any observed left
+    # free under partial conditioning. Stan models carry their own
     # record_template, if any.
     if pymc_build is not None:
-        record_template = model._record_template_for(pymc_build)
+        keep_names = list(model._conditioned_param_names(pymc_build))
+        record_template = model._record_template_for(pymc_build, keep_names)
     else:
+        keep_names = None
         record_template = getattr(model, "record_template", None)
 
     trace = nutpie.sample(
@@ -63,11 +66,10 @@ def condition_on_nutpie(
         chains=num_chains, seed=random_seed, **kwargs,
     )
 
-    # PyMC: extract chain columns in ``_param_names`` order so they line
-    # up with the template (nutpie sorts ``posterior.data_vars``
+    # PyMC: extract chain columns in `keep_names` order so they line up
+    # with the template (nutpie sorts ``posterior.data_vars``
     # alphabetically, which would otherwise mislabel draws). Stan: take
     # trace order.
-    keep_names = list(model._param_names) if pymc_build is not None else None
     chains, param_names = _extract_chains(trace, num_chains, keep_names=keep_names)
 
     return make_posterior(
--- a/probpipe/inference/_nutpie.py
+++ b/probpipe/inference/_nutpie.py
@@ -48,11 +48,9 @@ def condition_on_nutpie(
 
     compiled, pymc_build = _compile_for_nutpie(model, data)
 
-    # Resolve the PyMC param names + template from the same conditioned
-    # build before sampling, so a dynamic-RV or non-concrete model fails
-    # fast with a clear error rather than an opaque KeyError during chain
-    # extraction. The names cover canonical params plus any observed left
-    # free under partial conditioning. Stan models carry their own
+    # Resolve the PyMC inferred names + template from the conditioned
+    # build before sampling (fail fast on a dynamic-RV / non-concrete
+    # model); extraction shares the names. Stan models carry their own
     # record_template, if any.
     if pymc_build is not None:
         keep_names = list(model._conditioned_param_names(pymc_build))
@@ -120,12 +118,9 @@ def _extract_chains(
         Number of chains to extract.
     keep_names : list of str or None
         If given, extract exactly these variables, in this order, instead
-        of ``posterior.data_vars`` order. nutpie orders ``data_vars``
-        alphabetically, so PyMC callers pass the model's ``_param_names``
-        to keep chain-column order aligned with the ``record_template``
-        field order (otherwise the flat chain is split positionally and
-        draws are silently mislabeled). Stan callers pass ``None`` and
-        take trace order.
+        of ``posterior.data_vars`` order (which nutpie sorts
+        alphabetically). PyMC callers pass the param names so chain
+        columns align with the template; Stan callers pass ``None``.
 
     Returns
     -------
--- a/probpipe/inference/_nutpie.py
+++ b/probpipe/inference/_nutpie.py
@@ -5,12 +5,11 @@ from __future__ import annotations
 import logging
 from typing import Any
 
-import jax.numpy as jnp
-
 from ..core._registry import MethodInfo
 from ..core.node import workflow_function
 from ..custom_types import ArrayLike
 from ._approximate_distribution import ApproximateDistribution, make_posterior
+from ._inference_utils import extract_chain_columns
 from ._registry import InferenceMethod
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


probpipe/modeling/_pymc.py
```diff
@@ -6,6 +6,7 @@
 from __future__ import annotations
 
 import logging
+from collections.abc import Iterator
 from typing import Any, Callable
```
SoymilkWinsAgain: I remember that the style guide says `Callable` be imported from `collections.abc` too

jhuggins: Good catch — fixed in 818c8d2. Merged it into the adjacent `collections.abc` import: `from collections.abc import Callable, Iterator`, leaving `from typing import Any`. (STYLE_GUIDE §4.5.)

jhuggins:
## Review fixes (commit 818c8d2)

Addressed both review comments from @SoymilkWinsAgain:

**1. `Callable` import** — moved to `from collections.abc import Callable, Iterator` per STYLE_GUIDE §4.5. (Replied inline.)

**2. Chain-column order vs template field order** (the inline thread appears to have been removed, so recording the resolution here) — this was a real, confirmed silent-mislabeling bug, and I've fixed it:

- **Cause:** nutpie sorts `posterior.data_vars` **alphabetically**. `_extract_chains` concatenated columns in that order, while `record_template_for` builds fields in `_param_names` (declaration) order. `ApproximateDistribution` splits the flat chain positionally by template order, so for a model declaring `zeta, alpha, mu`, `draws["zeta"]` returned alpha's samples, etc.
- **Scope:** only the **nutpie** path. The PyMC-native path (`pymc_nuts`/`pymc_advi`) already extracts by name in `_param_names` order, so it was correct. The bug is **pre-existing** (not introduced by this PR) — `main` had the same alphabetical-vs-declaration mismatch — but this PR made the ordering contract explicit, which surfaced it.
- **Fix:** `_extract_chains` gains `keep_names=`; `condition_on_nutpie` passes `model._param_names` for PyMC so chain columns and template fields share one ordering. Stan passes `None` and keeps trace order. The nutpie path now structurally matches the PyMC-native one.
- **Tests:** a mock unit test (`keep_names` overrides `data_vars` order) and an end-to-end nutpie test declaring `zeta/alpha/mu` (non-alphabetical) with distinct priors, asserting each field recovers its own prior mean — verified to **fail before the fix**, pass after.

Full suite: `pytest tests/inference/test_nutpie.py tests/modeling/test_pymc_model.py` → 34 passed.

probpipe/modeling/_pymc.py
SoymilkWinsAgain: `_param_rvs()` silently skips missing canonical params in the conditioned build and also ignores free RVs added in the conditioned build.

probpipe/modeling/_pymc.py
```diff
@@ -142,15 +142,13 @@ def _param_rvs(self, model: Any) -> Iterator[tuple[str, Any]]:
...
-        return (total,)
+        """Total number of scalar free parameters (observed excluded).
+
+        Derived from :attr:`record_template` so the flat size and the
+        per-field template can never disagree; consequently this raises
+        on a non-concrete free-RV shape, just as the template does.
+        """
+        return (self.record_template.flat_size,)
 
     def record_template_for(self, model: Any) -> NumericRecordTemplate:
```
SoymilkWinsAgain: Is it a public method? If so, NumPy-style Parameters / Returns should be added. Or it should be named as `_record_temple_for()`

probpipe/inference/_nutpie.py
```diff
@@ -48,25 +47,29 @@ def condition_on_nutpie(
...
         compiled, draws=num_results, tune=num_warmup,
         chains=num_chains, seed=random_seed, **kwargs,
     )
 
     # PyMC: extract chain columns in ``_param_names`` order so they line
-    # up with ``record_template_for`` (nutpie sorts ``posterior.data_vars``
+    # up with the template (nutpie sorts ``posterior.data_vars``
     # alphabetically, which would otherwise mislabel draws). Stan: take
     # trace order.
     keep_names = list(model._param_names) if pymc_build is not None else None
```
SoymilkWinsAgain: From CodeX:

`_extract_chains()` concatenates columns in the order of `posterior.data_vars`, but the PyMC template now splits columns in the order of `model._param_names` / `record_template_for()`. `ApproximateDistribution` interprets the flat chain according to the template field order; if the ArviZ/nutpie variable order is inconsistent with `_param_names`, it will silently assign parameter values ​​to the wrong fields.

Recommendation: Explicitly extract the chain using `model._param_names` for PyMC/nutpie paths, or at least assert that the order of `posterior.data_vars` matches the template field order.

tests/modeling/test_pymc_model.py
SoymilkWinsAgain, line 209: This seems to conflict with the PR: the title and body still say to cache `_last_conditioned_model`, but here it explicitly requires `not hasattr(model, "_last_conditioned_model")`.

SoymilkWinsAgain (changes requested):
Sorry, I went out just now and didn't finish writing. Here is the complete review. Including a suggestion from AI (just for reference, as there are many codes involved, and I can't quickly verify it locally).

probpipe/modeling/_pymc.py
jhuggins: Both directions are now handled (463a0cc). The subtractive case — a canonical parameter missing from the conditioned build — already raised in `_param_rvs`. The additive case you flag was a real silent gap: a free RV introduced *only* in the conditioned build isn't in `_param_names`, and since extraction filters to `_param_names` it was dropped from both the template and the chain — so the user's parameter vanished from the posterior with no error. `_param_rvs` now also raises on that, so both directions fail fast with the #232 message. Added `test_additive_dynamic_rv_set_rejected`. (The unconditioned path is unaffected: there the free RVs are exactly `_param_names` ∪ observed names, so the check is a no-op.)

probpipe/modeling/_pymc.py
jhuggins: Renamed to `_record_template_for` (private) in 463a0cc. It takes a raw `pm.Model` build and is only called by the PyMC inference backends, so it's an internal seam rather than user API — the user-facing surface is the `record_template` property. The NumPy-style `Parameters`/`Returns` sections are retained on it.

probpipe/inference/_nutpie.py
jhuggins: Already addressed (thanks, and thanks CodeX). `condition_on_nutpie` extracts with `keep_names=list(model._param_names)` and `_extract_chains(..., keep_names=keep_names)` (now at `_nutpie.py:70-71`) — i.e. exactly the recommended fix: chain columns are pulled in `_param_names` (= template field) order rather than nutpie's alphabetical `posterior.data_vars`. That landed in 818c8d2.

PR #234 generalizes your fallback ("assert the orders match") into a structural guarantee: `ApproximateDistribution`/`make_posterior` gain a `field_order` parameter so columns are aligned to template fields by name regardless of the backend's column order. Once that merges, the per-caller `keep_names` pre-sort is belt-and-suspenders rather than load-bearing.

tests/modeling/test_pymc_model.py
jhuggins: The title and body were updated to the final stateless design — they no longer describe caching `_last_conditioned_model` (the body now has a paragraph explaining the cache was *removed* to fix a concurrency race). So the `not hasattr(model, "_last_conditioned_model")` assertion is consistent with the PR: it's a regression guard against reintroducing that cache.

probpipe/modeling/_pymc.py
SoymilkWinsAgain: What's the use of this? It appears to have never been used.

SoymilkWinsAgain: To allow no-data builds, `_param_rvs()` excludes `self._observed_names` from the extra free RV check. However, if the user misses an observed name when passing in a dict/record, that variable will become `free_RV` in the conditioned build, and because it belongs to `_observed_names`, it will be excluded from `_param_names`, template, and chain extraction. The result is that the sampler might sample this variable, but the field is silently dropped in the posterior.

probpipe/modeling/_pymc.py
```diff
@@ -150,7 +172,7 @@ def event_shape(self) -> tuple[int, ...]:
         """
         return (self.record_template.flat_size,)
 
-    def record_template_for(self, model: Any) -> NumericRecordTemplate:
+    def _record_template_for(self, model: Any) -> NumericRecordTemplate:
```
SoymilkWinsAgain: The function name is now right, but the descriptions in PR haven't been changed.

SoymilkWinsAgain (changes requested):

probpipe/modeling/_pymc.py
jhuggins: Removed in d0f9cd4 — `all_free` was computed and never read. Good catch.

jhuggins: Good catch, and agreed it shouldn't error — partial conditioning is now **supported** (d0f9cd4). The fix isn't to exclude that case but to *infer* it: the per-build parameter set is resolved by a new `_conditioned_param_names(model)` that returns the canonical params **plus any observed names left free** in the conditioned build. So if you condition on a subset and leave an observed variable unsupplied, it becomes a free RV and is included in both the template and the chain — it shows up in the posterior instead of being silently dropped.

True dynamic RVs are still rejected: a canonical param the build dropped, or a new *non-observed* free RV it introduced. Observed-left-free is explicitly allowed.

Added `test_partial_conditioning_includes_unsupplied_observed` and `test_partial_conditioning_via_inference` (condition on `y` only for a model with observed `(X, y)` → posterior has both `mu` and `X`).

probpipe/modeling/_pymc.py
jhuggins: Updated in d0f9cd4 — the PR title and body now say `_record_template_for` and describe the per-build name resolution (dynamic-RV rejection + partial-conditioning support).

SoymilkWinsAgain (approved):
Everything looks fine!

SoymilkWinsAgain (approved):
This round only involves modifying the comments. Everything looks fine.

Issue #233, opened 2026-06-05
Name-keyed posterior assembly: align chain columns to template fields by name, not position


jhuggins:
## Summary

`ApproximateDistribution` assembles a posterior by splitting the concatenated flat chain into named fields **positionally** — by `record_template.fields` order and per-field flat sizes (`_approximate_distribution.py` ~L110–122), with only a total-flat-size sanity check (L104). It never aligns columns to fields **by name**. Correctness therefore depends on an unwritten contract: *every caller must concatenate chain columns in exactly `record_template.fields` order.*

That contract is currently honored by per-caller discipline:
- PyMC-native (`pymc_nuts`/`pymc_advi`): pass `dist._param_names` to both extraction and the template.
- nutpie: pass `keep_names=model._param_names` to extraction (added in #231 after this exact mismatch shipped — nutpie sorts `posterior.data_vars` alphabetically, silently mislabeling draws for non-alphabetical parameter declarations).

The nutpie bug (#231) proves the fragility: a new or changed backend that orders variables differently re-arms the silent-mislabeling trap, and the only guard (total-size check) passes whenever the sizes happen to sum correctly.

## Proposed fix

Make posterior assembly **name-keyed** so column order from any backend is irrelevant:

1. Thread per-column field names from extraction (where names exist) through `make_posterior` to `ApproximateDistribution`. Concretely, add an optional `column_fields: list[str] | None` (one entry per flat column, naming the field it belongs to) to `ApproximateDistribution.__init__` / `make_posterior`.
2. When `column_fields` is provided alongside a multi-field `record_template`, assemble each template field by selecting its columns by name (in template order), rather than slicing positionally. Raise a clear error if a template field has no matching columns or counts disagree.
3. Keep the existing positional path as the default (`column_fields=None`), so every non-PyMC backend (TFP, BlackJAX NUTS/RWMH/ESS/SGMCMC, sbijax) is unchanged.
4. Route the three PyMC paths through the named path; `extract_chain_columns` already iterates a name list and can emit the matching `column_fields`.

The positional total-size check then becomes a redundant assertion rather than the only thing between a reordering backend and scrambled draws. The per-caller `keep_names` pre-sorting in #231 can stay (harmless) or be retired once assembly is name-safe.

### Tradeoffs

- Additive optional parameter keeps the blast radius to the assembly layer + PyMC call sites; other backends opt in later if desired.
- Adds a small amount of per-field bookkeeping at assembly time (negligible — once per inference run).
[...]

PR #234, opened 2026-06-05
Name-keyed posterior assembly via field_order (mechanism)

jhuggins:
Refs #233.

## Problem

`ApproximateDistribution` splits the concatenated flat chain into template fields **positionally** — by `record_template.fields` order and per-field flat sizes, with only a total-flat-size sanity check. Correctness has rested on an unwritten contract: *every caller must concatenate chain columns in `record_template.fields` order.* A backend whose trace orders variables differently (nutpie sorts `posterior.data_vars` alphabetically) silently mislabels draws unless the caller pre-sorts — exactly the bug fixed per-caller in #231.

## Change (mechanism only)

Adds an optional `field_order: list[str] | None` to `ApproximateDistribution.__init__` and `make_posterior`. It names the field each contiguous column-block of the chain occupies. When provided, the flat chain's columns are permuted into `record_template` order (new `_column_permutation` helper) **before** the existing positional split, so columns are aligned to fields by name instead of by position.

- **Default `field_order=None` is unchanged** — every current backend keeps the positional contract byte-for-byte (the reorder is a no-op preprocessing step that only runs when `field_order` is passed).
- Reordering the flat columns up front (rather than building a per-field Record) means all downstream consumers — `draws()`'s `unflatten`, `mean`, the `chains` property — see template-ordered columns and stay correct.

## Validation (when `field_order` is given)

Hardened so misuse fails loudly rather than silently corrupting draws:

- **Requires a template** — `field_order` without `record_template` raises (it names template fields).
- **Permutation-checked for any field count** — a wrong/typo'd name raises, including for single-field templates.
- **Chain width validated before the gather, for any field count** — `c[..., perm]` would otherwise silently drop extra columns (too-wide chain) or clamp out-of-bounds gather indices (too-narrow), then pass the post-gather size check.
- `ApproximateDistribution`'s `Parameters` docstring now documents `record_template` and `field_order` (it's public via mkdocstrings).

## Scope / follow-up

This PR lands the **assembly mechanism + validation** only. Wiring the PyMC inference paths to pass `field_order` (retiring the #231 `keep_names` pre-sort) is **#236**, stacked on this. Branched off `main`.

## Tests (`tests/inference/test_inference.py`)

- `test_field_order_reassembles_by_name` — non-template column order is reassembled by name.
- `test_field_order_none_is_positional` — default path matches the historical layout.
- `test_field_order_must_be_permutation` / `test_field_order_single_field_invalid_permutation_raises` — non-permutation raises (multi- and single-field).
- `test_field_order_chain_too_wide_raises` / `test_field_order_chain_too_narrow_raises` / `test_field_order_single_field_width_mismatch_raises` — width mismatches raise.
- `test_field_order_without_template_raises` — `field_order` without a template raises.

## Test plan

- [x] `pytest tests/inference/test_inference.py` — 95 passed (incl. positional backward-compat)
- [x] `pytest tests/inference/ tests/modeling/` — green
- [ ] CI green

### DIFF
--- a/probpipe/inference/_approximate_distribution.py
+++ b/probpipe/inference/_approximate_distribution.py
@@ -91,6 +91,20 @@ class ApproximateDistribution(RecordEmpiricalDistribution):
         Optional per-sample importance weights (across all chains).
     name : str or None
         Distribution name for provenance.
+    record_template : RecordTemplate or None
+        If given, names the posterior's fields: the concatenated chain is
+        split into per-field arrays (multi-field) so :meth:`draws`,
+        :meth:`_mean` / :meth:`_variance`, etc. return Records keyed by
+        the template fields. ``None`` leaves the posterior a single
+        unnamed numeric block.
+    field_order : list of str or None
+        Names the field each contiguous column-block of *chains* belongs
+        to, in the order they appear. Default (``None``) assumes the
+        columns are already in ``record_template.fields`` order. Pass this
+        when the chain's column order may differ (e.g. a backend that
+        sorts variable names) so columns are aligned to fields by name
+        rather than position. Requires *record_template*, and must be a
+        permutation of its fields.
 
     Notes
     -----
@@ -138,17 +152,17 @@ class ApproximateDistribution(RecordEmpiricalDistribution):
             # fields (raises otherwise) for any field count, so a
             # single-field typo or wrong name is also caught.
             perm = _column_permutation(record_template, field_order)
+            # Validate width for any field count, *before* the gather:
+            # ``c[..., perm]`` would otherwise silently drop extra columns
+            # (too-wide chain) or clamp out-of-bounds indices (too-narrow),
+            # then pass the post-gather total-size check.
+            for c in self._chains:
+                if c.shape[-1] != len(perm):
+                    raise ValueError(
+                        f"chain last dim ({c.shape[-1]}) doesn't match "
+                        f"the template total flat size ({len(perm)})."
+                    )
             if len(record_template.fields) > 1:
-                # Validate width *before* the gather: ``c[..., perm]``
-                # would otherwise silently drop extra columns (too-wide
-                # chain) or clamp out-of-bounds indices (too-narrow),
-                # then pass the post-gather total-size check.
-                for c in self._chains:
-                    if c.shape[-1] != len(perm):
-                        raise ValueError(
-                            f"chain last dim ({c.shape[-1]}) doesn't match "
-                            f"the template total flat size ({len(perm)})."
-                        )
                 self._chains = [c[..., perm] for c in self._chains]
                 self._concatenated = None
 
--- a/probpipe/inference/_approximate_distribution.py
+++ b/probpipe/inference/_approximate_distribution.py
@@ -128,14 +128,29 @@ class ApproximateDistribution(RecordEmpiricalDistribution):
         # order. The positional split below (and ``draws()`` unflatten)
         # then map each column to the right field by name rather than by
         # position, so callers don't have to pre-sort. See issue #233.
-        if (
-            field_order is not None
-            and record_template is not None
-            and len(record_template.fields) > 1
-        ):
+        if field_order is not None:
+            if record_template is None:
+                raise ValueError(
+                    "field_order requires a record_template; it names "
+                    "template fields and is meaningless without one."
+                )
+            # Validates field_order is a permutation of the template
+            # fields (raises otherwise) for any field count, so a
+            # single-field typo or wrong name is also caught.
             perm = _column_permutation(record_template, field_order)
-            self._chains = [c[..., perm] for c in self._chains]
-            self._concatenated = None
+            if len(record_template.fields) > 1:
+                # Validate width *before* the gather: ``c[..., perm]``
+                # would otherwise silently drop extra columns (too-wide
+                # chain) or clamp out-of-bounds indices (too-narrow),
+                # then pass the post-gather total-size check.
+                for c in self._chains:
+                    if c.shape[-1] != len(perm):
+                        raise ValueError(
+                            f"chain last dim ({c.shape[-1]}) doesn't match "
+                            f"the template total flat size ({len(perm)})."
+                        )
+                self._chains = [c[..., perm] for c in self._chains]
+                self._concatenated = None
 
         flat = self._concat_chains()
         # Track whether the user explicitly supplied a template; we use
--- a/probpipe/inference/_approximate_distribution.py
+++ b/probpipe/inference/_approximate_distribution.py
@@ -20,6 +20,56 @@ __all__ = ["ApproximateDistribution", "make_posterior"]
 
 
 # ---------------------------------------------------------------------------
+# Column ordering
+# ---------------------------------------------------------------------------
+
+
+def _column_permutation(
+    record_template: RecordTemplate, field_order: list[str],
+) -> list[int]:
+    """Column-index permutation mapping a *field_order*-laid-out flat chain
+    into ``record_template.fields`` order.
+
+    *field_order* names the field each contiguous column-block of the flat
+    chain occupies. The returned ``perm`` satisfies: ``flat[..., perm]``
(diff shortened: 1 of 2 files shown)


probpipe/inference/_approximate_distribution.py
```diff
@@ -64,13 +114,29 @@ def __init__(
         weights: ArrayLike | Weights | None = None,
         name: str | None = None,
         record_template: RecordTemplate | None = None,
+        field_order: list[str] | None = None,
     ):
         if not chains:
             raise ValueError("Must provide at least one chain")
 
         self._chains = [jnp.asarray(c) for c in chains]
```
SoymilkWinsAgain, line 136: The later on code checks if `flat.shape[-1] == total`. If the original chain here has an extra column, `c[..., perm]` will directly select 3 columns and discard the extra column, and subsequent validation will pass. This will silently drop data.

Before reordering, it should be verified that the last dimension of each original chain equals `len(perm) / template total size` before executing the index.

probpipe/inference/_approximate_distribution.py
```diff
@@ -64,13 +114,29 @@ def __init__(
...
         self._concatenated: Array | None = None
 
+        # When the caller's chain columns are laid out in a different
+        # field order than the template — e.g. a backend whose trace
+        # sorts variable names — permute them into ``record_template``
+        # order. The positional split below (and ``draws()`` unflatten)
+        # then map each column to the right field by name rather than by
+        # position, so callers don't have to pre-sort. See issue #233.
+        if (
+            field_order is not None
```
SoymilkWinsAgain: `field_order` is silently ignored when `record_template is None` or when it's a single-field template. Passing this parameter indicates that the caller considers field semantics to exist; a `ValueError` should be thrown if no template is provided. Single-field templates should also validate permutations.

tests/inference/test_inference.py
```diff
@@ -182,6 +182,55 @@ def test_without_template_returns_array(self):
...
+        """A field_order that isn't a permutation of template fields raises."""
+        template = RecordTemplate(a=(), b=())
+        chain = jax.random.normal(jax.random.PRNGKey(0), (5, 2))
+        prior = MultivariateNormal(loc=jnp.zeros(2), cov=jnp.eye(2), name="z")
+        with pytest.raises(ValueError, match="not a permutation"):
+            make_posterior(
+                [chain], parents=(prior,), algorithm="test",
+                record_template=template, field_order=["a", "c"],
+            )
+
```
SoymilkWinsAgain, line 233: The test is missing a few direct coverages: 
1. field_order + excessively wide chain should raise an error;
2.  excessively narrow chain should give a clear ValueError; 
3. field_order without record_template should raise an error; 
4. invalid field_order for a single field template should raise an error.

SoymilkWinsAgain (changes requested):

probpipe/inference/_approximate_distribution.py
jhuggins: Fixed in b73f53a. You're right — and it's actually worse for the too-narrow case: JAX gather *clamps* out-of-bounds indices rather than raising, so a too-narrow chain would duplicate the last column silently. Both are masked because the gather always yields exactly `len(perm) == total` columns, so the later size check passes.

Now each chain's last dim is validated against `len(perm)` (the template total) *before* the gather, raising a clear `ValueError` on either too-wide or too-narrow. Covered by `test_field_order_chain_too_wide_raises` and `test_field_order_chain_too_narrow_raises`.

probpipe/inference/_approximate_distribution.py
jhuggins: Fixed in b73f53a. The guard is restructured so that whenever `field_order` is given it's validated regardless of field count:

- `field_order` + `record_template is None` → `ValueError` (it names template fields and is meaningless without one).
- The permutation is validated via `_column_permutation` for *any* field count, so a single-field wrong/typo'd name now raises instead of being silently ignored.
- The column reorder + width check still run only for multi-field templates (a single field needs no reorder).

Covered by `test_field_order_without_template_raises` and `test_field_order_single_field_invalid_permutation_raises`.

tests/inference/test_inference.py
jhuggins: Added all four in b73f53a:
1. `test_field_order_chain_too_wide_raises`
2. `test_field_order_chain_too_narrow_raises`
3. `test_field_order_without_template_raises`
4. `test_field_order_single_field_invalid_permutation_raises`

(The existing `test_field_order_must_be_permutation` already covers the multi-field invalid-permutation case.) Full inference suite: 284 passed, 1 skipped.

probpipe/inference/_approximate_distribution.py
```diff
@@ -64,13 +114,44 @@ def __init__(
...
+            if record_template is None:
+                raise ValueError(
+                    "field_order requires a record_template; it names "
+                    "template fields and is meaningless without one."
+                )
+            # Validates field_order is a permutation of the template
+            # fields (raises otherwise) for any field count, so a
+            # single-field typo or wrong name is also caught.
+            perm = _column_permutation(record_template, field_order)
+            if len(record_template.fields) > 1:
```
SoymilkWinsAgain, line 165: From CodeX:

The width check is still placed inside `if len(record_template.fields) > 1:`. This way, single-field templates only check the field name permutation, not whether the last dimension of the chain equals the template's flat size.

Example: `RecordTemplate(a=(2,)) + chain.shape == (5, 3) + field_order=["a"]` will pass the construction; subsequently, `draws()` will unflatten columns according to the template, consuming only the first two columns, and extra columns may be silently discarded. This is the same type of data integrity issue as the recently fixed multi-field issue where "extra columns are discarded during gather".

Recommendation: If `field_order` is not None and `_column_permutation(...)` succeeds, perform a `c.shape[-1] == len(perm)` check on each chain; the reordering itself can continue to be performed only in multi-field cases.

probpipe/inference/_approximate_distribution.py
SoymilkWinsAgain, line 77: From CodeX:

`ApproximateDistribution` is a public class, exposed in the documentation via `::: probpipe.ApproximateDistribution`. However, the class documentation only lists `chains/weights/name` in its parameters, omitting the existing `record_template` and the newly added `field_order`. The `make_posterior` documentation already includes `field_order`, and the constructor should also be synchronized, specifically specifying that it is a "contiguous field-block order", not an arbitrary per-column map.

tests/inference/test_inference.py
```diff
@@ -182,6 +182,104 @@ def test_without_template_returns_array(self):
     def test_record_template_property(self, posterior_with_template, template):
         assert posterior_with_template.record_template is template
 
+    def test_field_order_reassembles_by_name(self):
```
SoymilkWinsAgain, line 185: It is recommended to update the PR description to include the newly added scope.

SoymilkWinsAgain (changes requested):

probpipe/inference/_approximate_distribution.py
jhuggins: Fixed in 884b80c — good catch (and CodeX). Moved the width check out of the `if len(record_template.fields) > 1:` guard so it runs for any field count; a single-field template with `field_order` now also validates that the chain's last dim equals the template's flat size (`len(perm)`). The reorder itself stays multi-field-only (a single field's permutation is the identity). Added `test_field_order_single_field_width_mismatch_raises`.

probpipe/inference/_approximate_distribution.py
jhuggins: Fixed in 884b80c. Added `record_template` and `field_order` to the `ApproximateDistribution` class docstring `Parameters` (both were missing; `record_template` predates this PR). They now render under `::: probpipe.ApproximateDistribution`.

tests/inference/test_inference.py
jhuggins: Will do — updating the PR description now to cover the added validation scope (field_order width check, no-template error, single-field validation).

SoymilkWinsAgain (approved):
Looks good!

PR #236, opened 2026-06-05
Wire PyMC paths through name-keyed field_order (closes #233)

### LINKED ISSUE #233: Name-keyed posterior assembly: align chain columns to template fields by name, not position

jhuggins:
Closes #233. **Stacked on #234** — until that merges, the diff below includes #234's `field_order` mechanism commits; review the wiring commit (`69545af`) or merge #234 first.

## What

#234 added the name-keyed assembly *mechanism* (`field_order` on `ApproximateDistribution` / `make_posterior`). This routes the three PyMC inference paths (`nutpie_nuts`, `pymc_nuts`, `pymc_advi`) through it, completing the #233 goal: **column order from any backend is now structurally safe**, retiring the per-caller `keep_names` pre-sort introduced in #231.

Each PyMC path now:
- builds the `record_template` in canonical field order (`_conditioned_param_names`);
- extracts chain columns in the trace's **natural** variable order (new `posterior_var_order` helper — nutpie sorts `data_vars` alphabetically);
- passes that order as `make_posterior(field_order=...)`, so `ApproximateDistribution` realigns columns to the template fields **by name**.

Previously (#231) the extraction was force-sorted into template order via `keep_names`; correctness relied on that per-caller discipline. Now the positional split only ever sees columns already realigned by name, so a backend whose trace orders variables differently can't silently mislabel draws. `posterior_var_order` still filters to the param set, preserving the drop of any transformed/observed vars.

## Why this is the structural fix

The nutpie alphabetical-`data_vars` bug (#231) showed the fragility of "every caller must pre-sort columns into template order." This makes the alignment a property of the assembly layer (by name) rather than caller discipline.

## Tests

No new tests required — the existing value tests already exercise this and are now genuine `field_order` regression guards (extraction is now natural-order, so they fail without the wiring). Verified: removing `field_order` from the nutpie path makes `test_multiparam_draws_not_mislabeled` fail (alphabetical extraction vs canonical template), and restoring it passes. The mechanism itself is unit-tested in #234 (`test_field_order_*`).

## Test plan

- [x] `pytest tests/inference/ tests/modeling/` — 405 passed, 4 skipped
- [x] Confirmed `field_order` wiring is load-bearing (revert → mislabel test fails)
- [ ] CI green (after #234 merges / base settles)

### DIFF
--- a/probpipe/inference/_inference_utils.py
+++ b/probpipe/inference/_inference_utils.py
@@ -114,9 +114,26 @@ def posterior_var_order(trace: Any, keep: Iterable[str]) -> list[str]:
     (:func:`make_posterior`), so name-keyed assembly realigns columns to
     the template regardless of the backend's variable order — nutpie, for
     instance, sorts ``data_vars`` alphabetically.
+
+    Raises
+    ------
+    ValueError
+        If any name in *keep* is absent from ``trace.posterior``. Failing
+        here names the missing variables, rather than letting them surface
+        later as a cryptic "not a permutation" error in
+        :func:`make_posterior`'s column realignment.
     """
-    keep = set(keep)
-    return [name for name in trace.posterior.data_vars if name in keep]
+    keep = list(keep)
+    available = list(trace.posterior.data_vars)
+    missing = [name for name in keep if name not in available]
+    if missing:
+        raise ValueError(
+            f"trace posterior is missing expected variable name(s) "
+            f"{missing}; available posterior variables are {available}. "
+            f"Every parameter being assembled must be present in the trace."
+        )
+    keep_set = set(keep)
+    return [name for name in available if name in keep_set]
 
 
 # ---------------------------------------------------------------------------
--- a/probpipe/inference/_inference_utils.py
+++ b/probpipe/inference/_inference_utils.py
@@ -23,7 +23,7 @@ utilities shared across the backend modules; not re-exported through
 
 from __future__ import annotations
 
-from collections.abc import Callable
+from collections.abc import Callable, Iterable
 from typing import TYPE_CHECKING, Any
 
 if TYPE_CHECKING:
@@ -50,6 +50,7 @@ __all__ = [
     "build_target_log_prob",
     "build_target_log_prob_flat",
     "extract_chain_columns",
+    "posterior_var_order",
     "get_init_state",
     "get_prior",
     "extract_record_template",
@@ -72,8 +73,9 @@ def extract_chain_columns(
     For each chain ``c`` and each variable in *names* (in that order),
     pulls ``trace.posterior[name].values[c]``, flattens trailing axes to
     2-D ``(draws, -1)``, and concatenates across variables. Columns are
-    laid out in *names* order, to match the ``record_template`` field
-    order.
+    laid out in *names* order; pass the same order as ``field_order`` to
+    :func:`make_posterior` so assembly aligns them to the template fields
+    by name.
 
     Parameters
     ----------
@@ -101,6 +103,20 @@ def extract_chain_columns(
             chain_arrays.append(jnp.asarray(vals))
         chains.append(jnp.concatenate(chain_arrays, axis=-1))
     return chains
+
+
+def posterior_var_order(trace: Any, keep: Iterable[str]) -> list[str]:
+    """Variable names in *trace*'s ``posterior`` natural order, filtered
+    to *keep*.
+
+    Pass the result as both the extraction order
+    (:func:`extract_chain_columns`) and ``field_order``
+    (:func:`make_posterior`), so name-keyed assembly realigns columns to
+    the template regardless of the backend's variable order — nutpie, for
+    instance, sorts ``data_vars`` alphabetically.
+    """
+    keep = set(keep)
+    return [name for name in trace.posterior.data_vars if name in keep]
 
 
 # ---------------------------------------------------------------------------
--- a/probpipe/inference/_pymc_method.py
+++ b/probpipe/inference/_pymc_method.py
@@ -6,7 +6,7 @@ from typing import Any
 
 from ..core._registry import MethodInfo
 from ._approximate_distribution import ApproximateDistribution, make_posterior
-from ._inference_utils import extract_chain_columns
+from ._inference_utils import extract_chain_columns, posterior_var_order
 from ._registry import InferenceMethod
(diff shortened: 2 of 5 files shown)


probpipe/inference/_pymc_method.py
```diff
@@ -111,20 +114,22 @@ def execute(self, dist: Any, observed: Any, **kwargs: Any) -> ApproximateDistrib
...
         record_template = dist._record_template_for(model, param_names)
         with model:
             approx = pm.fit(n=num_iterations, method=vi_method, random_seed=random_seed)
             trace = approx.sample(num_results)
 
-        # ADVI's approx.sample yields a single chain of `num_results` draws.
-        chains = extract_chain_columns(trace, param_names, num_chains=1)
+        # ADVI's approx.sample yields a single chain of `num_results`
+        # draws; extract in natural order and realign by name.
+        order = posterior_var_order(trace, param_names)
```
SoymilkWinsAgain, line 127: The `pymc_advi` path now also integrates the new mechanism via `posterior_var_order(..., param_names)` and `field_order=order`, but current tests only cover the `field-order realignment` of `pymc_nuts` and `nutpie_nuts`. It's recommended to add a small regression test specifically for ADVI, or create a lighter unit test using mock traces. The trace source for ADVI is `approx.sample(...)`, which is not entirely the same as for NUTS/nutpie, and deserves separate verification.

probpipe/inference/_inference_utils.py
```diff
@@ -103,6 +105,20 @@ def extract_chain_columns(
     return chains
 
 
+def posterior_var_order(trace: Any, keep: Iterable[str]) -> list[str]:
```
SoymilkWinsAgain, line 108: This shared helper needs a test, too.

From CodeX: Currently, missing variables will eventually be exposed as an error like "not a permutation" in `make_posterior` / `_column_permutation`. It's recommended to add direct testing and consider throwing a more explicit `ValueError` here, such as `trace posterior missing expected variable names`.

SoymilkWinsAgain (changes requested):

probpipe/inference/_pymc_method.py
jhuggins: Added `test_advi_field_order_realignment` (`tests/modeling/test_pymc_model.py`) in d7440b7. It runs the full `pymc_advi` path end-to-end, so the trace comes from `approx.sample(...)` rather than a NUTS run, on a mixed scalar+vector model. The names are chosen so the backend's alphabetical column order (`alpha` before `intercept`) differs from the template order (`intercept` defined first), and the shapes differ (scalar vs `(3,)`) — a positional split would shape-scramble instead of realigning by name. Asserts `draws.fields == ("intercept", "alpha")` with shapes `(25,)` and `(25, 3)`. Passing.

probpipe/inference/_inference_utils.py
jhuggins: Done in d7440b7. `posterior_var_order` now raises an explicit `ValueError` that names the missing variables (e.g. `trace posterior is missing expected variable name(s) ['sigma']; available posterior variables are [...]`) instead of letting them surface later as the cryptic "not a permutation" error in `make_posterior`/`_column_permutation`.

Added `TestPosteriorVarOrder` (`tests/inference/test_inference_utils.py`) with 5 unit tests on mock traces: trace-order (not keep-order) wins; the backend's alphabetical order is returned so `field_order` can realign by name; unkept trace vars are ignored; a missing kept name raises `ValueError`; and the message names the missing var.

SoymilkWinsAgain (approved):
Everything looks good

PR #240, opened 2026-06-08
Drop sbijax SBI backend (upgrade plan step 0)

jhuggins:
## Summary

Step 0 of the integrated **PyMC 6 / ArviZ 1.0 upgrade + SBI backend replacement** plan: **drop sbijax**. The sbijax dependency pins an older jax / jaxlib floor that blocks the rest of the ecosystem upgrade, so it comes out first as a pure removal — **no replacement code**. The SBI capability is being re-platformed onto **pyabc** (SMC-ABC), **BayesFlow** (amortized NPE / FMPE / CMPE), and **sbi** (NLE / NRE) in later steps.

This PR changes **no runtime version pins**. The jax / jaxlib `<0.9` and arviz `<1.0` caps that sbijax forced are retained here and lifted in their own isolated PRs (Steps 1–2: the jax-0.10 floor bump and the arviz-1.x ceiling lift).

## What's removed

**Public API (breaking):**

- The `[sbi]` extra (`pip install probpipe[sbi]`) and its `sbijax>=0.3.6` dependency.
- `sbi_learn_conditional` / `sbi_learn_likelihood` (top-level `probpipe` and `probpipe.inference`).
- `DirectSamplerSBIModel` (`probpipe.inference`), the `method=` selectors (`npe` / `fmpe` / `cmpe` for the direct sampler, `nle` / `nre` for the emulated-likelihood path), and the `network_factory=` hook.
- The `sbijax_smcabc` inference method (`SbiSMCABCMethod`, priority 5) and its registration.

`from probpipe import sbi_learn_conditional` now raises `ImportError` rather than returning an install-prompt stub.

**Internal / infra:**

- `probpipe/inference/_sbijax.py`.
- `tests/inference/test_sbijax.py` and the `sbi` pytest marker.
- The CI `pip install --no-deps sbijax` shims and the stale "pymc 6.0 isn't out" comment in `ci.yml`.

**Docs / notebooks:**

- Removed the now-broken `:::` autodoc directives and the stale `sbijax_smcabc` registry rows from `docs/api/inference.md`, plus the sbijax references in `CONTRIBUTING.md`, `README.md`, `STYLE_GUIDE.md`, `docs/user_guide.md`.
- `docs/examples/05_external_backends.ipynb` (CI-executed): re-platforming note added; re-executed cleanly.
[...]

PR #241, opened 2026-06-08
Bump jax/jaxlib floor to >=0.9, blackjax to >=1.4 (upgrade plan step 1)

jhuggins:
## Summary

Step 1 of the integrated **PyMC 6 / ArviZ 1.0 upgrade**: raise the core **jax / jaxlib floor to `>=0.9`** and **blackjax to `>=1.4`**. With sbijax gone (Step 0, #240), the `<0.9` jax / jaxlib cap — which existed only to keep sbijax's hard `jax==0.8.1` pin satisfiable — is lifted, and the floor moves up. Isolated as its own PR for bisectability and soak: the jax bump has the broadest RNG / numerics blast radius across every JAX backend.

> ⚠️ **Stacked on #240 (`dev/drop-sbijax`).** This PR's base is `dev/drop-sbijax`, so the diff below is *only* the Step-1 delta (3 files). **Merge after #240.** Once #240 lands on `main`, retarget this to `main` (or rebase) — at which point CI runs. **Note:** `ci.yml` only triggers on PRs targeting `main`, so no checks run while the base is `dev/drop-sbijax`; the local verification below stands in until then.

## Changes

**`pyproject.toml`:**

- `jax>=0.8.1,<0.9` → `jax>=0.9`
- `jaxlib>=0.8.1,<0.9` → `jaxlib>=0.9`
- `blackjax>=1.3` → `blackjax>=1.4` (held at `1.3` only to spare sbijax's jax 0.8.1 environment)
- `tfp-nightly` and `arviz>=0.13,<1.0` **unchanged** — the arviz `<1.0` ceiling lift and the `pymc>=6` bump land in their own isolated PRs (Steps 2–3).

**`.github/workflows/ci.yml`:** pin the verified known-good pair `jax==0.10.1` / `jaxlib==0.10.1` / `tfp-nightly==0.26.0.dev20260608` in both the test and notebook install blocks. A fresh tfp-nightly can lag jax's internal API under 0.10+ (DeprecationWarnings, occasionally hard errors); the pin keeps CI reproducible. The comment notes to refresh the pin when the nightly is purged from PyPI.

**`CHANGELOG.md`:** `### Changed` entry under `[Unreleased]`.

## Verification

Built a fresh isolated venv on Python 3.12 with `pip install -e ".[dev,nutpie]"` — byte-identical to what CI installs — which resolved to **jax 0.10.1 / jaxlib 0.10.1 / blackjax 1.5 / tfp-nightly 0.26.0.dev20260608 / numpy 2.4.6**.

- **Full suite: `2774 passed, 7 skipped, 0 failed`** (139s). The 7 skips match the jax-0.8 baseline exactly.
- `2774` *is* the CI test count. The dev machine's extra ~72 tests are pymc (29) + stan (34) + parametrization, whose backends CI does not install (`.[dev,nutpie]` has neither). Those are non-JAX (pytensor/numba, C++) and unaffected by the jax version; diffrax is unused by probpipe (0 references in source/tests).
- Every JAX-backend test — blackjax NUTS/HMC/RWMH/ESS/SGLD/SGHMC, TFP NUTS/HMC, all distributions and random functions, nutpie — ran and passed under jax 0.10.1.
- Python 3.13 wheels confirmed available for all three pinned packages (jaxlib `cp313`; jax / tfp-nightly pure-python), so the 3.13 CI leg won't fail on the pin.

The pymc/stan suites were not run locally under jax 0.10 (CI doesn't install those backends, and they're orthogonal to the jax version); the broader pymc-6 stack was separately exercised green under jax 0.10.1 during the upgrade readiness check.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -179,6 +179,11 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install -e ".[dev,nutpie]"
+          # Pin the verified jax / tfp-nightly pair. tfp-nightly tracks jax's
+          # internal API, and a fresh nightly can lag the current jax release
+          # (DeprecationWarnings, occasionally hard errors) under jax 0.10+.
+          # Refresh this pair when the nightly is purged from PyPI.
+          pip install "jax==0.10.1" "jaxlib==0.10.1" "tfp-nightly==0.26.0.dev20260608"
 
       - name: Run tests (changed files only)
         if: steps.changed.outputs.skip != 'true' && steps.changed.outputs.run_all == 'false'
@@ -387,6 +392,11 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install -e ".[dev,nutpie]"
+          # Pin the verified jax / tfp-nightly pair. tfp-nightly tracks jax's
+          # internal API, and a fresh nightly can lag the current jax release
+          # (DeprecationWarnings, occasionally hard errors) under jax 0.10+.
+          # Refresh this pair when the nightly is purged from PyPI.
+          pip install "jax==0.10.1" "jaxlib==0.10.1" "tfp-nightly==0.26.0.dev20260608"
 
       - name: Execute affected docs/examples notebooks
         if: steps.nb_check.outputs.skip != 'true'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,17 +31,20 @@ jobs:
           fi
 
           CHANGED=$(git diff --name-only --diff-filter=ACMRT "$BASE"...HEAD -- '*.py' || true)
+          CHANGED_ALL=$(git diff --name-only --diff-filter=ACMRT "$BASE"...HEAD || true)
           echo "Changed .py files:"
           echo "$CHANGED"
 
-          # If foundational files changed, run the full suite.
+          # If foundational files changed, run the full suite. Scan ALL
+          # changed files (not just .py): a dependency bump in pyproject.toml
+          # touches no .py file but must still trigger the suite.
           # Also force the full suite on direct pushes to main (post-merge
           # safety net — targeted tests are for PRs only).
           if [ "${{ github.event_name }}" = "push" ]; then
             RUN_ALL=true
           else
             RUN_ALL=false
-            for f in $CHANGED; do
+            for f in $CHANGED_ALL; do
               case "$f" in
                 pyproject.toml|setup.cfg|setup.py|tests/conftest.py|probpipe/__init__.py)
                   RUN_ALL=true ;;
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
SoymilkWinsAgain, line 33: From CodeX: Here, only *.py files are placed in CHANGED, but the foundational check for `pyproject.toml` is at :46, so the changes to `pyproject.toml` in this PR will never be seen. Subsequently, :52-55 will skip tests directly because CHANGED is empty. This PR only modifies .github/workflows/ci.yml, CHANGELOG.md, and pyproject.toml, so even if retargeting to main, the test job will not verify the JAX bump.

.github/workflows/ci.yml
```diff
@@ -179,6 +179,11 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install -e ".[dev,nutpie]"
```
SoymilkWinsAgain: Here, and in lines 394-399, we first run `pip install -e ".[dev,nutpie]"`, then pin JAX/TFP separately. This means the first step already performs a resolver transaction using unpinned metadata. It's recommended to change this to a single resolver transaction, or use a shared constraints file.

SoymilkWinsAgain (changes requested):

.github/workflows/ci.yml
jhuggins: Good catch — fixed in 8e35055. Added a separate `CHANGED_ALL` scan (all changed files, not just `*.py`) for the foundational-file check, and moved the `RUN_ALL` branch ahead of the empty-`CHANGED` skip so a `pyproject.toml`-only PR (like this one) runs the full suite instead of skipping.

.github/workflows/ci.yml
jhuggins: Fixed in 8e35055 — folded the jax/jaxlib/tfp-nightly pins into the project install so the whole tree resolves in a single transaction (`pip install -e ".[dev,nutpie]" "jax==..." "jaxlib==..." "tfp-nightly==..."`), instead of installing unpinned and re-pinning afterward. Applied to both the test job and the notebook job.

.github/workflows/ci.yml
SoymilkWinsAgain, line 222: From CodeX:

The notebook CI is skipping dependency-only PRs. The current PR is a JAX/TFP/BlackJAX dependency upgrade, but the notebook detector only looks at `docs/examples/*.ipynb` and `probpipe/`, ignoring `pyproject.toml`. Locally, I reproduced that the notebook trigger path for `origin/main...origin/dev/bump-jax` is empty; notebook jobs on GitHub pass in 5 seconds, essentially skipping the process. It's recommended that foundational changes like `pyproject.toml` trigger all example notebooks, or at least one set of JAX-heavy notebooks.

SoymilkWinsAgain (changes requested):

.github/workflows/ci.yml
jhuggins: Good catch — fixed in 64d9bd2. The notebook job had the same blind spot as the test detector: it diffed only `docs/examples/*.ipynb` and `probpipe/`, so a dependency-only PR produced an empty change set and the notebook pass was skipped in seconds. Added the matching `CHANGED_ALL` foundational scan (`pyproject.toml` / `setup.*`), evaluated ahead of the empty-`CHANGED` skip, so a foundational change runs the full notebook set.

Verified locally on `origin/main...origin/dev/bump-jax`: the old path returns an empty change set (skip), the new one flags `pyproject.toml` → executes all 11 `docs/examples` notebooks.

PR #242, opened 2026-06-08
Cut over to arviz 1.x + pymc 6; drop arviz 0.x straddle (upgrade plan step 2+3)

jhuggins:
## Summary

Combined **Step 2 (arviz `<1.0`→`<2.0` ceiling lift)** and **Step 3 (pymc-6 bump)** of the integrated upgrade into a single ecosystem cutover. Per the decision to **floor arviz at the 1.x line** rather than carry the 0.x/1.x straddle: pymc 5.x hard-caps `arviz<1.0`, so flooring `arviz>=1.1` *requires* `pymc>=6` (the first arviz-1.x-compatible pymc). The result is the §2 readiness-probe-verified stack — **pymc 6.0.1 + arviz 1.1.0 + jax 0.10.1** — and the 0.x straddle code is deleted.

> ⚠️ **Stacked on #241 (`dev/bump-jax`), which is stacked on #240.** Base is `dev/bump-jax`, so the diff is only the Step-2/3 delta. **Merge after #241.** CI (`ci.yml`) only triggers on PRs targeting `main`, so no checks run while the base is a `dev/*` branch; the local verification below stands in until this is retargeted to `main`.

## Changes

**`pyproject.toml`:**
- `arviz>=0.13,<1.0` → `arviz>=1.1,<2.0` — **drops arviz 0.x entirely**. The 1.x metapackage pulls `arviz-base` / `arviz-stats` / `arviz-plots`, so a separate `[diagnostics]` extra would be redundant and is omitted.
- `[pymc]`: `pymc>=5.28` → `pymc>=6`, plus `matplotlib` (the pymc 6 sampler progress bar imports it).

**Source — arviz bound by name, 0.x straddle removed:**
- `build_mcmc_datatree`: `arviz_base.from_dict` (positional groups dict); the `from_dict` signature probe is gone.
- `_cmdstan_method`: `arviz_base.from_cmdstanpy` (was bare `import arviz`).
- `ApproximateDistribution.warmup_samples` reads `aux.children` directly — every auxiliary is an arviz 1.x `xarray.DataTree` now (verified: the JAX-backend datatree *and* pymc 6's `return_inferencedata=True` trace are both `DataTree`s). Docstrings rebind `arviz_stats` by name.
- `_pymc.py`: `pm.sample_prior_predictive(samples=)` → `draws=` (the `samples=` kwarg was dropped in pymc 6).

[...]

PR #243, opened 2026-06-08
Reclaim pymc_nuts multi-core sampling via spawn (follow-up to #242)

jhuggins:
## Summary

Follow-up to #242 (split out per request): **`pymc_nuts` reclaims multi-core sampling**, which #242 deliberately deferred pending a fork-safety investigation.

> ⚠️ **Stacked on #242 → #241 → #240.** Base is `dev/lift-arviz`. **Merge after #242.** CI (`ci.yml`) only triggers on PRs targeting `main`, so no checks run while the base is a `dev/*` branch; the local verification below stands in until this is retargeted to `main`.

## Background

`pymc_nuts` pinned `cores=1` with the comment *"avoid os.fork() which deadlocks with JAX threads."* The hazard is real: ProbPipe's process holds live JAX worker threads, and pymc's default multiprocessing start method on Linux is **`fork`** — forking with a thread lock held hangs the worker. The deadlock is **intermittent** (a lock-held-at-fork race), which is exactly why leaving it to chance is unsafe.

## Fix (`probpipe/inference/_pymc_method.py`)

- `cores` defaults to one worker per chain, capped at `os.cpu_count()`, overridable via a `cores=` kwarg.
- **`mp_ctx="spawn"` is forced whenever `cores>1`.** Spawned workers start clean (no inherited threads), so the fork/threads deadlock cannot arise on any platform. Spawn requires the model to pickle — which PyMC models do, verified against ProbPipe's `PyMCModel`s (all 29 existing pymc tests pass, plus the new one).

## Investigation (pymc 6.0.1 / jax 0.10.1 env)

| config | result |
|---|---|
| `cores=1` (old) | PASS — 5.6 s (2 chains, 500 draws) |
| `cores=2`, `mp_ctx="spawn"` | PASS — **3.8 s** |
| `cores=2`, `mp_ctx="fork"` | passed *this run* (3.4 s) — but the deadlock is intermittent, so fork is not relied upon |

spawn is **not slower** than the old single-core path (faster here) and is deterministically safe, so multi-core is reclaimed as the default via spawn. macOS already defaults to spawn; the explicit `mp_ctx="spawn"` is what protects the Linux/CI `fork` default.

## Test

`tests/modeling/test_pymc_model.py::test_condition_on_multicore_spawn` spins up JAX's worker threads, then runs `condition_on(model, data, method="pymc_nuts", num_chains=2, cores=2)` and asserts both chains are returned — exercising the spawn path against the deadlock condition.

## Verification

Full suite **`2853 passed, 8 skipped, 0 failed`** under arviz 1.1.0 / pymc 6.0.1 / jax 0.10.1 (Python 3.12) — the existing 29 pymc tests plus the new multi-core test pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **`pymc_nuts` reclaims multi-core sampling.** The method previously
+  forced `cores=1` to avoid an `os.fork()` deadlock against JAX's worker
+  threads. It now samples one worker per chain (capped at the CPU count,
+  overridable via a `cores=` kwarg) using the **`spawn`** multiprocessing
+  start method — clean worker processes with no inherited threads, so it
+  is deadlock-free on every platform (POSIX `fork`, the deadlock-prone
+  default on Linux, is never used). Empirically `cores=2` spawn is no
+  slower than the old single-core path; a new test exercises the
+  multi-core path after spinning up JAX's threads to reproduce the
+  hazard.
+
 - **Ecosystem cutover to arviz 1.x and pymc 6 (breaking).** The core
   `arviz` pin moves `>=0.13,<1.0` → `>=1.1,<2.0`, **dropping arviz 0.x
   entirely**, and the `[pymc]` extra moves `pymc>=5.28` → `pymc>=6`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,8 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `aux.children`). `[pymc]` additionally requires `matplotlib` (the
   pymc 6 sampler progress bar imports it). Internal pymc-6 fix:
   `pm.sample_prior_predictive(samples=)` → `draws=` (the `samples=`
-  kwarg was dropped in pymc 6); `pymc_nuts` keeps `cores=1` pending a
-  dedicated fork-safety review (the multicore reclaim is deferred). The
+  kwarg was dropped in pymc 6). The
   ArviZ 1.0 defaults — credible interval 0.94 → 0.89 and HDI → ETI —
   are adopted as-is: no affected statistic is called in ProbPipe
   source, so no `rcParams` pin is needed, and a frozen-fixture suite
--- a/probpipe/inference/_pymc_method.py
+++ b/probpipe/inference/_pymc_method.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import os
 from typing import Any
 
 from ..core._registry import MethodInfo
@@ -45,6 +46,10 @@ class PyMCNutsMethod(InferenceMethod):
         num_results = kwargs.get("num_results", 1000)
         num_warmup = kwargs.get("num_warmup", 500)
         num_chains = kwargs.get("num_chains", 4)
+        # Multi-core sampling forces the "spawn" start method: this process
+        # holds live JAX threads, and pymc's POSIX-"fork" default deadlocks
+        # when a forked worker inherits a held thread lock.
+        cores = kwargs.get("cores", min(num_chains, os.cpu_count() or 1))
         random_seed = kwargs.get("random_seed", 0)
 
         model = dist._pymc_model(data=observed)
@@ -57,7 +62,8 @@ class PyMCNutsMethod(InferenceMethod):
                 draws=num_results,
                 tune=num_warmup,
                 chains=num_chains,
-                cores=1,  # avoid os.fork() which deadlocks with JAX threads
+                cores=cores,
+                mp_ctx="spawn" if cores > 1 else None,
                 random_seed=random_seed,
                 return_inferencedata=True,
             )
--- a/probpipe/inference/_pymc_method.py
+++ b/probpipe/inference/_pymc_method.py
@@ -48,7 +48,10 @@ class PyMCNutsMethod(InferenceMethod):
         num_chains = kwargs.get("num_chains", 4)
         # Multi-core sampling forces the "spawn" start method: this process
         # holds live JAX threads, and pymc's POSIX-"fork" default deadlocks
-        # when a forked worker inherits a held thread lock.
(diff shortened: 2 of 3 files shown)


tests/modeling/test_pymc_model.py
```diff
@@ -142,6 +142,31 @@ def test_condition_on(self, model):
...
+        # Spin up JAX's worker threads first, so the POSIX fork start method
+        # would be exposed to the deadlock this path avoids.
+        _ = jnp.ones(1000).sum().block_until_ready()
+
+        data = np.random.randn(50)
+        result = condition_on(
+            model, {"y": data}, method="pymc_nuts",
+            num_results=50, num_warmup=50, num_chains=2, cores=2, random_seed=0,
+        )
+        assert isinstance(result, ApproximateDistribution)
```
SoymilkWinsAgain, line 185: There's no assertion here that `pm.sample(..., mp_ctx="spawn")` is called. The PR body itself states that `fork "passed this run"`, but deadlocks are intermittent, so this test cannot reliably prevent future changes to `spawn` from being lost.

probpipe/inference/_pymc_method.py
```diff
@@ -45,6 +46,10 @@ def execute(self, dist: Any, observed: Any, **kwargs: Any) -> ApproximateDistrib
         num_results = kwargs.get("num_results", 1000)
         num_warmup = kwargs.get("num_warmup", 500)
         num_chains = kwargs.get("num_chains", 4)
+        # Multi-core sampling forces the "spawn" start method: this process
+        # holds live JAX threads, and pymc's POSIX-"fork" default deadlocks
+        # when a forked worker inherits a held thread lock.
+        cores = kwargs.get("cores", min(num_chains, os.cpu_count() or 1))
```
SoymilkWinsAgain, line 55: Note that `num_chains` bu default is 4. This will change the default pymc_nuts from single-process to spawn multi-process and require that the PyMC model be serializable by the spawn path.

CHANGELOG.md
SoymilkWinsAgain: The previous text stated that this PR reclaimed multicore, which contradicts this statement.

SoymilkWinsAgain (changes requested):

tests/modeling/test_pymc_model.py
jhuggins: Addressed in 941976c. Added two tests that patch `pymc.sample`, record the kwargs, and assert `mp_ctx == "spawn"` and `cores >= 2` — one for the explicit `cores=2` path and one for the default path (`num_chains` defaults to 4, so `cores` defaults to one worker per chain > 1). The patched sampler delegates to a real `cores=1` run so the downstream extraction/posterior pipeline still produces a genuine `InferenceData`, but the assertion is on the kwargs production passes, so it's deterministic. Verified: reverting `mp_ctx="spawn"` → `None` makes both fail with `assert None == 'spawn'`, independent of whether a real fork run happens to deadlock — exactly the regression you flagged. The original `test_condition_on_multicore_spawn` is kept as a complementary real-spawn smoke test.

probpipe/inference/_pymc_method.py
jhuggins: Intentional — reclaiming the multi-core default is the point of the PR, so `cores` defaults to `min(num_chains, cpu_count)` and spawn is selected for `cores > 1`.

On the serializability requirement: spawn does pickle the model, but it's a *soft* requirement, not a hard failure. PyMC catches pickle failures internally (`pymc/sampling/mcmc.py`) and falls back to single-process sampling with a warning. I verified this by running a model that closes over a `threading.Lock` under `cores=2`: it logged "Sequential sampling (2 chains in 1 job)" and returned a valid result — no error reached the caller. So I documented the behavior in the `cores=` comment rather than adding a `try/except` (which would be dead code given PyMC's own fallback). Commit 941976c.

CHANGELOG.md
jhuggins: Fixed in 941976c — dropped the contradictory clause from the #242 "Ecosystem cutover" entry (`pymc_nuts keeps cores=1 ... the multicore reclaim is deferred`). Since #243 stacks on #242 and both land under the same `[Unreleased]` block, the deferred state never ships on its own, so the `pymc_nuts reclaims multi-core sampling` entry above is now the single source of truth.

tests/modeling/test_pymc_model.py
```diff
@@ -142,6 +162,75 @@ def test_condition_on(self, model):
...
+
+        data = np.random.randn(50)
+        with _captured_pm_sample_kwargs() as captured:
+            _ = condition_on(
+                model, {"y": data}, method="pymc_nuts",
+                num_results=20, num_warmup=10, random_seed=0,
+            )
+
+        assert captured["chains"] == 4               # the documented default
+        assert captured["cores"] >= 2                # one worker per chain, > 1
```
SoymilkWinsAgain: The default-path test still implicitly assumes that the runner has at least 2 CPUs. In the implementation, `cores = min(num_chains, os.cpu_count() or 1)`, so in a single-core environment, `cores=1` and `mp_ctx=None` are valid. However, the test asserts that `cores >= 2` and `mp_ctx == "spawn"`. It is recommended to patch `probpipe.inference._pymc_method.os.cpu_count` in this test, for example, to return 4, and then assert the precise `cores == 4` and `mp_ctx == "spawn"`.

SoymilkWinsAgain (changes requested):

tests/modeling/test_pymc_model.py
jhuggins: Good catch — fixed in 3aad8b6. Pinned `probpipe.inference._pymc_method.os.cpu_count` to 4 in the default-path test and tightened the assertions to the precise `cores == 4` / `mp_ctx == "spawn"`, so it no longer implicitly assumes the runner has >= 2 cores.

I also added a companion `test_single_core_default_skips_spawn` that pins `cpu_count` to 1 and asserts `cores == 1` / `mp_ctx is None` — exactly the single-core case you flagged as valid — so both branches of the `cores > 1 -> spawn` conditional are now exercised deterministically, independent of the host's real core count.

SoymilkWinsAgain (approved):
Looks fine

Issue #117, opened 2026-04-07
Provenance Tracking Improvements


arob5:
Currently, provenance tracking is implemented so that the entire history of a distribution is maintained in memory. This is beneficial, in that the provenance graph can be traversed to obtain the state of a distribution at any point since its creation. However, this will become infeasible for larger workflows as the memory footprint increases.

We need to expand the current system:

1. A sensible default that scales to modestly large workflows.
2. Customization and extensibility: for example, it may be useful to enable full provenance tracking when debugging or mocking a small test workflow. But for larger workflows, can easily change so that only certain lightweight metadata is tracked.
3. A good example for inspiration is xarray's provenance tracking system. This issue will thus likely overlap with #113

erozercan:
## Design Investigation: Provenance Tracking + Prefect Caching + Restart Behavior

Before proposing any implementation changes, I did a deeper investigation into
this issue based on feedback from @jhuggins in our last meeting. Summary:

---

### What we discussed

When I brought this issue up, @jhuggins raised a few related questions that
expanded the scope of what needs to be thought about here. His main points were:

- Before designing new provenance infrastructure, we should ask: **what does
  Prefect already give us for free?** Flow run history, task-level artifacts,
  lineage - does Prefect already handle some of this so we don't duplicate it?
- **Restart and recovery**: if a workflow step fails mid-run, what happens? Can
  Prefect automatically pick up from where it left off, or does it restart from
  scratch?
- **Avoiding duplicate computation**: if you run a pipeline, it succeeds, and
  then you change only the final step — can you re-run and have only that step
  re-execute? This was part of why we use Prefect in the first place.
- He suggested stepping back and thinking about provenance tracking, restarts,
  and re-runs together as one problem, and **proposing design changes before
  implementing anything**.

So this comment is that proposal.

---

### Breaking down what needed to be investigated

These questions split into two separate investigations:

**Part 1 — What does Prefect provide?**
- Does Prefect record what ran, when, with what inputs?
- Is there a caching mechanism that skips already-completed tasks on re-run?
- What happens on failure — can it resume from the failed step?
- Does Prefect have an artifact/lineage system we could use?

**Part 2 — What is the actual memory problem in our provenance system?**
- Why does the current system keep everything in memory?
- Does `to_dict()` help with this?
- What would a "lightweight" alternative look like structurally?

---

### Part 1 findings: what Prefect provides vs. what we use

[...]

PR #244, opened 2026-06-08
feat: add ProvenanceMode enum with lightweight and off tracking modes

erozercan:
## Summary

Addresses the memory footprint problem described in the provenance tracking [issue](https://github.com/TARPS-group/prob-pipe/issues/117). The current system stores live Python references to parent Distribution/Record/RecordArray objects in every `Provenance` node, which means the entire ancestry chain stays in memory for as long as the final result is alive. For workflows that produce large distributions (MCMC samples, broadcast outputs), this can keep tens to hundreds of MB of intermediate data alive unnecessarily.

This PR introduces a `ProvenanceMode` enum that controls how much history is retained, with a new default (`LIGHTWEIGHT`) that drops live references once a step completes.

## Changes

**`probpipe/core/provenance.py`**
- Add `ProvenanceMode` enum: `FULL` / `LIGHTWEIGHT` / `OFF`
- Add `ParentInfo` dataclass: lightweight descriptor storing `type_name`, `name`, and a `fingerprint` field (reserved for the Prefect caching layer, PR 2)
- Update `Provenance.parents` to accept either live references (`FULL`) or `ParentInfo` descriptors (`LIGHTWEIGHT`)
- Update `to_dict()`, `provenance_ancestors()`, and `provenance_dag()` to handle both cases gracefully

**`probpipe/core/config.py`**
- Add `ProvenanceConfig` class and `provenance_config` singleton, following the same pattern as `PrefectConfig`
- Default mode is `LIGHTWEIGHT`; `FULL` and `OFF` are available as explicit opt-ins

**`probpipe/core/node.py`** and **`probpipe/core/_workflow_sweep.py`**
- Update all three provenance assembly sites to check `provenance_config.mode` and build `ParentInfo` descriptors, live references, or `None` accordingly

**`probpipe/__init__.py`**
- Export `ProvenanceMode`, `ParentInfo`, `provenance_config` from the top-level package

**`tests/core/test_provenance.py`**
- Add `full_provenance_mode` fixture; apply to the 5 existing tests that require live reference traversal
- Add 8 new tests in `TestProvenanceModes` covering LIGHTWEIGHT and OFF behaviour

## Usage

```python
import probpipe
from probpipe import ProvenanceMode

# Default — lightweight descriptors, parent objects can be GC'd
probpipe.provenance_config.mode = ProvenanceMode.LIGHTWEIGHT

# Debug / small test workflow — full live-reference chain, full traversal
probpipe.provenance_config.mode = ProvenanceMode.FULL

# Maximum performance — no provenance attached at all
probpipe.provenance_config.mode = ProvenanceMode.OFF
```

## What is NOT in this PR

- Fingerprinting logic for the `ParentInfo.fingerprint` field — that requires a stable hash function for JAX arrays, Records, and Distributions, and is addressed in PR 2
- Prefect task result caching (`persist_result`, `cache_key_fn`) — depends on the fingerprinting system from PR 2
- Ray serialization fix for `Record.__setattr__` blocking pickle — separate pre-existing issue

## Test results

40/40 provenance tests passing. Two pre-existing failures in the broader suite (`test_output_values_correct` — JAX tracer issue in `_numeric_record.py`, `test_pymc_model.py` — broken `arviz` version) are unrelated to this PR and exist on main.


jhuggins:
## Review + a design proposal for `LIGHTWEIGHT` mode

Thanks for this — it's a clean implementation of the `#117` memory fix and CI is green. One design point on `LIGHTWEIGHT`, plus the usual convention/coverage items.

### Design proposal: keep the full DAG in `LIGHTWEIGHT`, and let `Provenance` own the mode

As written, `LIGHTWEIGHT` severs the lineage graph — `provenance_ancestors()` returns `[]` and `provenance_dag()` collapses to a single node, because `ParentInfo` drops the link to the parent's own provenance. I think we can keep the **full DAG structure** in `LIGHTWEIGHT` while still freeing the heavy objects, because every `Provenance` across the codebase stores only light metadata (operation strings + scalars / field-name lists); the heavy data (sample arrays, broadcast outputs) lives in the *parent objects*, never in their `Provenance`. So retaining the `Provenance` tree is cheap; only the parent *objects* need to be released.

**1. Unify on a single descriptor — `parents` is always `tuple[ParentInfo, ...]`.** The only thing the mode changes is whether the descriptor carries the live object:

```python
@dataclass(frozen=True)
class ParentInfo:
    type_name: str
    name: str
    source: "Provenance | None" = None                      # parent's own .source — kept in all modes (cheap pointer)
    fingerprint: str | None = None                          # reserved (PR 2)
    obj: "ProvenanceNode | None" = field(default=None, compare=False)  # live parent, FULL only
```

- **FULL** → `obj=p` (live object retained, as today)
- **LIGHTWEIGHT** → `obj=None`, but `source=p.source` keeps the lineage tree intact (recursively light, since the parent's own parents are `ParentInfo` too)
- **OFF** → no `Provenance` at all

This removes every `isinstance(p, ParentInfo)` fork — `__repr__`, `to_dict`, `provenance_ancestors`, `provenance_dag` all read `p.type_name` / `p.name` / `p.source` uniformly, and the DAG renders in every mode.

**2. Put the mode logic inside `Provenance` so call sites don't each branch.** A factory centralizes it; the raw `__init__` stays low-level:

```python
@classmethod
def create(cls, operation, parents=(), metadata=None):
    """Build provenance honoring provenance_config.mode (None in OFF)."""
    mode = provenance_config.mode
    if mode is ProvenanceMode.OFF:
        return None
    keep = mode is ProvenanceMode.FULL
    refs = tuple(
        ParentInfo(type_name=type(p).__name__, name=getattr(p, "name", None),
                   source=getattr(p, "source", None), obj=p if keep else None)
        for p in parents
    )
    return cls(operation, parents=refs, metadata=metadata or {})
```

Each site then passes **raw, already-filtered** parents and does no mode check (the *which-values-are-parents* filter is genuinely site-specific and stays put):

```python
# node.py __call__ — replaces the ~14-line mode fork:
candidates = [v for v in values.values() if hasattr(v, "source")]
provenance = Provenance.create(f"workflow.{name}", parents=candidates,
                               metadata={"func": name})
```

**3. Make `with_source(None)` a no-op** (return `self` unchanged). Then OFF flows uniformly with no site-level guard:

```python
result.with_source(Provenance.create(...))   # no-op when create() returns None
```

**4. Apply this to all ~13 `Provenance` sites, not just the 3 WorkflowFunction sites.** Right now `from_distribution`, `transformed`, conditioning (`_product` / `_joint_*` / `_sequential_joint`), `make_posterior`, and `transition` still store live parents unconditionally, so the memory win is partial and a mixed DAG results. With the factory it's a near-mechanical `Provenance(...)` → `Provenance.create(...)` at each site, so doing them all here is low-risk and makes the mode consistent everywhere.

**5. Structural prerequisite — break the import cycle.** `Provenance.create` reads `provenance_config`, so `provenance.py` must import from `config.py`; but the PR currently has `config.py` importing `ProvenanceMode` *from* `provenance.py`. Move **`ProvenanceMode` into `config.py`** (alongside `WorkflowKind`, with `ProvenanceConfig` / `provenance_config`). Then `config.py` is stdlib-only, `provenance.py` imports `ProvenanceMode` + `provenance_config` from it (one-directional), and `node.py` / `_workflow_sweep.py` can drop their `ParentInfo` / `ProvenanceMode` / `provenance_config` imports entirely. This also resolves the current asymmetry of one enum in `config` and one in `provenance`.

**6. `provenance_ancestors` / `provenance_dag`** then traverse via `.source` uniformly and return `ParentInfo` in all modes (FULL callers reach the live object via `.obj`). Note this changes the return type of `provenance_ancestors` — the currently-FULL-gated tests (`assert ancestors[0] is base`, etc.) would move to `.obj` / `ParentInfo` assertions.

Net: mode logic lives in exactly one method, every site gets FULL/LIGHTWEIGHT/OFF for free, `LIGHTWEIGHT` keeps the full traversable DAG while still freeing the heavy objects, and `node.py` / `_workflow_sweep.py` shrink.

### Other items (independent of the design above)

**Must fix**
- `ProvenanceMode`, `ParentInfo`, `provenance_config` are imported into `probpipe/__init__.py` but missing from its `__all__` (lines 162–309) — §9.1. (If `ProvenanceMode` moves to `config.py` per (5), re-export it from there.)

**Should fix**
- No CHANGELOG entry — this adds public API and changes default behavior; add Added (the three symbols) + Changed (default mode → `LIGHTWEIGHT`) entries.
- Docs untouched: `docs/api/provenance.md` should document `ProvenanceMode` + the default, and the provenance-traversal cells in the example notebooks (`03_broadcasting`, `07_sequential_updating`, `10_random_functions_and_emulators`, `getting_started`) now render under a different default — refresh or annotate.
- Test gaps: `to_dict()` over `ParentInfo` parents (incl. the `fingerprint` branch) is unasserted; and the **sweep** path (`make_sweep_provenance`) `LIGHTWEIGHT`/`OFF` branches are untested (`TestProvenanceModes` only exercises the broadcast `__call__` path; `test_pure_sweep_provenance` runs the sweep only under FULL).

**Minor**
- `provenance.py` is missing a trailing newline at EOF.
- Underscore-prefixed locals (`_prov_mode`, `_candidates`, `_parents`, `_array_candidates`, `_dist_candidates`) — the leading underscore is the module-private convention, not for locals. (Moot if the factory refactor lands, since these sites collapse.)
- Redundant string-quoted annotations added in `provenance.py` (`provenance_dag(dist: "Distribution")`, `_label`/`_visit`) — with `from __future__ import annotations` the bare name is already lazy and the file uses bare names elsewhere.
- `probpipe/__init__.py` now has two separate `from probpipe.core.provenance import ...` lines — consolidate (and alphabetize `ParentInfo`, `ProvenanceMode`).
- `config.py` defines no `__all__` though it now exposes public `ProvenanceConfig` / `provenance_config` (§9.1; pre-existing for `WorkflowKind` / `prefect_config` too).

### Naming (low stakes, separate)
Once `ParentInfo` can hold the live `obj` and the `.source` edge, it's really a node in the provenance graph rather than a lightweight info struct — `ParentRef` (with the live handle as `.obj`) reads better. Since the symbol isn't released yet, renaming now is free. Not blocking; flagging for a follow-up.

jhuggins (changes requested):
See my comment regarding requested changes

erozercan:
Thanks for the detailed review @jhuggins! Based on your feedback I've put together the following plan:

1. Move `ProvenanceMode` into `config.py` to break the import cycle, and introduce the `ParentInfo` descriptor with `source` and `obj` fields so the full lineage DAG is preserved in all non-OFF modes.
2. Update all three WorkflowFunction provenance assembly sites to always produce `ParentInfo`, add deduplication by object identity, and update `provenance_ancestors()` / `provenance_dag()` for uniform traversal via `.source`.
3. Add `Provenance.create()` factory to centralise the mode-check, make `with_source(None)` a no-op, and replace the ~14-line mode forks at the workflow sites with a single factory call.
4. Apply `Provenance.create()` to the remaining ~10 provenance sites (`from_distribution`, `TransformedDistribution`, `condition_on`, `make_posterior`, `transition`, etc.) so the memory win and consistent DAG are enforced everywhere.

Steps 1–3 are complete. Here's what changed:

- **Step 1** moved `ProvenanceMode` (and the new `ProvenanceConfig` / `provenance_config` singleton) into `config.py`, which is now stdlib-only. `provenance.py` imports from it one-directionally, resolving the cycle.
- **Step 2** redesigned `ParentInfo` with a `source` field (the parent's own `Provenance` chain, kept in all modes since it's cheap) and an `obj` field (live reference in FULL, `None` in LIGHTWEIGHT). All three assembly sites now always build `ParentInfo`, so `provenance_ancestors()` and `provenance_dag()` traverse both `ParentInfo` and live objects uniformly via `.source` — no `isinstance` forks. LIGHTWEIGHT now preserves the full traversable DAG while still freeing parent data arrays.
- **Step 3** added `Provenance.create()` as the single place that reads `provenance_config.mode`, constructs `ParentInfo` tuples, and returns `None` in OFF mode. `with_source(None)` is now a no-op, so call sites pass the factory result directly without an extra guard. `node.py` and `_workflow_sweep.py` no longer import `ProvenanceMode`, `provenance_config`, or `ParentInfo`.

Step 4 is next.

erozercan:
update: 

#### Step 4: Migrate all remaining provenance sites to `Provenance.create()`

Applied `Provenance.create()` consistently across the entire codebase — every site that previously called `Provenance(...)` directly now goes through the factory. This ensures OFF-mode short-circuits and the LIGHTWEIGHT/FULL `obj` distinction are enforced uniformly, not just in `WorkflowFunction`.

Sites updated (~15 across 11 files):

- `_distribution_base.py` — `renamed()`
- `distributions/transformed.py` — `TransformedDistribution.__init__`
- `distributions/_product.py, _sequential_joint.py, _joint_gaussian.py, _joint_empirical.py` — `condition_on`
- `converters/_probpipe.py, _tfp.py, _scipy.py` — converter provenance helpers
- `core/transition.py` — iterate and resample
- `inference/_approximate_distribution.py` — make_posterior

Also fixed the public API surface:

- Added `__all__` to config.py
- Added `ParentInfo`, `ProvenanceMode`, and `provenance_config` to the top-level `probpipe/__init__.py` `__all__`

Finally, updated the test suite to reflect the new contract: tests that previously asserted `parents == (live_obj,)` or `live_obj` in ancestors were updated to either check `parents[0].name` (for LIGHTWEIGHT, the default) or use the `full_provenance_mode` fixture and `.obj` (for tests that genuinely need identity checks).

jhuggins:
## Consolidated review (4 independent lenses)

I ran four independent review passes over this PR — ProbPipe-conventions, general correctness, a focused bug-hunt, and a test audit — and merged them; findings raised by more than one lens are noted as high-confidence. **Overall the design is solid and the hard parts are done well:** all ~15 assembly sites route through a single `Provenance.create()`, OFF mode is crash-safe, and the earlier design concerns are addressed (DAG preserved via `ParentInfo.source`, mode-checking centralized, `obj` live only in FULL). Two items I'd treat as merge-blockers.

### Must fix

**1. Diamond lineages double-count shared ancestors — a regression in `provenance_ancestors()` and `provenance_dag()`, in both LIGHTWEIGHT *and* FULL.** *(found independently by two lenses, with a repro)*
`Provenance.create()` mints a fresh `ParentInfo` per call, and both traversals dedup on `id(ParentInfo)`. When an ancestor `X` is reachable by two paths, each path holds a *distinct* `ParentInfo` for `X`, so it is emitted twice — `provenance_ancestors` returns duplicates and `provenance_dag` renders `X` as two nodes (no longer a DAG). `main` deduped on `id(live object)`, which collapsed these, so this is a behavioral regression, not a LIGHTWEIGHT tradeoff. A 4-node diamond reproduces it: `ancestors == ['A','B','X','X']` in both modes. `test_no_duplicates` doesn't catch it (it repeats one arg, which is deduped earlier at the call site).
**Fix:** dedup on a logical key — `id(p.obj)` when present, else `(type_name, name, id(source))` — and add a diamond test.

**2. No test pins the GC behavior the PR exists to deliver.**
The closest test asserts `parent is not n` / `parent.obj is None`, which a wrapper satisfies even if a strong reference were retained elsewhere. Since #117 is specifically about parents staying alive, there should be a `weakref` + `gc.collect()` test: drop the strong ref after the workflow and assert the weakref is dead under LIGHTWEIGHT and alive under FULL. That is the regression guard the feature needs.

### Should fix

**3. Breaking change, undocumented.** Default → LIGHTWEIGHT silently changes `provenance_ancestors(x)` from live objects to `ParentInfo` descriptors (with `obj is None`), so `x in prov.parents` / `anc.name` / `anc.samples` change or break. No CHANGELOG entry or `kind:breaking-change` label, and `docs/api/provenance.md` still describes the always-live-DAG model. Add the CHANGELOG `Added`/`Changed` entries (with a migration note — `p.obj` only under FULL), the label, and the doc updates.

**4. `ParentInfo` hash/eq.** `@dataclass(frozen=True)` with `source: Provenance` in the hash set makes `hash(parent_info)` raise (`Provenance` holds a `metadata` dict) for any non-root parent — and `ParentInfo` is now a public export, so `set(provenance_ancestors(d))` (the natural way to dedup #1) raises `TypeError`. Also `__eq__` includes `source` but excludes `obj`, so two distinct parents can compare equal. Exclude `source` from hashing, decide the eq semantics deliberately, and test it.

**5. `ParentInfo.name: str` is violated with `None`.** `create()` does `getattr(p, "name", None)`; unnamed parents store `None`, which renders as the literal `'None'` in labels / makes `__repr__` fall back to `"ParentInfo"`. No crash, so not blocking — but type it `str | None` and normalize, or rely on `p.name` (every parent is named per the framework's naming rule).

**6. Dead code + stale comments.** Now that every site emits `ParentInfo`, the "Live object — transitional until all sites use ParentInfo" branches in `to_dict` / `provenance_dag._visit_parent` and the matching `provenance_ancestors` docstring are unreachable/false — deleting them simplifies all three. Separately, `ParentInfo.fingerprint`'s "Reserved … (PR 2)" docstring references an unmerged PR — reword to describe the field's purpose, or defer the field until it is used.

**7. Other test gaps.** (a) `to_dict()` over `ParentInfo` parents (incl. the `fingerprint` branch) is unexercised — Codecov shows ~6 missed lines there. (b) OFF-mode downstream traversal (`provenance_ancestors → []`, single-node dag) and the `Record.with_source(None)` no-op path are untested. (c) Add an `autouse` reset for the `provenance_config.mode` global in `tests/conftest.py` — under `pytest-xdist`, a test that sets the mode and raises before its `finally` leaks the mode into other tests in that worker.

**8. Env-var parity.** `ProvenanceConfig.reset()` hardcodes LIGHTWEIGHT, whereas `PrefectConfig` honors `PROBPIPE_WORKFLOW_KIND`. Consider a `PROBPIPE_PROVENANCE_MODE` for deployment-time control of this memory/perf knob (the mode already reads lazily, so it composes).

### Minor / polish

`__all__` sits between imports in `config.py`; the `__init__.py` config import crosses 100 chars; `create`'s "deduplicated by the caller" docstring is only true at the broadcast/sweep/node sites; LIGHTWEIGHT still retains live distributions held in a parent's `Provenance.metadata` (bounded + pre-existing — worth scoping the GC docstring to *parent data arrays*); test polish (assert `type_name`, not just `name`, in converter provenance tests; brittle positional `ancestors[0]/[1]` assertions; consider `tests/core/test_config.py` for the `ProvenanceConfig` tests; add a jax-path mode test alongside the sequential ones).

### Verified correct

All ~15 provenance sites migrated to `Provenance.create()` (`git grep` finds no remaining direct construction); mode-checking centralized in the factory; OFF mode crash-safe (every `.source` read guarded by `is not None`, `with_source(None)` a no-op); mode read lazily per call so runtime changes apply uniformly; call-site dedup of a repeated arg is correct (distinct from #1); no circular import and the subpackage graph is intact; `__all__` exports present; the mode setter rejects non-enum values with a clear `TypeError`. The earlier design concerns are addressed — the diamond issue is a separate dedup defect, not a failure of the DAG-preservation design.

### Suggested order

(1) diamond dedup + test → (2) gc/weakref test → (3) hash/eq → (4) CHANGELOG + label + docs → (5) dead-branch/comment cleanup → (6) remaining test gaps + autouse reset. (1) and (2) are the load-bearing ones; the rest are quick.

---
*Generated by Claude (consolidated multi-lens review); relayed for triage — apply your own judgment on what to act on.*

PR #246, opened 2026-06-08
Add BayesFlow amortized-SBI backend (NPE/FMPE/CMPE) (upgrade plan step 5)

jhuggins:
## Summary

Step 5 of the SBI-backend replacement: a **BayesFlow amortized-SBI backend**, restoring the amortized half of the SBI layer dropped with sbijax — now jax-native (keras-on-JAX, no TensorFlow/PyTorch). `learn_amortized_posterior` trains an amortized neural posterior estimator and returns a `SupportsConditioning` distribution, so `condition_on(model, observed)` draws from `p(theta | observed)` in a single forward pass with **no MCMC, no gradient bridge, and no prior translation**.

Based on `main`; review feedback (@SoymilkWinsAgain's five comments), a max-effort code-review pass, and a conventions review are addressed in follow-up commits so each delta is reviewable in isolation.

## What's added

**`probpipe/inference/_bayesflow.py`:**

- `learn_amortized_posterior(prior, simulator, *, method="npe"|"fmpe"|"cmpe", num_simulations=, epochs=, batch_size=, sim_backend="jax", inference_network=None, num_results=, random_seed=, optimizer="adam", **fit_kwargs)` — simulates `(theta, y)` pairs offline (`theta` via the `sample` op, `y` via `simulator.generate_data`), builds a BayesFlow `Adapter` from the prior's `record_template`, trains a `ContinuousApproximator`, and returns a `BayesFlowModel`.
- `BayesFlowModel(Distribution, SupportsConditioning)` — the **joint model** `p(theta, y)`: bundles prior + simulator + trained estimator (`prior`/`simulator` exposed as properties; no direct sampler — `_sample` raises a guarded `NotImplementedError` pointing to `condition_on`). `_condition_on` runs one network forward pass and returns **named** draws via `make_posterior(record_template=, ...)`. Amortized: the same instance conditions on any observation with no retraining; honours `num_results`/`random_seed`, ignores MCMC-only kwargs. Exported at the `probpipe` top level alongside `learn_amortized_posterior`.
- The prior needs **no** TFP translation — it is only sampled to draw `theta`.

## Design points worth a look

- **Simulator contract.** `generate_data(params, ...)` receives the prior's **native per-draw structured sample** (named fields, `params["a"]`) — matching `SimpleGenerativeModel` / `PriorPredictiveCheck` — reconstructed per draw via `NumericRecordArray.unflatten`; the `jax.vmap` fast path is preserved.
- **Constrained priors via per-field bijectors.** Each field is trained in unconstrained space (`bijector_for(support).inverse`, applied at the field's *native event shape*) and draws are mapped back through `forward` at sample time. Identity for real-valued fields; handles positive (Gamma), intervals (Beta), simplexes (Dirichlet), and positive-definite matrices (Wishart). Discrete priors have no smooth bijector and are rejected with a clear error. NPE's coupling-flow ≥2-dimension minimum is counted in **unconstrained** dimensions (a 2-simplex contributes one), falling back to flow matching below it.
- **`sim_backend` (not a bool).** `"jax"` (default) vmaps a traceable simulator; `"sequential"` runs an eager per-draw loop for **non-JAX simulators**. Values mirror `WorkflowFunction`'s dispatch vocabulary, so future parallel backends slot in as `Literal` members.
- **Likelihood protocols relocated to `core/protocols.py`.** `Likelihood`, `ConditionallyIndependentLikelihood`, `GenerativeLikelihood` moved from `modeling/_likelihood.py` (re-exported for back-compat) so `inference/` can depend on them without a cycle; removes two reverse import edges. Documented in `STYLE_GUIDE.md` §7.2; the lazy `inference/ → distributions/` edge (`bijector_for`) is now documented in §6's exceptions.
- **`ProductDistribution.supports`** — new per-field constraint accessor (`{field: component.support}`), implementing the canonical `RecordDistribution` API that previously raised `NotImplementedError`; the backend consumes it rather than reaching into `.components`.
- **RNG hygiene.** Training seeds keras (`keras.utils.set_random_seed`) for full reproducibility, but snapshots and **restores the caller's global NumPy/Python RNG state**, so a call does not silently perturb unrelated random streams.
- **Lazy import:** `import probpipe` loads neither keras nor bayesflow; `KERAS_BACKEND=jax` pinned with a post-import backend assertion.

## Robustness guards

Train-time, each raising before any simulation: non-`RecordDistribution` prior (`TypeError`); unknown `method`/`sim_backend`, non-positive `num_simulations`/`batch_size`/`epochs`/`num_results`, discrete/unsupported-support prior (`ValueError`); multi-field prior lacking per-field `supports` (`TypeError`). Condition-time: observation-size mismatch and non-positive `num_results` (`ValueError`). `_sample` raises `NotImplementedError` pointing to `condition_on` (the probe-able signal `WorkflowFunction`'s dispatch fallback catches).

## Tests (`tests/inference/test_bayesflow.py`, 35 passing)

Recovery near truth, amortization across observations, the no-direct-sampler guard, the `ApproximateDistribution` contract, repr, npe/fmpe/cmpe smoke, vector-valued field, caller-supplied `inference_network`, single-field and one-parameter priors, a multi-field (scalar+vector+scalar) prior with 8-d observations, a strict **named-field-access simulator** (fails under flattening), **constrained-support round-trips** for positive (Gamma), interval (Beta), simplex (Dirichlet under NPE), and positive-definite (Wishart — draws verified symmetric PSD), a non-JAX simulator on the `sequential` path, **global-RNG-restoration**, the condition-time guards, **adapter-namespace isolation** (fields literally named `observation` / `inference_variables` train and condition — theta fields are re-keyed to positional internal adapter names, so no field name is reserved), and six train-time validation rejections.

**Calibration against a conjugate Gaussian** (analytic posterior) checks the posterior mean **and spread** for both network paths:

| Path | Mean error | Std ratio |
|------|-----------|-----------|
| 2-D coupling flow | ≤ 0.11 post-std | 1.01–1.03 |
| 1-D flow-matching fallback | ≤ 0.19 post-std | 0.92–1.18 |

Training is fully seeded (same-seed runs are bit-identical); tolerance margins cover cross-platform numerical drift.

## Packaging / CI / docs

- `[bayesflow] = ["bayesflow>=2.0,<3"]`, **Python 3.12–3.13 only** (BayesFlow caps `<3.14`); resolved into `uv.lock` behind a `python_version < 3.14` marker, leaving the verified jax/jaxlib `0.10.1` + tfp pins untouched.
- A separate **`bayesflow` CI leg** (3.12 & 3.13) syncs `dev,nutpie,bayesflow` via `uv sync --frozen`, runs the suite, and uploads its coverage (the test job doesn't install the extra, so the union keeps `_bayesflow.py` — ~96% — in the patch metric).
- New **"Amortized SBI"** section in `docs/api/inference.md` (strict mkdocs build verified); CHANGELOG, STYLE_GUIDE §6/§7.2, and CONTRIBUTING updated.

## Verification

Local: bayesflow 2.0.12 / keras 3.14.1 (jax) / jax 0.10.1 — **35 bayesflow tests** plus `tests/distributions` + `tests/core` (2056) pass; `import probpipe` stays lazy; `uv lock --check` consistent; ruff clean on changed files; strict docs build green.

### DIFF
--- a/probpipe/distributions/_product.py
+++ b/probpipe/distributions/_product.py
@@ -377,8 +377,22 @@ class ProductDistribution(
 
     @property
     def supports(self):
-        """Per-field support constraints -- each component's ``support``."""
-        return {name: comp.support for name, comp in self._components.items()}
+        """Per-leaf support constraints -- each leaf component's ``support``.
+
+        Nested components are keyed by slash-delimited paths
+        (``"outer/a"``), matching ``RecordTemplate.leaf_shapes``, so every
+        value is a ``Constraint``."""
+        out: dict = {}
+
+        def _walk(components: dict, prefix: str) -> None:
+            for name, comp in components.items():
+                if isinstance(comp, dict):
+                    _walk(comp, f"{prefix}{name}/")
+                else:
+                    out[f"{prefix}{name}"] = comp.support
+
+        _walk(self._components, "")
+        return out
 
     # -- Conditioning -------------------------------------------------------
 
--- a/probpipe/distributions/_product.py
+++ b/probpipe/distributions/_product.py
@@ -375,6 +375,11 @@ class ProductDistribution(
             return MappingProxyType(self._components)
         return self._components
 
+    @property
+    def supports(self):
+        """Per-field support constraints -- each component's ``support``."""
+        return {name: comp.support for name, comp in self._components.items()}
+
     # -- Conditioning -------------------------------------------------------
 
     def _condition_on(self, observed=None, /, **kwargs):
--- a/probpipe/inference/_bayesflow.py
+++ b/probpipe/inference/_bayesflow.py
@@ -33,9 +33,7 @@ from ..custom_types import Array, PRNGKey
 from ._approximate_distribution import ApproximateDistribution, make_posterior
 
 if TYPE_CHECKING:
-    # Type-only imports of the optional ``bayesflow`` dependency: stringized by
-    # ``from __future__ import annotations`` and never evaluated at runtime
-    # (bayesflow is loaded only inside ``_ensure_bayesflow``).
+    # Type-only: the optional bayesflow dependency loads at runtime in _ensure_bayesflow.
     from bayesflow import Adapter, ContinuousApproximator
     from bayesflow.networks import InferenceNetwork
 
@@ -139,8 +137,6 @@ def _field_bijectors(prior: Distribution, fields: tuple[str, ...]) -> dict[str, 
         supports = prior.supports
     except NotImplementedError:
         if len(fields) > 1:
-            # Spreading one support across heterogeneous fields would silently pick
-            # the wrong bijector; demand the per-field accessor instead.
             raise TypeError(
                 f"{type(prior).__name__} has {len(fields)} fields but does not "
                 "implement per-field `supports`; cannot infer per-field constraints "
@@ -185,17 +181,14 @@ def _simulate_offline(
     fields = template.fields
     k_theta, k_sim = jax.random.split(key)
     theta = _sample_op(prior, key=k_theta, sample_shape=(num_simulations,))
-    # ``theta.flatten()`` is the prior's canonical flat layout (record-template
-    # field order); ``unflatten`` lifts it back to named per-field arrays --
-    # uniform across record containers and single-field priors (whose raw draws
-    # are not field-indexable by name, but whose flat layout is well defined).
+    # Round-trip through the canonical flat layout: single-field priors' raw draws
+    # are not field-indexable by name, so unflatten gives uniform named access.
     theta_flat = jnp.asarray(theta.flatten()).reshape(num_simulations, -1)
     record = NumericRecordArray.unflatten(
         theta_flat, template=template, batch_shape=(num_simulations,)
     )
-    # Unconstrain each field at its *native* event shape -- matrix-valued bijectors
-    # (e.g. positive-definite: Chain([CholeskyOuterProduct, FillScaleTriL])) require
-    # (..., n, n) input, so the inverse must run before flattening for the adapter.
+    # Invert before flattening: matrix-valued bijectors (positive-definite) require
+    # the field's native (..., n, n) event shape, not the flat adapter layout.
     named = {}
     for f in fields:
         u = bijectors[f].inverse(jnp.asarray(record[f]))
@@ -203,9 +196,8 @@ def _simulate_offline(
     sim_keys = jax.random.split(k_sim, num_simulations)
 
     def _one(flat_row: Array, k: PRNGKey) -> Array:
-        # Reconstruct the prior's native per-draw structured params (a record with
-        # named fields, so the simulator may use ``params["a"]``) -- the same object
-        # SimpleGenerativeModel / PriorPredictiveCheck pass, not a raw flat vector.
+        # Per-draw structured params (named-field access), per the
+        # GenerativeLikelihood contract.
         params = NumericRecordArray.unflatten(flat_row, template=template, batch_shape=())
         return jnp.ravel(simulator.generate_data(params, 1, key=k)[0])
 
@@ -214,9 +206,8 @@ def _simulate_offline(
         y = jax.vmap(_one)(theta_flat, sim_keys)
     else:  # "sequential"
         # Non-traceable simulators (numpy / external code): one eager call per
-        # draw, so generate_data runs concretely rather than under a jax trace.
-        # Move theta to the host once, rather than indexing the device array
-        # (and paying a sync) on every draw; keys stay jax (they are PRNG keys).
+        # draw. Theta is hosted once -- per-draw device indexing would sync
+        # every iteration.
         theta_rows = np.asarray(theta_flat)
         y = jnp.stack([_one(theta_rows[i], sim_keys[i]) for i in range(num_simulations)])
     return named, np.asarray(y, dtype="float32")
@@ -260,18 +251,14 @@ class BayesFlowPosterior(Distribution, SupportsConditioning):
         self._data_dim = data_dim
         self._num_results = num_results
         self._random_seed = random_seed
-        # Per-field forward bijectors (R^d -> support) applied to the network's
-        # unconstrained draws; identity for real-valued priors. Missing/None entries
-        # are treated as identity.
+        # Forward bijectors (unconstrained -> support); missing entries mean identity.
         self._bijectors = bijectors or {}
         # The ``Distribution`` metaclass requires a non-empty name.
         self._name = f"BayesFlowPosterior({method})"
 
     def _sample(self, key: PRNGKey, sample_shape: tuple[int, ...] = ()) -> Any:
-        # Defined (rather than left absent) so callers probing sampleability --
-        # e.g. WorkflowFunction's dispatch fallback, which catches
-        # NotImplementedError but not AttributeError -- get the guarded signal
-        # plus an actionable message.
+        # Present so sampleability probes get NotImplementedError -- which
+        # WorkflowFunction's dispatch fallback catches -- not AttributeError.
         raise NotImplementedError(
             "BayesFlowPosterior is an amortized conditional estimator with no "
             "unconditional sampler; draw from it by conditioning on data: "
@@ -299,12 +286,8 @@ class BayesFlowPosterior(Distribution, SupportsConditioning):
             conditions={_OBSERVATION_KEY: obs_flat[None, :]},
             seed=random_seed,
         )
-        # ``out`` maps each field to ``(1, num_results, d_field)``; squeeze the
-        # observation axis, map each field's unconstrained draws back to its support
-        # via the forward bijector (identity for real-valued fields), flatten any
-        # matrix-valued output, and concatenate in record-template field order (the
-        # layout ``make_posterior`` lifts back to named draws). Stays in jnp
-        # end-to-end -- this is the latency-critical amortized path, so no per-field
+        # ``out`` maps each field to ``(1, num_results, d_field)``. Stays in jnp
+        # end-to-end: this is the latency-critical amortized path, so no per-field
         # host round-trips.
         cols = []
         for f in self._fields:
@@ -462,8 +445,8 @@ def learn_amortized_posterior(
             f"prior field name {_OBSERVATION_KEY!r} collides with the reserved "
             "observation key used internally by the BayesFlow adapter; rename the field."
         )
-    # Per-field unconstraining bijectors (identity for real-valued priors); this also
-    # rejects discrete / unsupported-support priors up front, before any simulation.
+    # Built up front: also rejects discrete / unsupported-support priors before
+    # any simulation runs.
     bijectors = _field_bijectors(prior, fields)
     # The network trains on *unconstrained* widths, which differ from the prior's
     # event sizes for dimension-shifting bijectors (a d-simplex contributes d-1).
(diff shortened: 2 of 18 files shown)


probpipe/inference/_bayesflow.py
```diff
@@ -0,0 +1,381 @@
...
+    fields = template.fields
+    k_theta, k_sim = jax.random.split(key)
+    theta = _sample_op(prior, key=k_theta, sample_shape=(num_simulations,))
+    # ``theta.flatten()`` is the prior's canonical flat layout (record-template
+    # field order); the simulator's ``generate_data(params, ...)`` consumes it
+    # positionally, and ``unflatten`` splits it back into per-field arrays for
+    # the adapter via the same canonical layout -- uniform across record
+    # containers and single-field priors (whose raw draws are not field-indexable
+    # by name, but whose flat layout is well defined).
+    theta_flat = jnp.asarray(theta.flatten()).reshape(num_simulations, -1)
```
SoymilkWinsAgain, line 207: Here, the prior sample is forcibly flattened before being passed to `simulator.generate_data(p, ...)`. This is inconsistent with the existing semantics of `GenerativeLikelihood[P, D]` / `SimpleGenerativeModel`, which passes the prior original sample object (usually a named Record for `ProductDistribution`). Therefore, any simulator that depends on `params["a"]` / named fields will fail. Either unflatten the prior flat view back to the original structure before calling the simulator, or explicitly narrow the API to a "flat-vector simulator" and perform explicit validation/docs/tests.

probpipe/inference/_bayesflow.py
```diff
@@ -0,0 +1,381 @@
...
+        If ``method`` is not one of ``"npe"`` / ``"fmpe"`` / ``"cmpe"``,
+        ``sim_backend`` is not ``"jax"`` / ``"sequential"``, or a prior field
+        collides with the reserved ``"observation"`` key.
+    TypeError
+        If ``simulator`` lacks ``generate_data``, or ``prior`` is not a
+        ``RecordDistribution`` (has no ``record_template``).
+    ImportError
+        If the ``[bayesflow]`` extra is not installed.
+    """
+    if method not in ("npe", "fmpe", "cmpe"):
```
SoymilkWinsAgain, line 453: Need to check `num_simulations <= 0`, `batch_size <= 0`, `epochs < 0`, `num_results <= 0`, etc.

probpipe/inference/_bayesflow.py
```diff
@@ -0,0 +1,381 @@
...
+    if sim_backend not in ("jax", "sequential"):
+        raise ValueError(
+            f"Unknown sim_backend: {sim_backend!r}. Supported: 'jax', 'sequential'."
+        )
+    if not hasattr(simulator, "generate_data"):
+        raise TypeError(
+            "simulator must be a GenerativeLikelihood with a generate_data method, "
+            f"got {type(simulator).__name__}"
+        )
+    record_template = getattr(prior, "record_template", None)
```
SoymilkWinsAgain, line 474: This function only checks the `record_template` and doesn't restrict prior support. However, `ContinuousApproximator` outputs continuous, unconstrained values, and `_condition_on()` directly reassembles the posterior draws. For constrained or discrete priors like Gamma/Beta/Uniform/Bernoulli, this might silently generate posterior draws outside the support or non-discrete values. The logic needs modification, or the boundaries need to be restricted.

probpipe/inference/_bayesflow.py
```diff
@@ -0,0 +1,381 @@
...
+            "record_template."
+        )
+    fields = record_template.fields
+    if _OBSERVATION_KEY in fields:
+        raise ValueError(
+            f"prior field name {_OBSERVATION_KEY!r} collides with the reserved "
+            "observation key used internally by the BayesFlow adapter; rename the field."
+        )
+    bf = _ensure_bayesflow()
+    key = jax.random.PRNGKey(random_seed)
```
SoymilkWinsAgain: I believe that Keras/BayesFlow networks do not have an explicit seed during initialization/training here.

tests/inference/test_bayesflow.py
```diff
@@ -0,0 +1,456 @@
...
+small, fast-training toy fixture so the suite stays tractable.
+"""
+
+from __future__ import annotations
+
+import os
+
+import pytest
+
+pytest.importorskip("bayesflow")
```
SoymilkWinsAgain: From CodeX:

Should set the environment first, then use importorskip. Otherwise, if there are no preset environment variables on the local machine, importorskip may have already imported BayesFlow/Keras and locked the backend.

SoymilkWinsAgain (changes requested):

jhuggins:
Thanks for the review — all five points are addressed in 944f4d9 (a follow-up commit, so the delta against your review is isolated and squashes on merge). CI is green.

**1. `generate_data` received a flattened vector, not the structured sample (`_bayesflow.py:148`).** `_simulate_offline` now reconstructs the prior's native per-draw record (`NumericRecordArray.unflatten`, `batch_shape=()`) and passes that, so a simulator can use `params["a"]` — matching `SimpleGenerativeModel` / `PriorPredictiveCheck`. The `"jax"` (vmap) fast path is preserved (the reconstruction traces under `vmap`). New `test_simulator_receives_named_record` uses strict named access and would fail under the old flattening.

**2. Missing numeric validation (`:335`).** `num_simulations` / `batch_size` / `epochs` / `num_results` must be positive integers, rejected up front (`test_rejects_nonpositive_counts`, parametrized).

**3. Constrained / discrete prior support (`:348`).** Each prior field is now trained in unconstrained space via its `bijector_for` reparameterization (inverse at train time, forward applied to draws at sample time); real-valued fields use the identity, so existing behavior is unchanged. Constrained-continuous priors (positive / interval / …) are properly handled — `test_constrained_prior_draws_respect_support` trains on a Gamma prior and asserts every posterior draw lands in the support. Discrete priors have no smooth bijector and are rejected with a clear error (`test_rejects_discrete_prior`).

**4. No explicit keras/BayesFlow seed (`:363`).** `keras.utils.set_random_seed(random_seed)` now seeds network init + fit; two same-seed runs are byte-identical. The calibration tests' tolerance comments were updated accordingly — their margins now reflect cross-platform / library-version drift rather than RNG nondeterminism. (I left the numeric tolerances as-is: determinism is within-environment only, so they still need to absorb macOS↔Linux numerical drift. Happy to tighten if you'd prefer.)

**5. `importorskip` before `KERAS_BACKEND` (`test_bayesflow.py`).** Reordered so `KERAS_BACKEND=jax` is set **before** `pytest.importorskip("bayesflow")`, so a bare local `pytest` can no longer lock keras to a non-jax backend.

Suite is 28 passing (+7 new tests). Ready for another look.

probpipe/inference/_bayesflow.py
```diff
@@ -0,0 +1,551 @@
...
+    def prior(self) -> Distribution:
+        """The prior over model parameters."""
+        return self._prior
+
+    @property
+    def simulator(self) -> GenerativeLikelihood:
+        """The generative likelihood (provides ``generate_data``)."""
+        return self._simulator
+
+    def _sample(self, key: PRNGKey, sample_shape: tuple[int, ...] = ()) -> tuple[Any, Any]:
```
SoymilkWinsAgain: `inference.md` and here all imply that `sample()` will return `(params, data)`. However, `sample` is actually a `WorkflowFunction`, and the raw tuple will be wrapped by the result boundary and will not be returned as a tuple. Coincidentally, only the private method `_sample` was tested in `test_bayesflow`, thus bypassing this issue. It is recommended to clearly define the boundary behavior of the public `sample`.

tests/inference/test_bayesflow.py
```diff
@@ -0,0 +1,644 @@
...
+            num_simulations=1500, epochs=3, batch_size=256,
+            num_results=200, random_seed=0, verbose=0,
+        )
+        post = condition_on(model, _observe(0.5, 0.0, 0))
+        assert post.algorithm == f"bayesflow_{method}"
+        draws = post.draws()
+        assert np.isfinite(np.asarray(draws["a"])).all()
+        assert np.asarray(draws["a"]).reshape(-1).shape[0] == 200
+
+    def test_vector_valued_field(self):
```
SoymilkWinsAgain, line 301: See my comment in `_bayesflow.py`

probpipe/inference/_bayesflow.py
```diff
@@ -0,0 +1,551 @@
...
+            f"Unknown amortized SBI method: {method!r}. Supported: 'npe', 'fmpe', 'cmpe'."
+        )
+    if sim_backend not in ("jax", "sequential"):
+        raise ValueError(
+            f"Unknown sim_backend: {sim_backend!r}. Supported: 'jax', 'sequential'."
+        )
+    for _name, _val in (
+        ("num_simulations", num_simulations), ("batch_size", batch_size),
+        ("epochs", epochs), ("num_results", num_results),
+    ):
```
SoymilkWinsAgain, line 464:
```python
if not isinstance(_val, int):
  raise TypeError(...)
```

probpipe/distributions/_product.py
```diff
@@ -375,6 +375,11 @@ def components(self):
             return MappingProxyType(self._components)
         return self._components
 
+    @property
+    def supports(self):
```
SoymilkWinsAgain, line 379: The newly added `ProductDistribution.supports` only handles flat components; it throws an `AttributeError: 'dict' object has no attribute 'support'` when encountering nested dict components. This conflicts with existing nested `ProductDistribution` support.

probpipe/inference/_bayesflow.py
jhuggins: Yeah, I decided include _sample is not a good idea, including for the reasons you mention. Now removed

probpipe/inference/_bayesflow.py
```diff
@@ -0,0 +1,551 @@
...
+        )
+    record_template = getattr(prior, "record_template", None)
+    if record_template is None:
+        raise TypeError(
+            "learn_amortized_posterior requires a RecordDistribution prior with "
+            "named parameter fields -- typically a ProductDistribution of named "
+            f"distributions -- but got {type(prior).__name__}, which has no "
+            "record_template."
+        )
+    fields = record_template.fields
```
SoymilkWinsAgain, line 482: From CodeX:

Here top-level fields are used to index `record_template.numeric_leaf_shapes`, but the numeric leaf keys of the nested template are slash paths, causing nested priors to fail. It is recommended to recursively generate leaf-path keyed supports, or explicitly reject nested priors in BayesFlow with a clear error; also, supplement the tests with supports for nested product distributions.

SoymilkWinsAgain (changes requested):

probpipe/inference/_bayesflow.py
jhuggins: Fixed in 10837a6 — the count parameters now require integers (`int` or `np.integer`; floats, strings, and `None` raise `TypeError` with the parameter name) before the existing `< 1` ValueError check. Parametrized test added (`test_rejects_non_integer_counts`).

probpipe/distributions/_product.py
jhuggins: Fixed in 10837a6 — `supports` now walks nested dict components recursively and keys leaves by slash-delimited paths (`"outer/a"`), matching the `RecordTemplate.leaf_shapes` convention, so every value is a leaf `Constraint` and flat products are unchanged. Tested against a nested product with heterogeneous leaf constraints (`test_supports_nested_components`).

probpipe/inference/_bayesflow.py
jhuggins: Fixed in 10837a6, both halves: `ProductDistribution.supports` is now recursively leaf-path keyed (see the thread on `_product.py`), and `learn_amortized_posterior` explicitly rejects nested priors up front with a clear `TypeError` — the per-field bijector/adapter machinery is flat, so this fails fast before any `numeric_leaf_shapes` indexing. Tests added for both (`test_supports_nested_components`, `test_rejects_nested_prior`).

SoymilkWinsAgain (approved):
Looks fine

PR #248, opened 2026-06-09
chore(infra): migrate dev + CI dependency management to uv

jhuggins:
## Summary

Migrates dev environment + CI dependency management from `pip` to [`uv`](https://docs.astral.sh/uv/). Adds a committed `uv.lock` as the single source of truth for resolved dependency versions; CI installs via `uv sync --frozen`, so the matrix legs cannot silently drift from the lockfile.

This is structurally a small change — the workflow steps shrink because the duplicated inline jax/jaxlib/tfp-nightly pin pair (and the "single resolver transaction" explanatory comment) collapses into uv.lock entries. uv.lock itself is large (~5000 lines, 255 packages), but it's a generated artifact; reviewers should focus on the workflow + CONTRIBUTING.md diff.

## What's in this PR

- **`uv.lock`** (new, ~5000 lines, generated) — full resolved tree. uv resolves the current `pyproject.toml` constraints to the same `jax==0.10.1` / `jaxlib==0.10.1` / `tfp-nightly==0.26.0.dev*` versions the inline pin in CI was using.
- **`.github/workflows/ci.yml`** — both jobs (test + notebooks):
  - `actions/setup-python@v5` + multi-line `pip install -e ".[...]"` with the inline jax/tfp pins → `astral-sh/setup-uv@v3` + `uv sync --frozen --extra ...`.
  - `pytest ...` → `uv run pytest ...`; `python -m nbconvert ...` → `uv run python -m nbconvert ...`.
  - The duplicated 8-line explanatory comment about the single-resolver-transaction workaround is removed (the workaround is no longer needed under uv's solver + lockfile model).
- **`.github/workflows/docs.yml`** — same `setup-uv` + `uv sync --frozen --extra docs` + `uv run mkdocs build`.
- **`CONTRIBUTING.md`** — Installation, Running Tests, Coverage, Documentation, and CI sections updated to reflect the uv-first flow. Adds an explicit "Updating dependencies" subsection so contributors know to commit `uv.lock` alongside `pyproject.toml` bumps.
- **`CHANGELOG.md`** — entry under `[Unreleased]` / `### Changed` noting the tooling switch.

## Linked issue(s)

Refs the CI-tightening conversation in the beta plan (`Beta hygiene` line in `probpipe-0.2.0-beta-plan.md` §2). This PR is the foundation for follow-ups (ruff in CI, coverage floor enforcement, pyright advisory mode, pre-commit) — those become cleaner once uv is in place.

## Test plan

- `uv lock` succeeded locally (255 packages resolved in ~4s); resulting `uv.lock` chooses jax/jaxlib 0.10.1 and a current tfp-nightly, matching what CI has been using.
- YAML validated locally on both workflow files.
- CI run on this PR is the real verification — it exercises the full install + test + notebook path with the new tooling on both 3.12 and 3.13 matrix legs.
- Local pip workflow still works: `pip install -e ".[dev]"` resolves and tests run (uv changes don't remove the pip-installable path; only the recommended way).

## Breaking changes

None for runtime users — `pip install probpipe` continues to work unchanged. Contributors are encouraged to switch to `uv sync` but pip remains supported.

## Documentation

- [x] Docstrings — no code touched.
- [x] User Guide / tutorials — not applicable.
- [x] CHANGELOG entry added.
- [x] CONTRIBUTING.md updated (Installation, Running Tests, Coverage, Documentation, CI sections + new "Updating dependencies" subsection).

## Checklist

- [x] PR title follows `<type>(<scope>): <subject>` — `chore(infra): ...`.
- [x] `area:infrastructure` label applied.
- [x] No issue to link (this is the dev-tooling switch we discussed; opening here as a standalone PR per CONTRIBUTING.md's "small standalone fix" exception).

## Notes for reviewers

- **Why uv now:** Astral's uv is genuinely faster (resolver in ms, installs in seconds), the lockfile model fixes the "two CI jobs with duplicated inline pins" smell already documented in the old workflow comments, and it sets up cleanly for the pending CI tightening (ruff lint job, pyright advisory mode, etc.).
- **Why `--frozen` in CI:** prevents the "CI passes locally but lockfile drifts on next push" failure mode. Any dep bump becomes an explicit `uv lock --upgrade*` step in a PR.
- **Lockfile review:** uv.lock is auto-generated; standard review approach is to verify the workflow + CONTRIBUTING.md diff and trust the lockfile contents. If you want a sanity check, `head -100 uv.lock` shows the top-level pins.
- **Workshop / R7 interaction:** Once the "New to Python?" install fork (R7 in the docs roadmap) lands, Route A becomes `uv tool install probpipe` — one command instead of "install Python via conda, then pip install". Worth coordinating R7 with this PR's merge.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -176,35 +176,28 @@ jobs:
             fi
           fi
 
-      - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v5
+      - name: Install uv
+        uses: astral-sh/setup-uv@v3
         with:
+          enable-cache: true
           python-version: ${{ matrix.python-version }}
 
       - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip
-          # Install the project together with the verified jax / tfp-nightly
-          # pins in a single resolver transaction. Installing the project
-          # first and re-pinning afterwards resolves the tree twice — the
-          # first time against unpinned metadata — which can leave the rest
-          # of the dependency set resolved against a different jax. tfp-nightly
-          # tracks jax's internal API, and a fresh nightly can lag the current
-          # jax release (DeprecationWarnings, occasionally hard errors) under
-          # jax 0.10+. Refresh this pair when the nightly is purged from PyPI.
-          # [pymc] is installed so pymc 6 is exercised on both matrix legs.
-          pip install -e ".[dev,nutpie,pymc]" \
-            "jax==0.10.1" "jaxlib==0.10.1" "tfp-nightly==0.26.0.dev20260608"
+        # Dependency versions (including jax/jaxlib/tfp-nightly) are pinned
+        # in uv.lock. --frozen makes the resolver fail if the lockfile is
+        # out of sync, so CI never silently drifts from the committed pins.
+        # [pymc] is included so pymc 6 is exercised on both matrix legs.
+        run: uv sync --frozen --extra dev --extra nutpie --extra pymc
 
       - name: Run tests (changed files only)
         if: steps.changed.outputs.skip != 'true' && steps.changed.outputs.run_all == 'false'
         run: |
-          pytest ${{ steps.changed.outputs.test_files }} --cov-report=xml --cov-report=html
+          uv run pytest ${{ steps.changed.outputs.test_files }} --cov-report=xml --cov-report=html
 
       - name: Run tests (full suite)
         if: steps.changed.outputs.skip != 'true' && steps.changed.outputs.run_all == 'true'
         run: |
-          pytest --cov-report=xml --cov-report=html
+          uv run pytest --cov-report=xml --cov-report=html
 
       - name: Upload coverage reports to Codecov
         if: matrix.python-version == '3.12' && steps.changed.outputs.run_all == 'true'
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -176,35 +176,28 @@ jobs:
             fi
           fi
 
-      - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v5
+      - name: Install uv
+        uses: astral-sh/setup-uv@v3
```
SoymilkWinsAgain: From CodeX:

`astral-sh/setup-uv@v3` does not support python-version input. The actual CI logs show `Unexpected input(s) 'python-version'`, and the `test (3.13)` entry shows `Using CPython 3.12.3 interpreter at: /usr/bin/python3`. Therefore, the current "3.13" check refers to the 3.13 in the name, but it actually runs version 3.12.

.github/workflows/ci.yml
SoymilkWinsAgain, line 50: From CodeX:

The changes to `uv.lock` and the workflow were not treated as foundational changes. The test jobs in this PR actually output `No .py files changed - skipping tests`, and the notebook job was also skipped because there were no changes to the notebook/source. This means that `uv.lock`-only dependency updates will pass CI but won't run pytest/notebooks, directly violating the lock file verification semantics that this PR aimed to establish.

README.md
```diff
@@ -116,6 +116,8 @@ cd prob-pipe
 pip install .
 ```
 
+ProbPipe also installs cleanly with [uv](https://docs.astral.sh/uv/) — substitute `uv pip install .` for any `pip install` example below, or use `uv sync` for a lockfile-managed dev env (see [CONTRIBUTING.md](CONTRIBUTING.md#installation)).
```
SoymilkWinsAgain: `uv pip install` typically requires an existing virtualenv or an explicit target environment. Directly replacing `pip install` from the current context of `README.md` may fail. Ithink it would be better to write "inside an active uv-managed environment, use uv pip install .", or provide the explicit paths to `uv venv && uv pip install . / uv sync`.

SoymilkWinsAgain (changes requested):

.github/workflows/ci.yml
jhuggins: Fixed in 4d93b09. Bumped to `astral-sh/setup-uv@v5` and dropped the `python-version` input that `@v3` ignored, pinning the interpreter explicitly via `uv sync --frozen --python ${{ matrix.python-version }}` on both test matrix legs (and `--python 3.12` on the notebook job). uv fetches a managed CPython when the runner lacks it, so the `3.13` leg now genuinely runs 3.13, and `uv run pytest` uses that synced venv. Good catch on the silent fallback.

.github/workflows/ci.yml
jhuggins: Fixed in 4d93b09. Added `uv.lock` (and `.github/workflows/ci.yml`) to the foundational-path `case` in both the test and notebook jobs, so a lockfile-only dependency bump now forces the full suite instead of hitting "No .py files changed — skipping tests." That restores the lockfile-verification semantics the migration is meant to provide.

Side effect: since `ci.yml` is now foundational too, this very push runs the full suite on both 3.12 and 3.13 — which also exercises the Python-pinning fix from the other thread.

README.md
jhuggins: Fixed in 4d93b09. Reworded the uv note in both `README.md` and `docs/index.md` (the same line appears in each). It now leads with `uv sync` for the managed path and clarifies that `uv pip install` installs into an active or explicitly targeted environment (`uv venv && source .venv/bin/activate`), not a global one — so it's no longer presented as a drop-in for `pip install`.

SoymilkWinsAgain (approved):
Looks good!

PR #249, opened 2026-06-10
chore(infra): CI hygiene bundle (3.14 matrix, concurrency, coverage floor, auto-labeling, dependabot)

jhuggins:
## Summary

Five small, independent CI-hygiene additions bundled in one PR. Each is a few lines on its own; bundling keeps review costs low.

### Python 3.14 in the CI matrix

`.github/workflows/ci.yml` matrix becomes `[3.12, 3.13, 3.14]`. `requires-python = ">=3.12"` in `pyproject.toml` is unchanged. README, `docs/index.md`, and CONTRIBUTING.md updated to reflect "tested on 3.12, 3.13, and 3.14".

Per SPEC 0 (the successor to NEP 29) — as of mid-2026, 3.12, 3.13, and 3.14 are all within the supported window:

| Python | Released | Age | SPEC 0 |
|---|---|---|---|
| 3.12 | Oct 2023 | 32 mo | support |
| 3.13 | Oct 2024 | 20 mo | support |
| 3.14 | Oct 2025 | 8 mo | support |

Verified before bundling: `uv lock` resolves all 255 packages cleanly on 3.14.5 — wheels exist for every dep (jax, jaxlib, tfp-nightly, pymc 6, nutpie, arviz 1.x, blackjax, etc.).

### Concurrency cancellation for PR runs

`.github/workflows/ci.yml` gets:
```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```
A new push to a PR cancels the prior in-progress CI run (saves CI minutes during iteration). Pushes to `main` are **never** cancelled — the merge-history gate stays solid.

`.github/workflows/docs.yml` gets the same pattern, but the original `concurrency: group: pages, cancel-in-progress: false` block is preserved for pushes to `main` (gh-pages allows only one in-flight deploy and that serialization is intentional). The new expression scopes PR builds by ref and cancels them:
```yaml
group: ${{ github.event_name == 'pull_request' && format('docs-pr-{0}', github.ref) || 'pages' }}
cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

`docs.yml`'s `astral-sh/setup-uv` is also bumped `@v3` → `@v5` to match `ci.yml` (the `@v3` step silently ignored its `python-version` input).

### Coverage floor enforced at 88% (full-suite runs)

[...]

PR #193, opened 2026-05-20
feat(diagnostics): DiagnosticsModule — unified Bayesian diagnostic pipeline workflow

YonghoLim:
## Overview

Adds a Bayesian **diagnostics subsystem** for ProbPipe posteriors. After
fitting (e.g. via `condition_on`), diagnostics are computed by in-place
`add_*` ops that attach their results to the posterior's `_auxiliary` DataTree
and are read back through a typed `posterior.diagnostics` view — no manual
draw extraction, no ArviZ boilerplate, and no resampling (everything reads the
existing `ApproximateDistribution` chain structure).

```python
from probpipe.diagnostics import add_mcmc_diagnostics, add_loo

add_mcmc_diagnostics(posterior)        # R-hat, ESS, MCSE — computed in place
posterior.diagnostics.mcmc             # MCMCView: per-parameter stats + warnings
print(posterior.diagnostics.summary_table())

add_loo(posterior, model=model, data=data)
posterior.diagnostics.loo              # LOOView: ELPD, p_loo, Pareto-k
```

## Public API (`probpipe.diagnostics`)

**MCMC convergence / mixing** (read the chain structure directly; no resampling):
- `add_rhat` — R-hat per parameter
- `add_ess` — bulk and tail effective sample size per parameter
- `add_mcse` — Monte Carlo standard error (mean and sd) per parameter
- `add_mcmc_diagnostics` — all of the above plus human-readable warnings

**Predictive checks:**
- `add_ppc` — posterior predictive check (pass `observed_data`) or prior
  predictive check (`observed_data=None` with `num_observations`); accepts one
  or more test statistics and reports one-sided p-values

**Cross-validation:**
- `add_loo` — PSIS-LOO; reports ELPD, `p_loo`, SE, and per-bucket Pareto-k
  counts with warnings

**Views** (returned by `posterior.diagnostics`): `DiagnosticsView` →
`MCMCView` / `PPCView` / `LOOView`, plus the generic `DiagnosticRunView` and the
`NotComputed` sentinel.

Vector- and matrix-valued parameters are handled throughout (event dimensions
are preserved; components are named `param[i]` / `param[i, j]`). The estimators
delegate to ArviZ, and results are stored in an ArviZ-compatible DataTree for
use with `az.plot_*`.

## Design principles

- **No resampling** — diagnostics read `posterior.chains` / `posterior.draws()`
  as stored by `condition_on`.
- **In-place, identity-preserving** — `add_*` ops attach results under
[...]

PR #250, opened 2026-06-10
ci: add advisory ruff lint job + pre-commit hooks

jhuggins:
## Summary

Turns on **ruff in CI (advisory)** and adds **local pre-commit hooks** — C1 + C4 from the CI-tightening plan.

The key finding that shaped this: ruff is configured in `pyproject.toml` but has never been enforced, so the tree currently has **~810 violations** (a backlog, not regressions). Rather than a repo-wide auto-fix sweep — which would merge-conflict with every heavy in-flight PR (#193 +6,980 / #67 +2,401 / #246 +1,231 / #204 +750) — this lands ruff in **advisory mode**: it annotates PRs but does not gate merges. It flips to blocking later, once the backlog is burned down and the big refactors land.

### CI: `lint (advisory)` job (`.github/workflows/ci.yml`)
A new job parallel to `test`/`notebooks` (on `astral-sh/setup-uv@v5`, matching the other jobs). It resolves the ruff version from the pre-commit config, then runs the check in advisory mode:
```yaml
- name: Resolve ruff version from .pre-commit-config.yaml
  id: ruffver
  run: |
    set -euo pipefail
    ver=$(grep -A2 'astral-sh/ruff-pre-commit' .pre-commit-config.yaml \
            | sed -nE 's/^[[:space:]]*rev:[[:space:]]*v?([0-9][0-9.]*).*/\1/p' | head -1)
    test -n "$ver"
    echo "ruff=$ver" >> "$GITHUB_OUTPUT"
- name: Ruff check (advisory — does not gate merges)
  continue-on-error: true
  run: uvx "ruff@${{ steps.ruffver.outputs.ruff }}" check . --output-format=github
```
`--output-format=github` posts inline annotations on the PR's Files tab; `continue-on-error: true` keeps the check non-gating. The resolve step *does* gate — it fails loudly if the rev can't be parsed (a real config error). **To enforce later:** drop `continue-on-error` and scope `.` to changed files (the repo already has a changed-files diff pattern in this same workflow).

### Local hooks (`.pre-commit-config.yaml`, new)
- `ruff` (lint, **no `--fix`** — flags but doesn't silently rewrite)
- file hygiene: `trailing-whitespace`, `end-of-file-fixer`, `check-yaml`, `check-toml`, `check-merge-conflict`, `check-added-large-files` (`--maxkb=2000` so `uv.lock` (~760 KB) passes while stray data/binaries are still caught)
- Install: `uvx pre-commit install`. Hooks run on **staged files only**, so the codebase ratchets clean file-by-file. (A one-off `pre-commit run --all-files` is *not* expected clean yet — it'll touch the pre-existing backlog; CONTRIBUTING.md says so. This PR's own files are hook-clean, including a trailing-newline fix to `pyproject.toml`.)

### Ruff config tuning (`pyproject.toml`)
- Ignore **`RUF001/2/3`** (ambiguous-unicode) — pure noise on the Greek letters, en-dashes, →, ×, ≤ that fill a Bayesian-stats codebase's docstrings; the homoglyph-attack surface they guard is negligible here.
- Per-file-ignore **`E402`/`E702`** for `*.ipynb` — notebooks legitimately import after narrative cells and group plotting calls with semicolons.
- Net effect: the advisory report drops from **810 → 678** real issues, so signal isn't buried in false positives.

### `ruff format` deliberately NOT adopted
It would reformat **155 of 187 files** (2,804 changed lines in `probpipe/` alone) in a black-like style that conflicts with the horizontal-packing conventions documented in CONTRIBUTING.md § Code formatting. Only `ruff check` runs, in both CI and pre-commit.

### Docs
- CONTRIBUTING.md: new **"Linting & pre-commit"** section + a CI bullet.
- CHANGELOG: entry under `[Unreleased]`.

## Version pinning

The pre-commit `rev` (`v0.15.16`) is the **single source of truth** for the ruff version: the CI `lint` job reads it from `.pre-commit-config.yaml` at run time (no separately-drifting inline pin), and `pre-commit autoupdate` / Dependabot bump it there. `uv.lock` pins ruff `0.15.16` too — kept in sync for local `uv run ruff` — and all three currently agree.

## Linked issue(s)

Refs the beta-hygiene CI-tightening discussion. Type checking (pyright, advisory) is a **separate follow-up PR** per the agreed split.

## Test plan

Verified locally before pushing:
- `uvx ruff@0.15.16 check .` → 678 issues with the tuned config (was 810); confirms the ignores apply.
- `uvx pre-commit validate-config` passes.
- `uvx pre-commit run check-yaml --all-files` and `check-toml --all-files` → **Passed** across the whole repo (mkdocs.yml, all workflows, issue templates, dependabot, pyproject all parse).
- `uvx pre-commit run ruff --files probpipe/_dtype.py` → installs the `v0.15.16` hook env and runs clean, confirming the pinned rev resolves.
- The CI `Resolve ruff version` extraction was verified to read `0.15.16` from `.pre-commit-config.yaml`.
- CI on this PR exercises the advisory job end-to-end (it will go green with annotations, by design).

## Breaking changes

None. The lint job doesn't gate; pre-commit is opt-in; no runtime/API change.

## Documentation

- [x] No docstrings touched (no source code changes).
- [x] CONTRIBUTING.md "Linting & pre-commit" section added.
- [x] CHANGELOG entry under `[Unreleased]`.

## Checklist

- [x] PR title follows `<type>(<scope>): <subject>` — `ci: ...`.
- [x] `area:infrastructure` label applied.
- [x] No linked issue — standalone CI hygiene per the CONTRIBUTING.md small-fix exception.

## Notes for reviewers

- **Branched from `main`, not stacked on #249.** Both touch `ci.yml`, but in different regions (#249 adds a top-level `concurrency:` block + 3.14 matrix; this adds a new `lint` job under `jobs:`), so they should auto-merge. The advisory-pyright follow-up also adds a job under `jobs:`, so it overlaps this one more directly — whichever of those two lands second will need a trivial conflict resolution.
- **Why advisory and not a cleanup sweep:** chosen explicitly to avoid conflicting with the in-flight refactors. The 678-issue backlog burns down naturally as pre-commit cleans touched files; we flip the CI gate to blocking once it's near zero.
- **Why no `ruff format`:** the 155-file reformat would fight the hand-tuned style the project documents and uses. If the team later wants automated formatting, that's a separate decision requiring reconciling ruff-format's style with CONTRIBUTING.md (or adopting ruff-format's style wholesale) — out of scope here.


CONTRIBUTING.md
```diff
@@ -171,6 +171,43 @@ line.
...
+```bash
+uvx pre-commit install      # or: pre-commit install
+```
+
+Thereafter `ruff` (lint) plus a few file-hygiene hooks run on your staged
+files at commit time. The hooks see only the files you're changing, so the
+codebase cleans up file-by-file rather than in one sweep. To run manually:
+
+```bash
+uvx ruff check .            # lint the whole tree
```
SoymilkWinsAgain: From CodeX:

It is recommended to change `uvx ruff check .`.

This command uses the remote uvx version that was resolved at the time, which may not be equal to `rev: v0.15.16` in `.pre-commit-config.yaml`, nor may it be equal to the ruff version in `uv.lock`. This weakens the "single source version/keep synchronized" design in the PR. A more reliable approach would be to write `uv run ruff check .`, since the previous requirement was `uv sync --extra test --extra dev`, which will use the lock file version.

.github/workflows/ci.yml
```diff
@@ -7,6 +7,44 @@ on:
     branches: [main]
 
 jobs:
```
SoymilkWinsAgain, line 9: From CodeX:

The new lint job currently only consumes the ruff revs in `.pre-commit-config.yaml`, but does not validate the pre-commit configuration itself.

If subsequent non-ruff hooked YAML or configurations are corrupted, CI may still pass. It is recommended to add a lightweight step to the lint job, such as `uvx pre-commit validate-config`. This can be blocking because it only validates the configuration structure and does not trigger the current full-scale lint debt.

SoymilkWinsAgain (changes requested):

CONTRIBUTING.md
jhuggins: Done in 90746e7 — switched to `uv run ruff check .`, which uses the uv.lock-pinned ruff (the dev env is synced per the Installation section). This matches the source-of-truth note already in `.pre-commit-config.yaml`. The CI lint job separately derives its ruff version from the pre-commit rev (7e8f784), so all three — CI, pre-commit, and local docs — now resolve to the same pin.

.github/workflows/ci.yml
jhuggins: Added in 90746e7 — a blocking `uvx pre-commit validate-config` step runs before the (advisory) ruff check. It validates the whole config structure, including the non-ruff hooks, but runs no hook, so it doesn't surface the lint backlog. A corrupted `.pre-commit-config.yaml` now fails CI.

SoymilkWinsAgain (approved):
Looks good. But there are code conflicts

PR #251, opened 2026-06-10
ci: add advisory pyright type-check job

jhuggins:
## Summary

Adds **pyright type checking in advisory mode** — C6 from the CI-tightening plan, the separate follow-up to #250 (ruff + pre-commit).

Same finding as ruff: the source carries a type-debt baseline — **324 issues** at pyright's `basic` mode with the scoped config — much of it JAX/TFP untyped-attribute noise (`reportAttributeAccessIssue` 121, `reportOperatorIssue` 39, `reportArgumentType` 76) rather than real bugs. So pyright lands **advisory** (reports, doesn't gate), to be tightened as the baseline burns down.

### CI: `typecheck (advisory)` job (`.github/workflows/ci.yml`)
```yaml
- name: Install dependencies
  run: uv sync --frozen --extra dev --extra nutpie --extra pymc
- name: Pyright (advisory — does not gate merges)
  continue-on-error: true
  run: uv run --with 'pyright[nodejs]==1.1.410' pyright
```
- Syncs the full dependency env first so imports resolve against the **real** jax/tfp/pymc (otherwise pyright drowns in unresolved-import noise).
- `pyright[nodejs]` bundles the Node runtime → no dependency on a runtime download in CI.
- Version pinned (`1.1.410`) for a reproducible baseline.
- `continue-on-error: true` → advisory. **To enforce later:** drop that line and tighten `typeCheckingMode`.

### `pyrightconfig.json` (new)
- `include: ["probpipe"]` — scoped to shipped code; tests/docs out of scope until the source baseline is clean.
- `typeCheckingMode: "basic"`, `pythonVersion: "3.12"`.
- `reportMissingTypeStubs: "none"` — jax / tfp-nightly / pymc ship incomplete or absent stubs; that noise isn't actionable (drops the count 366 → 324).
- `venvPath`/`venv` point at uv's `.venv` so import resolution works locally and in CI.

### Docs
- CONTRIBUTING.md: new **"Type checking"** subsection.
- CHANGELOG: entry under `[Unreleased]`.

## Linked issue(s)

Follows #250 (ruff + pre-commit) — the agreed "C6 as a separate PR" split. Same advisory-now/enforce-later philosophy.

## Test plan

Verified locally:
- `uv run --with 'pyright[nodejs]==1.1.410' pyright` resolves imports against the synced env and reports **324** issues with this config (vs 366 unscoped / stubs-on) — confirms the config applies and pyright reads it.
- `uv.lock` is **unchanged** — pyright is an ephemeral `--with` tool, not a project dependency.
- ci.yml YAML validated.
- CI on this PR runs the job end-to-end (goes green with the report, by design).

## Breaking changes

None. The job doesn't gate; no runtime/API change; pyright is not added to the locked dependency set.

## Documentation

- [x] No docstrings touched (no source changes).
- [x] CONTRIBUTING.md "Type checking" subsection.
- [x] CHANGELOG entry.

## Checklist

- [x] PR title `<type>(<scope>): <subject>` — `ci: ...`.
- [x] `area:infrastructure` label.
- [x] No linked issue — standalone CI hygiene (CONTRIBUTING small-fix exception); follows #250.

## Notes for reviewers

- **Expected trivial conflicts with #250**, since both are branched from `main` and touch the same three files additively:
  - `ci.yml`: #250 adds a `lint` job, this adds a `typecheck` job, both right after `jobs:`. Resolution: keep both.
  - `CONTRIBUTING.md`: #250 adds "Linting & pre-commit", this adds "Type checking", both before "### Documentation". Keep both.
  - `CHANGELOG.md`: both add a bullet under `[Unreleased]`. Keep both.

  Whichever merges second takes a ~30-second keep-both rebase. Kept independent (rather than stacked) so they can merge in either order.
- **Why scope to `probpipe` only:** type-checking tests adds noise with little safety value; the shipped package is what users type against. Expand to tests later if desired.
- **Why basic mode:** `standard`/`strict` would multiply the baseline. Start permissive, ratchet — the same path scikit-learn and others took.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
         uses: actions/checkout@v6
 
       - name: Install uv
-        uses: astral-sh/setup-uv@v8
+        uses: astral-sh/setup-uv@v7
         with:
           enable-cache: true
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,9 +13,14 @@ jobs:
   # unrelated work; advisory surfaces it without forcing a cleanup. Syncs
   # the full dependency env first so imports resolve against the real
   # jax/tfp/pymc. Config + scope live in pyrightconfig.json. To enforce
-  # later: drop continue-on-error and tighten typeCheckingMode. Pyright is
-  # pinned for a reproducible baseline; `[nodejs]` bundles the Node runtime
-  # so CI doesn't depend on a runtime download.
+  # later: drop continue-on-error and tighten typeCheckingMode.
+  #
+  # Pyright runs as an ephemeral `uv run --with` tool rather than a locked
+  # [dev] dependency: that keeps its per-platform Node binaries out of
+  # uv.lock. The version is pinned here (and referenced in CONTRIBUTING.md)
+  # for a reproducible baseline; `[nodejs]` bundles the Node runtime so CI
+  # doesn't depend on a runtime download. The final summary line is lifted
+  # into the job summary so the count is visible without opening the log.
   typecheck:
     name: typecheck (advisory)
     runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


SoymilkWinsAgain (approved):
This looks good

PR #259, opened 2026-06-11
feat(inference): jax-native NLE/NRE via BayesFlow (rescoped SBI step 6)

jhuggins:
## Summary

Rescoped **Step 6** of the SBI plan: **jax-native NLE and NRE** on the merged BayesFlow backend (#246), sampled by the existing BlackJAX/TFP NUTS machinery — replacing the originally planned sbi-torch default path with **zero new dependencies, zero new samplers, and zero registry changes**. Gated and de-risked by the Step-6a spike and Phase-0 gates (G1/G2/G3 all PASS): BayesFlow's keras-on-jax learned densities/ratios are `jax.grad`-transparent via a traceable bypass, finite-difference-exact, and NUTS recovers analytic posteriors — including discrete-observation + constrained-parameter cases for NRE.

## What's added

**`probpipe/inference/_bayesflow_likelihoods.py`:**

- `learn_amortized_likelihood(prior, simulator, *, num_simulations=, epochs=, batch_size=, sim_backend=, inference_network=None, dequantize=False, random_seed=, optimizer=, **fit_kwargs) -> BayesFlowLikelihood` — trains a conditional `CouplingFlow` for `p(y | θ)` on offline `(θ, y)` simulations (the same pipeline, adapter-namespace isolation, and RNG bracketing as NPE).
- `learn_amortized_ratio(...) -> BayesFlowRatio` — trains an NRE-C classifier (`RatioApproximator`; contrastive pairs are built internally by batch-shuffling θ, so the training data are identical to NLE's).
- Both wrappers implement **`ConditionallyIndependentLikelihood`** (+ `GenerativeLikelihood` passthrough): trained per-datum, summed over rows — so **datasets of any size work natively** (NPE's conditioning shape is fixed at training time), and `per_datum_log_likelihood` comes for free.
- The traceable score path is the spike-verified bypass (`standardizer.maybe_standardize` + `inference_network.log_prob` / `logits`), faithful to the public host-bound APIs and pure keras-jax ops end to end — `SimpleModel(prior, learned)` + `condition_on(model, data)` just works.
- **`dequantize=True` for integer-valued observations** — uniform dequantization (Theis et al. 2016, arXiv:1511.01844; Ho et al. 2019, Flow++ §3.1, arXiv:1902.00275): training adds `U[0,1)^d` jitter to the simulated `y` (the user's simulator stays untouched), and the wrapper scores integer data at the unit-cell midpoint `y + ½`, approximating the implied pmf `P(y|θ) = ∫_{[y,y+1)^d} p(u|θ)du`.

## Design points worth a look

- **Raw constrained θ as network input.** Unlike NPE (θ is the modeled *output*, needing bijector reparameterization), NLE/NRE consume θ as an input — `_simulate_offline` gains a `bijectors=None` mode returning raw constrained draws, matching what the MCMC log-density assembly passes at sampling time. Discrete-valued parameter fields are accepted (input, not density).
- **No d_y=1 NLE fallback — by measurement.** Pre-checked that `FlowMatching.log_prob` is **not reverse-mode differentiable** (adaptive ODE via dynamic-bound `while_loop`), so unlike NPE's sampling-only d=1 fallback, there is none for gradients: NLE rejects 1-D observations under the default net with an error pointing at `learn_amortized_ratio` (the MLP classifier has no minimum dimension and handles discrete observations — G3 verified Gamma–Poisson to 0.005 post-std).
- **Discrete observations, measured not folklore.** Fitting a continuous flow to atoms is ill-posed (the MLE collapses onto the data points); empirically the raw fit's posterior overdisperses as observations concentrate on few values (std ratios 1.13–1.27 across seeds on an atom-heavy Gamma–Poisson check) while `dequantize=True` calibrates it (0.93–0.96, mean err 0.11–0.17 posterior sds). The jitter rides a `fold_in` stream, so the existing simulation key path — and every previously measured tolerance — is untouched. NRE remains the no-flag native option for discrete or mixed rows (cf. MNLE, Boelts et al. 2022, for the torch-sbi approach).
- **`BayesFlowRatio` semantics documented loudly:** summed per-row log-ratios equal the joint log-likelihood up to a θ-independent constant — valid for conditioning (the constant cancels in MCMC), **invalid** for model comparison / LOO / absolute-likelihood reads.
- **Plain functions, deliberately not `@workflow_function`.** The workflow result boundary coerces returns into `Record`/`Distribution`, which would wrap (and break) `Likelihood` components — discovered the hard way when every fixture-based test failed with `Record(learn_amortized_likelihood=...)`. Same precedent as the rwmh/elliptical_slice de-workflowing (#203 convention); now codified as the §1.4 exception in STYLE_GUIDE.
- **Module structure made honest:** the BayesFlow code is now three modules — `_bayesflow_common.py` (the shared bridge: lazy import, validation, seeded training, offline simulation, adapter keying), `_bayesflow_posteriors.py` (NPE/FMPE/CMPE + `BayesFlowModel`; `git mv` from `_bayesflow.py`, history preserved), and `_bayesflow_likelihoods.py` (NLE/NRE) — no sibling cross-imports. `learn_amortized_posterior` is refactored onto the shared helpers with behavior unchanged (its 43-test suite, now `test_bayesflow_posteriors.py`, unchanged-green).

## Tests (`tests/inference/test_bayesflow_likelihoods.py`, 25 passing)

Bypass **faithfulness** vs public `log_prob`/`log_ratio`; **FD-verified, jit-stable gradients** (both wrappers); **end-to-end NUTS calibration** against analytic conjugate posteriors — NLE n=1, NLE n=8 (size generalization: trained on single rows, conditioned on eight — posterior matches the analytic 8-observation posterior), NRE n=1; a **constrained Gamma prior** through the registry's natural-space walk; **dequantized NLE on integer counts** against the analytic Gamma–Poisson posterior plus an exact shift-identity check against a non-dequantized wrapper of the same approximator; CIL membership + exact row-sum consistency; record/flat params coercion (the MCMC helper passes flat vectors, predictive paths pass records); `generate_data` passthrough; data/params width guards; 8 validation rejections; same-seed **training determinism**.

## Packaging / CI / docs

No new dependencies (rides the existing `[bayesflow]` extra and lockfile). The CI `bayesflow` job now runs both test files. `docs/api/inference.md`'s Amortized SBI section documents all three flavors (NPE / NLE / NRE) with the NRE caveat and the graded discrete-observation guidance; strict mkdocs build green. CHANGELOG updated. The measured-tolerance testing discipline used here (and in #246) is now codified in **STYLE_GUIDE §8.6** (independent baselines, location+spread checks, seed-measured documented bounds), with a pointer from CONTRIBUTING.

## Verification

Local (bayesflow 2.0.12 / keras 3.14.1-jax / jax 0.10.1): **25 new + 43 existing bayesflow tests pass**; ruff clean on changed files (zero delta on touched pre-existing files); `import probpipe` still loads neither keras nor bayesflow; strict docs build green, rendered API tables verified. Spike + gate evidence (scripts and numbers) archived in the planning memos; the notebook-restoration workstream (flexible_inference rewrite incl. an NRE-on-Poisson-counts section, `05_external_backends` decision-table rows) follows as its own PR per the plan.


CONTRIBUTING.md
SoymilkWinsAgain: From CodeX:

This section describes `learn_amortized_posterior` -> `BayesFlowPosterior`, but the current public class name is `BayesFlowModel`. The same section can also briefly explain: `NPE/BayesFlowModel` bypasses the registry through `SupportsConditioning`; `NLE/NRE`, on the other hand, uses the existing MCMC registry as a `SimpleModel` likelihood.

probpipe/inference/_bayesflow_likelihoods.py
```diff
@@ -0,0 +1,567 @@
...
+            t = params
+        t = jnp.ravel(jnp.asarray(t))
+        if t.shape[0] != self._theta_dim:
+            raise ValueError(
+                f"params has {t.shape[0]} values but the estimator was trained on "
+                f"{self._theta_dim}-dimensional parameters."
+            )
+        return t
+
+    def _data_rows(self, data: ArrayLike | np.ndarray) -> Array:
```
SoymilkWinsAgain, line 138: It will reject one-dimensional scalar observation datasets for `data_dim == 1`. For example, inputting `jnp.array([1., 2., 3.])` will be transformed into (1, 3) by `jnp.atleast_2d`, subsequently triggering the "trained on observations of size 1" error. Users must manually pass (n, 1) to bypass this, but this is not required by the API documentation.

probpipe/inference/_bayesflow_likelihoods.py
```diff
@@ -0,0 +1,567 @@
...
+            # "extra" stream would alias the simulation keys. Splitting only
+            # under the flag keeps the dequantize=False key path -- and every
+            # tolerance measured on it -- bit-identical.
+            key, k_jitter = jax.random.split(key)
+        named, y = _simulate_offline(
+            prior, simulator, num_simulations, key,
+            sim_backend=sim_backend, bijectors=None,   # raw theta: it is a net input
+        )
+        if dequantize:
+            y = np.asarray(jnp.asarray(y) + jax.random.uniform(k_jitter, y.shape),
```
SoymilkWinsAgain, line 361: It seems that this branch doesn't have the 2**23 upper bound.

probpipe/inference/_bayesflow_likelihoods.py
```diff
@@ -0,0 +1,567 @@
...
+        untouched and keeps emitting raw integers), making the target
+        absolutely continuous, and the returned wrapper scores integer data at
+        the unit-cell midpoint ``y + 1/2``, approximating the implied pmf
+        ``P(y | theta) = integral over [y, y+1)^d of p(u | theta) du``. This is
+        the fixed-``q`` special case of variational dequantization: the
+        cell-integral identity is exact, and the training objective is its
+        Jensen lower bound (Theis et al., 2016, arXiv:1511.01844; Ho et al.,
+        2019 "Flow++", arXiv:1902.00275, section 3.1). Pass raw integers as
+        data; do **not** pre-jitter or pre-shift. Counts must stay below
+        ``2**23`` (~8.4e6): the pipeline is float32, whose spacing reaches 1.0
```
SoymilkWinsAgain, line 446: Maybe we can add a test for this?

SoymilkWinsAgain (changes requested):

CONTRIBUTING.md
jhuggins: Fixed in e484ff1: BayesFlowPosterior -> BayesFlowModel (stale since the #246 rename), and the paragraph now states both routes -- NPE short-circuits the registry via SupportsConditioning, while the NLE/NRE likelihoods ride the existing registry as ConditionallyIndependentLikelihood components inside SimpleModel (typically blackjax_nuts; no new registry entries).

probpipe/inference/_bayesflow_likelihoods.py
jhuggins: Good catch -- fixed in e484ff1. For d_y == 1 a 1-D array is now read as n scalar observations (unambiguous there), so jnp.array([1., 2., 3.]) scores as 3 rows; (n, 1) is unchanged and d_y > 1 keeps the strict reading. Added a shape-semantics test (tiny untuned NRE on scalar observations) asserting the row-sum identity and 1-D == (n, 1) equivalence.

probpipe/inference/_bayesflow_likelihoods.py
jhuggins: Right -- the bound was documented but unenforced. e484ff1 adds an eager check on the simulated observations before the jitter (concrete numpy at that point, so it can raise cleanly): |y| >= 2**23 with dequantize=True now raises with a rescale-or-use-NRE message. Scoring-time data cannot be checked under trace, but data far outside the trained regime is out of contract anyway.

probpipe/inference/_bayesflow_likelihoods.py
jhuggins: Added in e484ff1: test_dequantize_rejects_counts_at_float32_cell_limit (TestValidation) -- a simulator emitting 2**23 counts trips the new guard before any training runs.

SoymilkWinsAgain (approved):
Looks good

PR #260, opened 2026-06-12
refactor(core): Isolate user namespaces

### LINKED ISSUE #254: Separate `WorkflowFunction` control options from user function kwargs
### Summary

`WorkflowFunction` should be able to wrap ordinary Python functions with minimal naming restrictions. Users should not have to avoid common parameter names just because they overlap with `ProbPipe` execution controls.

### Motivation

`workflow_function` currently uses ordinary keyword arguments for both `ProbPipe` control options and user function inputs.

At construction/decorator time:

```python
@workflow_function(dispatch="jax", seed=114514, name="score")
def score(x):
    ...
```

keywords such as `name`, `dispatch`, `seed`, `n_broadcast_samples`, `workflow_kind`, and `include_inputs` are reserved for `WorkflowFunction` configuration. This means users cannot conveniently bind same-named function parameters through the decorator API without falling back to less ergonomic mechanisms such as `bind={...}`.

At call time, the problem is more severe. `WorkflowFunction` currently reserves names such as `n_broadcast_samples`, `seed`, `include_inputs`, etc., for call-time overrides. Wrapped functions **cannot** declare these as normal parameters, even though names like seed are common in modeling and simulation APIs. Furthermore, because users have considerable freedom in actual use, it is difficult to debug during the testing phase.

In one sentence, this creates a namespace collision between `ProbPipe`’s control plane and the user’s function signature. 

`WorkflowFunction` should be able to wrap ordinary Python functions with minimal naming restrictions. Users should not have to avoid common parameter names just because they overlap with `ProbPipe` execution controls.

The API should preserve the simple decorator style while providing an explicit, ergonomic place for `ProbPipe`-specific options.

### Proposed Design

Introduce an explicit options API that separates `ProbPipe` controls from user kwargs.

Construction-time options:

```python
@workflow_function(
    name="score_model",
    dispatch="jax",
    n_broadcast_samples=1_000,
    seed=114514,
)
def score_model(x, seed, name):
    ...
```

Call-time overrides:

```python
result = score_model.with_options(
    seed=1919810,
    n_broadcast_samples=5_000,
    include_inputs=True,
)(x, seed=user_seed, name=model_name)
```

Core rule:

- `ProbPipe` controls live only in `@workflow_function(...)` and `@workflow_function.with_options(...)`.
- Common names such as `seed`, `name`, `dispatch`, etc. should be legal user function parameters.
- Bare decorator usage remains unchanged.

### Plan

1. Change `workflow_function(...)`
  - Equivalent to current `workflow_function(...)` configuration path.
  - Keep `@workflow_function` bare usage unchanged.
  - Prefer this new API in docs and examples.
2. Add `WorkflowFunction.with_options(...)`
  - Returns a lightweight callable wrapper or cloned `WorkflowFunction` with temporary call options.
  - Does not consume ordinary kwargs from the eventual user function call.
3. Update call resolution
  - Stop treating `seed`, `n_broadcast_samples`, and `include_inputs` as always-reserved user parameter names.
  - During transition, if the wrapped function explicitly declares one of these names, ordinary `wf(seed=...)` should bind to the user function.
  - `ProbPipe` call-time overrides should be expressed through `wf.with_options(...)`.
4. Deprecate old control kwargs
5. Remove reserved-name validation later

### Tasks

- [x] Add new API
- [x] Deprecate old control kwargs
- [ ] Remove reserved-name validation

SoymilkWinsAgain:
## Summary

This PR separates `ProbPipe` workflow controls from wrapped-function keyword arguments.

The new public API is:

- `@workflow_function` for the common no-options decorator case.
- `@workflow_function(...)` for definition-time `ProbPipe` controls such as `dispatch`, `seed`, `max_workers`, `n_broadcast_samples`, and `workflow_kind`.
- `workflow.with_options(...)(...)` for one-call controls such as `seed`, `n_broadcast_samples`, and `include_inputs`.

As a result, wrapped functions can safely use common parameter names such as `seed`, `dispatch`, `name`, `n_broadcast_samples`, and `include_inputs` without those names being intercepted by `WorkflowFunction`. Legacy decorator options and legacy call-time overrides remain temporarily supported with `DeprecationWarning` where they are unambiguous.

## Linked issue(s)

Refs: #254

## Test plan

- Added focused namespace-separation coverage:
  - `tests/core/test_workflow_options_namespace.py`
- Updated existing call-resolution and broadcasting tests for the new options API.
- Ran:
  - `pytest tests/core/test_workflow_options_namespace.py -x -v -p no:xdist -o "addopts="`
  - `pytest tests/core/test_workflow_call_resolution.py tests/core/test_broadcasting.py -x -v -p no:xdist -o "addopts="`
  - `ruff check probpipe/core/node.py probpipe/core/_workflow_call.py probpipe/core/ops.py probpipe/modeling/_glm.py tests/core/test_workflow_options_namespace.py tests/core/test_workflow_call_resolution.py tests/core/test_broadcasting.py tests/core/test_record_array_view.py tests/core/test_recordarray_broadcasting.py tests/inference/test_inference.py`
  - `git diff --check`

## Breaking changes

**Yes.**

Workflow call keyword arguments now belong to the wrapped user function whenever they can bind to that function. Use `workflow.with_options(...)(...)` for ProbPipe call-time controls.

Migration examples:

- `wf(x, n_broadcast_samples=10)` -> `wf.with_options(n_broadcast_samples=10)(x)`
- `wf(x, seed=114514)` -> `wf.with_options(seed=114514)(x)`

Legacy forms still emit `DeprecationWarning` during the transition window when they can be interpreted safely.

## Documentation

## Checklist

## Tasks

- [x] Add new API
- [ ] Remove reserved-name validation


probpipe/core/_workflow_call.py
```diff
@@ -1,19 +1,20 @@
...
 from __future__ import annotations
 
 import inspect
+import warnings
 from collections.abc import Callable, Mapping
 from dataclasses import dataclass
 from typing import Any, get_type_hints
 
-RESERVED_WORKFLOW_CALL_NAMES = frozenset(
+WORKFLOW_CALL_OPTION_NAMES = frozenset(
```
jhuggins: derive from `WorkflowCallOptions` rather than having to keep the two in sync

jhuggins (changes requested):

SoymilkWinsAgain:
Change from `@workflow_function.options(...)` to `@workflow_function(...)`.

jhuggins (approved):

PR #263, opened 2026-06-12
ci: fix Codecov misreporting on targeted PRs (#261)

### LINKED ISSUE #261: Codecov can fail on ordinary PRs that run the changed-files-only CI path
## Problem

Codecov can fail on ordinary PRs that run the changed-files-only CI path, even when all GitHub Actions jobs pass.

Example: PR #260 (`dev/isolate-namespaces`) had all Actions jobs green, but both external Codecov checks failed:

- `codecov/patch`: 60.93% of diff hit, target 70.00%
- `codecov/project`: 38.53%, a large apparent drop from base
- Codecov also reported: `HEAD has 1 upload less than BASE`

The coverage result is misleading rather than a real test failure.

## Root cause

The main test job only uploads coverage to Codecov when the full-suite path runs:

```yaml
if: matrix.python-version == '3.12' && steps.changed.outputs.run_all == 'true'
```

That means targeted PRs can upload only the BayesFlow coverage report, without the primary test coverage report. Codecov then computes project and patch coverage from an incomplete report set.

This appears to be an interaction between the CI hygiene coverage policy from #249 and the BayesFlow coverage upload added in #246.

## Proposed fix

Make the Codecov report set consistent.

Recommended approach:

- On pull requests, make the Python 3.12 main test job run the full suite and upload coverage.
- Keep Python 3.13 and 3.14 on the changed-files-only targeted path to preserve most of the matrix speed benefit.
- Keep the BayesFlow 3.12 coverage upload, so Codecov receives the same two report classes on PR heads as it does on base/main.

This makes Python 3.12 the coverage authority for PRs, while the other Python versions continue to provide compatibility signal cheaply.

Alternative approaches:
- Gate the BayesFlow coverage upload behind the same condition as the main coverage upload. This avoids incomplete Codecov reports, but may leave Codecov checks absent on targeted PRs if Codecov is required.
- Add a more explicit Codecov flag/carryforward configuration. This is more complex and likely unnecessary for now.

## Acceptance criteria

- A PR that does not touch foundational files still uploads primary Python 3.12 coverage to Codecov.
- Codecov no longer reports HEAD has 1 upload less than BASE for targeted PRs.
- codecov/project compares full primary coverage plus BayesFlow coverage, not BayesFlow alone.
- Python 3.13 and 3.14 can remain targeted on PRs.
- Pushes to main continue to run the full suite and enforce the coverage floor.

jhuggins:
Closes #261.

## Investigation verdict (the issue is half-right)

The mechanism is **real and reproducible**, but it's a **coverage-misreporting defect, not a merge-blocking bug**:

- **Confirmed in config:** the main test job uploads coverage only on `run_all=='true'`, while the BayesFlow job uploaded on every `3.12` run. So a *targeted* PR sent Codecov **only** the BayesFlow report (which runs `test_bayesflow.py` under `--cov=probpipe` → ~38% project), and "HEAD has 1 upload less than BASE" is literally `2 − 1`.
- **Reproduced by #260:** every Actions job green, only `codecov/patch` (60.93%) + `codecov/project` (38.53%) red. #260's diff is pure core/workflow code (`node.py`, `ops.py`, `_workflow_call.py`, `_glm.py`) with **zero** BayesFlow — yet the only uploaded report was BayesFlow's, so the diff read as uncovered.
- **Not blocking:** there are no required status checks and no rulesets on `main` (verified via the API), so the red `codecov/*` X's can't block a merge — #260's `BLOCKED` state is from the PR-review requirement, not Codecov. So the real cost is **trust erosion** (scary numbers on healthy PRs), not a broken gate.

**On the patch question specifically:** patch coverage was *not* correct either. Patch is scoped to the diff lines, but "covered" still comes from the uploaded reports — and the BayesFlow-only report doesn't exercise a non-BayesFlow diff, so those lines read uncovered (hence #260's 60.93%). Carryforward alone wouldn't fix patch, because the base commit has no coverage for newly-added lines; you need the report from the run that actually exercised the changed lines.

## Fix (preserves the changed-files-only speedup — no full suite on every PR)

1. **Main test job uploads coverage on the targeted path too** (gate relaxed `run_all=='true'` → `skip!='true'`), tagged flag `unit`. The targeted run covers exactly the changed files → **patch is accurate** and stays the enforced PR gate (`target: 70%`).
2. **`codecov.yml` project → `informational: true`.** A targeted PR's partial report can't give an authoritative project number, so it must not gate. The real **88% project floor is still enforced in-CI** on the full-suite run (`pytest --cov-fail-under=88`), which fires on pushes to `main` and on foundational PRs — i.e., coverage is enforced exactly when the full suite runs, as requested.
3. **BayesFlow leg gated** with the same RUN_ALL-style logic as the test job (push to main / foundational change / a BayesFlow file changed); otherwise skipped, and its flag carries forward.
4. **Per-flag `carryforward`** for `unit` and `bayesflow` so the project number stays sane and the "uploads less than BASE" warning stops.

Net: **patch** is the accurate, enforced PR gate; **project** is informational on PRs and a hard in-CI gate on full-suite runs.

## Test plan

- `ci.yml` and `codecov.yml` YAML validated locally.
- The in-job `--cov-fail-under=88` (full suite) / `=0` (targeted) split is preserved unchanged.
- Real verification is this PR's own Codecov checks: as a targeted PR, `codecov/patch` should now reflect the changed files accurately and `codecov/project` should appear as an informational (non-failing) status.

## Breaking changes

None — CI/coverage-reporting only.

## Documentation

- [x] CHANGELOG entry under `[Unreleased] / ### Fixed`.
- [x] Rationale captured inline in `codecov.yml` and the workflow comments.

## Checklist

- [x] Title `<type>(<scope>): <subject>` — `ci: ...`.
- [x] `area:infrastructure` auto-label applies (touches `.github/**`, `codecov.yml`).
- [x] Closes #261.

## Notes for reviewers

- **Honest caveat:** on a *targeted* PR the partial `unit` upload means Codecov's **project** number still shows low — that's why project is `informational`. It's not "fixed to look right" on targeted PRs; it's correctly marked non-authoritative there, with the real enforcement on full-suite runs. Patch is the metric that's both correct and gating per-PR.
- If you'd additionally like the full-suite coverage to be a **hard merge-blocker** (not just a failing job), that's a one-time branch-protection change (add the `test (3.12/3.13/3.14)` checks to required status checks) — say the word and I'll set it via `gh api`; it's outside this PR since it's a repo setting, not a file.
- `@SoymilkWinsAgain`'s original recommendation (run the full 3.12 suite on every PR) also works and gives a real project number per PR, but costs the full-suite run on every PR; this approach keeps the targeted-test speedup and makes patch the per-PR gate instead.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -310,11 +310,18 @@ jobs:
           uv run pytest --cov-fail-under=88 --cov-report=xml --cov-report=html
 
       - name: Upload coverage reports to Codecov
-        if: matrix.python-version == '3.12' && steps.changed.outputs.run_all == 'true'
+        # Upload on 3.12 for BOTH the targeted and full-suite paths (gated only
+        # on the test step having run). The targeted run covers exactly the
+        # changed files — which is what Codecov *patch* coverage needs — so the
+        # patch gate is accurate even on targeted PRs. The full-suite path
+        # additionally gives the authoritative *project* number. Tagged `unit`
+        # so the flag carries forward when a PR doesn't run it (see codecov.yml).
+        if: matrix.python-version == '3.12' && steps.changed.outputs.skip != 'true'
         uses: codecov/codecov-action@v7
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           slug: TARPS-group/prob-pipe
+          flags: unit
 
       - name: Upload coverage HTML report
         if: always() && matrix.python-version == '3.12' && steps.changed.outputs.run_all == 'true'
@@ -337,13 +344,52 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 0   # need full history for git diff
+
+      - name: Check if BayesFlow-relevant files changed
+        id: bf_check
+        # Gate the leg with the same logic the test job uses for RUN_ALL: run
+        # on pushes to main, on foundational changes (deps / this workflow),
+        # or when a BayesFlow file itself changed. Otherwise skip — and the
+        # `bayesflow` coverage flag carries forward from base (codecov.yml),
+        # so a skipped leg doesn't drag the coverage union down.
+        run: |
+          if [ "${{ github.event_name }}" = "pull_request" ]; then
+            BASE="origin/${{ github.base_ref }}"
+          else
+            BASE="${{ github.event.before }}"
+          fi
+          CHANGED=$(git diff --name-only --diff-filter=ACMRT "$BASE"...HEAD \
+            -- 'probpipe/inference/_bayesflow.py' 'tests/inference/test_bayesflow.py' || true)
+          CHANGED_ALL=$(git diff --name-only --diff-filter=ACMRT "$BASE"...HEAD || true)
+          RUN_BF=false
+          if [ "${{ github.event_name }}" = "push" ]; then
+            RUN_BF=true
+          else
+            for f in $CHANGED_ALL; do
+              case "$f" in
+                pyproject.toml|uv.lock|.github/workflows/ci.yml) RUN_BF=true ;;
+              esac
+            done
+            [ -n "$CHANGED" ] && RUN_BF=true
+          fi
+          if [ "$RUN_BF" = "true" ]; then
+            echo "skip=false" >> "$GITHUB_OUTPUT"
+            echo "Running the BayesFlow leg."
+          else
+            echo "skip=true" >> "$GITHUB_OUTPUT"
+            echo "No BayesFlow-relevant changes — skipping (coverage flag carries forward)."
+          fi
 
       - name: Install uv
+        if: steps.bf_check.outputs.skip != 'true'
         uses: astral-sh/setup-uv@v5
         with:
           enable-cache: true
 
       - name: Install dependencies
+        if: steps.bf_check.outputs.skip != 'true'
         # Versions (jax/jaxlib/tfp-nightly/keras/bayesflow) are pinned in uv.lock;
         # --frozen fails if the lockfile drifts. --python pins the interpreter to
         # the matrix leg. The bayesflow extra is locked behind a python_version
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -337,32 +344,73 @@ jobs:
...
+        # `bayesflow` coverage flag carries forward from base (codecov.yml),
+        # so a skipped leg doesn't drag the coverage union down.
+        run: |
+          if [ "${{ github.event_name }}" = "pull_request" ]; then
+            BASE="origin/${{ github.base_ref }}"
+          else
+            BASE="${{ github.event.before }}"
+          fi
+          CHANGED=$(git diff --name-only --diff-filter=ACMRT "$BASE"...HEAD \
+            -- 'probpipe/inference/_bayesflow.py' 'tests/inference/test_bayesflow.py' || true)
```
SoymilkWinsAgain: The BayesFlow-related paths here are hardcoded with the old single-file layout, which is incompatible with the ongoing BayesFlow split/NLE-NRE branch and could lead to similar skip errors in the future. 

Suggest change: `probpipe/inference/_bayesflow*.py`, `tests/inference/test_bayesflow*.py`. Then run `pytest tests/inference/test_bayesflow*.py`.

.github/workflows/ci.yml
```diff
@@ -337,32 +344,73 @@ jobs:
...
+        if: steps.bf_check.outputs.skip != 'true'
         # Versions (jax/jaxlib/tfp-nightly/keras/bayesflow) are pinned in uv.lock;
         # --frozen fails if the lockfile drifts. --python pins the interpreter to
         # the matrix leg. The bayesflow extra is locked behind a python_version
         # < 3.14 marker (BayesFlow caps <3.14), so it resolves only on these legs.
         run: uv sync --frozen --python ${{ matrix.python-version }} --extra dev --extra nutpie --extra bayesflow
 
       - name: Run BayesFlow backend tests
+        if: steps.bf_check.outputs.skip != 'true'
         run: uv run pytest tests/inference/test_bayesflow.py --cov-report=xml
```
SoymilkWinsAgain: Same as above

.github/workflows/ci.yml
```diff
@@ -337,32 +344,73 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 0   # need full history for git diff
+
+      - name: Check if BayesFlow-relevant files changed
+        id: bf_check
+        # Gate the leg with the same logic the test job uses for RUN_ALL: run
```
SoymilkWinsAgain, line 352: From CodeX:

The comment states that BayesFlow gating uses similar logic to the RUN_ALL of the test job, but the actual foundational list is narrower: it lacks `setup.cfg`, `setup.py`, `tests/conftest.py`, and `probpipe/__init__.py`. Changes to these files may affect the BayesFlow test environment or public export, but the current BayesFlow leg will skip them.

SoymilkWinsAgain (changes requested):

.github/workflows/ci.yml
jhuggins: Done in 2c4c47b — gating now matches `probpipe/inference/_bayesflow*.py` and the test run is `pytest tests/inference/test_bayesflow*.py`, so the NLE/NRE split is picked up automatically.

.github/workflows/ci.yml
jhuggins: Done in 2c4c47b — `pytest tests/inference/test_bayesflow*.py` (glob) so split test files are all collected.

.github/workflows/ci.yml
jhuggins: Good catch — aligned the foundational set with the test job's RUN_ALL in 2c4c47b: added `setup.cfg`, `setup.py`, `tests/conftest.py`, and `probpipe/__init__.py`, so the comment is now accurate and those files trigger the BayesFlow leg.

SoymilkWinsAgain (approved):
Everything looks great. I think it's ready for merge

PR #272, opened 2026-06-14
refactor(ci): deduplicate the AST import-graph builder shared by the test + notebooks jobs

### LINKED ISSUE #266: Deduplicate the AST import-graph builder shared by the test and notebooks CI jobs
## Summary

`.github/workflows/ci.yml` carries **two near-identical inline Python heredocs** that build a reverse-dependency import graph over `probpipe/` via AST parsing:

- the **`test`** job's `/tmp/expand_deps.py` — maps changed source files to the test modules that transitively import them, so a change to `provenance.py` also runs `tests/test_record.py`.
- the **`notebooks`** job's `/tmp/affected_notebooks.py` — the same reverse-dependency walk, used to pick which notebooks to execute when source changes.

Both define the same helpers (`file_to_module`, `module_to_file`), build the same `rdeps` reverse-dependency map with the same `ast.walk` over `ImportFrom`/`Import` nodes (including the relative-import level handling), and do the same BFS over the dependency graph. They have drifted into two copies of one algorithm.

## Why it matters

A fix or improvement to the import-graph logic — e.g. handling a relative-import edge case, namespace packages, or `import probpipe.x as y` aliasing — must be applied to **both** heredocs. If one lags, the test job and the notebook job silently disagree about which artifacts a given source change affects: a source edit could run the right tests but skip the notebook that depends on it (or vice versa), and CI would look green while under-covering the change.

This is pre-existing (it predates the docs/tutorials generalization in #265, which only re-globbed the notebook copy), so it is filed separately rather than folded into that PR.

## Proposed fix

Extract the shared import-graph builder into a single committed script — e.g. `scripts/ci/import_graph.py` — exposing:

- `reverse_dependency_map(package_root) -> dict[str, set[str]]`
- `affected_modules(changed_files, rdeps) -> set[str]`

Then both CI jobs `python scripts/ci/import_graph.py ...` instead of carrying inline heredocs. Benefits: one source of truth, unit-testable in isolation (the heredocs currently cannot be tested), and no YAML-embedded Python to maintain.

Acceptance:

- [ ] One shared, committed import-graph module replaces both inline heredocs.
- [ ] The `test` and `notebooks` jobs both call it; behavior unchanged (verify a source-only PR still selects the same tests/notebooks).
- [ ] A couple of unit tests for the graph builder (relative imports, `__init__` re-exports, alias imports).

## Pointers

- `test` job graph builder and `notebooks` job graph builder both live in `.github/workflows/ci.yml` (search for `expand_deps.py` and `affected_notebooks.py`).
- Surfaced during the consolidated `/code-review` of [#265](https://github.com/TARPS-group/prob-pipe/pull/265).

jhuggins:
## Summary

Deduplicates the AST import-graph builder that the `test` and `notebooks` CI jobs each
carried as a near-identical **inline Python heredoc** in `.github/workflows/ci.yml`. The two
copies had drifted into two versions of one algorithm — a fix to one risked the jobs silently
disagreeing about which artifacts a source change affects (right tests but the wrong notebooks,
or vice versa), with CI still green. The heredocs were also impossible to unit-test.

Closes #266.

## What changed

- **New `scripts/ci/import_graph.py`** — one committed, stdlib-only module (the change-detection
  step runs on the runner's system `python3`, before dependencies install). Public API:
  - `reverse_dependency_map(package_root="probpipe") -> dict[str, set[str]]`
  - `affected_modules(changed_files, rdeps) -> set[str]`
  - notebook-only: `symbol_definition_map()`, `notebook_affected()`
  - two CLI subcommands the jobs invoke:
    ```
    python3 scripts/ci/import_graph.py expand    <changed_files...>
    python3 scripts/ci/import_graph.py notebooks <nb_dir> <changed_files...>
    ```
- **`.github/workflows/ci.yml`** — both heredocs deleted; each job now calls the script
  (`+11 / -192`).
- **`tests/ci/test_import_graph.py`** (+ package `__init__.py` and a `conftest.py` that puts
  `scripts/ci/` on `sys.path`) — unit tests for relative imports, `__init__` re-exports, alias
  imports, the transitive BFS, and the `__init__`-skip rule.

## Behavior: fully preserved (no CI-selection change)

An earlier revision unified both jobs on skip-`__init__`. A follow-up review caught this as a real
coverage regression: root-level test files such as `tests/test_coverage_gaps.py` regression-test
*re-exported* classes and relied on the test job's `__init__`→root trigger, so unifying silently
stopped running them on targeted PRs.

The current revision keeps the single shared module but restores each job's exact historical
behavior via a `skip_init` flag — `expand` (test job) passes `skip_init=False` (keeps `__init__`,
preserving the root-test trigger); `notebooks` passes `skip_init=True` (unchanged; that job
resolves package imports via `symbol_definition_map`, not the reverse graph). Both subcommands'
output is now **byte-identical to the old heredocs** — a pure, behavior-preserving extraction.

## Verification

- **Parity vs the old heredocs** (reconstructed from `origin/main`, run against the real tree):
  both `expand` and `notebooks` output is **byte-identical** to the originals across representative
  inputs — `core/_numeric_record_distribution.py`, `core/node.py`, `distributions/transformed.py`,
  `_weights.py` for `expand`, and `docs/examples` + `docs/tutorials` for `notebooks`. (The
  single-pass scan and module-scope symbol map preserve selection.)
- **Unit tests**: `tests/ci/test_import_graph.py` — 8 passed (relative/alias imports, transitive
  BFS, both `skip_init` values, module-level symbol scope, and `expand`/`notebooks` CLI smoke tests).
- **Lint/format**: `ruff check` + `ruff format --check` clean (ruff 0.15.16, the CI-pinned version).
- This PR edits `ci.yml`, a foundational path, so its own CI run executes the full test matrix and
  all notebooks — exercising both CLI subcommands end-to-end.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -163,66 +163,11 @@ jobs:
             echo "run_all=false" >> "$GITHUB_OUTPUT"
             echo "No .py files changed — skipping tests."
           else
-            # Write an import-graph expander: given changed source files, it
-            # uses AST parsing to find all probpipe modules that *transitively*
-            # depend on them, so a change to provenance.py also runs
-            # tests/test_record.py (because record.py imports provenance.py).
-            cat > /tmp/expand_deps.py << 'PYEOF'
-          import ast, sys
-          from pathlib import Path
-          from collections import defaultdict
-
-          def file_to_module(p):
-              return str(Path(p).with_suffix('')).replace('/', '.')
-
-          def module_to_file(m):
-              path = m.replace('.', '/')
-              for candidate in [path + '.py', path + '/__init__.py']:
-                  if Path(candidate).exists():
-                      return candidate
-              return None
-
-          # Build reverse import graph: module -> {modules that directly import it}
-          rdeps = defaultdict(set)
-          for fpath in Path('probpipe').rglob('*.py'):
-              importer = file_to_module(str(fpath))
-              try:
-                  tree = ast.parse(fpath.read_text())
-              except Exception:
-                  continue
-              for node in ast.walk(tree):
-                  if isinstance(node, ast.ImportFrom):
-                      lvl = node.level or 0
-                      if lvl == 0:
-                          if not node.module or not node.module.startswith('probpipe'):
-                              continue
-                          imported = node.module
-                      else:
-                          parts = importer.split('.')
-                          base = parts[:-lvl] if lvl < len(parts) else []
-                          imported = '.'.join(base + node.module.split('.')) if node.module else '.'.join(base)
-                      rdeps[imported].add(importer)
-                  elif isinstance(node, ast.Import):
-                      for alias in node.names:
-                          if alias.name.startswith('probpipe'):
-                              rdeps[alias.name].add(importer)
-
-          # BFS: expand changed source files to all transitively dependent modules
-          changed_sources = [f for f in sys.argv[1:] if f.startswith('probpipe')]
-          seen = set(file_to_module(f) for f in changed_sources)
-          queue = list(seen)
-          while queue:
-              m = queue.pop(0)
-              for dep in rdeps.get(m, []):
-                  if dep not in seen:
-                      seen.add(dep)
-                      queue.append(dep)
-
-          for m in sorted(seen):
-              f = module_to_file(m)
-              if f:
-                  print(f)
-          PYEOF
+            # Expand changed source files to their transitively-dependent
+            # modules via the shared, unit-tested import-graph builder
+            # (scripts/ci/import_graph.py; tests in tests/ci/) — so a change to
+            # provenance.py also runs tests/core/. Replaces the former inline
+            # heredoc (issue #266).
 
             # Split CHANGED into test files (include directly) and source
             # files (expand via import graph before mapping to test folders).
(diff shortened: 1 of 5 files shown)


jhuggins:
Addressed the consolidated `/code-review` + `/review` + `/review-pr` findings in f2e88c8:

1. **Coverage regression (critical).** The earlier unconditional skip-`__init__` dropped the test job's `__init__`→root trigger, so targeted PRs to re-exported classes stopped running the root-level regression tests (notably `tests/test_coverage_gaps.py`). Restored each job's exact behavior via a `skip_init` flag on `affected_modules` (`expand`=`False`, `notebooks`=`True`). `expand` and `notebooks` output is now **byte-identical** to the old heredocs — verified on representative inputs incl. `core/_numeric_record_distribution.py`.
2. **`package_root`.** Threaded through the import-prefix checks so the parameter is fully functional rather than half-wired to the global `PACKAGE`.
3. **Docs.** Documented the `scripts/ci/import_graph.py` helper + `tests/ci/` in `CONTRIBUTING.md` (CI section).
4. **Tests.** Grouped into `Test*` classes; added coverage for both `skip_init` values, module-level symbol scope, and `expand`/`notebooks` CLI smoke tests (8 passing).
5. **Cleanups.** Single-pass `_scan_package()` (the notebooks path previously parsed every file twice) and module-level (`tree.body`) symbol scope so nested/local defs no longer leak into the symbol map — parity-verified to leave notebook selection unchanged.

Net: this is now a pure behavior-preserving extraction. PR description updated accordingly.

SoymilkWinsAgain (changes requested):
There's a minor issue with the PR body: the PR body's verification includes "ruff format --check", but CONTRIBUTING.md explicitly states that the project does not use the ruff format.

Issue #229, opened 2026-06-04
Implement PyMCModel._log_prob (joint log-density evaluation)


jhuggins:
## Summary

`PyMCModel` currently does not implement `_log_prob` and is therefore not `SupportsLogProb`. Calling `log_prob(pymc_model, …)` fails with `TypeError: PyMCModel does not support log_prob`. There's no way to evaluate the joint log-density `log p(θ, y)` of a PyMC-defined model at a fixed point, even though PyMC itself provides this via `pm.logp`.

## Motivation

Joint log-density evaluation is a primitive ProbPipe operation — `SimpleModel` and `StanModel` both support it via `log_prob(model, …)`. For an external academic seminar / tutorial, swapping the modeling backend from `SimpleModel` to `PyMCModel` should leave all downstream ops (`mean`, `log_prob`, `condition_on`, `predictive_check`, …) working uniformly. `log_prob` is the one that's currently missing.

## Proposed approach

Implement `PyMCModel._log_prob(value: Record) -> Array` by wrapping PyMC's `pm.logp` on the conditioned model:

```python
def _log_prob(self, value: Record) -> Array:
    """Joint log-density at a fixed (params, data) point."""
    import pymc as pm
    model = self._pymc_model(data=value)        # rebuild with full Record as kwargs
    # Sum log-prob of every observed and free RV at the assigned value
    # (PyMC's pm.logp evaluates RV-by-RV; need to bundle into a single joint log-density)
    ...
```

Implementation detail to nail down: `pm.logp(rv, value)` evaluates a single RV's log-prob; the joint is a sum across all free RVs (parameters) plus all observed RVs (data). The conditioning data flows in through `_pymc_model(data=...)`, the same path inference uses.

`record_template` already exists on PyMCModel (introspected from `_unconditioned_model.free_RVs` — though see #224 for the data-size-dependent-RV bug that interacts here). The kwarg form `log_prob(pymc_model, intercept=…, slope=…, X=…, y=…)` falls out automatically once `_log_prob` is implemented and the op-layer shim from #228 lands.

## Relationship to other open issues

- **Depends on #228** for the Record-as-canonical `_log_prob` contract. Landing #228 first means this issue's `_log_prob` already has the right signature shape and the kwarg form works without further op-layer changes.

## Task checklist

- [ ] Add `SupportsLogProb` to `PyMCModel`'s bases.
[...]

Issue #228, opened 2026-06-04
Unified kwarg form for log_prob-family ops; switch _log_prob contract to Record-as-canonical


jhuggins:
## Summary

Add kwarg form to the `log_prob`-family of ops (`log_prob`, `prob`, `unnormalized_log_prob`, and the `random_log_prob` / `random_unnormalized_log_prob` callables) so users can write

```python
log_prob(model, intercept=0.0, slope=0.5, X=X_obs, y=y_obs)
log_prob(prior, intercept=0.0, slope=0.5)
log_prob(Normal(name='x'), x=1.5)
```

instead of having to know each `Distribution` class's bespoke value layout (a 2-tuple `(params, data)` for `SimpleModel`, a single `Record` for `ProductDistribution`, a `dict` for `SequentialJointDistribution`, a raw scalar for `Normal`, …). Also clean up an existing silent-misroute bug in `condition_on`.

cc @arob5

## Motivation

`condition_on` already supports named data kwargs:

```python
condition_on(model, X=bootstrap["X"], y=bootstrap["y"])   # works today
```

…but the symmetric form for `log_prob` and its siblings doesn't exist. The op layer is a thin passthrough (`probpipe/core/ops.py:93-99`) and the underlying `_log_prob(value)` contract varies by class. This forces users to learn per-class value layouts that don't appear anywhere in user-facing docs.

A second pain point on the `condition_on` side: a case-mismatched kwarg (e.g., `condition_on(model, x=...)` when the field is `X`) silently routes the kwarg to `inference_kwargs` and is ignored downstream by NUTS. **Wrong result, no error.**

## Status quo

| Call site | Status |
|---|---|
| `condition_on(model, observed_record)` | ✅ works (positional) |
| `condition_on(model, X=..., y=...)` (kwargs matching `dist.fields`) | ✅ works |
| `condition_on(model, x=...)` (lowercase, field is `X`) | ⚠️ silently routes to `inference_kwargs`; **wrong result, no error** |
| `log_prob(dist, value)` | ✅ works (positional passthrough) |
| `log_prob(dist, **kwargs)` | ❌ `TypeError: got an unexpected keyword argument …` |
| `log_prob(SimpleModel, Record(all_fields=...))` | ❌ `ValueError: too many values to unpack (expected 2)` |
| `log_prob(PyMCModel, …)` | ❌ `PyMCModel` is not `SupportsLogProb` — tracked separately in #229 |

## Root cause

[...]

PR #278, opened 2026-06-15
feat(core): keyword form for log_prob-family ops via _pack_value (#228 PR 1)

jhuggins:
## Summary

Implements **PR 1 of #228** — the atomic core: a keyword form for the
`log_prob`-family ops, backed by a per-class `_pack_value` adapter, so users can
write

```python
log_prob(model, intercept=0.0, slope=0.5, X=X_obs, y=y_obs)
log_prob(prior, intercept=0.0, slope=0.5)
log_prob(Normal(name="x"), x=1.5)
```

instead of having to know each distribution's bespoke value layout. The
positional form is unchanged and still broadcasts.

Part of #228. (Remaining: PR 2 `condition_on` case-mismatch, PR 3 StanModel
Tier 2, PR 4 `RandomMeasure` kwarg form; follow-up #229.)

## What changed

- **Protocols genericized** — `SupportsLogProb[T]` / `SupportsUnnormalizedLogProb[T]`
  are now generic in the sample type `T`, with `_log_prob(value: T | ArrayLike)`.
  Documents the value contract; runtime-checkable behavior is unchanged.
- **`Distribution._pack_value(**kwargs) -> T`** (new, on the base) — builds one
  draw from named fields: single-field → bare value (`T = Array`); multi-field →
  `Record` in field order. Raises clearly on missing/unexpected fields.
- **Op layer rewritten as outer wrapper + inner `@workflow_function`** for
  `log_prob`, `prob`, `unnormalized_log_prob`, `unnormalized_prob`, and the two
  `random_*_log_prob` ops. The outer wrapper resolves the positional-or-keyword
  value and forwards the WF controls (`seed` / `n_broadcast_samples` /
  `include_inputs`) explicitly, so a field named like a reserved control can't be
  silently swallowed. Broadcasting still happens — the inner `_<op>_impl` is the
  WorkflowFunction.
- **Reserved-name warning** — constructing a distribution whose field collides
  with `{seed, n_broadcast_samples, include_inputs}` warns (cheap, side-effect-free
  check in the metaclass; points users to the positional `Record` escape hatch).
- **`SimpleModel._log_prob` accepts a `Record` *or* a `(params, data)` tuple.**
  The Record branch (kwarg form) splits into the prior's value (repacked via the
  prior's own `_pack_value`) and a data sub-record from the likelihood's
  `data_template`. See "Deviation" below.
- **`SequentialJointDistribution`** already handled `Record` at runtime
  (`_eval_log_prob` converts it); annotations tightened to say so.
- **`StanModel` / `_UnconstrainedStanView` Tier 1** — `_pack_value` accepts a
  single `parameters=` flat array (the form `_log_prob` already consumes). See
  "Not runtime-verified" below.

## Deviation from the issue plan (justified)

The plan said "tuple form *gone*" for `SimpleModel`. In practice a likelihood
without a `data_template` (bare-array data, e.g. the `GaussianLikelihood` test
fixture) has **no named data fields**, so its data can't be expressed as kwargs;
the `(params, data)` tuple is the only way. `SimpleModel._log_prob` therefore
**dual-accepts** Record and tuple. Consequences:

- The headline kwarg form works for likelihoods with a `data_template` (GLM).
- The tuple form remains valid, so the **"5 tuple-form call sites"** in the
  checklist need **no change** — `_inference_utils.py`'s bare-target path and the
  existing `test_simple_model.py` / `test_sbijax.py` tuple tests keep working.
  (`build_target_log_prob` already bypasses `SimpleModel._log_prob` for the
  structured MCMC path, so inference is unaffected either way.)

## Test plan

- **New:** `tests/core/test_log_prob_kwargs.py` — kwarg form across scalar dists
  (Normal/Beta/MVN), `ProductDistribution`, `JointGaussian`, `SimpleModel` (GLM,
  asserting kwarg == positional-Record == tuple), `StanModel._pack_value`
  (in isolation), error paths (missing / extra / positional+kwargs / neither),
  `prob` / `unnormalized_log_prob` kwarg forms, and the reserved-name warning.
- **Updated:** `test_ops.py` (ops are now thin wrappers over inner
  `_<op>_impl` WorkflowFunctions; positional-only `value`+kwargs collision);
  `test_distribution_array.py` (one call switched from keyword `value=` to
  positional — see Breaking changes).
- **Regression:** full `tests/{core,distributions,modeling,inference,converters}`
  run; my branch introduces **no new failures** vs `origin/main` (verified by
  diffing the full failure sets — the only failures are pre-existing
  Prefect-detection flakes in `test_workflow_parallel.py`, identical on both).

## Breaking changes

`log_prob` / `prob` / `unnormalized_log_prob` / `unnormalized_prob` /
`random_*_log_prob` now take `dist` and `value` as **positional-only** (so field
names are free for the keyword form). The only in-repo keyword-`value=` call site
was one test, updated to positional. `log_prob(dist, value)` (positional) is
unchanged. Labeled `kind:breaking-change`.

## Not runtime-verified

`StanModel` requires `bridgestan` (not installed in CI / locally), so its
`_pack_value` is tested in isolation via `object.__new__` (per STYLE_GUIDE §8.4)
but the full `log_prob(stan_model, parameters=...)` path is not exercised here.

## Documentation

- [x] Docstrings updated for changed public APIs (`log_prob` & friends,
  `_pack_value`, `SimpleModel._log_prob`, protocol contracts)
- [ ] User Guide / tutorials — none needed for this layer
- [x] CHANGELOG / release-notes — noted in #228

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,9 +39,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   single draw via the new `Distribution._pack_value` (single-field → the bare
   field value; multi-field → a `Record`). The positional form is unchanged and
   still broadcasts. Per-call ProbPipe controls use `with_options` —
-  `log_prob.with_options(seed=0)(dist, x)` — consistent with the dispatch ops,
-  so a field whose name matches a control (`seed` / `n_broadcast_samples` /
-  `include_inputs`) is addressed normally by the keyword form.
+  `log_prob.with_options(seed=0)(dist, value)`, delegating to the op's inner
+  WorkflowFunction — so a field whose name matches a control (`seed` /
+  `n_broadcast_samples` / `include_inputs`) is addressed normally by the bare
+  keyword form.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,10 +39,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   single draw via the new `Distribution._pack_value` (single-field → the bare
   field value; multi-field → a `Record`). The positional form is unchanged and
   still broadcasts. Per-call ProbPipe controls use `with_options` —
-  `log_prob.with_options(seed=0)(dist, x)` — consistent with the dispatch ops.
-  Constructing a distribution whose field name collides with a reserved
-  WorkflowFunction control (`seed` / `n_broadcast_samples` / `include_inputs`)
-  now warns.
+  `log_prob.with_options(seed=0)(dist, x)` — consistent with the dispatch ops,
+  so a field whose name matches a control (`seed` / `n_broadcast_samples` /
+  `include_inputs`) is addressed normally by the keyword form.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   dataset). The cell is a no-op in local Jupyter, the docs build, and CI, so
   notebook execution elsewhere is unaffected.
 
+- **Keyword form for the `log_prob`-family ops (#228).** `log_prob`, `prob`,
+  `unnormalized_log_prob`, `unnormalized_prob`, and the `random_*_log_prob`
+  ops accept named field arguments —
+  `log_prob(model, intercept=0.0, slope=0.5, X=X_obs, y=y_obs)` — built into a
+  single draw via the new `Distribution._pack_value` (single-field → the bare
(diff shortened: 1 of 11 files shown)


jhuggins:
Superseded by #282 — same feature (keyword form for the `log_prob`-family ops, issue #228 PR-1), reimplemented cleanly:

- The ops stay **plain `WorkflowFunction`s** (no wrapper / `_DensityOp`), resolving the keyword form in their body via `Distribution._pack_value` — consistent with how `condition_on` handles named data kwargs.
- Branched off **current `main`** (includes #280), so the suite is fully green (0 failures, 17 skipped).
- Non-breaking: `value=` still works; the keyword form is purely additive.

This branch's history churned through `_DensityOp` (introduced → de-smelled → removed). Closing in favor of the clean #282, which carries the full feature.

Issue #237, opened 2026-06-08
Proposal: two-distribution split (`probpipe-core` / `ProbPipe`) for a lean install option


jhuggins:
## Summary

ProbPipe is gaining several heavyweight inference backends — `pyabc` (SMC-ABC), BayesFlow (jax-native NPE/NLE/NRE, landed in #259), `pymc` + `nutpie` (PPL/MCMC backends), `stan`, and `sbi` (PyTorch, bonus). The question is which belong in a bare `pip install probpipe` vs. behind extras, and whether a lean distribution should exist at all.

**Decision (governing principle):** `pip install probpipe` must run **every example and tutorial** in `docs/`. Batteries is defined by the docs' import closure, not by taste. A lean **`probpipe-core`** carries only the JAX base; the friendly **`probpipe`** name is a **batteries metapackage** bundling everything the docs exercise. This is the shape ArviZ and `langchain-core` converged on (see *Prior art*).

The import name `probpipe` is unchanged for all users in both cases.

## Governing principle: batteries = the docs' import closure

A scan of the **code cells** (not markdown — an earlier all-cell grep overcounted) of every `docs/examples/` + `docs/tutorials/` notebook for the backend each actually *executes*:

| Backend (executed in code) | Notebook | Guarded? |
|---|---|---|
| BayesFlow (`learn_amortized_*`) | `flexible_inference` | **no — required to run** |
| `pymc` (`PyMCModel`) | `getting_started` | **no — required to run** |
| `pyabc` (`pyabc_smcabc`) | `flexible_inference` | yes — section self-skips when absent |
| `stan` (`StanModel`/`cmdstanpy`) | `05_external_backends` | yes — section self-skips when absent |
| `nutpie` | *(none — no code cell executes it)* | — included on its own merits |
| `sbi` (PyTorch) | *(none)* | — pure extra |

So the only backends a notebook *needs to run* are **BayesFlow** and **`pymc`** (the unguarded ones). `pyabc` and `stan` are guarded — their notebooks run fine without them and execute those sections only when the extra is installed. `nutpie` is exercised by no notebook; it is a batteries pick on its own merits (fast NUTS). `sbi` (PyTorch) is used nowhere → pure extra, and bundling it would drag in a second multi-hundred-MB array framework.

[...]

PR #290, opened 2026-06-17
feat(packaging): lean probpipe-core distribution + probpipe metapackage (#237)

jhuggins:
## Summary

Implements the two-distribution split from #237: a minimal **`probpipe-core`** and
a **`probpipe`** metapackage (the core plus all the inference backends the docs
use), so users can choose a minimal JAX base or a one-command install that runs
every example and tutorial. The `probpipe` **import** name is unchanged in both.

This is the **packaging half** of #237 (per the agreed split); the CI to build /
publish both distributions, the version-sync guard, and the minimal / 3.14
install jobs follow in a separate PR.

## What changed

- **Root distribution renamed `probpipe` → `probpipe-core`** (`pyproject.toml`).
  Minimal deps only (jax, jaxlib, tfp, blackjax, arviz); every backend stays an
  optional extra. The `probpipe/` import package and all extras are unchanged; the
  `dev` self-reference and `uv.lock` are updated to the new name (a 2-line lock
  diff — resolution is otherwise identical). `probpipe/__init__.py` now reads its
  version from the `probpipe-core` distribution.
- **New code-less metapackage** `packaging/probpipe/` (`name = "probpipe"`).
  Pins `probpipe-core==0.1.0` and adds the backends the docs use — `pymc`,
  `nutpie`, and marker-guarded `bayesflow` (`python_version < "3.14"`). Builds a
  metadata-only wheel.
- **Extras not already bundled are re-exported** (`prefect`/`viz`/`stan` →
  `probpipe-core[...]`), so `pip install "probpipe[prefect]"` keeps working.
  `pymc`/`nutpie`/`bayesflow` are bundled already, so they are *not* re-exported
  (that would be redundant); `probpipe-core` users add any backend via
  `pip install "probpipe-core[<extra>]"`.
- **Source / Colab installs point at `probpipe-core`.** Since the git root now
  builds `probpipe-core`, the tutorials' Colab setup cells install
  `probpipe-core[...] @ git+...` (a `probpipe @ git+<root>` install would fail the
  name check), and the backend-missing error hints (`_bayesflow_common`, the
  nutpie example in `STYLE_GUIDE.md`) point at `probpipe-core[<extra>]`.
- **Authorship.** Both distributions' `authors` metadata is a single entry —
  `ProbPipe Development Team <huggins@bu.edu>` — with the full contributor list
  in `AUTHORS`. The academic-citation files (`CITATION.cff`, the README /
  `docs/cite.md` BibTeX) keep the named authors, ordered Huggins, Roberts, Zhu,
  Erozer, Lim.
- **README + docs install matrix** describing the two distributions and the 3.14
  BayesFlow caveat (plain description, ArviZ-style — no marketing label).

## Scope notes (deferred, per #237)

- **`pyabc`** is not bundled yet — its SMC-ABC backend hasn't landed, and it's
  only used in a *guarded* notebook section that self-skips, so the docs run
  without it. It joins `probpipe` (one dep + a guarded registration) when the
  backend ships.
- **`sbi-torch`** stays out (no backend; used by no notebook).
- **BayesFlow 3.14:** marker-guarded hard dep (the recommended option in #237), so
  `pip install probpipe` resolves on 3.14 — just without neural SBI until upstream
  lifts the cap.
- Publishing to PyPI is pending, so the matrix is documented with a "install from
  source for now" note.

## Test plan

- Both distributions build: `uv build` → `probpipe_core-0.1.0` and `probpipe-0.1.0`
  (the latter a metadata-only wheel; verified contents are dist-info only, and the
  metadata carries `Author-email: ProbPipe Development Team <huggins@bu.edu>`).
- Metapackage `METADATA` verified: `Requires-Dist: probpipe-core==0.1.0`, `pymc>=6`,
  `matplotlib`, `nutpie>=0.12`, `bayesflow<3,>=2.0; python_version < "3.14"`, plus the
  three re-exported extras (`prefect`/`viz`/`stan` only).
- `uv sync --frozen` succeeds and `import probpipe` works with `__version__ = 0.1.0`,
  verified in a clean env where only `probpipe-core` is installed.
- Notebook edits are textual (install string only); all stored outputs preserved.
- Full CI green on the branch (re-running after the latest wording/metadata commit).
- Full cross-dist *resolution* (`pip install probpipe` pulling `probpipe-core`) is
  exercised once both are published — that lands with the publish workflow in the
  CI follow-up PR.


docs/index.md
SoymilkWinsAgain: From CodeX:

`README.md` and `docs/index.md` describe `uv pip install .` / `pip install .` as the default source install, then direct users to Getting Started. But the root project is now `probpipe-core`, and `getting_started.ipynb` has an unguarded PyMC section. A fresh source install from the documented path can therefore fail before users reach the intended tutorial experience.

Suggested fix: make the tutorial-oriented source install explicit, for example `pip install ".[pymc,nutpie,viz]"`, and keep bare `pip install .` documented as the lean-core install.

docs/index.md
```diff
@@ -62,6 +62,19 @@ uv users can substitute `uv pip install .` (into an active `uv venv`), or
 `uv sync` for a lockfile-managed dev environment (see
 [CONTRIBUTING.md](https://github.com/TARPS-group/prob-pipe/blob/main/CONTRIBUTING.md#installation)).
 
+### Two distributions: `probpipe` and `probpipe-core`
+
+ProbPipe ships as two distributions that share the same `probpipe` import name:
+
+| Install | What you get |
+|---|---|
+| `pip install probpipe` | **Recommended.** The minimal core plus the inference backends the docs use — PyMC, nutpie, and BayesFlow — so every example and tutorial runs out of the box. |
```
SoymilkWinsAgain: The metapackage correctly omits BayesFlow on Python 3.14, but here still say `pip install probpipe` runs every example/tutorial. `flexible_inference.ipynb` contains unguarded BayesFlow tutorial cells, so that statement is only true for Python 3.12-3.13 unless the tutorial is guarded.

Suggested fix: qualify the claim as “all tutorials on Python 3.12-3.13” and explicitly note that BayesFlow tutorial sections are unavailable on Python 3.14, or guard those notebook sections.

SoymilkWinsAgain (changes requested):
Also `CONTRIBUTING.md` doesn't describe the new packaging layout.

docs/index.md
jhuggins: Good catch — fixed in d64a9f3. The source-install quickstart (both the uv and pip paths, in `README.md` and `docs/index.md`) now installs `.[pymc,nutpie,viz]` so the linked Getting Started tutorial (unguarded PyMC in §11) runs from the documented path, and bare `pip install .` is documented as the minimal `probpipe-core` install (add backends as extras).

docs/index.md
jhuggins: Fixed in d64a9f3. The install matrix now qualifies the claim — `pip install probpipe` runs every example and tutorial **on Python 3.12–3.13** — and the note below it states explicitly that the BayesFlow neural-SBI sections of the *Flexible Inference* tutorial are unavailable on 3.14 (where the metapackage drops BayesFlow). I took the "qualify the claim" option rather than guarding the cells, so the SBI content stays intact on the supported Pythons.

jhuggins:
Addressed the CONTRIBUTING.md point in 01af765: added a **Distributions: `probpipe-core` and `probpipe`** subsection under *Package Structure* (root `probpipe-core` vs the code-less `packaging/probpipe` metapackage, the version-lockstep rule, and `uv build` commands for each), plus a pointer from *Installation* noting that `uv sync` / `pip install -e .` build `probpipe-core`.

SoymilkWinsAgain (changes requested):
`11_prefect_scalability.ipynb` contains `from prefect.testing.utilities import prefect_test_harness`. However bare ProbPipe installation will fail because it won't install `prefect`. Either skip this step or add `prefect` to the installation.

jhuggins:
> `11_prefect_scalability.ipynb` contains `from prefect.testing.utilities import prefect_test_harness`. However bare ProbPipe installation will fail because it won't install `prefect`. Either skip this step or add `prefect` to the installation.

This is fine for now and can get taken care of more optimally during the broader docs cleanup.

Issue #235, opened 2026-06-05
Proposal: Distribution, Record, and EventTemplate hierarchy redesign


arob5:
> **📌 Standing directive for all implementation sessions.** Before implementing any phase of this
> plan, read and follow **`CONTRACTS.md`** at the repo root. Every major abstraction's contract
> (API, shapes, return types, error cases, invariants, and variable names) must be **crystal clear
> before you implement** and **fully documented in docstrings when you do** — this contract
> documentation seeds the future user guide. `CONTRACTS.md` holds the directives, the per-PR
> checklist, and the canonical naming table. (Added with PR #311; it lands on `main` when that
> merges — until then it lives on the active phase branch.) **Before this issue is closed, `CONTRACTS.md` must be deleted** — it is a guide for implementing this plan, not lasting documentation (the contracts it tracks live on in docstrings and, later, the user guide).
>
> **📌 Closing requirement — make the documentation conventions persistent.** `CONTRACTS.md` is
> deleted when this issue closes, but the *general documentation conventions* it carries are lasting
> and must not be lost. Before closing, migrate them to a permanent home in the codebase. In
> particular, the docstring-authoring rules — **lead with the API / contract** (what the abstraction
> is, what it accepts/returns, how to use it), written as **precise, clear, low-jargon** reference
> text, with **design reasoning, motivation, and trade-offs deferred** to a later **Notes** section —
> together with the "document the precise contract (types/shapes/orderings/raises/invariants)" and
> "code documentation stands on its own (no PR/issue refs in docstrings)" rules. These must be
> **laid out in one canonical place** (e.g. a developer/contributor guide) and **referenced from the
> other relevant places** (e.g. the `review-pr` skill, and a short pointer from `CONTRIBUTING`/docs)
> so contributors and reviewers apply them after this issue and `CONTRACTS.md` are gone.

---

# Proposal: Distribution / Record / EventTemplate hierarchy — the wrapped-value design

> **Revision note.** This body has been rewritten to the **wrapped-value design** and supersedes
> the earlier "native value types, wrap only when forced" version. The central principle
> *inverted*. Two things drove it: (1) names/provenance/aux are core to ProbPipe, and anonymous
> raw arrays/functions can't carry them; (2) on inspection, **current ProbPipe already wraps op
> outputs at the `WorkflowFunction` boundary** (`_coerce_output` → `_wrap_as_record`) — so
> "raw by default" was the departure, not the status quo. Earlier review comments that reference
[...]

PR #292, opened 2026-06-18
Phase 1a (#235): rename RecordTemplate → EventTemplate + LeafSpec representation

arob5:
## PR-1a — `RecordTemplate` → `EventTemplate` + `LeafSpec` representation

**First slice of Phase 1 of #235.** This is a *handoff/draft* PR: the branch is seeded with an
empty commit; an independent session implements it against this brief. Scope and policy decided
with @arob5:

- **Scope:** the **rename + the leaf-spec representation only**. The new template *methods*
  (`numeric_subset` / `to_vector` / `from_vector` / `make_batch` / `vectorize`) and
  `RandomFunction` / `RandomMeasure` population are **out of scope** → later PRs (1b / 1c).
- **Rename policy:** **hard rename, no deprecation alias.** Matches the issue's pre-stable stance.

### Goal — two coupled changes, *no new behavior*

**1. Rename** (hard, repo-wide; 442 occurrences across 41 files incl. tests on `main`):
`RecordTemplate → EventTemplate`, `NumericRecordTemplate → NumericEventTemplate`,
`Distribution.record_template → event_template`.

**2. Replace the leaf representation** `_LeafSpec = "tuple[int, ...] | None"`
(`record.py:575`) with a closed sum of **frozen, hashable** dataclasses:

```python
@dataclass(frozen=True)
class ArraySpec:       shape: tuple[int, ...]; dtype=None; support=None   # numeric array leaf
@dataclass(frozen=True)
class OpaqueSpec:      meta: Hashable = None                              # non-array Python object
@dataclass(frozen=True)
class DistributionSpec: inner: "EventTemplate"                            # leaf holds a Distribution
@dataclass(frozen=True)
class FunctionSpec:    input: "EventTemplate"; output: "EventTemplate"    # leaf holds a callable
```

### Key decisions (flag on the PR if you disagree)

- **`ArraySpec.dtype` / `support` are OPTIONAL, default `None`.** The current rep stores *only*
  shape (the auto-build is `RecordTemplate(**{name: event_shape})`, `_numeric_record_distribution.py:260`),
  so PR-1a must not force call sites to supply dtype/support. Distributions populate them later;
  `ArraySpec((3,))` is valid now.
- **Retain construction-time input sugar.** `EventTemplate(x=(3,), label=None, sub=EventTemplate(...))`
  must still work: a bare shape tuple auto-converts to `ArraySpec(shape=...)`, `None` → `OpaqueSpec()`,
  a nested template stays nested. This keeps construction sites unchanged (only the *class name*
  changes) and makes the representation swap mostly internal. New code may also pass explicit specs.
- **`__getitem__` returns the spec object** (`ArraySpec`/`OpaqueSpec`/…), *not* a bare tuple/None.
  Shape-shaped access stays available via `leaf_shapes` / `event_shapes` / `field_event_shape`
  (unchanged return types). **Audit** existing `template[name]` consumers (see below).
- **`DistributionSpec` / `FunctionSpec` are defined but not yet *produced*.** Nothing emits them in
  PR-1a; they exist so the sum is complete and `NumericEventTemplate` validation can reject them.
  Producing them is PR-1c.

### Old → new leaf mapping

| current (`_FieldSpec`) | new |
|---|---|
| `tuple[int,...]` (numeric leaf shape) | `ArraySpec(shape=tuple, dtype=None, support=None)` |
| `None` (opaque leaf) | `OpaqueSpec(meta=None)` |
| nested `RecordTemplate` | nested `EventTemplate` |
| — (none today) | `DistributionSpec`, `FunctionSpec` (defined, not yet produced) |

### Consumers to migrate — all in `probpipe/core/record.py` unless noted (line numbers on `main`)

- `_LeafSpec` / `_FieldSpec` typealiases (`:575-576`) → redefine as the sum.
- `RecordTemplate` (`:633`): `__new__` auto-promotion (`:676`), `__init__` validation (`:691`),
  `_post_validate` (`:723`), `__getitem__` (`:807`), `__repr__`, `pack` (`:793`), `from_record` (`:840`).
- `leaf_shapes` (`:746`), `event_shapes` (`:764`), `field_event_shape` (`:776`) → derive from specs;
  **preserve return types** (`tuple|None` / `tuple`) so downstream is behavior-preserving.
- `__eq__` / `__hash__` + `_spec_key` (`:815`, `:824`) → specs are frozen/hashable, so `_spec_key`
  simplifies to `hash(spec)`; keep the **order-sensitive** eq/hash contract (#124).
- `NumericRecordTemplate` (`:921`): `_post_validate` (`:941`) → require every leaf be `ArraySpec`
  (or nested `NumericEventTemplate`); reject `OpaqueSpec` / `DistributionSpec` / `FunctionSpec`.
  `_compute_flat_size` / `flat_size`, `numeric_leaf_shapes`.
- `_all_numeric` (used by `__new__`), `_spec_size` (`:999`), `_unpickle_record_template` (`__reduce__`).
- Rename of `Distribution.record_template → event_template` and the `event_template` invariant in
  `_record_distribution.py` (`_RecordDistributionMeta`); the auto-build at
  `_numeric_record_distribution.py:260`.
- **Repo-wide sweep:** `git grep -n "RecordTemplate\|NumericRecordTemplate\|record_template"` (the
  442/41 rename), then `git grep -n "\.leaf_shapes\|\.event_shapes\|\.field_event_shape\|\.flat_size"`
  and template `[name]` indexing to find spec/shape-access consumers needing review.

### Hashability (hard requirement)

All four spec classes are `@dataclass(frozen=True)` with hashable fields, so `EventTemplate` stays
hashable for jit treedef caching and keeps the order-sensitive `__eq__`/`__hash__`.
`ArraySpec.support` / `OpaqueSpec.meta` must be hashable (or `None`).

### Out of scope (do NOT do here)

- `numeric_subset`, `to_vector` / `from_vector`, `make_batch`, `vectorize` → **PR-1b**.
- Populating `event_template` on `RandomFunction` / `RandomMeasure` via `FunctionSpec` /
  `DistributionSpec` → **PR-1c**.
- Any `*Array → *Batch`, the `Tracked` / `Annotated` mixins, or `DistributionBatch` work →
  **Phases 2–5**.
- No deprecation alias for the old names.

### Definition of done

- Repo compiles; `RecordTemplate` / `NumericRecordTemplate` / `record_template` no longer appear
  (`git grep` clean).
- New rep in place; `EventTemplate(x=(), y=(3,))` auto-promotes to `NumericEventTemplate`;
  `EventTemplate(label=None, x=())` stays a plain `EventTemplate`.
- `leaf_shapes` / `event_shapes` / `field_event_shape` / `flat_size` return the **same values** as
  before for equivalent templates (behavior-preserving).
- `EventTemplate` instances are hashable and order-sensitively equal/unequal as today.
- Full existing test suite passes after the mechanical rename of test files.

### New tests

- spec sum: each `*Spec` is frozen, hashable, value-equal.
- construction sugar: tuple → `ArraySpec`, `None` → `OpaqueSpec`, nested template preserved;
  explicit specs accepted.
- auto-promotion to `NumericEventTemplate` iff all leaves `ArraySpec` (recursively);
  `NumericEventTemplate` rejects non-`ArraySpec` leaves.
- hash/eq: equal templates hash-equal; field-order-sensitive inequality.
- back-compat: `leaf_shapes` / `event_shapes` / `field_event_shape` / `flat_size` unchanged for
  representative templates.

---
Design rationale and the full multi-phase plan live in #235 (Chapter 1 = `EventTemplate`;
Chapter 6 = migration). This brief scopes only PR-1a.


arob5:
## Review — Phase 1a: `RecordTemplate → EventTemplate` + `LeafSpec` representation

### Summary
A faithful, high-quality, behavior-preserving implementation that matches the plan (#235 + the brief above) on every point checked. The representation swap and the repo-wide hard rename are clean, the new test file is thorough (81 tests), and all out-of-scope items were correctly excluded. **CI is currently red**, but only because two test assertions were missed in the `__getitem__`-returns-spec consumer audit (one caught by the standard test job, one by the Stan job), plus a docs-build failure that looks like infra. The core code is essentially mergeable once the two stale test assertions are fixed.

### Plan alignment — all ✅
- **Hard rename, no alias:** 0 residual `RecordTemplate`/`NumericRecordTemplate`/`record_template` in code, tests, **and** docs; no deprecation alias.
- **`ArraySpec.dtype`/`support` optional (default `None`).**
- **Construction sugar retained:** `_to_spec` converts `tuple→ArraySpec`, `None→OpaqueSpec`, keeps nested templates and explicit specs.
- **`__getitem__` returns the spec**; `leaf_shapes`/`event_shapes`/`field_event_shape`/`flat_size` preserve return types & values (verified against back-compat tests).
- **Frozen + hashable specs; order-sensitive `__eq__`/`__hash__` (#124)** — `_spec_key` correctly simplified to direct hashing.
- **Out-of-scope correctly excluded:** `numeric_subset`/`to_vector`/`from_vector`/`make_batch`/`vectorize`, `Tracked`/`Annotated`, `*Batch` — all absent. (`is_numeric`/`is_multi_field` also not added — consistent with the minimal slice.)
- Consumer adaptations (`_spec_size`, `template[name]`→`field_event_shape` switches in `_numeric_record_distribution.py`, opaque-spec checks in `_approximate_distribution.py`) are correct and minimal. The `#233` column-ordering code in `_approximate_distribution.py` is **pre-existing on `main`** — not scope creep.
- Bonus: convention docs (`STYLE_GUIDE`, `CONTRIBUTING`), `docs/`, notebooks, and `CHANGELOG` all updated.

### Findings

#### 🔴 Critical (blocks green)
1. **Two stale `template[name] == <tuple>` test assertions** — the direct consequence of the (correct, planned) spec-returning `__getitem__`; the audit missed two backend-gated sites:
   - `tests/modeling/test_pymc_model.py:325` — `assert tpl_c["alpha"] == (N,)` now fails (`ArraySpec(shape=(50,)) == (50,)`). Fails **`test (3.12/3.13/3.14)`** (1 failed, 2977 passed). The same test updated its *earlier* assertions to `== ArraySpec(...)` but missed this line.
   - `tests/modeling/test_stan_model.py:~325` — `{f: ...event_template[f]...} == {'L': (2,2), 'theta': (3,)}` fails the **`stan`** job (only runs with real BridgeStan → invisible to a local run).
   - A full-tree sweep confirms these are the **only** two missed sites.

#### 🟡 Recommended (fix / verify)
2. **`build` job fails in "Build docs"** with `× No migration path exists – existing projects cannot be upgraded` — a mkdocs/tooling error with no ProbPipe symbol; almost certainly **infra/pre-existing**. Please confirm it also fails on `main`; if so it's out of this PR's scope.
3. **`codecov/patch` fails** — patch-coverage gate, likely downstream of the test failures and/or an uncovered new branch (e.g., the `_to_spec` unsupported-type `TypeError` path). Re-check after fixing #1; add a one-line test for `_to_spec`'s reject path if still uncovered.

#### 🟢 Minor (nice to have)
4. **Spec classes not surfaced at top level.** `ArraySpec`/`OpaqueSpec`/`DistributionSpec`/`FunctionSpec`/`LeafSpec` are in `record.py.__all__` but not re-exported from the top-level `probpipe` namespace (unlike `EventTemplate`/`NumericEventTemplate`). Worth a deliberate choice — surface for consistency, or intentionally keep at `probpipe.core.record` (defensible since sugar means they're rarely constructed directly).
5. **Grammar artifacts from the rename** ("a record" → "an event"): "a EventTemplate"/"a event_template" at `_record_distribution.py:306`, `_approximate_distribution.py:147` (a **user-facing error message**), `test_protocols.py:262`, `test_inference.py:129/262`, `test_nutpie.py:297`. The error-message fix at `_approximate_distribution.py:147` is coupled to the regex in `test_inference.py:266` (`match="requires a event_template"`) — fix both together.
6. `ArraySpec.dtype`/`support` are typed `Any` and unvalidated for hashability; the frozen-dataclass `__hash__` fails lazily rather than at construction. Acceptable this phase; tighten when dtype/support get populated later.
7. Cosmetic: a redundant empty seed commit shares the implementation commit's headline — squash on merge.

### Suggested changes
1. `tests/modeling/test_pymc_model.py:325` → `assert tpl_c["alpha"] == ArraySpec((N,))` (or, cleaner, `assert tpl_c.field_event_shape("alpha") == (N,)`).
2. `tests/modeling/test_stan_model.py:~325` → wrap the expected dict values in `ArraySpec(...)`, or compare against `event_template.event_shapes` / `field_event_shape(f)`.
3. Verify the `build`/docs failure on `main`; note or fix accordingly.
4. (Optional) Re-run `codecov/patch` after 1–2; add a `_to_spec` reject-path test if flagged.
5. (Optional) Decide on top-level exports for the spec classes (item 4).
6. (Optional) Fix the "a event_template" → "an event_template" artifacts, keeping `_approximate_distribution.py:147` and `test_inference.py:266` in sync.

Items **1–2 are the only things blocking green** (both one-liners). Everything else is optional polish or verification.

probpipe/core/record.py
```diff
@@ -568,29 +578,123 @@ def __hash__(self) -> int:
...
-# Leaf spec: shape tuple for numeric fields, None for opaque leaves.
-_LeafSpec: TypeAlias = "tuple[int, ...] | None"
-_FieldSpec: TypeAlias = "_LeafSpec | RecordTemplate"
+# ---------------------------------------------------------------------------
+# Leaf specs — the closed sum describing what one leaf of a template is.
+# ---------------------------------------------------------------------------
+#
+# Each spec is a frozen, hashable dataclass so that an ``EventTemplate`` (which
+# stores them) stays hashable for jit treedef caching and keeps the
+# order-sensitive ``__eq__`` / ``__hash__`` contract (#124).
```
jhuggins: remove #124 reference

probpipe/core/record.py
```diff
@@ -568,29 +578,123 @@ def __hash__(self) -> int:
...
+class ArraySpec:
+    """A numeric array leaf: a fixed event ``shape`` plus optional metadata.
+
+    ``dtype`` and ``support`` are optional (default ``None``); the current
+    auto-build path stores only shape, and distributions populate the extra
+    fields in a later phase. Both must be hashable (or ``None``).
+    """
+
+    shape: tuple[int, ...]
+    dtype: Any = None
```
jhuggins: tighten up dtype and support hints? At least make them Hashable but that seems looser than necessary.

probpipe/core/record.py
```diff
@@ -568,29 +578,123 @@ def __hash__(self) -> int:
...
+    """
+
+    input: EventTemplate
+    output: EventTemplate
+
+
+# A leaf spec is one of the four above; a *field* spec is a leaf spec or a
+# nested template (an internal node, kept as-is rather than wrapped).
+LeafSpec: TypeAlias = ArraySpec | OpaqueSpec | DistributionSpec | FunctionSpec
+_LeafSpec: TypeAlias = LeafSpec
```
jhuggins: Why keep this type alias?

probpipe/core/record.py
```diff
@@ -568,29 +578,123 @@ def __hash__(self) -> int:
...
+@dataclass(frozen=True)
+class DistributionSpec:
+    """A leaf whose value is a ``Distribution``.
+
+    ``inner`` is the :class:`EventTemplate` of one draw from that
+    distribution. Defined for completeness of the sum; not yet produced in
+    this phase.
+    """
+
+    inner: EventTemplate
```
jhuggins: The meaning of `inner` is not clear. 

I'm also not sure what "Defined for completeness of the sum" means.

probpipe/core/record.py
```diff
@@ -568,29 +578,123 @@ def __hash__(self) -> int:
...
+
+def _to_spec(spec: Any) -> "_FieldSpec":
+    """Normalise a constructor input to a stored field spec.
+
+    Construction-time sugar (preserved): a bare shape ``tuple`` becomes an
+    :class:`ArraySpec`, ``None`` becomes an :class:`OpaqueSpec`, and a nested
+    :class:`EventTemplate` is kept as-is. Already-built specs pass through, so
+    new code may supply explicit ``ArraySpec(...)`` / ``OpaqueSpec(...)`` etc.
+    """
+    if isinstance(spec, (EventTemplate, ArraySpec, OpaqueSpec, DistributionSpec, FunctionSpec)):
```
jhuggins: how about `isinstance(spec, _FieldSpec)`?

probpipe/core/record.py
```diff
@@ -568,29 +578,123 @@ def __hash__(self) -> int:
...
-        if spec is None:
-            return False
-        if isinstance(spec, RecordTemplate) and not isinstance(
-            spec, NumericRecordTemplate,
-        ):
-            return False
-        if not isinstance(spec, (tuple, RecordTemplate)):
-            # Leave validation of unsupported spec types to __init__.
-            return False
+        if isinstance(spec, tuple):
```
jhuggins: consolidate to a single `instance` call

probpipe/core/record.py
```diff
@@ -647,26 +751,35 @@ class RecordTemplate:
...
-        RecordTemplate(physics=RecordTemplate(force=(), mass=()), obs=())
+        EventTemplate(x=(), y=(3,))                    # -> NumericEventTemplate
+        EventTemplate(label=None, x=())                 # -> EventTemplate (mixed)
+        EventTemplate(physics=EventTemplate(force=(), mass=()), obs=())
 
     Notes
     -----
-    Calling ``RecordTemplate(...)`` directly auto-promotes to a
-    :class:`NumericRecordTemplate` when every spec is numeric (and
+    Leaves are stored as the frozen, hashable spec objects above
```
jhuggins: make sure documentation is self-contained and doesn't reference placement in file

jhuggins (changes requested):

probpipe/core/record.py
arob5: Done in b601e1a — removed the `#124` reference here and from the `__eq__`/`__hash__` comments, keeping the plain explanation that insertion order is part of the template identity.

probpipe/core/record.py
arob5: Done in b601e1a — tightened to `dtype: jnp.dtype | None` and `support: Constraint | None`, matching the existing distribution API (`dtypes -> dict[str, jnp.dtype]`, `supports -> dict[str, Constraint]`). No import cycle: `constraints` only imports jax + `custom_types`.

probpipe/core/record.py
arob5: Removed (in 11a5cc3). `_LeafSpec` was dead — referenced nowhere. The public union is `LeafSpec`; the internal annotation alias is `_FieldSpec` (`LeafSpec | EventTemplate`).

probpipe/core/record.py
arob5: Renamed for clarity in b601e1a: `DistributionSpec.inner -> inner_template`, and `FunctionSpec.input/output -> input_template/output_template`. Also dropped the "defined for completeness of the sum" phrasing — the docstrings now just say these leaf kinds are not produced yet in this phase.

probpipe/core/record.py
arob5: Tried this, but it does not play well with pyright. `_FieldSpec` is a string forward-ref (`"LeafSpec | EventTemplate"`, since `EventTemplate` is defined further down), so it cannot be used in `isinstance` directly. Using the runtime union — `isinstance(spec, (LeafSpec, EventTemplate))` — works at runtime, but pyright will not *narrow* `spec` through a union-alias inside an `isinstance` tuple: the subsequent `return spec` stays typed as the wider input alias (`_FieldSpecInput`), which trips an advisory pyright error.

So I kept an explicit class tuple in 9dfb5e8 — `isinstance(spec, (ArraySpec, OpaqueSpec, DistributionSpec, FunctionSpec, EventTemplate))` — which is still a single `isinstance` call and which pyright narrows correctly. Happy to switch to the alias + a `cast(_FieldSpec, spec)` instead if you prefer the shorter form.

probpipe/core/record.py
arob5: Done in b601e1a — `_all_numeric` now uses a single guard: `if not isinstance(spec, (tuple, ArraySpec, NumericEventTemplate)): return False`.

probpipe/core/record.py
arob5: Done in b601e1a — reworded the `EventTemplate` Notes to name the spec types directly (`ArraySpec`/`OpaqueSpec`/`DistributionSpec`/`FunctionSpec`) instead of referring to "the spec objects above".

probpipe/core/record.py
jhuggins: @arob5 looks like this didn't actually get implemented

jhuggins (approved):

Issue #238, opened 2026-06-08
PyMC 6 / ArviZ 1.0 / jax-0.10 migration + SBI backend replacement (sbijax → pyabc + BayesFlow + sbi)


jhuggins:
## Summary

Two coupled workstreams, both unblocked by the same action — **dropping `sbijax`**:

- **(A) Ecosystem upgrade** to PyMC 6 / PyTensor 3 / ArviZ 1.0 / jax-0.10. **Empirically verified:** a clean target env runs the full test suite green (**2807 passed, 8 skipped, 0 failed**) after a **single one-line source fix**. The upgrade's code surface is ~1 line plus packaging.
- **(B) SBI backend replacement** — retire `sbijax` (single maintainer, `jax==0.8.1`-pinned, dm-haiku / tfp-nightly tail) and replace its functionality with better-maintained backends.

The two are **orthogonal** and parallelizable once sbijax is gone. `sbijax` is the *sole* blocker to a single co-installable `jax 0.10 / pymc 6 / arviz 1.x` environment (it hard-pins `jax==0.8.1` and `arviz==1.0.0`, which cannot co-resolve with pymc6's `arviz>=1.1`).

## Verified target environment

`pymc 6.0.1 · pytensor 3.0.4 · numba 0.65.1 · arviz 1.1.0 · jax/jaxlib 0.10.1 · blackjax 1.5 · tfp-nightly (dev20260605) · nutpie 0.16.10`. The only required source change is `pm.sample_prior_predictive(samples=)` → `draws=` (`probpipe/modeling/_pymc.py`).

## SBI backend ownership (decided)

| Capability | Backend | Install |
|---|---|---|
| SMC-ABC | **pyabc** | **core** — guarantees basic SBI on a bare `pip install probpipe` |
| NPE / FMPE / CMPE (amortized) | **BayesFlow** (keras-on-jax) | `[bayesflow]` extra (jax-native, no torch) |
| NLE / NRE (+ MCMC) | **sbi** (PyTorch) | `[sbi-torch]` extra (heavy) |

[...]

PR #293, opened 2026-06-18
feat(inference): add pyabc SMC-ABC backend (Step 4 of #238)

XinyueLi1012:
## Summary

Adds a `pyabc_smcabc` SMC-ABC method for `SimpleGenerativeModel` — Step 4 of #238. It builds a pyabc prior from the prior's TFP marginals, runs SMC-ABC, and returns the weighted particles. Brings back likelihood-free SMC-ABC after sbijax was dropped, and works for 1-D params too. `make_posterior` gets a `weights=` arg so the weighted particles come through without resampling.

## Linked issue(s)

Part of #238

## Test plan

- New `tests/inference/test_pyabc.py`: recovery (1-D and 2-D), auto-dispatch, weights.
- `make_posterior(weights=)` cases added to `test_inference.py`

## Breaking changes

None — `weights=` is optional; [pyabc] extra, bundled by the probpipe metapackage.

## Documentation

One thing to flag: I defaulted the sampler to `SingleCoreSampler` since pyabc's multicore ones fork and can deadlock with JAX's threads. Multicore still works via `sampler=` — let me know if we'd rather default to pyabc's.


jhuggins:
## Consolidated review (4 independent lenses)

Nice work — this is a clean, well-scoped restoration of SMC-ABC. I ran four independent review passes (ProbPipe-conventions, general correctness, a focused bug-hunt, and a test audit) and merged them; findings raised by more than one lens are noted as high-confidence. **The core design is sound and most of it checks out independently:** the `make_posterior(weights=)` plumbing is correct end-to-end, the TFP→pyabc mapping is faithful for its four families, auto-dispatch routes correctly, and defaulting to `SingleCoreSampler` (to dodge pyabc's fork vs. JAX-threads deadlock) is the right call. A few things to address before merge.

### Must fix

**1. `check()` reports feasible for priors that `execute()` then crashes on.**
`check()` gates on `hasattr(prior, "_tfp_dist")`, but `_build_pyabc_prior` requires `prior.components` (a `ProductDistribution`). A `SimpleGenerativeModel` whose prior is a bare scalar marginal — e.g. `SimpleGenerativeModel(Normal(0., 1., name="theta"), lik)` — passes `check()` (a `Normal` *has* `_tfp_dist`) and then raises `TypeError` inside `execute()` (a `Normal` has no `.components`). The same happens for a nested-dict product prior. Because `pyabc_smcabc` is the sole auto-dispatch pick for `SimpleGenerativeModel`, this means auto-dispatch *crashes* rather than cleanly reporting infeasible. The tests all wrap even single params in `ProductDistribution`, so the path is uncovered. **Fix:** gate `check()` on what `execute()` actually needs (`hasattr(prior, "components")` + each component translatable), sharing one probe between `check` and `_build_pyabc_prior`.

**2. The weighted-particle feature — the headline of this PR — isn't actually tested.**
`test_posterior_is_weighted` asserts only `post.weights is not None` (always true: `make_posterior` always builds a `Weights`, uniform or not) and `num_atoms == 100`. The recovery tests do go through the weighted mean, but the final-population particles are already concentrated near the truth, so an *unweighted* mean would very likely pass the same `abs=0.5` band — a broken-weighting regression would slip through. The one genuinely discriminating test (`test_make_posterior_forwards_weights`, 8.0 vs 5.0) exercises `make_posterior`'s single-field path, not the pyabc backend. **Fix:** assert the pyabc posterior's weights are non-uniform (`not is_uniform`), or that its weighted mean differs from the unweighted mean of the same `draws()`.

### Should fix

**3. `random_seed` doesn't reach pyabc → runs aren't reproducible.** `random_seed` seeds only the simulator's `np.random.default_rng`; pyabc's own proposal/perturbation RNG is never seeded, so two `random_seed=0` runs differ — unlike every other backend, which threads the seed into the sampler. This also makes the recovery tests non-deterministic, which (with the loose tolerances) is fragile. Seed pyabc's RNG too (snapshot/restore numpy global state, as the other backends do), or — if pyabc can't be cleanly seeded — rename/document the param so it doesn't over-promise.

**4. `_marginal_to_pyabc_rv` reinvents an existing mapping and silently covers fewer families.** It hand-rolls a TFP→scipy parameterization map that already lives in `converters/_scipy.py` (the Gamma `rate→1/scale` and Uniform `high→scale` reparams are duplicated verbatim), reaches into the private `_tfp_dist`, and raises `NotImplementedError` for `Exponential`/`LogNormal`/`Cauchy`/`Laplace` — which the converter registry already handles. Prefer routing marginals through the converter registry (→ scipy frozen → `pyabc.RV`), or at minimum widen the families and add a comment so the two maps don't drift.

**5. Test quality below the project bar.** Recovery tests check only the posterior **mean**, at a **single seed**, with **guessed** tolerances (`abs=0.5/0.6`) and no measured-range comment. The model has a known analytic conjugate posterior — assert the **std** too, document the observed cross-seed range, and parametrize ≥2 seeds. Also untested: the `check()` "prior must be TFP-backed" branch; an end-to-end error for an untranslatable prior; the **`SingleCoreSampler` default** (a regression to a forking sampler could *hang* CI, not just fail it); and the numeric correctness of the family mapping (`test_supported_families` checks only the `pyabc.RV` *names*, not that e.g. `Gamma(rate=1.5)` maps to the right scale).

**6. Document `execute()`'s kwarg surface.** It has no docstring, yet defines the user-facing API via `kwargs`: `n_particles`, `max_populations` (+ an undocumented `n_rounds` alias), `eps_alpha`, `random_seed`, `summary_fn`, `distance_fn`, `sampler`. Document them NumPy-style, and surface the multicore `sampler=` escape hatch and its fork-deadlock rationale where users can find it.

**7. Docs/framing.** Add the `pyabc_smcabc` row to the built-in-methods table in **CONTRIBUTING.md** (only `docs/api/inference.md` was updated — the tables are duplicated). And reconcile the PR body's "pyabc added to core deps" with the code: it's actually the **`[pyabc]` extra, bundled by the `probpipe` metapackage** (matching the CHANGELOG) — which still gives `pip install probpipe` users SBI out of the box while keeping `probpipe-core` lean. Just a wording fix.

### Minor

`vec` is hard-cast to `float32`, ignoring `jax_enable_x64`; a 1-D param's `mean(post)["theta"]` comes out shaped `(1,)` rather than scalar `()` while `draws()` is `(n,)`; drop or document the `n_rounds` alias; a single `_ensure_pyabc()` helper (STYLE_GUIDE §4.4) instead of four scattered `import pyabc`; the `# Tier 1-10` priority comment is a dev-phase label (describe by behavior); the `mkdocs-jupyter` floor bump is an unrelated drive-by (drop or note); a few call sites break after `(` against the horizontal-packing convention.

### Verified correct
- `make_posterior(weights=)` is correct end-to-end: forwards through `ApproximateDistribution` (single- and multi-field) → `RecordEmpiricalDistribution` → `Weights` (validates non-negativity + positive sum, normalizes); the weighted mean is genuinely used; no resampling on the weighted path.
- TFP→pyabc mapping is faithful for the four supported families (Uniform/Normal/Beta/Gamma) with correct scipy reparameterizations.
- Auto-dispatch routing is correct: `SimpleGenerativeModel` isn't `SupportsConditioning`, so it reaches the registry; higher-priority methods' `check()` correctly reject it; pyabc is the sole pick at priority 6.
- Registry contract, naming (`pyabc_smcabc` / `PyABCSMCMethod`), the `try/except ImportError` registration guard, prior name/order → column alignment, the `SingleCoreSampler` default + `sampler=` override, `weights=` backward-compat, and pyabc imposing no Python-version cap (so the 3.12–3.14 matrix is unaffected) all check out.

### Suggested order
(1) fix `check()` to match `execute()`'s precondition → (2) add a discriminating weighting test → (3) seed pyabc (or document the limit) → (4) route through the converter registry / widen families → (5) spread + seeds + the untested branches → (6–7) docstrings, the CONTRIBUTING table, and the dep wording. (1)–(3) are the load-bearing ones.

---
*Generated by Claude (consolidated multi-lens review); relayed for triage — apply your own judgment on what to act on.*

jhuggins (changes requested):
Update the flexible inference notebook to reflect that abc is now available again.

probpipe/inference/_pyabc.py
```diff
@@ -0,0 +1,229 @@
...
+from ..core._registry import MethodInfo
+from ..core.record import RecordTemplate
+from ._approximate_distribution import ApproximateDistribution, make_posterior
+from ._registry import InferenceMethod
+
+# Key for the simulated/observed vector in pyabc's summary-statistic dict.
+_DATA_KEY = "y"
+
+
+def _ensure_pyabc() -> Any:
```
jhuggins: tighten up type hints (throughout)

probpipe/inference/_pyabc.py
```diff
@@ -0,0 +1,229 @@
...
+        ) from e
+    return pyabc
+
+
+def _marginal_to_pyabc_rv(leaf: Any, pyabc: Any) -> Any:
+    """Wrap a ProbPipe marginal as a ``pyabc.RV`` backed by the distribution
+    itself — ``sample`` for ``rvs``, ``prob`` for ``pdf`` — with no scipy
+    round-trip. ``cdf`` is unused by SMC-ABC.
+    """
+    from probpipe import prob, sample
```
jhuggins: hoist imports to the top level unless there's a good reason not to

probpipe/inference/_pyabc.py
```diff
@@ -0,0 +1,229 @@
...
+
+    class _ProbPipeRV(pyabc.RV):
```
jhuggins: this should be defined at the top level, not inside the function

probpipe/inference/_pyabc.py
```diff
@@ -0,0 +1,229 @@
...
+        def __init__(self, dist: Any):
+            self._dist = dist
+            self.distribution = None  # what RV.__getattr__ proxies to
+            self.name = getattr(dist, "name", "rv")
+            self.args, self.kwargs = (), {}
+
+        def rvs(self, *args: Any, **kwargs: Any) -> float:
+            key = jax.random.PRNGKey(int(np.random.randint(0, 2**31 - 1)))
```
jhuggins: thread a JAX key properly (split, don't reseed from numpy)

probpipe/inference/_pyabc.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+def _summarize(data: Any, summary_fn: Any) -> np.ndarray:
+    """Apply ``summary_fn`` (if any) and flatten to a 1-D vector."""
+    if summary_fn is not None:
+        data = summary_fn(data)
+    return np.asarray(data, dtype=float).ravel()
+
+
+def _euclidean_distance(x: dict, x0: dict) -> float:
```
jhuggins: no need to implement this – already in numpy / jax

probpipe/inference/_pyabc.py
```diff
@@ -0,0 +1,229 @@
...
+    return float(np.sqrt(np.sum((x[_DATA_KEY] - x0[_DATA_KEY]) ** 2)))
+
+
+class PyABCSMCMethod(InferenceMethod):
```
jhuggins, line 117: class and methods missing documentation

CONTRIBUTING.md
```diff
@@ -604,6 +604,7 @@ Built-in methods:
 | 75 | `blackjax_elliptical_slice` | BlackJAX | `SimpleModel` + Gaussian prior + JAX-traceable likelihood |
 | 55 | `blackjax_rwmh` | BlackJAX | Any `SupportsLogProb` (eager fallback for non-traceable targets) |
 | 45 | `blackjax_sgld` | BlackJAX | `SimpleModel` + `ConditionallyIndependentLikelihood` + `batch_size=` |
+| 6 | `pyabc_smcabc` | pyabc | `SimpleGenerativeModel` + TFP-backed product prior (requires the `[pyabc]` extra) |
```
jhuggins: doesn't actually require a TFP-backed prior

docs/tutorials/flexible_inference.ipynb
```diff
@@ -23,29 +25,54 @@
     "- **NPE** (`learn_amortized_posterior`): amortized direct posterior sampling.\n",
     "- **NLE** (`learn_amortized_likelihood`): an emulated likelihood that any MCMC backend can drive.\n",
     "- **NRE** (`learn_amortized_ratio`): an emulated likelihood-to-evidence ratio, also MCMC-driven.\n",
-    "- **SMC-ABC** through the inference registry (shown below; runs once the `pyabc` backend is installed).\n",
+    "- **SMC-ABC** through the inference registry (shown below).\n",
     "\n",
-    "> **Prerequisites:** [Getting Started tutorial](getting_started.ipynb). The neural SBI engines (NPE/NLE/NRE) require `pip install probpipe-core[bayesflow]`; SMC-ABC requires the `pyabc` backend (a later release)."
+    "> **Prerequisites:** [Getting Started tutorial](getting_started.ipynb). The neural SBI engines (NPE/NLE/NRE) require `pip install probpipe-core[bayesflow]`; SMC-ABC uses the `pyabc` backend — the `[pyabc]` extra, bundled by the `probpipe` metapackage (`pip install probpipe`, or `probpipe-core[pyabc]`)."
```
jhuggins: this can get cleaned up to reflect the symmetry of BayesFlow and pyabc: both are extras to `probpipe-core`, but included with `probpipe`

docs/tutorials/flexible_inference.ipynb
```diff
@@ -23,29 +25,54 @@
...
+    "    _IN_COLAB = True\n",
+    "except ImportError:\n",
+    "    _IN_COLAB = False\n",
+    "\n",
+    "if _IN_COLAB:\n",
+    "    import subprocess, sys\n",
+    "\n",
+    "    subprocess.run(\n",
+    "        [sys.executable, \"-m\", \"pip\", \"install\",\n",
+    "         \"probpipe-core[viz,bayesflow] @ git+https://github.com/TARPS-group/prob-pipe@main\"],\n",
```
jhuggins: need to include `pyabc` extra here

docs/tutorials/flexible_inference.ipynb
```diff
@@ -65,9 +92,9 @@
     ")\n",
     "from probpipe.modeling import Likelihood, GenerativeLikelihood\n",
     "\n",
-    "# SMC-ABC arrives with the optional pyabc backend (a later release): the cells\n",
-    "# that use it run only when the `pyabc_smcabc` method is registered, and\n",
-    "# self-skip otherwise. Everything else is functional with probpipe-core[bayesflow].\n",
+    "# SMC-ABC uses the optional pyabc backend (the `[pyabc]` extra, bundled by the\n",
```
jhuggins: now that pyabc is always included, you can get rid of this checking code

docs/tutorials/flexible_inference.ipynb
```diff
@@ -617,27 +644,268 @@
...
+     "iopub.status.idle": "2026-06-19T02:34:24.564551Z",
+     "shell.execute_reply": "2026-06-19T02:34:24.564271Z"
     }
    },
    "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "ABC.History INFO: Start <ABCSMC id=1, start_time=2026-06-18 22:31:36>\n"
```
jhuggins: it would be nice to suppress all of this extra output; might require tweaking the `silence` context manager

probpipe/inference/_pyabc.py
```diff
@@ -0,0 +1,229 @@
...
+        df, weights = history.get_distribution(m=0, t=history.max_t)
+        particles = jnp.asarray(df.reindex(columns=names).to_numpy(dtype=float))
+        weights = np.asarray(weights, dtype=float)
+
+        # Name the columns by component so draws()/mean() return Records keyed by
+        # parameter name (the prior's marginals are scalar).
+        template = RecordTemplate(**{n: tuple(prior.components[n].event_shape) for n in names})
+
+        return make_posterior(
+            [particles], parents=(prior,), algorithm="pyabc_smcabc",
```
jhuggins: Only recording only the prior as provenance parent, not the model. Every other backend passes the model (dist), so the lineage here drops the likelihood/data

jhuggins (changes requested):
I'm not sold on the per-marginal `_ProbPipeRV` / `pyabc.Distribution(**rvs)` approach, because of the restrictions it puts on the prior. Building the pyabc prior as one RV per named component forces the prior to be a *product of independent scalar marginals*: pyabc's `Distribution` samples each key independently and its `pdf` is the product of the per-key marginal pdfs. So:

- correlated priors (a `JointGaussian` with off-diagonal covariance, any `MultivariateNormal`)
  can't be represented — they'd be silently wrong as a product, and `_build_pyabc_prior` rejects
  anything without `.components` anyway;
- a vector-valued component slips through `check()` (it only rejects non-products and nested dicts)
  and then crashes in `rvs` on `float(<vector>)` — a feasible-then-crash, the same class of bug the
  `check()` guard was meant to close;
- and we're maintaining a hand-rolled RV with `rvs`/`pdf`/`pmf`/`cdf`/`copy`.

I think a single distribution over the **flattened** parameter vector is cleaner and strictly more general. Keep it as a `pyabc.Distribution` (that's the supported, typed prior interface, and it stays dict-like / picklable), but make it *joint*: register `d` keys = the flat positions of the parameter vector, and override the sampling/density rather than using per-key independent RVs.

- `__init__` → register the `d` flat-position keys with placeholder per-position RVs. They exist only
  to supply the parameter names (`get_parameter_names()`); they're never sampled or scored directly,
  because we override the next two:
- `rvs()` → one `sample(prior, key=...)`, flattened, returned as `Parameter(**{flat_key_i: v_i})`;
- `pdf(x)` → reassemble the flat vector and return the **joint** prior density (`prob` /
  `exp(log_prob)`), not the product of the placeholder marginals.

Because it's still a `Distribution`, `get_parameter_names()`, dict access, pickling, and the perturbation kernel (which fits on the population's flat scalar columns) all keep working. And it maps onto our existing flat machinery: `prior.as_flat_distribution()` already returns a `FlatNumericRecordDistribution` that samples a flat vector and carries the joint `log_prob`, and `as_record_distribution(template)` lifts the final particles back to name-keyed Records.

The big benefit of this approach is that it removes both the independence and the scalar-marginal restrictions (correlated and multivariate priors just work), eliminates the feasible-then-crash cases, and lets `check()` simplify to "can-sample + has-a-density + can-flatten" instead of probing `.components`.

A bonus is that collapsing to one `sample(prior, key=...)` per draw makes proper JAX-key threading straightforward (one split key per particle) — which also resolves the reproducibility point I left inline.

One thing to consider: For constrained supports (positive, simplex, …) the flat perturbation will propose out-of-support  points that get density 0, which is correct but wasteful. The obvious follow-up (not required here) is to perturb in **unconstrained** space via the constraint bijectors, the same way the BayesFlow path does.

jhuggins (approved):

PR #294, opened 2026-06-18
Bug fix(linalg): 3 small algebra bugs

SoymilkWinsAgain:
## Summary

Bug 1: `RootLinOp.diag()` returned an incorrect value for a `DiagonalLinOp` root. It directly returned `self.root.diagonal`, but `RootLinOp` represents `S @ S.T`, so diagonal-root entries must be squared.

Bug 2: `CholeskyLinOp.cholesky(lower=False)` returned a type that did not satisfy its own subsequent interfaces. The Cholesky representation now preserves both lower-root (`L @ L.T`) and upper-root (`U.T @ U`) operator semantics across `matvec`, `rmatvec`, `matmat`, `rmatmat`, `diag`, `to_dense`, and `solve`.

Bug 3: `JointGaussian.condition_on` used explicit matrix inversion. It now uses `jnp.linalg.solve`, avoiding explicit inverse formation while preserving the same conditional Gaussian algebra.

Bug 4: `SumLinOp.matmat` / `rmatmat` returned a 1-D vector for single-column matrix input. They now preserve the matrix contract and return `(n, 1)` / `(m, 1)` results.

Weights hardening: `Weights(log_weights=...)` now rejects `NaN` entries and zero-total-mass inputs such as all `-inf`, while still allowing individual `-inf` entries for zero-weight atoms.

Hygiene: `linear_operator.py` also includes small import/exception modernizations, and `CHANGELOG.md` now records the algebra, SumLinOp, and Weights fixes under `[Unreleased] / Fixed`.

## Test plan

```bash
.venv/bin/python -m pytest -p no:xdist -o "addopts=" --tb=short -q tests/linalg/test_linop.py tests/linalg/test_operations.py tests/distributions/test_joint_gaussian.py tests/test_weights.py
git diff --check
.venv/bin/ruff check --select F,E9 probpipe/linalg/linear_operator.py tests/linalg/test_linop.py CHANGELOG.md
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,6 +100,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Linear-algebra and Gaussian-conditioning edge cases on the algebra bug-fix
+  branch.** `RootLinOp.diag()` now squares diagonal roots; `CholeskyLinOp`
+  keeps lower-root (`L @ L.T`) and upper-root (`U.T @ U`) representations
+  consistent across `cholesky`, `to_cholesky_representation`, `matvec`,
+  `rmatvec`, `matmat`, `rmatmat`, `diag`, `to_dense`, and `solve`;
+  `JointGaussian.condition_on` uses linear solves instead of forming explicit
+  covariance inverses; and `SumLinOp.matmat` / `rmatmat` preserve the `(n, 1)`
+  matrix shape for single-column inputs.
+
+- **Invalid log-space weights are rejected before normalization.**
+  `Weights(log_weights=...)` now rejects `NaN` entries and zero-total-mass
+  inputs such as all `-inf`, avoiding downstream `nan` normalized weights while
+  still allowing individual `-inf` entries for zero-weight atoms.
+
 - **`StanModel` now works against a real BridgeStan backend.** Two bugs at the
   BridgeStan boundary were hidden by the mocked tests: construction passed a
   `data=` keyword that `bridgestan.StanModel.from_stan_file` does not accept,
--- a/probpipe/_weights.py
+++ b/probpipe/_weights.py
@@ -77,6 +77,13 @@ def _validate_to_log_weights(
                 f"log_weights shape {log_weights.shape} does not match "
                 f"number of items {n}."
             )
+        if jnp.any(jnp.isnan(log_weights)):
+            raise ValueError("log_weights must not contain NaN.")
+        total_log_weight = jax.scipy.special.logsumexp(log_weights)
+        if not jnp.isfinite(total_log_weight):
+            raise ValueError(
+                "log_weights must define a positive finite total weight."
+            )
         return log_weights, False
 
     return None, True
@@ -594,5 +601,4 @@ jax.tree_util.register_pytree_node(
     lambda w: w.tree_flatten(),
     lambda aux, children: Weights.tree_unflatten(aux, children),
 )
-
 
--- a/probpipe/distributions/_joint_gaussian.py
+++ b/probpipe/distributions/_joint_gaussian.py
@@ -7,11 +7,10 @@ from __future__ import annotations
 
 from types import MappingProxyType
 
-import jax
 import jax.numpy as jnp
 
-from .._dtype import _as_float_array, _promote_floats
-from ..core._record_distribution import RecordDistribution, _build_record_template
+from .._dtype import _promote_floats
+from ..core._record_distribution import _build_record_template
 from ..core.distribution import (
     NumericRecordDistribution,
     _mc_expectation,
@@ -219,7 +218,7 @@ class JointGaussian(NumericRecordDistribution, SupportsSampling, SupportsLogProb
 
     def _condition_on_impl(
         self, observed_leaves: dict[KeyPath, ArrayLike],
-    ) -> "JointGaussian":
+    ) -> JointGaussian:
         """Condition on observed component values using exact Gaussian formulas.
 
         Returns a new :class:`JointGaussian` over the remaining (unobserved)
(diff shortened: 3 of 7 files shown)


SoymilkWinsAgain:
2 more bug fixed.

1. `SumLinOp.matmat` / `rmatmat` has a shape bug for single-column matrices. When `X.shape[1] == 1`, matvec/rmatvec is called, which returns `(n,)`, but matmat should return `(n, 1)`.
2. `Weights(log_weights=[-inf, -inf])` will produce `normalized == [nan, nan]`. The original weights=0 will be rejected, but the log path does not reject cases where the total weight is 0 or NaN.

jhuggins:
## Review (ProbPipe `/review-pr`)

All three documented bugs are **fixed correctly** — I verified the algebra term-by-term, there are no dependency-graph violations, and the imports dropped from `_joint_gaussian.py` are genuinely dead (0 remaining references). The findings below are about hygiene and test coverage, not correctness.

### Critical (must fix)
None.

### Recommended (should fix)

1. **The PR does more than its "3 algebra bugs" title, and none of it is in the CHANGELOG.** The branch also (a) hardens `_weights.py` (`_validate_to_log_weights` now rejects NaN and all-`-inf`/zero-total log-weights — your commit message notes the `[-inf, -inf] → [nan, nan]` bug) and (b) fixes a fourth bug: `SumLinOp.matmat`/`rmatmat` returned a 1-D vector instead of an `(n, 1)` column for single-column input. The PR *description* still lists only the three (the commit messages do capture the others). Please update the PR body to mention the `_weights` and `SumLinOp` fixes (and the import modernizations), and add a `[Unreleased] / ### Fixed` CHANGELOG entry covering all four bugs plus the new `Weights` rejection behavior.

2. **The new `CholeskyLinOp` operator methods aren't directly tested.** `test_cholesky_upper_factor_preserves_operator` exercises `to_dense` and `solve` of the upper representation — but the bulk of the added code is the new `matvec` / `rmatvec` / `matmat` / `rmatmat` / `diag` overrides (which branch on `root.lower` to compute `UᵀU` vs `LLᵀ`, and pick `axis=0` vs `axis=1` for `diag`). A wrong branch/axis there wouldn't be caught. Add direct assertions against the dense `expected`, for both the upper rep and the lower-root `op` — e.g. `approx(upper_rep.matvec(x), expected @ x)`, `approx(upper_rep.diag(), jnp.diag(expected))`, `approx(upper_rep.matmat(X), expected @ X)`.

### Minor (nice to have)

3. The new `CholeskyLinOp` methods have no docstrings. This matches the terse, undocumented operator-method style already in the file, so it's low priority — but the class docstring still says only "`A = L @ L.T`", which is now incomplete; a one-line note that an upper root represents `A = UᵀU` would help. A non-diagonal-root `RootLinOp.diag` test (the diagonal-root case is covered; the general `einsum` path isn't) would also round out coverage.
4. The `linear_operator.py` import modernizations (`FrozenSet`→`frozenset`, `typing.Iterable`→`collections.abc.Iterable`, dropping the unused `import math`, `raise … from e`, EOF newline) are all good and convention-aligned — just unrelated to the bug fixes; worth a line in the PR body so the diff isn't surprising.
5. `_weights.py`'s new guards are value-dependent (`jnp.any(jnp.isnan(...))`, `if not jnp.isfinite(...)`), so `_validate_to_log_weights` isn't JAX-traceable — but that exactly matches the pre-existing `if jnp.any(weights < 0)` / `if total <= 0` checks right above, so it's the established eager-construction pattern, not a regression. Noting only.

### Verified correct
*(called out since this is a correctness PR)*
- **Bug 1 — `RootLinOp.diag()`:** for a diagonal root `S = diag(d)`, `diag(S Sᵀ) = d²`, so `self.root.diagonal ** 2` is right — and it now matches the already-correct `RootLinOp.trace()` (which squares) and the general `einsum('ij,ij->i', S, S)` row-sum path.
- **Bug 2 — `CholeskyLinOp`:** `cholesky(lower=False)` now returns a genuine `TriangularLinOp(Lᵀ, lower=False)` (was a `TransposedLinOp`, which broke its return contract *and* made `to_cholesky_representation(lower=False)` raise inside `CholeskyLinOp.__init__`). The added overrides correctly compute `LLᵀ` (lower root) vs `UᵀU` (upper root) for `matvec`/`matmat`/`diag`/`to_dense`/`solve` — verified including the `diag` axis choice and the two-stage triangular `solve` for each orientation.
- **Bug 3 — `JointGaussian._condition_on_impl`:** the `jnp.linalg.solve` form is mathematically identical to `inv(Σ_oo) @ ·` for both the mean and the cross-covariance (`Σ_oo⁻¹ Σ_ou`), and faster/stabler; the `monkeypatch` test that fails if `inv` is ever formed is a clean way to pin it.
- **`SumLinOp` shape fix is complete** (no `shape[1] == 1` shortcut remains anywhere), and the class semantics (`RootLinOp = S Sᵀ`, `CholeskyLinOp = L Lᵀ / Uᵀ U`, `DiagonalRootLinOp = diag(s²)`) are mutually consistent.

### Suggested changes
1. **PR body + `CHANGELOG.md`** — describe the `_weights.py` and `SumLinOp` fixes in the body; add a `[Unreleased] / ### Fixed` entry for all four bugs and the new `Weights` NaN / zero-total-mass rejection.
2. **`tests/linalg/test_linop.py`** (`test_cholesky_upper_factor_preserves_operator`) — add direct `matvec`/`rmatvec`/`matmat`/`rmatmat`/`diag` assertions against the dense `expected`, for both the upper rep and the lower-root `op`.
3. **`probpipe/linalg/linear_operator.py`** — extend the `CholeskyLinOp` class docstring to cover the upper-root (`UᵀU`) case and add brief docstrings to the new operator methods; and add a non-diagonal-root `RootLinOp.diag` test.

Items 1–2 are the non-optional ones; the algebra itself is good to merge.

---
*Generated by Claude (`/review-pr`); relayed for triage — apply your own judgment on what to act on.*

jhuggins (commented):
Looks good. Claude had a few minor suggestions that seemed worth implementing

jhuggins (approved):

Issue #303, opened 2026-06-22
Add laplace / laplace_calibrated inference methods (Laplace foundation)


jhuggins:
## Context

Part of the **calibrated-SGLD UQ initiative**. The MAP + curvature machinery the
calibrated SG-MCMC methods need *offline* is, by itself, a **Laplace
approximation** — so build it once as a first-class method family and reuse it
(it is also the init + preconditioner for the SGLD methods, sibling issue, and a
fast UQ method in its own right).

Theory: Negrea et al. (arXiv:2207.12395), Wang–Ding–Huggins (arXiv:2606.00293).

## Scope — in `probpipe`

- [ ] **MAP** `θ̂` via optax — use the Campbell et al. 2026 tuning-free panel
  (Adam 1e-4, or the 5-optimizer ensemble); full-data or `M`-subsample. (Declare
  `optax` in `pyproject.toml`.)
- [ ] **Pluggable curvature** (`curvature=`), first cut **full + diagonal**
  (GGN-Fisher / low-rank / K-FAC / last-layer are the documented roadmap).
- [ ] **`laplace`** → Gaussian `N(θ̂, Ĵ⁻¹)` (model-based BVM / standard-posterior
  asymptotic covariance). Needs only a differentiable joint.
- [ ] **`laplace_calibrated`** → Gaussian with the **sandwich** `Ĵ⁻¹ Î Ĵ⁻¹`
  (frequentist-valid, misspecification-robust). `Î` is the empirical Fisher
  `(1/N)Σ ∇ℓᵢ∇ℓᵢᵀ`, built from `per_datum_log_likelihood` scores — needs
  `ConditionallyIndependentLikelihood`.
  - **Correctness invariant:** estimate `Ĵ` (model Hessian / GGN) and `Î`
    (empirical score variance) **separately**. The robustness *is* the gap
    `Ĵ ≠ Î`; approximating `Ĵ` by the empirical Fisher collapses the sandwich to
    `Ĵ⁻¹` and erases the correction. Under correct specification `Ĵ = Î` and the
    two methods coincide (a useful test).
- [ ] Register in the inference registry (`laplace` ~ parametric tier;
  `laplace_calibrated` opt-in). Return an `ApproximateDistribution` (draws +
  closed-form moments).

## Acceptance — uses the validation metrics from #301

- On the analytic conjugate Gaussian, `laplace` recovers the posterior mean +
  covariance within tolerance.
[...]

Issue #302, opened 2026-06-22
Create probpipe-benchmark: heavy inference benchmark suite (separate repo)


jhuggins:
## Context

ProbPipe needs a **standing, cross-method, cross-dataset inference benchmark** —
a place to ask "how well does method *X* recover the posterior on target *T*?"
for *any* method (NUTS / RWMH, SG-MCMC, VI, external Stan / PyMC / nutpie,
BayesFlow amortized SBI, bagged posteriors, GP-emulator surrogates, the coming
Laplace family) against *any* standardized target with a trusted reference. The
value is general and long-lived:

- **method comparison** — apples-to-apples leaderboards across methods on shared
  targets;
- **regression detection** — catch accuracy/calibration regressions as inference
  internals change;
- **onboarding** — a new method proves itself on the suite before it is
  recommended;
- **reproducible evaluation** — one registry, standardized references, scorecards.

The heavyweight side of this — reference posteriors at scale, large datasets, a
stratified target grid, heavy plotting/leaderboards — is split into a **separate
`probpipe-benchmark` repo**, keeping `probpipe` itself light. This matches
near-universal PPL precedent (posteriordb is separate from Stan; sbibm from the
`sbi` library; MCMCBenchmarks.jl, bdl-benchmarks, uncertainty-baselines,
inference_gym are all separate from their libraries), and it honors the project
principle that `pip install probpipe` stays light and runs every
example/tutorial.

The reusable, dependency-light **metrics + references + SBC** live in
`probpipe.validation` (sibling issue #301); this repo is the heavy harness that
consumes them. The **calibrated-SGLD initiative** (#303/#304) is the first
consumer that motivates standing the repo up — and it contributes the SG-MCMC /
calibration track below — but the suite is permanent, method-agnostic
infrastructure, not an SGLD artifact.

## Scope — new repo `probpipe-benchmark`, depends on `probpipe`

- [ ] Repo scaffolding (CI, packaging) with `probpipe` as an (editable)
  dependency.
- [ ] **Adapt existing suites — don't reinvent targets/references:**
  - **posteriordb + subsampledposteriordb** (Campbell et al. 2026,
    `github.com/trevorcampbell/subsampledposteriordb`): reference draws +
    term-indexed (minibatch-enabled) log-densities → ProbPipe targets;
  - **inference_gym** (TFP spinoff, JAX backend): standardized Bayesian targets
    *with precomputed ground-truth posterior means/variances + standard errors* —
[...]

Issue #301, opened 2026-06-22
Inference validation utilities: metrics, references, SBC (in-repo)


jhuggins:
## Context

ProbPipe spans **many inference methods** — NUTS / RWMH (blackjax), SG-MCMC,
external Stan / PyMC / nutpie, BayesFlow amortized SBI (NPE/NLE/NRE), bagged
posteriors, GP-emulator surrogates — with more on the way (an approximate-MAP /
Laplace foundation, calibrated SG(L)D: #303/#304). What the library lacks is a
**method-agnostic way to score a posterior approximation against a trusted
reference** — calibration, mean/covariance accuracy, and full distributional
quality.

That capability is broadly valuable, well beyond any single method: it lets us
**validate a new inference method** when it lands, **regression-test inference
quality** as the library evolves, **compare methods** on the same target, and it
is the **dependency-light scoring core beneath `probpipe-benchmark`** (#302). The
calibrated-SGLD initiative is one concrete consumer that motivates building it
now (it needs the calibration and covariance-accuracy checks), not the purpose.

This issue is the **in-repo, dependency-light slice**: comparison metrics, a
small reference container, simulation-based calibration, and a `quantile` op. The
heavy reference suite (posteriordb / inference_gym, large datasets, leaderboards)
is #302.

Conventions: `STYLE_GUIDE.md` §8.6 (independent baselines, measured tolerances,
check both location *and* spread).

## Home: a separate `probpipe.validation` subpackage (reuse #193's diagnostics)

PR #193 builds `probpipe.diagnostics` — ESS/R-hat/MCSE (`add_mcmc_diagnostics`),
PPC/SPC, LOO, sensitivity, an ArviZ bridge (`extract_draws`/`to_arviz_dataset`),
a `posterior.diagnostics` view. **The distinction:** *diagnostics* operate on one
fitted posterior ("did this fit converge / fit the data?"); **validation** scores
a method/approximation against a known reference ("does this method recover the
right posterior?"). So these method-validation utilities live in a **separate
`probpipe.validation` subpackage**, conceptually distinct from diagnostics, and
**reuse #193's MCMC diagnostics / arviz bridge (no reimplementation of
ESS/R-hat/PPC)**. #301 proceeds **in parallel** with #193 — the additions need
only `flat_samples` + `condition_on`, both on `main` — and merges after it.

### A. Comparison metrics — `probpipe/validation/_comparison.py`

Dep-light JAX. Moment metrics consume the reference's high-precision
[...]

PR #306, opened 2026-06-22
feat(validation): posterior-vs-reference comparison metrics + quantile op (#301 PR 1)

jhuggins:
## Summary

PR 1 of #301 — the in-repo, dependency-light scoring core for **method
validation**: score a posterior approximation against a trusted reference
(mean/covariance accuracy and distributional quality). It is method-agnostic
infrastructure — used to validate and regression-test any inference method, and
as the scoring core beneath `probpipe-benchmark` (#302) — with the
calibrated-SGLD initiative (#303/#304) as one motivating consumer. It lives in a
separate `probpipe.validation` subpackage, distinct from the per-fit diagnostics
in #193 (which it will reuse for ESS/R-hat in later PRs).

## What's in this PR

**`probpipe.validation._comparison`** — plain, jit-compatible JAX scoring
functions (typed via a `DrawsLike` = array-or-`flat_samples` alias):

- **`Reference`** — analytic / long-NUTS / sandwich references behind one
  container (high-precision `(mean, cov)`, `draws`, and/or `score_fn`);
  `from_moments` validates the `(d,)` / `(d, d)` shapes.
- **`standardized_mean_error`** = `‖Σ_ref^{-1/2}(μ̂ − μ_ref)‖₂` — Mahalanobis
  mean error, in posterior-σ units.
- **`relative_cov_error`** = `‖I − Σ_ref^{-1/2} Σ̂ Σ_ref^{-1/2}‖₂` — operator norm
  of the symmetrically whitened covariance deviation, which equals the
  worst-direction variance-ratio error `max_d |1 − λ_d|`.
- **`std_ratios`**, **`sliced_wasserstein`** (random projections → *exact*
  order-statistic 1-D `W₂`, handling unequal sample sizes), **`mmd`** (unbiased
  RBF), **`ksd`** (IMQ kernel Stein discrepancy), and **`score_posterior`** (the
  aggregator / scorecard reused by the future `probpipe-benchmark`; skips
  metrics whose reference pieces are absent). The moment metrics check
  approximation/reference dimension compatibility.

**`quantile` op** — a new `SupportsQuantile` protocol + `quantile(dist, q)` op
(parallel to `mean`/`variance`/`cov`) + `RecordEmpiricalDistribution._quantile`,
weight-aware via the midpoint-CDF (Hazen, type-5) quantile used for both uniform
and non-uniform weights, so the estimator is continuous in the weights.
Registered in the workflow distribution-hint allowlist so the op keeps its
distribution argument; rejects `q ∉ [0, 1]` (eager, skipped under `jit`).

**Docs + changelog** — a new `docs/api/validation.md` page (wired into the nav),
the `quantile` op added to `docs/api/operations.md`, and a CHANGELOG entry.

## Tests

- `tests/validation/test_comparison.py` — each metric vs an independent baseline
  (closed form / known identity / clearly-separated case), with measured
  tolerances per STYLE_GUIDE §8.6; the optimal-transport metrics are
  cross-checked against **POT** (`ot.wasserstein_1d` for the exact 1-D `W₂`
  including unequal `n`, and `ot.sliced_wasserstein_distance` for the projection
  pipeline). Also covers the non-diagonal `relative_cov_error` case, the
  float-bandwidth `mmd` / `ksd` paths, jit-compatibility, input rejection
  (dimension mismatch / `< 2` draws / 3-D), distribution inputs, and the
  `score_posterior` skip / selection paths.
- `tests/core/test_quantile_op.py` — uniform (vs NumPy Hazen) and weighted (vs an
  analytic `√q` and an independent NumPy baseline) quantiles, scalar/vector `q`,
  multi-field, the single-field return type, and `TypeError` / out-of-range-`q`
  errors.
- Full `tests/core` + `tests/validation` pass (1391); `ruff` clean.

## Scope notes

- **No new runtime dependencies** — the core install stays light. **POT** (Python
  Optimal Transport) is added as a **dev-only** test baseline for the
  optimal-transport metrics. SBC + interval coverage are PR 2; the shared
  `conftest` + references + the NUTS / vanilla-SGLD validation suite are PR 3
  (per #301).
- **Distinct from #193** (`probpipe.diagnostics`): validation scores a *method*
  against a reference; diagnostics diagnose a single *fit*. This PR reuses none
  of #193 yet (only `flat_samples` + the ops already on `main`).

Part of #301.


probpipe/validation/_comparison.py
```diff
@@ -0,0 +1,352 @@
...
+    _require(ref, "mean", "cov")
+    mu_hat = _as_draws(approx).mean(axis=0)
+    diff = mu_hat - ref.mean
+    chol = jnp.linalg.cholesky(ref.cov)
+    z = jax.scipy.linalg.solve_triangular(chol, diff, lower=True)
+    return jnp.linalg.norm(z)
+
+
+def relative_cov_error(approx: Any, ref: Reference) -> Array:
+    r"""Operator-norm covariance error: ``‖I − Σ_ref⁻¹ Σ̂‖₂``.
```
SoymilkWinsAgain: The implementation of `relative_cov_error` is inconsistent with the description of "whitened covariance deviation / worst-direction variance-ratio error".

In `_comparison.py`, the current calculation is $\Sigma_{ref}^{-1} \Sigma_{hat}$, followed by $I - ...$. This matrix is ​​typically not symmetric: if the index semantics are the whitened covariance error in reference covariance coordinates, $L^{-1} \Sigma_{hat} L^{-T}$ should be calculated before taking $|I - ...|$. Current tests only cover unit/diagonal commutative scenarios and fail to capture this difference. It is recommended to modify it to a symmetric whitening form and add a test for a non-diagonal $\Sigma_{ref}$.

probpipe/core/ops.py
```diff
@@ -252,6 +254,23 @@ def cov(dist: SupportsCovariance) -> Array:
...
+@workflow_function
+def quantile(dist: SupportsQuantile, q: Any) -> Any:
+    """Compute quantile(s) of ``X ~ dist`` at probability level(s) ``q``.
+
+    ``q`` is a scalar or array of probabilities in ``[0, 1]``; the return is
+    ``T``-shaped per field (finite-sample distributions return the weight-aware
+    empirical quantile via ``_quantile``).
+
+    Requires the distribution to implement :class:`SupportsQuantile`.
+    """
```
SoymilkWinsAgain, line 269: Should add check if q is in [0, 1].

SoymilkWinsAgain (changes requested):
Also from CodeX:

The statement in the PR description that "uniform weights reduce to jnp.quantile" is outdated/inaccurate.

The actual implementation uses Hazen/type-5 midpoint-CDF, and testing is performed using `np.quantile(..., method="hazen")`. It is recommended to update the PR description; ideally, the finite-sample quantile convention should also be explicitly stated in the user-visible docstring for quantile.

probpipe/validation/_comparison.py
jhuggins: Thanks — good catch on the diagonal-only coverage. Switched to the symmetric whitened form `‖I − Σ_ref^{-1/2} Σ̂ Σ_ref^{-1/2}‖₂` (computed as `L⁻¹ Σ̂ L⁻ᵀ` via two triangular solves, `L = chol(Σ_ref)`) in f7c50a2. With the whitened matrix symmetric, the operator norm now **equals** the worst-direction variance-ratio deviation `max_d |1 − λ_d|` exactly, rather than the non-symmetry-inflated upper bound the previous form gave, and the docstring is updated to match. Added `test_relative_cov_error_non_diagonal`, which checks this against an independent NumPy eigendecomposition on a non-commuting `Σ̂`.

Two clarifications for the record:

- The previous `‖I − Σ_ref⁻¹ Σ̂‖₂` wasn't actually incorrect — it's a valid discrepancy and a genuine upper bound on the variance-ratio error, with equality exactly when `Σ̂` and `Σ_ref` commute (which is why the diagonal tests agreed). The symmetric form is just the cleaner, exact quantity, so we've adopted it.
- The left/right multiplication order doesn't change the value: `Σ̂ Σ_ref⁻¹ = (Σ_ref⁻¹ Σ̂)ᵀ`, so the operator norm is identical either way.

probpipe/core/ops.py
jhuggins: Done in 5e9fcb7 — `quantile` now raises `ValueError` when a concrete `q` falls outside `[0, 1]`. The check is guarded so it is skipped when `q` is traced (e.g. under `jit`), keeping the op jit-compatible. Covered by `test_raises_on_out_of_range_q`.

probpipe/validation/_comparison.py
SoymilkWinsAgain: I'm not sure if this will cause some minor deviations in numerical calculations. This kind of problem can sometimes occur when using low-precision matrix calculations (though not necessarily here), and I've encountered similar issues when using low precision in `PyTorch`. Perhaps the docstrings should at least be consistent?

SoymilkWinsAgain (approved):
Everything looks fine!

PR #308, opened 2026-06-22
Phase 1b (#235): EventTemplate numeric query & projection methods

arob5:
## PR-1b — EventTemplate numeric query & projection methods

Second slice of Phase 1 of #235 (follows #292 — rename + LeafSpec representation, merged).
**Purely additive**: new methods on `EventTemplate`; no renames, no breaking changes, no
dependency on later-phase batch types.

### Scope — add four methods to `EventTemplate` (`probpipe/core/record.py`), per #235 Chapter 1
- `is_numeric -> bool` (property): True iff every reachable leaf is an `ArraySpec` (recurse into
  nested templates). Implement as an explicit recursive leaf check — the source of truth; it should
  agree with `isinstance(self, NumericEventTemplate)` under the auto-promotion invariant, but the
  recursive form is robust to hand-built mixed nestings.
- `is_multi_field -> bool` (property): True iff the template describes **more than one reachable
  leaf** (recurse into nested sub-templates) — `len(self.leaf_shapes) > 1`. Note this counts leaves,
  not top-level fields, so a single top-level field nesting several leaves (e.g. leaves `a/b` and
  `a/c`) is multi-field. (Refined from #235 Chapter 1's `len(self.fields) > 1` per PR review — a
  top-level-field count is unintuitive when one field nests multiple leaves.)
- `non_numeric_fields() -> tuple[str, ...]`: top-level field names whose leaf is non-numeric
  (`OpaqueSpec`/`DistributionSpec`/`FunctionSpec`, or a nested template that isn't all-numeric).
  Powers clear inference error messages.
- `numeric_subset() -> NumericEventTemplate`: project to the `ArraySpec`-leaf subtemplate — keep
  numeric leaves (recursing into nested templates), drop opaque/distribution/function leaves, prune
  any nested template that becomes empty. Path-stable (surviving leaves keep their `/`-paths).
  Returns a `NumericEventTemplate` so `flat_size` / `numeric_leaf_shapes` are available.

### Decisions (flag if you disagree)
- `numeric_subset()` on an already-all-numeric template → returns an equal `NumericEventTemplate`
  (idempotent).
- `numeric_subset()` when **no** numeric leaves survive → raise `ValueError` (an `EventTemplate`
  needs ≥1 field; an empty numeric subset is meaningless); name the dropped fields.
- Nested fields contribute `nested.numeric_subset()`; if that would be empty, the field is dropped.
- `is_numeric` via recursive leaf check, not just `isinstance(self, NumericEventTemplate)`.

### Out of scope (do NOT do here)
- `to_vector` / `from_vector` and the `NumericRecord.flatten → to_vector` rename + consumer audit
  → **PR-1c**.
- `make_batch` / `vectorize` → later, with the `Batch` ABC / `*Array→*Batch` work (Phase 2/3); they
  consume/produce batch containers and shouldn't bind to the soon-to-be-renamed `*Array` types.
- Rewiring inference (`_inference_utils.py` / `_tfp_mcmc.py`) to call `numeric_subset()` → separate
  change (also wants `to_vector`/`from_vector`). Keep 1b additive — no behavior change to existing
  paths.
- `RandomFunction`/`RandomMeasure` population; `Tracked`/`Annotated` mixins; any `*Batch` work.

### Definition of done
- Four methods added with NumPy-style docstrings; all existing behavior unchanged (additive only).
- Full test suite green (nothing renamed → no consumer audit).
- New tests added (below).
- New code conforms to the repo's **ruff lint + format** standard (`[tool.ruff]` in
  `pyproject.toml`: line-length 100, isort import order; ruff version pinned in
  `.pre-commit-config.yaml`). The CI `lint & format` job enforces **both** `ruff check .` and
  `ruff format --check .`, so match both: run `uv run ruff check --fix` and `uv run ruff format`
  on your changed files (or `pre-commit run --files <changed>`). The repo is fully
  `ruff format`-clean — earlier guidance to skip `ruff format` was stale and produced a red CI
  check; format your files to match.

### New tests (extend `tests/core/test_event_template.py`)
- `is_numeric`: True for all-`ArraySpec` (incl. nested all-numeric); False for mixed, for a
  `DistributionSpec`/`FunctionSpec` leaf, and for a nested mixed template.
- `is_multi_field`: True/False by reachable-leaf count, incl. nested (a single top-level field with
  leaves `a/b`, `a/c` is multi-field; a single nested leaf is not) and a single opaque leaf.
- `non_numeric_fields`: exactly the non-numeric top-level names; empty for all-numeric.
- `numeric_subset`: drops opaque/dist/func; keeps numeric; recurses nested; prunes an emptied nested
  field; path-stable; idempotent on all-numeric; returns `NumericEventTemplate` with correct
  `flat_size`; raises when no numeric leaves survive.

Rationale: #235 Chapter 1 (EventTemplate API) and §7 (why inference uses `numeric_subset`). This PR
adds the methods only; inference adoption follows later.

---
*Draft/handoff PR — to be implemented by a separate session per this brief.*

### DIFF
--- a/probpipe/core/record.py
+++ b/probpipe/core/record.py
@@ -974,14 +974,42 @@ class EventTemplate:
         """
         return len(self._specs) > 1
 
+    def numeric_fields(self) -> tuple[str, ...]:
+        """Top-level field names whose leaf is numeric.
+
+        A top-level field is numeric if its spec is an :class:`ArraySpec` or a
+        nested :class:`EventTemplate` that is itself all-numeric (see
+        :attr:`is_numeric`). The exact complement of :meth:`non_numeric_fields`:
+        together the two partition :attr:`fields` (each in insertion order).
+
+        Note that this reports *top-level* fields only, so a field is numeric
+        here iff its nested sub-template is *entirely* numeric — it does not
+        descend to count partially-numeric nestings. Use :meth:`numeric_subset`
+        for the recursive, path-stable projection to numeric leaves.
+
+        Returns
+        -------
+        tuple of str
+            Names of the numeric top-level fields, in insertion order. Empty
+            when every top-level field is non-numeric.
+        """
+        result: list[str] = []
+        for name, spec in self._specs.items():
+            if isinstance(spec, ArraySpec) or (
+                isinstance(spec, EventTemplate) and spec.is_numeric
+            ):
+                result.append(name)
+        return tuple(result)
+
     def non_numeric_fields(self) -> tuple[str, ...]:
         """Top-level field names whose leaf is non-numeric.
 
         A top-level field is non-numeric if its spec is an
         :class:`OpaqueSpec` / :class:`DistributionSpec` / :class:`FunctionSpec`
         leaf, or a nested :class:`EventTemplate` that is not itself all-numeric
-        (see :attr:`is_numeric`). Used to build clear inference error messages
-        when a template cannot be projected to a numeric subset.
+        (see :attr:`is_numeric`). The exact complement of :meth:`numeric_fields`:
+        together the two partition :attr:`fields`. Used to build clear inference
+        error messages when a template cannot be projected to a numeric subset.
 
         Returns
         -------
--- a/probpipe/core/record.py
+++ b/probpipe/core/record.py
@@ -995,9 +995,7 @@ class EventTemplate:
         """
         result: list[str] = []
         for name, spec in self._specs.items():
-            if isinstance(spec, ArraySpec) or (
-                isinstance(spec, EventTemplate) and spec.is_numeric
-            ):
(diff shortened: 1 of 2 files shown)


probpipe/core/record.py
```diff
@@ -930,6 +930,150 @@ def pack(self, **field_kwargs: Any) -> Record:
...
+            if isinstance(spec, EventTemplate):
+                if not spec.is_numeric:
+                    return False
+                continue
+            # Opaque / distribution / function leaf — not numeric.
+            return False
+        return True
+
+    @property
+    def is_multi_field(self) -> bool:
```
jhuggins, line 967: This would return true if there were fields `a/b` and `a/c`, which is not intuitive. I would expect this to return true iff there is one leaf

arob5: good catch

jhuggins (approved):

PR #318, opened 2026-06-23
feat(core): add stable content fingerprinting for provenance parents

erozercan:
> **Maintainer update — rebased + review fixes.** This branch is rebased onto
> `dev/record-named-collection` (the leaf-keyed `Record`/`EventTemplate`
> redesign, #326/#327), so **this PR is now stacked on #327** and targets that
> branch — merge after #327 lands. The `/code-review` findings are addressed in
> a follow-up commit:
>
> - **WorkflowFunction** now hashes `co_names`/`co_varnames`/`co_freevars` +
>   closure cells + defaults, so a body that changes only the name called
>   (`jnp.sin` vs `jnp.cos`), a captured value, or a default is detected
>   (previously collided).
> - **Records** are hashed by their leaf-keyed `items()` — the old `fields` +
>   index pattern *crashed* on nested records under the new API.
> - **EmpiricalDistribution** uses the public `.samples`/`.is_uniform`/
>   `.log_weights` (the Record-backed subclass has no `_samples`), and `Weights`
>   are hashed by content — so reweighted (IS/SMC) posteriors and weighted
>   mixtures no longer collide.
> - **Cross-process determinism**: `set`/`frozenset` order-independent, numpy
>   scalars by dtype+bytes, `float` `-0.0`/NaN canonicalized, object-dtype arrays
>   hashed element-wise, repr fallback no longer truncated to 500 chars.
> - **Arrays**: over-cap sampling now covers the tail and slices device-side to
>   bound the host transfer.
> - `ParentInfo.fingerprint` is excluded from equality/hashing (identity is
>   structural), and `_is_*` guards catch `ImportError` only.
>
> **Deferred (open):** whether to gate fingerprinting off the default
> `LIGHTWEIGHT` path until a `cache_key_fn` consumes it, and a full
> `jax.tree_util`-based encoder generalization. The strategy table below is the
> original design; the array cap is **256 MB**.

## Summary

Adds `probpipe/core/_fingerprint.py`, a stable content hashing utility, and
wires it into the provenance system so that every `ParentInfo` descriptor now
carries a populated `fingerprint` field.

The `fingerprint(obj) -> str` function returns a 16-character hex digest
(64-bit SHA-256 prefix) that is deterministic across processes for the same
inputs. It dispatches on ProbPipe's type hierarchy:

| Input type | Strategy |
|---|---|
| `jax.Array` / `numpy.ndarray` | shape + dtype + raw bytes (256 MB cap; samples tail-covering, device-sliced windows for large arrays; object-dtype hashed element-wise) |
| Python scalars (`int`, `float`, `bool`, `str`, `None`) | direct binary encoding via `struct.pack` |
| `Record` | field names + values, hashed recursively in insertion order |
| `Distribution` | class name + distribution name + TFP parameter dict; falls back to `vars()` for non-TFP types such as `EmpiricalDistribution` |
| `WorkflowFunction` | user function bytecode (`__code__.co_code` + `co_consts` + source location), **not** the Prefect wrapper closure — so changes to the user function body are detected reliably |
| Unknown / unhashable | `repr()` truncated to 500 characters |

All ProbPipe type imports inside `_fingerprint.py` are deferred to call time
to avoid a circular import (`_distribution_base.py` already imports from
`provenance.py` at module load time). The call from `Provenance.create()` is
also deferred for the same reason.

The `fingerprint` field on `ParentInfo` was introduced in the previous PR
(which added `ProvenanceMode`, `ParentInfo`, `ProvenanceConfig`, and
`Provenance.create()`). That field was always `None` until now. This PR
populates it during provenance creation so the lineage descriptor carries a
stable identity for each ancestor.

---

## Context and three-PR roadmap

This PR is the second of three planned PRs for the provenance scalability
issue. The full plan is described in the linked tracking issue.

**PR 1 (merged)** — `ProvenanceMode` enum (`FULL` / `LIGHTWEIGHT` / `OFF`),
`ParentInfo` descriptor, `ProvenanceConfig` singleton, `Provenance.create()`
factory. Solved the core memory problem: in `LIGHTWEIGHT` mode, parent
Distributions are no longer held alive by the provenance chain.

**PR 2 (this PR)** — Stable content fingerprinting. Builds the utility that
a future Prefect `cache_key_fn` can delegate to, without touching any Prefect
code yet. Populates `ParentInfo.fingerprint` for all three provenance modes
(except `OFF`, where no `ParentInfo` is created).

**PR 3 (planned)** — Prefect result caching. Wire `persist_result=True` and a
custom `cache_key_fn` (using the fingerprinting from this PR) into the task
decorators in `_workflow_execution.py`. Depends on a serialization audit of
ProbPipe result types and alignment on remote storage configuration for
multi-machine scenarios.

---

## Linked issue(s)

Refs #117 
---

## Test plan

New test file: `tests/core/test_fingerprint.py` (41 tests).

- **Return format** — output is always a 16-character lowercase hex string.
- **Python primitives** — int, float, bool, str, None: same value → same
  digest; different values → different digest; bool and int hash differently
  even when numerically equal.
- **Array hashing** — same JAX array → same digest; different values, shapes,
  or dtypes → different digest. Also covers `numpy.ndarray`.
- **Record hashing** — same fields and values → same digest; any difference
  in field names, values, or number of fields → different digest.
- **Distribution hashing** — same parameters and name → same digest;
  different `loc`, `scale`, name, or distribution type → different digest.
  Covers both TFP-backed parametrics (`Normal`, `Beta`) and sample-based
  distributions (`RecordEmpiricalDistribution`).
- **WorkflowFunction hashing** — same function → same digest; different
  function body or different literal constants → different digest.
- **Integration with `Provenance.create()`** — `ParentInfo.fingerprint` is
  set and non-`None` after `Provenance.create()`; stable across multiple
  `create()` calls for the same parent; two parents with different parameters
  produce different fingerprints; fingerprint appears in `to_dict()` output;
  `OFF` mode produces `None` (no `Provenance` object created at all).

Existing provenance tests (`tests/core/test_provenance.py`, 58 tests) all
pass unchanged.

---

## Breaking changes

None. `ParentInfo.fingerprint` was already part of the public dataclass
(introduced in PR 1); it was previously always `None`. Callers that read
`fingerprint` now receive a string instead of `None` — this is the intended
behavior and is additive.

---

## Documentation

- [x] `_fingerprint.py` module docstring describes supported types and design
      constraints (deferred imports, size cap, digest length)
- [x] CHANGELOG noted in the linked issue

---

## Checklist

- [x] `area:provenance`, `area:core` labels applied
- [x] Linked to tracking issue above

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,8 +67,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **`ParentInfo` descriptor** (new public export).  A frozen dataclass carrying
   `type_name`, `name`, `source` (the parent's own `Provenance`, kept in all
   non-OFF modes so the ancestry DAG remains traversable), `fingerprint`
-  (reserved for a future caching layer), and `obj` (the live parent object,
-  set only in FULL mode).
+  (a 16-character stable content hash — see below), and `obj` (the live parent
+  object, set only in FULL mode).
+
+- **`ParentInfo.fingerprint` — stable content hashing for provenance parents.**
+  Every `ParentInfo` descriptor now carries a `fingerprint: str` — a
+  16-character hex digest (64-bit SHA-256 prefix) that stably identifies the
+  parent's content across processes.  The hash covers the full value: numeric
+  parameters for TFP-backed distributions, field-by-field content for Records,
+  and actual user-function bytecode for WorkflowFunctions (not the Prefect
+  wrapper closure, so changes to the function body are detected reliably).
+  Large arrays (> 1 MB) are sampled at evenly-spaced offsets rather than read
+  in full.  The fingerprint is visible in `to_dict()` output and is the
+  foundation for a future Prefect `cache_key_fn` that will enable cross-run
+  task caching and failure recovery.
 
 - **`Provenance.create()` factory classmethod.**  Centralises mode-checking:
   reads `provenance_config.mode`, wraps each parent in a `ParentInfo`, and
--- a/docs/api/provenance.md
+++ b/docs/api/provenance.md
@@ -84,6 +84,37 @@ assert any(a.obj is prior for a in ancestors)
 ancestors[0].obj.sample(key, (10,))
 ```
 
+## Content fingerprints
+
+Every `ParentInfo` descriptor carries a `fingerprint` — a 16-character hex
+string that stably identifies the parent's content across processes.  It is
+populated automatically by `Provenance.create()` and visible in `to_dict()`
+output:
+
+```python
+prior = Normal(loc=0.0, scale=1.0, name="prior")
+posterior = wf(prior)
+
+anc = provenance_ancestors(posterior)[0]
+print(anc.fingerprint)   # e.g. "8d86780c50cea472"
+
+d = posterior.source.to_dict()
+print(d["parents"][0]["fingerprint"])   # same digest
+```
+
+The fingerprint covers the full content of the parent:
+
+| Parent type | What is hashed |
+|---|---|
+| TFP-backed distribution (`Normal`, `Gamma`, …) | class name + distribution name + all TFP constructor parameters |
+| `EmpiricalDistribution` | class name + name + sample arrays |
+| `Record` | field names + values, recursively |
+| `WorkflowFunction` | user function bytecode (not the Prefect wrapper closure) |
+| JAX / NumPy array | shape + dtype + raw bytes (large arrays are sampled) |
+
+The fingerprint is intended as the foundation for a future Prefect
+`cache_key_fn` that will enable cross-run task caching and failure recovery.
+
 ## API reference
 
 ::: probpipe.ProvenanceMode
--- a/docs/api/provenance.md
+++ b/docs/api/provenance.md
@@ -107,7 +107,7 @@ The fingerprint covers the full content of the parent:
 | Parent type | What is hashed |
 |---|---|
 | TFP-backed distribution (`Normal`, `Gamma`, …) | class name + distribution name + all TFP constructor parameters |
-| `EmpiricalDistribution` | class name + name + sample arrays |
+| `EmpiricalDistribution` | class name + name + sample arrays + log-normalised weight array |
 | `Record` | field names + values, recursively |
 | `WorkflowFunction` | user function bytecode (not the Prefect wrapper closure) |
 | JAX / NumPy array | shape + dtype + raw bytes (large arrays are sampled) |
--- a/probpipe/core/_fingerprint.py
+++ b/probpipe/core/_fingerprint.py
@@ -7,7 +7,8 @@ same inputs.  Intended for provenance tracking (populating
 
 Supported types
 ---------------
-- ``jax.Array`` / ``numpy.ndarray`` — shape + dtype + bytes (capped at 1 MB)
+- ``jax.Array`` / ``numpy.ndarray`` — shape + dtype + bytes (up to
+  ``max_array_bytes``; sampled beyond that)
 - Python scalars (``int``, ``float``, ``bool``, ``str``, ``None``)
 - ``Record`` — field names + values, hashed recursively
 - ``Distribution`` — class name + distribution name + numeric parameters
@@ -16,6 +17,21 @@ Supported types
 
 All imports of ProbPipe types are deferred to call time so this module can
 be imported early in the package without triggering circular-import issues.
+
+Notes
+-----
+**Collision semantics:** the digest is a 64-bit prefix of SHA-256.  At
+ProbPipe scale a collision is negligible in probability, but when one does
+occur it produces a *wrong cached result*, not a cache miss.
+
+**JAX x64 mode:** ``str(dtype)`` is used for array dtype, so digests for
+the same array differ between ``jax_enable_x64=False`` (default, float32)
+and ``jax_enable_x64=True`` (float64).  Digests are stable within a fixed
+JAX configuration.
+
+**Python version portability:** bytecode (``WorkflowFunction``) is not
+portable across Python minor versions — a 3.12-compiled digest differs from
+a 3.13 one for the same source.
 """
 
 from __future__ import annotations
@@ -26,13 +42,25 @@ from typing import Any
 
 __all__ = ["fingerprint"]
 
-# Arrays larger than this threshold are hashed by sampling bytes at
-# evenly-spaced offsets instead of reading the whole buffer.  1 MB is
-# large enough that most parameter arrays are hashed completely.
-_FULL_HASH_BYTES = 1 << 20  # 1 MB
+import numpy as _np
+
+try:
+    import jax as _jax
+
+    _JAX_ARRAY_TYPE = _jax.Array
+except ImportError:
+    _JAX_ARRAY_TYPE = type(None)  # type: ignore[assignment,misc]
+
+_NP_ARRAY_TYPE = _np.ndarray
+
+# Default cap: arrays up to 256 MB are hashed in full (zero-copy via
+# memoryview).  Arrays beyond this are sampled at evenly-spaced offsets so
+# large posteriors are still discriminated without a full buffer read.
+# Pass ``max_array_bytes=None`` to always hash the full buffer.
+_DEFAULT_MAX_ARRAY_BYTES: int = 256 << 20  # 256 MB
 
 
-def fingerprint(obj: Any) -> str:
+def fingerprint(obj: Any, *, max_array_bytes: int | None = _DEFAULT_MAX_ARRAY_BYTES) -> str:
     """Return a 16-character hex digest that stably identifies *obj*'s content.
 
     The digest is deterministic within a Python session and across processes
@@ -44,6 +72,11 @@ def fingerprint(obj: Any) -> str:
     obj:
         Any ProbPipe object or Python primitive.  Unknown types fall back
         to a ``repr()``-based hash.
+    max_array_bytes:
+        Arrays whose byte size is at or below this threshold are hashed in
+        full (zero-copy via ``memoryview``).  Larger arrays are sampled at
+        evenly-spaced offsets.  Pass ``None`` to always hash the full buffer.
+        Default: 256 MB.
 
     Returns
     -------
@@ -51,7 +84,7 @@ def fingerprint(obj: Any) -> str:
         A 16-character hex string (64-bit prefix of a SHA-256 digest).
     """
     h = hashlib.sha256()
-    _update(h, obj, depth=0)
+    _update(h, obj, depth=0, max_array_bytes=max_array_bytes)
     return h.hexdigest()[:16]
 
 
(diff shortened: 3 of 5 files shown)


jhuggins (changes requested):
I dug into whether the 1 MB array cap (`_FULL_HASH_BYTES`) makes sense, since the sampling it triggers is the one place in the array path where two different arrays can silently collide. Short version: I'd raise it a lot and make it user-settable.

**Cost.** SHA-256 here runs at ~2.45 GB/s (hardware-accelerated), linear in size:

| array | full hash |
|---|---|
| 1 MB | 0.4 ms |
| 64 MB | 27 ms |
| 256 MB | 110 ms |
| 1 GB | ~0.4 s |
| 10 GB | ~4 s |

Fingerprinting runs once per provenance-creation, not in a JIT/hot loop, so tens of ms is a small cost. (The device→host transfer the comment flags is free on the CPU backend — it only costs anything on GPU/TPU.)

**Two issues with the cap sitting at 1 MB:**

1. It triggers far too early. A 4-chain × 1000-draw posterior crosses 1 MB at only ~32 parameters, so most real posteriors land on the sampling path — exactly where a collision returns the wrong cached result down the line. We're trading correctness for ~25 ms.
2. The sampling branch calls `arr_np.tobytes()` on the *whole* array before sampling, so it pays the full memcpy anyway and only skips the (accurate, cheap) SHA pass. At 256 MB the sampling path is ~27 ms, ~25 of which is just that copy — so it isn't buying the speed it appears to.

**Suggestion:**

- Default the cap to **256 MB**, and make it a parameter with `None` = no limit (always hash the full buffer). 256 MB is ~110 ms worst-case here and covers many realistic problems. Though I'd be open to increasing it to 1 GB if we want to increase safety at the cost of some performance. 
- Hash zero-copy from the buffer instead of `.tobytes()`, so full hashing becomes the cheap default:

```python
_DEFAULT_MAX_ARRAY_BYTES = 256 << 20  # 256 MB; None = no cap (always full hash)

def fingerprint(obj, *, max_array_bytes: int | None = _DEFAULT_MAX_ARRAY_BYTES) -> str:
    ...  # threaded through _update -> _update_array

def _update_array(h, arr, max_bytes):
    buf = np.ascontiguousarray(np.asarray(arr))      # (keep the existing try/except guard)
    h.update(f"array:{buf.shape}:{buf.dtype}:".encode())
    if max_bytes is None or buf.nbytes <= max_bytes:
        h.update(memoryview(buf).cast("B"))          # full, zero-copy
    else:
        # sample by slicing buf (a strided view) — O(sample), not O(n); no full copy
        ...
        h.update(struct.pack(">Q", buf.nbytes))
```

Also, the sliced sample should be approximately equal in size other cap, so there's as much precision as possible.

jhuggins:
Nice, tightly-scoped PR — the deferred-import circular break, type-tagged dispatch, and OFF-mode handling are all clean, and the simple-parametric / Record / primitive paths are correct. The issues below are concentrated in the `repr()`-fallback and bytecode paths, where the "stable, content-complete digest" contract doesn't yet hold for several real ProbPipe types. None is load-bearing today (nothing consumes `fingerprint` until the planned caching PR), but each becomes a *wrong cached result* or a *silent cache miss* there.

## Must fix

**1. `WorkflowFunction` fingerprint is not stable across processes.** `_fingerprint.py:267` _(reproduced)_
`repr(code.co_consts)` embeds a memory address (`<code object <lambda> at 0x…>`) whenever the user function body contains a nested `def`, `lambda`, or generator expression (and comprehensions on CPython ≤ 3.11, before they were inlined). Across two separate processes the *same source* hashed to `585802782dc03fb8` vs `dee326273929efc0`, while a flat-body function was stable. This breaks the module's stated "deterministic across processes" guarantee and means the future cache key never hits for non-trivial functions — the common case. While here, also drop `co_filename` + `co_firstlineno` (`:268-269`): an absolute path and line number make the digest machine- and position-dependent, defeating cross-machine caching. Recurse into nested code objects instead of repr-ing them, e.g.:

```python
import types
def _hash_code(h, code):
    h.update(code.co_code)
    for c in code.co_consts:
        if isinstance(c, types.CodeType):
            _hash_code(h, c)        # recurse: no repr, no addresses
        else:
            _update(h, c, 0)
```
…and note in the docstring that bytecode is not portable across Python minor versions.

**2. `repr()`-fallback collisions — content that distinguishes two objects is dropped.** Two confirmed cases, same root cause (the object reaches the `repr:` branch at `_fingerprint.py:112` and its repr elides values):

- **EmpiricalDistribution weights** — `_fingerprint.py:218-228` _(reproduced)_. `_w` is a `Weights` object whose `__repr__` (`_weights.py:592`) is just `Weights(n=3, uniform)`. Two empiricals with identical samples but weights `[0.7, 0.2, 0.1]` vs `[0.1, 0.2, 0.7]` both hash to `eeec05969a1f5429` — a reweighted posterior (IS/SMC) is indistinguishable from the original. (`Weights.__hash__` at `_weights.py:505` already folds in `log_weights.tobytes()`; the fingerprint path bypasses it.)
- **Composite / nested TFP distributions** — `_fingerprint.py:216` _(reproduced)_. When a TFP `parameters` dict holds a nested distribution/bijector (KDE → `MixtureSameFamily`, `TransformedDistribution`, joints), it hits the repr fallback, whose TFP repr is shape/dtype-only. `KDEDistribution([0,1,2])` and `KDEDistribution([0,1,99])` both hash to `e0e70a773e336f5c`.

Fix: hash EmpiricalDistribution samples **and** weight arrays explicitly (retiring the `vars()`/`_SKIP` fallback at `:218-223` — it's untested, and three of its four skip-names aren't even attributes of the class that uses the branch), and recurse into nested distribution/bijector parameters via `_update` rather than letting them repr. Then update the docs table at `docs/api/provenance.md:110`, which currently claims empiricals hash "sample arrays."

**3. Stale docstring (1 line).** `provenance.py:52-54`
`ParentInfo.fingerprint`'s docstring still reads "currently always `None`" — this PR is precisely what populates it.

## Should fix

**4. `fingerprint()` crashes on Python ints outside the signed-64-bit range.** `_fingerprint.py:90-91` _(reproduced)_
`struct.pack(">q", obj)` raises `struct.error` for `2**70`. `Provenance.create()` swallows it to `None`, but the utility recurses into Records/lists/dicts, so a big int anywhere raises — `fingerprint([1, 2**100])` confirmed to raise. Use `obj.to_bytes(n, signed=True)` with the byte length folded in, or hash `str(obj)`.

**5. Over-broad silent `try/except Exception → None`.** `provenance.py:166-169`
Swallows every fingerprint error with no log, so a real bug degrades to `fingerprint=None`, indistinguishable from "disabled" — which in the caching PR surfaces as "caching mysteriously never fires." (It does *not* catch the collisions in #2 — those return wrong-but-valid digests.) Narrow the catch and emit a `logging.debug`/warning.

**6. Tests are green only because they miss the buggy paths.** `tests/core/test_fingerprint.py`
That's how #1–#4 slipped through. Add coverage for: nested-body `WorkflowFunction` across two processes (#1), non-uniform-weight empirical (#2a), a composite distribution such as KDE (#2b), big-int input (#4), and the `try/except → None` branch (#5, via monkeypatch); plus the currently-untested `depth > 32` guard (`:64`), `list`/`tuple`/`dict` dispatch (`:100-111`), and unknown-type `repr` fallback (`:112`). Separately, the scalar "stable" tests (`test_int_stable`, `test_float_stable`, …, and especially `test_numpy_different_from_jax`) compare a value to *itself*, so they would pass even if the digest embedded `id(obj)` — rewrite them to construct two distinct-but-equal inputs (the array/Record/distribution stability tests already do this correctly).

## Minor / polish

- **64-bit collision semantics** deserve a docstring line: as a cache key, a collision returns a *wrong* result, not a miss (~50% at ~5 B distinct keys — fine at ProbPipe scale, worth stating). `_fingerprint.py:36`
- **dtype is hashed as `str(dtype)`**, so the digest depends on JAX x64 mode; stable in-repo, but a one-clause caveat would be honest. `_fingerprint.py:138`
- **Distribution `name` is hashed twice** (directly, and again inside the TFP `parameters` dict), and that dict also pulls in `validate_args`/`allow_nan_stats` — construction flags, not content. `_fingerprint.py:206-216`
- **`_update` re-imports numpy/jax on every recursive node**; hoist for deep Records. `_fingerprint.py:68-79`
- **`fingerprint` participates in `ParentInfo.__eq__`/`__hash__`**, so a swallowed-`None` vs a real digest makes two descriptors for the same ancestor compare unequal. DAG dedup is unaffected (it uses `_parent_key`), but worth awareness. `provenance.py`
- **`hashlib._Hash` annotation** isn't a runtime attribute — it works only under `from __future__ import annotations` (and resolves under mypy via typeshed). Cosmetic; `Any` would be cleaner. `_fingerprint.py:63`
- **Cheap missing edge tests:** `False` vs `0`, NaN/inf floats, empty containers, and a `WorkflowFunction` with no `_func`/`__code__`.

> 🤖 **AI-assisted code review by Claude**. Items tagged _(reproduced)_ were confirmed by running code against `de04076`; digests are included as evidence. The large-array size cap is covered in the separate comment above and omitted here.

jhuggins:
@erozercan I think the underlying issue here is using `__repr__` to get a fingerprint. Instead, take advantage of more precise mechanisms like pyre registration (which will soon be improved by @arob5's in-progress work).

erozercan:
Thanks for the detailed review @jhuggins ! Addressed everything below.

**Array cap & sampling path (large comment + #1 in must-fix)**

Raised the default from 1 MB to 256 MB (`_DEFAULT_MAX_ARRAY_BYTES = 256 << 20`) and made it a `max_array_bytes` keyword arg threaded all the way through `fingerprint()` → `_update()` → `_update_array()`. Pass `None` for no cap. Also fixed the sampling branch — it was calling `arr_np.tobytes()` on the whole array before sampling, which paid the full copy cost anyway. Replaced it with `memoryview(buf).cast("B")` slices so both the full and sampled paths are now zero-copy.

**WorkflowFunction not stable across processes (#1 must-fix)**

Added `_hash_code()` which recurses into nested `types.CodeType` objects instead of repr-ing them. `repr(co_consts)` was embedding memory addresses for any nested `def`, `lambda`, or generator, which made the digest process-dependent. Also dropped `co_filename` and `co_firstlineno` since those are machine- and line-position-dependent. Added a note in the docstring about Python version portability.

**repr() fallback collisions (#2 must-fix)**

- **EmpiricalDistribution weights:** Now explicitly hashes `_samples` and `_w.log_normalized` instead of falling through to the `vars()`/repr path. A reweighted posterior (IS/SMC) now produces a different fingerprint from the original.
- **Nested TFP distributions (KDE etc.):** Added `_is_tfp_object()` / `_update_tfp_object()` that duck-types on `.parameters` + `.log_prob`/`.forward` and recurses into nested params via `_update` instead of repr. Updated the docs table to note that empiricals now hash weights too.

**Stale docstring (#3 must-fix)**

Updated `ParentInfo.fingerprint` docstring — removed "currently always None", replaced with an accurate description.

**Big int crash (#4 should-fix)**

Replaced `struct.pack(">q", obj)` with `str(obj).encode()`. Confirmed `fingerprint(2**100)` and `fingerprint([1, 2**100])` no longer raise.

**Silent try/except (#5 should-fix)**

Added `import logging` and `logger = logging.getLogger(__name__)` to `provenance.py`. The except block now calls `logger.warning()` with the type name, dist name, and exception before setting `fp = None`.

**Tests (#6 should-fix)**

Rewrote the scalar "stable" tests to construct two distinct-but-equal inputs (e.g. `fingerprint(int("42")) == fingerprint(int("42"))`) so they'd fail if `id(obj)` were embedded. Added coverage for:
- Nested-lambda `WorkflowFunction` stability across two separate subprocesses
- Non-uniform weights empirical (same samples, different weights → different digest)
- KDE distribution (different data → different digest)
- Big int (`2**100` in scalars and inside a list)
- Depth > 32 guard
- `list`/`tuple`/`dict` dispatch
- Unknown-type repr fallback
- `try/except → None` branch via monkeypatch (verifies warning is emitted)

**Polish**

- Added collision semantics note and JAX x64 caveat to the module docstring
- Filtering `_TFP_SKIP = {"name", "validate_args", "allow_nan_stats"}` to avoid double-hashing the distribution name and construction flags
- Hoisted `import numpy` and `jax.Array` type to module level so they're not re-imported on every recursive call

jhuggins:
Rebased this onto `dev/record-named-collection` and applied the fingerprint review fixes (full list in the updated description). Summary for @erozercan / @arob5:

**Why the rebase.** The fingerprint dispatch was written against the pre-#326 `Record` API. On the leaf-keyed model, `record[<interior node>]` raises, so nested-record fingerprinting *crashed* — the record path had to become a leaf-keyed `items()` walk, which only makes sense on top of #326. So **this PR is now stacked on #327** and targets `dev/record-named-collection`; please merge it after #327.

**Fixes applied** (one follow-up commit + regression tests; `ruff` clean, 1543 `tests/core/` pass):
- **WorkflowFunction** — hash referenced names (`co_names`/`co_varnames`/`co_freevars`) + closure cells + defaults, so `jnp.sin` vs `jnp.cos`, captured values, and defaults are detected (were colliding).
- **Empirical / Weights** — public `.samples`/`.is_uniform`/`.log_weights` (the Record-backed subclass has no `_samples`) + content-hashed `Weights`, so reweighted IS/SMC posteriors and weighted mixtures no longer collide.
- **Cross-process determinism** — `set`/`frozenset` order-independent, numpy scalars by dtype+bytes, `float` `-0.0`/NaN canonicalized, object-dtype arrays element-wise, repr fallback no longer truncated.
- **Arrays** — over-cap sampling now covers the tail and slices device-side to bound the host transfer.
- **`ParentInfo.fingerprint`** excluded from equality/hashing (structural identity); `_is_*` guards catch `ImportError` only.

jhuggins (approved):

PR #327, opened 2026-06-26
feat(records): Record/EventTemplate as a leaf-keyed named collection (#326)

### LINKED ISSUE #326: Record/EventTemplate as a named collection of objects (follow-up to #317)
# `Record` / `EventTemplate` as a named collection of objects (follow-up to #317)

**Status.** Design specification for a dedicated, breaking follow-up to #317. It
depends on #317 (which splits `EventTemplate` into its own module and makes every
`Record` carry an authoritative `EventTemplate`) having landed first. This document
is written to be implemented by a contributor or session that was not part of the
discussion, so it spells the design out in full.

**How to read this.** Sections describe the agreed design in detail. All of #326's
own open questions have been resolved and folded into the sections below; the final
*Open questions* section indexes them and what each resolved to. Two boxes in §6 are
marked **➡️ Deferred** rather than open: they are collection-view questions that are
**out of scope for #326** and belong to future phases of the #235 refactor, kept here
(and in the Bookkeeping note) only so a future implementer sees them. (The
construction contract in §5 — once the most delicate part, since its edge-case rules
determine correctness — is now fully specified.)

**Bookkeeping (do not skip).** When the PR implementing this issue lands, **update
the #235 master plan** — specifically its "Update (2026-06) — what PR #317 landed,
and the #326 redesign" section — to record what this PR shipped (the
collection-of-objects public API, the `_tree` rename, path-keyed construction, the
removed/renamed surface) and to reconcile any items that were resolved or that
remain deferred (e.g. the batched analog of `from_field_values` — decompose/reconstruct
a batch's per-element field values, tracked in #235 as batched
`to_leaf_list`/`from_leaf_list` — the distribution-side shape reshape, and the open
value-model questions). **Two collection-view questions are explicitly out of scope
here and must be carried into the relevant #235 phases** (they are stated in full in
the deferred boxes in §6): how a `RecordDistribution` exposes its collection view over
its support (the distribution chapter), and how the batch types layer the collection
view over the batch axis with dual integer/string indexing (the `*Array → *Batch`
redesign). Do **not** close #235; it spans several further phases.

---

### 1.1 The viewpoint

A `Record` is most naturally understood as **an ordered collection of named
objects**. For example, one draw from a model might be a collection holding an
object named `weights`, an object named `bias`, and an object named
`noise_scale`. The objects are *named* (each has a label) and *ordered* (they have
a stable, insertion order). A user thinks of a `Record` much like a labelled struct
or an ordered dictionary: "a bag of values, each with a name."

Sometimes the objects are organized hierarchically. A model might group two of its
objects under `physics` — `physics/force` and `physics/mass` — alongside a
top-level `observation`. To represent that grouping, a `Record` stores its objects
in the **leaves of a tree** whose interior nodes are themselves `Record`s
(sub-collections). But the tree is an organizational convenience: conceptually the
`Record` is still just the collection of named objects that live at the leaves. The
objects we actually read, transform, and serialize are the leaf objects; the
interior nodes exist only to express grouping.

When a `Record` has no grouping (a *flat* `Record`), its leaves are exactly its
top-level entries, so "the collection of objects" and "the top-level entries"
coincide. Nesting is the only thing that makes them differ.

### 1.2 The problem this fixes

Today the public, dictionary-style surface of `Record` and `EventTemplate`
(`keys`, `values`, `items`, iteration, membership, `len`) ranges over the
**top-level** entries of the storage tree, while the operations that do real work
(`map`, serialization to/from vectors, equality, validation) range over the
**leaves**. For a flat object the two are the same, so the discrepancy is invisible;
for a nested object they disagree, and nothing in the API signals which level a
given operation acts on. This split — a top-level dictionary view sitting next to a
leaf-level operational surface, with no label distinguishing them — is the source
of recurring confusion.

### 1.3 The change, and how it touches each class

The change is to make the **primary public API present the collection-of-objects
view directly**: the standard mapping operations (`keys`/`values`/`items`/iteration/
`len`) range over the objects (the tree's leaves), each addressed by its full path
(e.g. `"physics/mass"`). The tree structure is **not** hidden, though — it is kept
as a *labelled, secondary* surface: a one-level `children` mapping and a separate
`at_path` navigator (`obj.at_path("physics")` → the subtree). The mapping operators
themselves (`[]`, `in`, iteration) stay strictly leaf-keyed, so they behave as an
ordinary mapping; reaching an interior node is a deliberate, separately named step.
The original failure this fixes was the
*absence of a label* distinguishing the leaf view from the one-level view, not the
existence of two views; so the design keeps both, clearly named, rather than
deleting either. Three layers are involved, and it is important throughout this
document to be precise about which one is meant:

- **`Record` (values).** A `Record`'s objects are **data values** — arrays,
  scalars, strings, arbitrary opaque Python objects, or (at interior nodes) nested
  `Record`s. `Record` adds the value-specific operations (mapping over values,
  numeric conversion, equality of data, hashing, JAX pytree registration).

- **`EventTemplate` (specs).** An `EventTemplate` is the *schema* for such a
  collection. Its objects are **leaf specs** — `ArraySpec` / `OpaqueSpec` /
  `DistributionSpec` / `FunctionSpec` — and its interior nodes are nested
  `EventTemplate`s. It is therefore also a named, ordered collection (of specs,
  not values), with the same hierarchy mechanics. `EventTemplate` adds the
  schema-specific operations (numeric queries and projection, shape/vector-size
  queries on the numeric specialization, schema inference).

- **`_NamedTree` (shared mechanics).** Both classes inherit their structural
  behavior from the `_NamedTree` mixin: how a field is named and addressed by path,
  how the storage tree is traversed to its leaves, and how a stored child is
  recognized as an interior node versus a leaf. The mixin operates on a tree and is
  agnostic about whether the leaves hold values or specs, so it is written in tree
  vocabulary ("leaf", "node"). The collection vocabulary ("field") belongs to the
  `Record`/`EventTemplate` surface; the mixin is where "field" is realized in terms
  of "leaf".

For a flat collection, none of the observable behavior changes (paths are just
names, and the leaves are the top-level entries). Only nested collections gain the
flattened, path-addressed object view.

### 1.4 What stays out (and why this is just #317 continued)

`#317` already made each `Record` carry an authoritative `EventTemplate` and fixed
the internal-node test so a `Record`'s leaves agree with its template's leaves.
This follow-up builds directly on that: it reorganizes the *public surface* around
the leaves that #317 already made authoritative. The batch types (`RecordArray` /
`NumericRecordArray`) and the distribution types (`RecordDistribution` and
friends) are touched only as described in §6: #326 brings them the leaf-keyed
*field-navigation* surface but leaves their *batch axis* for #235 (see "Transitional
state of the batch / distribution surface" and the deferred boxes in §6).

---

## 2. Terminology

These terms are used precisely throughout. Several **redefine** words from #317;
the redefinitions are intentional and are summarized in the change table below.

- **Field.** One named object in the collection. A field has a **name** (its full
  path, e.g. `"physics/mass"`; for a flat collection, a single label like
  `"weights"`) and a **value**. A field's value is a *data object* in a `Record`
  and a *leaf spec* in an `EventTemplate`. *(This redefines #317's "field", which
  meant a top-level entry. The decision to use the word "field" — rather than, say,
  "component" — is settled.)*

- **Path.** The `/`-joined sequence of names that addresses a position in the
  storage tree, read from the root downward (e.g. `"physics/mass"`, or the
  single-segment `"observation"`). A path may address **either** a field (a leaf)
  **or** an interior node. A **partial path** is one that stops at an interior node
  rather than a field (e.g. `"physics"`). Paths are what the `at_path` navigator
  (§3.0) accepts.

- **Key.** A path that addresses a **field** (a leaf) specifically. Keys are the
  paths returned by `keys()`, and they are exactly the paths the mapping operators
  (`obj[key]`, `key in obj`, iteration) accept. Every key is a path; a partial path
  is a path but **not** a key. This distinction is the one rule the rest of the
  contract follows from: *the mapping surface is keyed by fields; the `at_path`
  navigator ranges over all paths.*

- **Collection.** A `Record` or `EventTemplate` viewed as its set of fields. "The
  collection" is the object as a whole.

- **Storage tree (the tree).** The internal nested representation of a collection.
  Interior nodes are sub-collections of the same class (a `Record` inside a
  `Record`, an `EventTemplate` inside an `EventTemplate`); the field objects live at
  the leaves. The storage tree is held in the `_tree` attribute (§4).

- **Subtree.** An interior node of the storage tree — itself a `Record` (within a
  `Record`) or an `EventTemplate` (within an `EventTemplate`).

- **Children (the one-level view).** The immediate entries of a collection — its
  storage tree at depth one. Exposed publicly as the `children` mapping
  (`local_name → child`, where a child is a leaf object or a subtree). This is the
  *labelled* top-level view that the design keeps alongside the leaf-keyed
  collection view, so the two strata are never confused (see §3). It is the
  successor to #317's `fields` property.

- **Leaf / node.** Tree-structure words used **only inside `_NamedTree`**. A *leaf*
  is where a field object sits; a *node* is a subtree. The public `Record` /
  `EventTemplate` surface says "field", not "leaf" — with two deliberate exceptions
  that read better in tree language: the `children` view, and the
  `_NamedTree.is_leaf(path)` predicate (aliased as `is_field(path)` on `Record` /
  `EventTemplate`, where the field vocabulary fits the public surface).

- **Canonical order.** The order in which fields are enumerated: a depth-first walk
  of the storage tree that visits children in insertion order. For a flat
  collection this is simply insertion order. Every field-wise operation
  (`keys`/`values`/`items`, `map`, vector serialization) uses this one order.

### 2.1 Naming change table

| #317 / current | Follow-up | Scope |
|---|---|---|
| `field` means a top-level entry | `field` means a named object, addressed by its full path | concept (both classes) |
| `_fields` (on `Record`), `_specs` (on `EventTemplate`) — the storage attribute | `_tree` (both classes use the same name) | storage |
| `_field_map()` — hook method returning the storage mapping | removed; the mixin reads `self._tree` directly | `_NamedTree` |
| `keys()` / `values()` / `items()` / iteration / `len` / `in` range over top-level entries | range over fields (objects keyed by path) | shared protocol |
| `leaf_paths` — public property listing leaf paths | replaced by `keys()`; use `tuple(obj.keys())` for a tuple | shared (kept internal to `_NamedTree`) |
| `fields` — property returning top-level names | replaced by **`children`** — a one-level mapping `local_name → (leaf \| subtree)`; top-level names are `tuple(obj.children)` | `_NamedTree` (both classes) |
| `in` / `[]` reach leaves *and* subtrees | **`in` / `[]` are leaf-only** — `key in obj ⇔ obj[key] works ⇔ key in keys()` (a clean mapping); subtree navigation moves to the new `at_path` (next row). `is_leaf(path)` / `is_field(path)` are retained as a readable spelling of leaf membership (equal to `key in obj`) | `_NamedTree` (+ alias on consumers) |
| (none — subtree reached via `[]`) | **`at_path(path)`** — the navigator: returns the field value for a key, or the **subtree** for a partial path; accepts the same `/`-string or tuple forms as `[]` | `_NamedTree` (both classes) |
| `to_dict()` — nested dict (on `Record`) | **`to_nested_dict()`** — nested dict, on `_NamedTree` (works for both classes); the flat path-keyed dict is the builtin `dict(obj)` | `_NamedTree` |
| (none — a nested `dict` was opaque data) | **`from_nested_dict(d, event_template=None)`** — inverse of `to_nested_dict()`; reads every `dict` level as a subtree, or (with a template) lets the template decide leaf-vs-node so opaque `dict` data is preserved (§5.4) | `_NamedTree` |
| `map(fn)` (on `Record`) | `map(f, /, *args, **kwargs)` — maps over leaf objects, structure-preserving; **moved to `_NamedTree`** | `_NamedTree` |
| `map_with_names(fn)` — passed the local name | `map_with_keys(f, /, *args, **kwargs)` — passes the full key path; **moved to `_NamedTree`** | `_NamedTree` |
| `Record.to_leaf_list()` (added in #317) | **removed** — `list(record.values())` | `Record` |
| `EventTemplate.to_leaf_list(value)` (added in #317) | **removed** — `[value[p] for p in tpl.keys()]`, kept only as an internal helper for `to_vector` | `EventTemplate` |
| `from_leaf_list(values)` (added in #317) | **`EventTemplate.from_field_values(values)`** — template-driven reconstruction from ordered field values (§3.2) | `EventTemplate` |
| `replace` / `merge` / `without` (top-level only) | path-aware structural edits; **moved to `_NamedTree`**; a `Record` threads its template by applying the same edit to `event_template` (§3.1) | `_NamedTree` |
| `select` / `select_all` (top-level only) | path-aware; still a value-only `dict` projection (remap form kept); not a template-threading update | `Record` |
| `to_vector` / `from_vector` | unchanged | `NumericRecord` / `NumericEventTemplate` |
| `_node_type()` (added in #317) | unchanged | `_NamedTree` |
| (none) | `_walk_leaves()` / `_rebuild_from_leaves()` — new shared recursion primitives | `_NamedTree` |

**Resolved — the `to_leaf_list` / `from_leaf_list` helpers.** The two `to_leaf_list`
helpers are removed: extracting a collection's own leaves in canonical order is now
`list(record.values())`, and extracting a foreign value's leaves through a template is
the one-liner `[value[p] for p in tpl.keys()]`, kept only as an internal helper that
`to_vector` uses. The reconstruction survives, renamed to
`EventTemplate.from_field_values(values)` (§3.2). It is genuinely distinct from
path-keyed construction (§5): the caller supplies only an ordered sequence of values,
with no names and no structure, so the operation **must** be driven by a template —
the template is the sole source of the field paths and the tree shape. That is why it
lives on `EventTemplate` (a template method) rather than as a bare `Record`
constructor, and why `from_vector` builds on it (§3.2). The asymmetry with
`to_vector` / `from_vector` (which keep both directions) is deliberate: a vector
needs both a non-trivial ravel and its inverse, whereas the "to a list of field
values" direction is pure extraction (`list(record.values())`) and so earns no
separate method. *(The path-tuple sub-question is also resolved: top-level names are
`tuple(obj.children)` and field paths are `tuple(obj.keys())`, so no extra named
accessor is needed.)*

---

## 3. Public API contract

Both classes expose the **same collection protocol**, differing only in what a
field's value is. The protocol is defined once here; the per-class subsections then
state what the objects are and list the operations unique to each class.

### 3.0 The shared collection protocol

Let a collection's fields, in canonical order (§2), have paths `p₀, p₁, …` and
objects `o₀, o₁, …`.

- **`keys()`** → an iterator over the field paths `p₀, p₁, …`. This is the canonical
  enumeration of the collection and **replaces** #317's `leaf_paths`. (Use
  `tuple(obj.keys())` when a tuple is wanted.)
- **`values()`** → an iterator over the field objects `o₀, o₁, …`.
- **`items()`** → an iterator over `(pᵢ, oᵢ)` pairs.
- **`__iter__`** (i.e. `for k in obj`) → the field paths, matching `keys()`.
- **`__len__`** (i.e. `len(obj)`) → the number of fields (leaves) — *not* the number
  of top-level entries.
- **`key in obj`** → membership over fields only: `True` if and only if `key` is one
  of `keys()` — equivalently, if and only if `obj[key]` succeeds. A **partial path is
  not a member** (`"physics" in obj` is `False`), because a partial path is not a
  key. This makes `in` / `[]` / iteration all coincide, exactly as for an ordinary
  `dict`. To test for a navigable interior node, use `obj.children` (one level) or
  `at_path` (any depth).
- **`obj[key]`** → the field object at `key` (for `Record`, the value; for
  `EventTemplate`, the spec). `key` must address a field; a missing key — **or a
  partial path that stops at a subtree** — raises `KeyError` (reach a subtree with
  `at_path`). A tuple key is identical to its `/`-joined string form:
  `obj["physics", "mass"]` equals `obj["physics/mass"]`.
- **`obj.at_path(path)`** → the navigator over the whole tree: returns the field
  value when `path` is a key, or the **subtree** (a `Record` / `EventTemplate`) when
  `path` is a partial path. It accepts the same `/`-string and tuple forms as `[]`,
  and is the **one** operator that reaches interior nodes — the mapping operators
  never do. `obj.at_path(key)` and `obj[key]` agree on every key, so `[]` is exactly
  the leaf-restricted `at_path`.
- **`obj.is_field(path)`** (aliasing `_NamedTree.is_leaf(path)`) → `True` iff `path`
  is a key (resolves to a field, not a subtree). With leaf-only membership this is
  exactly `path in obj`; it is kept as a readable spelling for call sites that want
  to say "is this specifically a field?" explicitly.
- **`obj.children`** → a read-only ordered mapping `{local_name → child}` for the
  **immediate** entries (depth one), where a child is a leaf object or a subtree.
  This is the labelled top-level view (successor to #317's `fields`); top-level
  names are `tuple(obj.children)`. It makes subtrees enumerable (`children.values()`
  yields leaves *and* subtrees) — the leaf-keyed `keys`/`values`/`items` never
  yield subtrees.
- **`dict(obj)`** → a flat dictionary `{p₀: o₀, p₁: o₁, …}` keyed by field path (a
  consequence of the leaf-keyed protocol — `dict(x)` calls `x.keys()`/`x[k]`). It is
  **value-only and lossy w.r.t. the schema** (and, for a `Record`, the concrete
  class): see §5 for what round-trips and what does not.
- **`obj.to_nested_dict()`** → a nested `dict` mirroring the storage tree (subtrees
  become nested dicts). This is the tree-shaped export (the renamed #317
  `Record.to_dict`), defined on `_NamedTree` so it works for both classes. Distinct
  from the flat builtin `dict(obj)`; the two names signal the two shapes.

For both `obj[path]` and `at_path`, a path that reaches nothing raises `KeyError`, a
path that tries to descend through a field (a leaf) rather than a subtree also raises
`KeyError`, and a key that is neither a string nor a tuple of strings raises
`TypeError`.

> **Note — the mapping surface is an ordinary mapping; navigation is separate.**
> Because `[]`, `in`, and iteration are all keyed by fields, the three coincide
> exactly as they do for a plain `dict` (`key in obj ⇔ obj[key] works ⇔ key in
> list(obj)`), and a `Record` / `EventTemplate` can be treated as an honest mapping
> (`dict(obj)`, `{**obj}`, and mapping-based dispatch all behave). Reaching an
> interior node is the one thing the mapping surface deliberately does *not* do; that
> is `at_path`'s job. This is what separates the two concerns the old API conflated —
> *which fields are in the collection* (the mapping) versus *how the tree is
> navigated* (`at_path`) — by giving them two differently named operators rather than
> two meanings of `[]`. The price is paid once: a caller that previously reached a
> subtree through `obj["physics"]`, or tested `"physics" in obj`, moves to
> `obj.at_path("physics")` / `obj.children` (the sites are listed in §6).
>
> *Why `at_path`, not `at`.* `at` collides with JAX's functional-update idiom
> (`array.at[i].set(...)`), which reads as a write; for a JAX-native library a
> read-only `record.at[...]` would be a trap. `at_path` is unambiguous and pairs with
> the key-vs-path vocabulary of §2.

For a **flat** collection, every path is a single name, so `keys()`/`values()`/
`items()` produce exactly what #317's top-level versions produced — flat callers
observe no change.

**Shared transformations — `map` and `map_with_keys`.** Both classes inherit two
structure-preserving transformations from `_NamedTree`:

- **`map(f, /, *args, **kwargs)`** → a new collection of the **same class and the
  same structure** (identical names and nesting), with `f` applied to every leaf
  object in canonical order (§2). `f` is called as `f(leaf, *args, **kwargs)`, where
  `leaf` is a *value* for a `Record` and a *spec* for an `EventTemplate`; any extra
  `*args` / `**kwargs` are forwarded to `f` unchanged and are **constant across
  leaves** (not varied per leaf). `f` is positional-only because `map` takes no
  options of its own — the structure is fixed, the template is rebuilt (below), and a
  `Record`'s `name` / `source` are reset (it is a semantic transformation).
- **`map_with_keys(f, /, *args, **kwargs)`** → identical, except `f` is called as
  `f(key, leaf, *args, **kwargs)`, where `key` is the leaf's full key path (§2). This
  replaces #317's `map_with_names`, which passed only the local name at each level.
  The key is the universal handle: anything else one might want per leaf — the leaf's
  spec, path-dependent logic — is one lookup away (`obj.event_template[key]`), so no
  separate `map_with_template` variant is needed.

**The structure-preservation guarantee.** `map` and `map_with_keys` *never* change
the tree's shape: every leaf slot stays a leaf and every interior node stays an
interior node. This is enforced, not hoped for — the traversal reuses the existing
structure rather than re-deriving it from `f`'s outputs, so an `f` output is always
placed in the leaf slot it came from and is never re-interpreted as nesting (a plain
`dict` returned by `f` is an opaque leaf, exactly as in construction, §5.1). If `f`
returns an instance of the node type — a `Record` from `Record.map`, an
`EventTemplate` from `EventTemplate.map` — that would introduce nesting, so it raises
`ValueError` naming the key. (Reusing the structure is also what lets the operation
*catch* this, rather than silently re-nest.)

**What each class does with the result.** Because the structure is fixed, only the
leaves change, and each class finishes differently:

- **`Record`:** the new leaf values are stored, and the result's `event_template` is
  rebuilt by keeping `self`'s structure and **re-inferring each leaf's spec** from its
  new value. Re-inference is lossy (it cannot recover an `ArraySpec`'s `dtype` /
  `support`, an `OpaqueSpec`'s `meta`, etc.), so even an identity `map` may produce a
  record that does not compare equal to the original — inherent, because `f` is opaque
  and its outputs carry no spec to preserve. The result is `type(self)`, so
  `NumericRecord.map` returns a `NumericRecord` and therefore **requires `f` to return
  numeric leaves**; a non-numeric result raises rather than silently demoting to a
  base `Record`.
- **`EventTemplate`:** the new leaves *are* specs, so there is no separate template to
  infer — but each `f` output must be **coercible to a spec**. That check belongs to
  `_to_spec` (it accepts a leaf spec or the shape-tuple / `None` sugar and raises
  `TypeError` otherwise), not to the traversal, which stays agnostic about what a leaf
  is. Numeric auto-promotion re-applies, so mapping every spec to a numeric one yields
  a `NumericEventTemplate`.

> **Note — co-mapping is a separate, future operation.** The `*args` / `**kwargs` here
> are *constant* arguments to `f`, which differs from `jax.tree_util.tree_map`, where
> extra positional arguments are *additional trees mapped leaf-by-leaf*. A leaf-wise
> co-map over two same-shaped collections is a genuinely different operation and is
> intentionally **not** part of `map`; it may be added later under its own name.

### 3.1 `Record` (a collection of data values)

A `Record`'s field objects are **data values**: arrays, scalars, strings, `xarray`
/ `pandas` objects, arbitrary opaque Python objects, or nested `Record`s at
interior nodes. `values()` yields these data objects; `items()` yields
`(path, value)`.

Record-specific operations:

- **`map` / `map_with_keys`** → inherited from `_NamedTree` (§3.0, "Shared
  transformations"). For a `Record` they map over field **values** and rebuild the
  result's `event_template` by re-inferring each leaf's spec; the result is
  `type(self)`, so `NumericRecord.map` returns a `NumericRecord` and requires `f` to
  return numeric leaves.
- **`replace` / `merge` / `without`** → immutable, path-aware updates inherited from
  `_NamedTree`, each returning a new collection of the **same class**. They are
  structural tree edits, so they apply equally to a `Record` (a tree of values) and an
  `EventTemplate` (a tree of specs); a `Record` additionally **threads its template by
  applying the same edit to `self.event_template`** (detailed in the Resolved note
  below). In brief: `without(*paths)` drops the named keys or subtrees
  (`r.without("physics/mass")` keeps `physics/force`; `r.without("physics")` drops the
  group); `merge(other)` returns the union (deep, path-aware); `replace(updates)`
  overwrites the objects at existing paths.
- **`select`** → a **dict** projection helper for splatting into calls
  (`predict(**params.select("r", "K"))`), path-aware, with a remap form
  (`params.select(growth_rate="r")` → `{"growth_rate": <value of r>}`). It returns a
  value-only `dict`, **not** a `Record`, so it carries no `event_template` and is not
  part of the update family above. (Its view-returning behavior on the batch and
  distribution subclasses is part of the deferred §6 redesign.)
- **`to_numeric()`** → conversion to `NumericRecord` (unchanged from #317).
- **Numeric serialization** (`to_vector`, and `vector_size`) lives on
  `NumericRecord` and is unchanged.
- **Equality and hashing** are unchanged in semantics from #317 (equality requires
  equal class, equal `event_template`, and field-by-field equal data; hashing is
  structural). They are re-expressed on the shared recursion primitive (§4) but
  observe the same contract.
- **JAX pytree registration** is unchanged: a `Record`'s children are its field
  values; `name` / `source` / `event_template` remain runtime-only metadata. *(Note:
  this means the pytree round-trip `tree_unflatten(tree_flatten(r))` is itself
  schema- and class-lossy — it re-infers the template and rebuilds a base `Record`
  — exactly like `dict(r)`; see §5.)*
- **`children`, `at_path`, `is_field`, `to_nested_dict`** are inherited from
  `_NamedTree` (§3.0): the one-level mapping, the whole-tree navigator, the
  leaf-test predicate, and the tree-shaped dict.
- **`dict(record)`** is the flat path→value view — value-only, so it does **not**
  round-trip on its own (see §5 for the conditional round-trip).

**Resolved — template handling for `map` / `map_with_keys`.** These reuse `self`'s
structure exactly and re-infer only the per-leaf specs of the result, as specified in
§3.0. Re-inference is lossy (an identity `map` may not compare equal to the original),
which is inherent: `f` is opaque, so its outputs carry no spec to preserve. There is
nothing to "thread" beyond the structure, which is already fixed.

**Resolved — template handling for the value updates.** Each update applies the same
structural edit to the value tree and to `self.event_template`, so the fields the edit
leaves in place keep their authoritative specs and only genuinely-new objects are
re-inferred. This makes "the fields provably left unchanged" precise — they are
exactly the fields the structural edit preserves — and is the reason `replace` /
`merge` / `without` live on `_NamedTree` (the same edit is meaningful on a tree of
specs). Per operation:

- **`without(*paths) -> C` — fully threaded, no inference.** Drops the given keys
  (leaves) or partial paths (whole subtrees). A missing path raises `KeyError`;
  removing every field raises `ValueError` (a collection needs at least one field);
  surviving fields keep their order. Every survivor is unchanged, so the new template
  is `self.event_template.without(*paths)` — nothing is inferred. On an
  `EventTemplate` this *is* the operation (the pruned specs are the result).

- **`merge(other: C) -> C` — fully threaded, no inference.** Returns the union,
  **deep and path-aware**: `{"physics/mass": …}` merged with `{"physics/force": …}`
  yields both under a combined `physics` group. Collisions use the construction rules
  (§5.1): a leaf key present in both inputs, or a field-versus-prefix clash (one side
  has `a` as a field while the other has `a/…`), raises `ValueError`. Order is
  `self`'s fields followed by `other`'s new fields, merging into existing subtrees
  where prefixes coincide. Both inputs carry authoritative templates and no field
  changes, so the new template is `self.event_template.merge(other.event_template)`.
  (This generalizes #317's shallow top-level merge, which is the flat case.)

- **`replace(_updates: Mapping[str, Any] | None = None, /, **updates) -> C` — partial
  threading.** Overwrites the objects at the given paths, mirroring the constructor's
  input form: a path-keyed positional mapping (`r.replace({"physics/mass": m})`) or
  flat keywords (`r.replace(obs=y)`), not both. Every path must already exist
  (`KeyError` otherwise — `replace` edits, it does not add; adding is `merge`'s job).
  A partial path replaces a whole subtree, and replacing is permissive about kind: a
  leaf may be replaced by a subtree value (a `Record`) or vice versa, since `replace`
  is a deliberate, targeted edit (unlike `map`, which forbids any structure change).
  Every untouched path keeps its authoritative spec; each replaced path takes the
  **new value's** spec — its authoritative `event_template` when the new value is a
  `Record` (no loss), otherwise an inferred leaf spec (lossy, since a raw value
  carries none, exactly as in `map`). So the new template is
  `self.event_template.replace({p: spec_of(new_value)})` over only the replaced paths.

**Resolved — nested vs. flat dict exports.** There are two distinct dict shapes and
they now have two distinct names, so neither is a footgun:
- `dict(obj)` (builtin) → the **flat** `{path → object}` view. This shape is *forced*
  by the leaf-keyed protocol (`dict(x)` calls `x.keys()`/`x[k]`), not a free choice.
- `obj.to_nested_dict()` → the **nested** dict mirroring the storage tree (the
  renamed #317 `Record.to_dict`), moved to `_NamedTree` so both classes have it.

### 3.2 `EventTemplate` (a collection of leaf specs)

An `EventTemplate`'s field objects are **leaf specs** (`ArraySpec`, `OpaqueSpec`,
`DistributionSpec`, `FunctionSpec`); its interior nodes are nested
`EventTemplate`s. `values()` yields specs; `items()` yields `(key, spec)`;
`template[key]` yields a spec; `template.at_path(partial_path)` yields a nested
`EventTemplate`. The shared `children` / `at_path` / `is_field` / `to_nested_dict`
surface (§3.0) is inherited from `_NamedTree` (here `children` values are leaf specs
or nested templates, and `to_nested_dict` yields a nested dict of specs).

`map` / `map_with_keys` are also inherited (§3.0, "Shared transformations"): for an
`EventTemplate` they map over **specs**, and each `f` output must be coercible to a
spec via `_to_spec` (a leaf spec or the shape-tuple / `None` sugar, else `TypeError`).
The mapped specs *are* the new template — nothing is inferred — and numeric
auto-promotion re-applies. The structural edits `replace` / `merge` / `without` (§3.1)
are inherited too and operate on specs (a `Record` uses exactly these to thread its
own template); `replace`'s new values are specs, normalised through `_to_spec`. There
is no template to re-infer here — for an `EventTemplate`, the edited specs are the
result.

EventTemplate-specific operations (semantics unchanged from #317 except for being
re-keyed by path):

- **`is_numeric`**, **`numeric_subset()`** — numeric queries and the
  mixed→numeric projection.
- **`infer_from(value)`** — best-effort, lossy schema inference (a `Record`
  returns its stored template; a mapping is inferred field by field). It must build
  the same nested structure the shared construction step (§5.1) produces.
- On the numeric specialization **`NumericEventTemplate`**: **`leaf_shapes`**,
  **`vector_size`**, **`to_vector`** / **`from_vector`** — unchanged in semantics.
- **`from_field_values(values)`** — reconstruct a `Record` from an ordered sequence
  of field values, taking the names and the tree shape from this template (the values
  fill the leaves in canonical order, §2). It raises `ValueError` when the number of
  values does not equal the number of fields (`len(self)`); per-leaf shape/dtype is
  not checked, exactly as construction validates structure only (§5.2). The resulting
  `Record` carries `self` as its **authoritative** `event_template` — nothing is
  inferred — so the round-trip is faithful: `tpl.from_field_values(list(r.values()))
  == r` whenever `r.event_template == tpl`. The export side is just `list(r.values())`,
  which is why only the reconstruction earns a method (§2.1). On
  `NumericEventTemplate` the result is a `NumericRecord`, and `from_vector` builds on
  this: it slices the 1-D vector into per-leaf arrays (using `leaf_shapes`), reshapes
  each, and then calls `from_field_values` to assemble them.
- **Equality and hashing** — order-sensitive over the `(path, spec)` items
  (unchanged contract from #317; re-expressed on the shared primitive).

The numeric 1-D (de)serialization (`to_vector` / `from_vector`) stays on the
numeric subclasses (`NumericRecord` / `NumericEventTemplate`) and is unchanged: it
ravels and concatenates the numeric field objects in canonical order, which is a
different operation from the general object collection described here.

---

### 4.1 The `_tree` storage attribute

Both classes store their data in a single attribute named **`_tree`** (replacing
#317's `_fields` on `Record` and `_specs` on `EventTemplate`). `_tree` is an
ordered mapping from a **local name** (one path segment, never containing `/`) to a
**child**, where a child is either a field object (a leaf — a value for `Record`, a
spec for `EventTemplate`) or a subtree (a nested same-class collection). The
nesting in `_tree` is exactly the storage tree; the flat, path-addressed view is
computed from it.

### 4.2 The mixin hooks

`_NamedTree` provides all the shared mechanics over `self._tree`:

- The data hook `_field_map()` from #317 is **removed**. The mixin reads
  `self._tree` directly. *(This is the rename agreed in discussion: "the mixin
  attribute that stores the data is `_tree`.")*
- **`_node_type()`** (added in #317) is unchanged: a classmethod each family
  narrows to itself (`Record` → `Record`, `EventTemplate` → `EventTemplate`) so a
  collection only ever descends into a child of its own kind. A child is an
  interior node if and only if it is an instance of `_node_type()`; otherwise it is
  a leaf (a field object). This is what guarantees a `Record`'s field paths agree
  with its `EventTemplate`'s field paths.
- The mixin implements the entire shared protocol (§3.0) and path machinery
  (parsing `/`-paths and tuple paths, descending through subtrees, raising the
  right errors) **once**, over `self._tree` and `_node_type()`. `Record` and
  `EventTemplate` do not re-implement the protocol. *(The batch and distribution
  types currently override parts of it; see §6.)*
- New shared members on the mixin: **`children`** (read-only ordered view of
  `self._tree`), **`at_path(path)`** (the whole-tree navigator: descend `self._tree`
  by `/`-string or tuple path, returning a leaf object or a subtree, raising
  `KeyError` past the end or through a leaf), **`is_leaf(path)`** (now exactly
  `path in self`, since membership is leaf-only — `True` iff `path` is a key), and
  **`to_nested_dict()`** (recurse the tree, subtrees → nested dicts). The mapping
  operators are the leaf-restricted face of `at_path`: `__getitem__` calls `at_path`
  and raises `KeyError` if the result is a subtree, and `__contains__` is `True` iff
  that succeeds. `Record` and `EventTemplate` add a thin **`is_field`** alias for
  `is_leaf` so the public surface reads in field vocabulary; `children` / `at_path` /
  `to_nested_dict` keep their names on both (tree vocabulary that reads naturally).

### 4.3 Shared recursion primitives

To avoid the duplicated leaf recursion that exists today (each of `map`,
vector serialization, equality, and validation hand-rolls its own descent), the
mixin defines two primitives that all field-wise operations are expressed on:

- **`_walk_leaves()`** → yields `(path, leaf_object)` for every field in canonical
  order. This *is* the implementation of `items()`.
- **`_rebuild_from_leaves(values)`** → builds a new collection with the same nested
  shape as `self`, taking field objects from `values` in canonical order. It **reuses
  `self`'s structure** rather than re-deriving it, so it places each value in the leaf
  slot it came from and raises `ValueError` if a value is a node-type instance (which
  would introduce nesting); the per-class leaf handling differs (a `Record` stores the
  value and re-infers its leaf spec; an `EventTemplate` runs the value through
  `_to_spec`). This is the engine behind `map` / `map_with_keys` (§3.0),
  `from_field_values`, `from_vector` (slice the vector into per-field arrays, then
  rebuild), and the value side of path-keyed construction.

Re-expressing `map`, the (de)serialization helpers, and the field-by-field portion
of equality on these two primitives replaces several independent recursions with
one traversal and one rebuild to define, test, and maintain.

The mixin keeps tree vocabulary internally (`_walk_leaves`, `_node_type`, "leaf",
"node"); the field vocabulary is the `Record` / `EventTemplate` surface. There is
no separate translation layer beyond the fact that the public method names
(`keys`/`values`/`items`) are already neutral and the few field-named methods live
on the subclasses.

---

## 5. Constructors

The construction contract is now fully specified. Both classes build their storage
tree (§4.1) the same way — through one shared unflattening step (§5.1) — and then
add their own concerns: a `Record` resolves `name` and its authoritative
`event_template` (§5.2); an `EventTemplate` normalises spec sugar and auto-promotes
to `NumericEventTemplate` (§5.3). Path-keyed construction lets a collection be built
from a flat `{path: object}` mapping — a genuine ergonomic that matches how
JAX/optax/flax users think about flattened parameter trees — but it is **not** a
faithful bijection from `dict(r)`, because a value-only mapping drops the schema and
the concrete class (§5.2). Two named convenience constructors (§5.4) cover the cases
the deliberately-restrictive bare constructor does not.

### 5.1 Shared construction mechanics (`_unflatten`)

Both constructors accept a single positional mapping (or keyword fields) and turn it
into the nested storage tree through one shared step on `_NamedTree`. Recall the
key-vs-path distinction (§2): a **key** addresses a field (a leaf); a **path** may
also address an interior node. Construction speaks in keys that may be *path-shaped* —
a key containing `/` denotes nesting.

**How nesting may be expressed — exactly two ways.** A collection nests only when the
caller writes one of:

- a **path key** in the positional mapping (`{"physics/mass": m}`), or
- a **pre-built node value** (`physics=Record(mass=m)` / `physics=EventTemplate(...)`).

Nothing else nests. In particular a plain `dict` **value** is *not* structural: in a
`Record` it is an ordinary opaque-leaf value (a `dict` is legitimate data), and in an
`EventTemplate` the spec normaliser rejects it. The unflattening step therefore reads
only the *keys*; it never looks inside a value to decide structure. (The convenience
constructor in §5.4 is the opt-in way to treat a nested `dict` as structure.)

**The unflattening step.** `_unflatten` turns a positional mapping into a nested
ordered mapping by exploding `/`-delimited keys:

1. **Validate each key** in input order: a key must be a non-empty `str` (a non-`str`
   key raises `TypeError`); it is split on `/`, and every segment must be non-empty
   (so `""`, `"a/"`, `"/a"`, and `"a//b"` each raise `ValueError`). Construction keys
   are **strings only** — tuple keys are an access convenience (§3.0), not a
   construction form; restricting construction to strings removes any chance of the
   same path arriving in two spellings.
2. **Group by first segment**, preserving the order in which each distinct first
   segment first appears. (The *first segment* of `"physics/mass"` is `"physics"`.) A
   segment that appears with no remainder is a **complete key**; a segment that
   appears as the start of a longer key is a **prefix**.
3. For each first segment: if it is a complete key, its value occupies that slot
   directly; if it is a prefix, recurse on the remainders (`{"force": f, "mass": m}`
   after stripping `physics/`) to build the nested sub-mapping.

**The field-xor-prefix rule (covers prefix collisions and mixed forms).** A first
segment may be used **either** as a complete key **or** as a prefix of longer keys,
never both. The check looks only at the keys, not the values, so all of these raise
`ValueError` (*"name 'a' is used both as a field and as a path prefix"*):

```
{"a/b": x, "a": 5}        {"a/b": x, "a": {"b": y}}        {"a/b": x, "a": Record(b=y)}
```

We never merge whatever sits at the complete key with its path siblings; the
ambiguity (which value wins, in what order, how conflicts resolve) is reported rather
than guessed.

**Canonical order after unflattening.** The field order is the first-appearance order
of each distinct first segment, applied recursively within each subtree. So

```
Record({"physics/force": f, "observation": y, "physics/mass": m}).keys()
    == ("physics/force", "physics/mass", "observation")
```

— `observation` follows the whole `physics` subtree because the *prefix* `physics`
first appears before `observation`, even though `physics/mass` is written last. This
order is observable through `keys()`, so it is part of the contract.

**Keyword fields are always literal.** Path unflattening applies **only** to the
positional mapping. Keyword names are Python identifiers and cannot contain `/`, and
a `/` smuggled through `**{"a/b": x}` is rejected (the existing `_check_no_path_sep`
guard). Nesting via keywords is therefore expressed only by passing a nested node
value, never a path.

`EventTemplate` and `Record` share this entire step unchanged; they differ only in
what a leaf is and in the per-class steps that follow.

### 5.2 `Record` constructor

`Record(_fields=None, /, *, name=None, event_template=None, **fields)` resolves the
positional-or-keyword input, unflattens it (§5.1), materialises each slot — a leaf
value stored as-is, a sub-mapping built into a child `Record` — and then settles two
`Record`-specific concerns:

- **`name`** defaults as in #317 (a derived `"record(...)"` label) and is otherwise
  taken from the keyword.
- **`event_template`** is either supplied or inferred. When inferred, it is built from
  the materialised tree via `infer_from` (lossy — it cannot recover an `ArraySpec`'s
  `dtype` / `support`, an `OpaqueSpec`'s `meta`, or a `DistributionSpec` /
  `FunctionSpec`). When supplied, construction validates **structure only** — field
  names and tree shape (which slots are leaves vs. subtrees, recursively); per-leaf
  shape/dtype/kind is the producing generator's responsibility and is **not** checked.
  A mismatch raises `ValueError` naming the offending `/`-path.

**Subtrees carry their slice of the authoritative template.** A supplied
`event_template` already describes the whole tree, all the way to the leaves, so each
child `Record` stores the corresponding slice (`event_template.at_path(name)`) rather
than re-inferring its own — no information is dropped at any depth. It guarantees the
consistency invariant, for any partial path `p`:

```
record.at_path(p).event_template == record.event_template.at_path(p)
```

The invariant holds in the inferred case too: with no template supplied, each child
infers its own and the parent infers *from* its children (`infer_from` reads each
child's `event_template`), so the parent's slice for a subtree *is* that child's
template. Navigating the value and navigating the schema always agree on subtrees.
(This settles the template handling for *construction*; the separate question of how
the value-transforming operations `map` / `replace` / `merge` / `without` handle the
template remains open — §3.1.)

**The round-trip is conditional — `dict(r)` is value-only.** `dict(r)` yields
`{path → value}` and carries no schema and no class, so `Record(dict(r))` re-infers a
lossy template and builds a base `Record`. For any record whose template carries
`dtype` / `support` / `meta` / a `Distribution` / a `Function`, and for any
`NumericRecord`, `Record(dict(r)) != r` — unavoidable from a value-only export. The
faithful round-trip is the schema- and class-bearing one:

```
type(r)(dict(r), event_template=r.event_template) == r
```

### 5.3 `EventTemplate` constructor

`EventTemplate(_field_specs=None, /, **field_specs)` mirrors §5.1 over specs instead
of values, so that

```
EventTemplate({"physics/force": ArraySpec(...), "physics/mass": ArraySpec(...),
               "observation": ArraySpec(...)})
    == EventTemplate(physics=EventTemplate(force=ArraySpec(...), mass=ArraySpec(...)),
                     observation=ArraySpec(...))
```

It carries neither `name` nor `event_template`; its only per-class steps are:

- **Spec sugar.** Each leaf value is normalised by `_to_spec`: a bare shape tuple
  becomes an `ArraySpec`, `None` becomes an `OpaqueSpec`, an already-built leaf spec
  passes through, and a nested `EventTemplate` value is a subtree. Anything else
  raises `TypeError`, naming the field. (This is why a plain `dict` cannot reach an
  `EventTemplate` leaf — there is no spec for it.)
- **Numeric auto-promotion.** When every field is numeric the instance is promoted to
  `NumericEventTemplate` (the existing `__new__` hook). The hook inspects the input
  values, which path keys do not change — only their grouping — so promotion is
  unaffected by unflattening; nested numeric subtrees, built through the same
  constructor, auto-promote independently, so a numeric subtree is itself a
  `NumericEventTemplate`.

Because a template and the value it describes share this one construction path, a
`Record` and its `EventTemplate` always have identical field paths. `infer_from` and
`from_field_values` build templates through the same step.

### 5.4 Convenience constructors

The bare constructor is deliberately restrictive (a plain `dict` is opaque data, and
splatting path keys through `**` is rejected). Two cases this makes awkward have named
entry points instead, each the inverse of an export shape (§3.0):

- **A flat dict of path keys needs no new constructor.** The positional form already
  *is* the path-keyed entry point: `Record({"physics/mass": m, "observation": y})`.
  Splatting (`Record(**flat)`) is intentionally unsupported because path keys are not
  identifiers; pass the mapping positionally instead. This is the inverse of the flat
  `dict(r)` export, conditional on schema + class as in §5.2.

- **A genuinely nested dict uses `from_nested_dict`** — the inverse of
  `to_nested_dict()`, defined on `_NamedTree` so both classes have it. On its own it
  walks the dict treating **every** `dict` level as structure (a subtree) and every
  non-`dict` value as a leaf. The deliberate restriction is that, used this way, it
  cannot also carry a `dict` as opaque leaf data — that is the price of reading every
  `dict` as structure.

  **`from_nested_dict(d, event_template=tpl)` lifts that restriction.** When a template
  is supplied, the template — not the value's Python type — decides structure at each
  position: a slot the template marks as an interior node is recursed into, and a slot
  it marks as a leaf keeps its value verbatim (so a `dict` there is preserved as opaque
  data). Construction then validates structurally and carries the template slices down
  exactly as in §5.2. This is the same principle as the bare constructor — the
  template is the authority on structure — applied to a nested-dict input.

---

## 6. Scope, migration, and blast radius

The observable changes are: (a) `keys()` / `values()` / `items()` / iteration /
`len` flip from top-level entries to **leaf fields by path**; (b) the public
`fields` property is replaced by **`children`**; (c) `leaf_paths` is replaced by
`keys()`; and (d) `in` / `[]` become **leaf-only**, with subtree navigation moving
to the new **`at_path`** method. For a *flat* collection (a)–(d) are all no-ops —
every path is a single name and every name is a key — so flat callers are
unaffected; only *nested* collections observe a change.

**Note — the two sites that reach a subtree through `[]` / `in` now migrate.** The
clean leaf-only mapping reintroduces exactly the two sites an earlier #326 draft
worried about, and they now need a one-line change each:
- `probpipe/core/_record_distribution.py:196` — `if key not in template:`, where
  `key` is a single top-level name that may be an interior node. Move to
  `if key not in template.children:` (one-level membership), and the error message's
  `template.fields` to `template.children`.
- `tests/distributions/test_product.py:1205` — the assertion block
  `assert "physics" in s; … s["physics"] …`, where `"physics"` is an interior node.
  Move the interior-node access to `at_path`:
  ```python
  phys = s.at_path("physics")        # was s["physics"]; "physics" in s is now False
  assert isinstance(phys, Record)
  assert "mass" in phys              # unchanged — "mass" is a key of the subtree
  assert "force" not in phys
  ```
  The leaf accesses inside the subtree (`"mass" in phys`, `phys["mass"]`) are
  unchanged, because they are keys; only the interior-node hops move to `at_path`.
  **Note that here `s` is the *batch* type** (`sample_shape=(5,)` → a `RecordArray`),
  so this migration requires `at_path` to exist on `RecordArray`. That is why
  `at_path` is part of the minimal batch-surface ship (see "Transitional state of the
  batch / distribution surface" below); without it this test would not run green.

**What actually needs migrating.** Two things, both mechanical:
- **Iteration / `keys` / `values` / `items` / `len` on *nested* collections** now
  range over leaf fields, not top-level entries. The #317-era audit found flat
  callers unaffected and almost no source site that iterates a *nested* collection's
  protocol; the few that want one level move to `children`.
- **`.fields` → `children`** — the real bulk. There are ~144 `.fields` references in
  `probpipe/` (plus tests); most are one-level work (`for f in x.fields`,
  `set(template.fields)`, `{f: s[f] for f in s.fields}`) and migrate to `children`
  (iterating a mapping yields its keys = top-level names). The one rough spot is
  positional `s.fields[0]` → `next(iter(s.children))`.

### Transitional state of the batch / distribution surface

This is the one place #326 knowingly ships an inconsistency, so it is stated plainly
here and the line is drawn deliberately. After this PR, a single `Record` /
`EventTemplate` is leaf-keyed, while the batch type `RecordArray` /
`NumericRecordArray` and the distribution type `RecordDistribution` still range their
mapping protocol over **top-level** entries. Left unmanaged that re-creates, across
the single↔batch boundary, exactly the unlabelled level-confusion this whole change
removes. The resolution rests on one observation and one minimal ship.

**The observation — a batch has two orthogonal axes.** A single `Record` has only a
*field axis* (its named-leaf collection). A batch has that *and* a *batch axis* (its
N elements). No single mapping protocol can mean both, which is the real reason a
batch must be a distinct type from a single value (`len` alone is irreconcilable:
field count vs. batch size). The two axes map onto the surface as follows, and this
is the **end-state design** a future implementer should build toward:

| Surface | single `Record` | batch (`RecordBatch`, end state) | Axis |
|---|---|---|---|
| `at_path`, `children`, string-keyed `[]` (leaf-only), `is_field` | field navigation | **identical** — field navigation | field axis (orthogonal to the batch axis) |
| `len`, `__iter__`, integer `[]` | leaves / leaf keys / (n/a) | batch size / elements / one element | batch axis |
| `keys` / `values` / `items`, `in` | the field mapping (leaf-keyed) | **the genuinely open question** (deferred — see below) | — |

The first row is *axis-orthogonal*: it navigates fields without touching the batch
axis, so it is identical on a single value and on a batch, and it carries no
contradiction. The middle row is where a batch legitimately differs, and that
difference is **labelled** (it is the batch axis), not the unlabelled level-confusion
§1.2 describes. Only the last row is genuinely undecided for a batch (it is deferred
OQ — see the box below).

**Why the field surface lives on `_NamedTree`, and what that buys later.** Because
`at_path` / `children` / `is_field` are defined on the `_NamedTree` mixin (§4.2), not
on `Record`, they can be shared by a type that is *not* a `Record`. In the end state,
`RecordArray` is replaced by a `RecordBatch` that **no longer inherits `Record`** (it
should not carry single-value semantics — the `float` / `jnp.asarray` / `.shape`
shims, single-value equality, `to_numeric`): it takes the field-navigation surface
from `_NamedTree` and the batch axis from the planned `Batch` ABC
(`len` / `__iter__` / `batch_shape` / indexing). Placing the field surface on
`_NamedTree` in #326 is therefore a *precondition* for that collapse, not merely
compatible with it.

**Minimal ship with #326.** Bring the **field-navigation surface** to the batch type
now, and defer the batch axis:
- `RecordArray` / `NumericRecordArray` *is-a* `Record` today, so it **inherits**
  `at_path` / `children` / `is_field` from the #326 `_NamedTree` surface for free. The
  one deliberate change is to make its string-keyed `[]` **leaf-only** (so
  `arr["physics"]` raises and `arr.at_path("physics")` returns the subtree), matching
  a single `Record`. This makes nested *field* access consistent across single↔batch
  and is what the migrated `test_product.py:1205` needs (it calls `at_path` on a
  batch). The surface is **permanent** — it survives the `RecordBatch` collapse
  unchanged (only what `at_path` *returns* evolves with the batch representation), so
  it is not throwaway work.
- **Deferred to #235 Chapter 2:** the batch-axis operators — `len`, `__iter__`,
  integer `[]` — and the meaning of `keys` / `values` / `items` / `in` on a batch.
  These keep their current behavior for now. In particular `RecordArray.__len__` is
  **not** changed to leaf count here: the end state is `len == batch_shape[0]`, so
  leaf-keying it now would be a value we immediately re-change (field count → leaf
  count → batch size). Leaving it is one change, owned by #235, instead of two. The
  stale comment at `_record_array.py` (which says `__len__` "matches
  `Record.__len__`") is now false and should be updated to record this transitional
  inconsistency and point at #235 Chapter 2.
- `RecordDistribution` stays **fully deferred**. It is a `Distribution`, not a
  `_NamedTree`, so `at_path` / `children` would be net-new work there, and its
  collection view over its support is itself a deferred open question (the
  distribution chapter — see the box below). Nothing forces it now: the migrated test
  operates on *samples* (values), not on the distribution. Its top-level field
  surface stays as-is, documented as transitional.

The residual, documented inconsistency after #326 is therefore narrow: on
`RecordArray` (and the whole of `RecordDistribution`) `keys` / `values` / `items` /
`len` / `__iter__` / `in` still range top-level, while the field-navigation surface
and the single `Record` are leaf-keyed. That residue is confined to exactly the
batch-axis-entangled operators that #235 Chapter 2 owns.

**The `.fields` → `children` sweep is the prerequisite migration.** A `.fields`
call-site sweep (the ~144 references) should be done up front, alongside
re-confirming the `keys`/`values`/`items` audit. Most sites are one-level and map
directly to `children` (which preserves the top-level view `.fields` provided, now
as a mapping). The one genuine nested traversal at
`probpipe/inference/_approximate_distribution.py:229–258` recurses via `.fields` +
indexing; it can stay a one-level recursion over `children`, or be simplified onto
the shared `_walk_leaves` / `_rebuild_from_leaves` primitives.

**Note — the JAX pytree round-trip is also schema/class-lossy (a #317 carry-over).**
`_record_flatten` stores only field *names* in the pytree aux and `_record_unflatten`
re-infers the template and rebuilds a base `Record` — so `tree_unflatten(tree_flatten
(r))` is lossy for the same reason `dict(r)` is (§5). This is pre-existing #317
behavior, but #317's PR description claims "the template lives in the pytree aux,"
which the code contradicts; reconcile that wording (and decide whether the template
*should* live in the aux) as a small side item when implementing this.

Everything else — the many `.items()` / `.keys()` / `.values()` uses across
`probpipe/core/` and `probpipe/distributions/` — operates on private backing dicts
or on dict-returning properties, and is unaffected.

**Resolved — `len(observed_data)` at `_predictive_check.py:88` is unaffected.**
`observed_data` is a generic type parameter `D`, and `len(observed_data)` is the
fallback for `num_observations` — the number of observations (the leading-axis count)
of the dataset. That is array / batch semantics, not `Record` mapping semantics: a
`Record` is *one* structured value, so its `len` (a field count under either #317 or
this design) was never the observation count, independent of the leaf-keyed change
here. The site therefore needs no change. A structured or `Record`-valued dataset
should pass `num_observations` explicitly, which the function already supports (the
`len` fallback applies only when `num_observations` is `None`).

> **➡️ Deferred to a future #235 phase — `RecordDistribution`'s collection view over its support.**
> This is **out of scope for #326** and recorded here so a future implementer sees it.
> A `RecordDistribution` is a distribution over a (possibly nested) collection of
> fields, and should respect that nested structure exactly as a `Record` does. This
> is **orthogonal to probabilistic dependence**: a joint distribution still has a
> field-collection *support*; dependence between fields is a separate matter and does
> not change the support's structure. What is undecided is what the distribution's
> collection view should *yield* — for instance, whether `values()` returns per-field
> marginal distributions, conditional distributions, or structural sub-distribution
> views — and how a subclass that requires a flat support enforces that through its
> `EventTemplate`. `RecordDistribution` is the basis for many joint distributions, so
> this should be designed deliberately as part of the distribution chapter of #235.

> **➡️ Deferred to a future #235 phase — the batch-axis operators on a batch collection.**
> This is **out of scope for #326** and recorded here so a future implementer sees it.
> The two-axis framing and what #326 ships now are written up under "Transitional state
> of the batch / distribution surface" above; #326 brings the *field-navigation*
> surface (`at_path` / `children` / `is_field` / leaf-only string-`[]`) to
> `RecordArray` / `NumericRecordArray`, leaving only the **batch axis** for this phase.
> What remains genuinely open: with indexing dual (a **string path selects a field**,
> an **integer selects a batch element**), `len` / `__iter__` should range over the
> batch axis (`batch_shape[0]`, numpy-aligned), but the meaning of `keys` / `values` /
> `items` / `in` on a batch is undecided — making them range over fields would break
> the clean `in ⇔ keys ⇔ iter` coincidence that the single `Record` enjoys (because
> `iter` is the batch axis there). Resolving this — likely by exposing the field
> collection through a named view (e.g. a `fields` mapping) rather than the bare
> mapping protocol — is the substantive task, to be coordinated with the
> `*Array → *Batch` redesign (where `RecordBatch` takes the field surface from
> `_NamedTree` and the batch axis from the `Batch` ABC, and no longer inherits
> `Record`).

---

## 7. Test plan

The implementing PR should add tests covering at least:

- **Conditional round-trip.** `type(r)(dict(r), event_template=r.event_template)
  == r` for flat, nested, mixed (numeric + opaque), and fully numeric records (and
  the analogue for `EventTemplate`). **And the negative:** for a record whose
  template carries `dtype`/`support`/`meta`/`Distribution`/`Function`, or for a
  `NumericRecord`, `Record(dict(r)) != r` (value-only is lossy) — so the lossiness
  is pinned, not accidentally "fixed" later.
- **Flat coincidence (regression guard).** For a flat collection, the new
  `keys`/`values`/`items` produce exactly the #317 top-level results, proving flat
  callers are unaffected.
- **Construction unflattening and every §5 rule**, with the error cases written
  first: the field-xor-prefix collision (all three forms in §5.1), malformed keys
  (empty/leading/trailing/doubled `/`, non-`str` key), keyword keys with `/`
  rejected, the canonical order after unflattening (the `physics`/`observation`
  example), and a plain `dict` value kept as an opaque `Record` leaf (not exploded).
- **Subtree template consistency.** `record.at_path(p).event_template ==
  record.event_template.at_path(p)` for every partial path `p`, in both the
  supplied-template and inferred-template cases — so subtrees carry their authoritative
  slice and nothing re-infers a lossy one.
- **Convenience constructors.** `from_nested_dict(r.to_nested_dict())` reproduces the
  structure (conditional on schema + class, as in §5.2); `from_nested_dict` with no
  template reads every `dict` as a subtree, while `from_nested_dict(d,
  event_template=tpl)` keeps a `dict` at a template-leaf position as opaque data.
- **Boundary rules (§3.0).** `len` / iteration / `keys` range over leaf fields;
  `in` / `[]` / iteration coincide (a key is a member and indexes; a **partial path
  is neither** a member nor indexable — `"physics" in obj` is `False` and
  `obj["physics"]` raises `KeyError`); `obj.at_path(partial_path)` returns the
  subtree and `obj.at_path(key)` equals `obj[key]`; `obj.is_field(path)` is `True`
  only for keys (equal to `path in obj`); `obj.children` is the one-level mapping
  (its `.values()` include subtrees).
- **The two migrated sites (regression guard).** `_record_distribution.py:196` (now
  `key not in template.children`, where `template` is an `EventTemplate` and so has
  `children`) and `test_product.py:1205` (now reaching the interior node via
  `at_path` on a *batch* sample) behave correctly.
- **Batch field-navigation surface (§6, "Transitional state").** On a `RecordArray` /
  `NumericRecordArray`: `at_path(partial_path)` returns the subtree and
  `at_path(key)` the batched leaf; string-keyed `[]` is leaf-only (`arr["physics"]`
  raises, `arr.at_path("physics")` does not); `children` / `is_field` match a single
  `Record`. The batch-axis operators are unchanged this phase: integer `[]` still
  selects a batch element, and `len` / `__iter__` / `keys` keep their current
  top-level behavior (the deferred #235 work), so this test also pins the residual
  inconsistency rather than letting it drift.
- **`to_nested_dict` round-trip.** `to_nested_dict()` mirrors the storage tree and
  is distinct from the flat `dict(obj)`; both shapes asserted on a nested case.
- **Field-path agreement (carried from #317).**
  `tuple(record.keys()) == tuple(record.event_template.keys())` for nested and
  cross-type cases.
- **Shared-primitive equivalence.** `map`, the (de)serialization helpers, and
  equality produce results identical to the pre-refactor implementations, so the
  consolidation onto `_walk_leaves` / `_rebuild_from_leaves` is behavior-preserving.
- **`map` / `map_with_keys` structure preservation (§3.0).** A nested case: the
  result has the same names and nesting as the input; `f` returning a plain `dict`
  yields an opaque leaf (no re-nesting); `f` returning a node-type instance (a
  `Record` / `EventTemplate`) raises `ValueError` naming the key; `*args` / `**kwargs`
  reach `f` unchanged; `map_with_keys` passes the full key path. For `Record`, leaf
  specs are re-inferred (an identity `map` need not compare equal when the template
  carried `dtype` / `support`), and `NumericRecord.map` returning a non-numeric leaf
  raises. For `EventTemplate`, a non-spec-coercible `f` output raises `TypeError`, and
  mapping every spec to a numeric one auto-promotes to `NumericEventTemplate`.
- **Value updates and template threading (§3.1).** `without` and `merge` thread the
  template with no inference: a nested `without` keeps the surviving fields' exact
  specs (`dtype` / `support` / `meta` preserved), and `merge` preserves both inputs'
  specs; `replace` keeps the untouched fields' specs and re-infers only the replaced
  paths (and takes a replacing `Record`'s authoritative template). Error cases:
  `without`/`replace` on a missing path raise `KeyError`; `without` of every field
  raises `ValueError`; `merge` raises `ValueError` on a duplicate leaf key or a
  field-versus-prefix clash; `replace` of a non-existent path raises `KeyError`.
  Deep-`merge` of `{"physics/mass"}` with `{"physics/force"}` yields the combined
  `physics` group. The same edits on an `EventTemplate` produce the edited specs
  directly.

---

## Open questions (index) — resolved, with two items deferred to #235

Each is described in full at the cited section. *(Resolved during review and now
folded into the design above: the membership rule (leaf-only `in` / `[]`, with the
separate `at_path` navigator and the key-vs-path distinction of §2), the top-level
accessor (`children`), the `to_dict`/flat-dict split (`to_nested_dict` vs.
`dict(obj)`), the round-trip semantics (conditional on schema + class), **the
whole construction contract (§5)** — the shared `_unflatten` step, the field-xor-prefix
rule, canonical order, malformed-key errors, string-only construction keys,
positional-only path keys, subtrees carrying their authoritative template slice, and
the `from_nested_dict` convenience constructor — and **the list helpers (§2.1, §3.2)**:
the two `to_leaf_list` helpers are removed and the reconstruction is renamed to
`EventTemplate.from_field_values`, the template-driven inverse that `from_vector`
builds on; **`map` / `map_with_keys` (§3.0, §3.1)** — promoted to `_NamedTree`,
structure-preserving with a node-type-return guard, `(f, /, *args, **kwargs)` with
constant forwarding, and per-leaf spec re-inference (so there is nothing to thread
beyond the fixed structure); **the value updates (§3.1)** — `replace` / `merge` /
`without` are path-aware structural edits on `_NamedTree` that thread the template by
applying the same edit to `event_template` (`without` / `merge` need no inference,
`replace` re-infers only the overwritten paths), while `select` is reclassified as a
value-only `dict` projection outside the template question; and **the
`_predictive_check.py:88` type (§6)** — `len(observed_data)` is a leading-axis
observation count, not `Record` mapping semantics, so the leaf-keyed change does not
affect it.)*

**All of #326's own open questions are now resolved.** Two collection-view questions
surfaced during this work are **out of scope for #326** and deferred to future #235
phases; they are stated in full in the deferred boxes in §6 and recorded in the
Bookkeeping note so they are carried into #235:

1. **`RecordDistribution`'s collection view over its support** (§6) — what
   `values()` etc. yield, and flat-support restriction; orthogonal to dependence.
   Belongs to the distribution chapter of #235. (Eased by `children`:
   subtrees/sub-distributions are now enumerable.)
2. **Batch types' dual indexing** (§6) — string path → field, integer → batch
   element; coordinate with the `*Array → *Batch` redesign.

arob5:
## Summary

Implements the [#326](https://github.com/TARPS-group/prob-pipe/issues/326) redesign: `Record` / `EventTemplate` are now a **named, ordered collection of fields** whose mapping protocol (`keys` / `values` / `items` / iteration / `len` / `in` / `[]`) is **leaf-keyed by `/`-path**. The tree structure is a secondary, clearly-labelled surface — `children` (one-level view) and `at_path` (navigate to a leaf **or** an interior subtree). `Record` / `EventTemplate` and the batch types share a `_NamedTree` mixin. Breaking, pre-stable-API.

Closes [#326](https://github.com/TARPS-group/prob-pipe/issues/326). (Depends on #317, landed.)

## Relationship to the consolidated design reference ([#332](https://github.com/TARPS-group/prob-pipe/pull/332))

The consolidated design reference landed after this PR was opened and is now the **source of truth** for ProbPipe's contracts. This PR is the first large step toward its **Part II.1** (`NamedTree`) and **Part III.1 / III.3** (`EventTemplate`, `Record`, `NumericRecord`): it establishes exactly the leaf-keyed collection surface the reference specifies (fields keyed by path, `children` / `at_path` navigation, canonical order, path-keyed construction).

It intentionally stops short of full alignment in a few places. This is deliberate — the remaining work is a multi-PR refactor, and doing it here would balloon the diff and entangle unrelated concerns (identity, specs, the distribution layer). Each remaining delta is now a scheduled follow-up:

| Delta from the design reference | Where it lands |
|---|---|
| `_NamedTree` is private → `NamedTree` (public); `is_leaf` removed in favor of `is_field`; `is_multi_field` moved onto the tree; add `with_names` and `_leaf_type`; `Record` auto-promotes to `NumericRecord` | **#338** (NamedTree/Record alignment) |
| Identity is ad-hoc (`name` / `source`) → the `Tracked` (`name` + `provenance` + `name_is_auto`) and `Annotated` (`annotations`) mixins | **#336** (Tracked/Annotated) |
| Leaf specs have no common base → a `ValueSpec` base with `is_valid`; `DistributionSpec.inner_template → event_template` | **#337** (ValueSpec) |
| `fields` / `to_dict` temporary aliases; distribution-layer `fields → children`; `RecordDistribution` removal | [#328](https://github.com/TARPS-group/prob-pipe/issues/328) |
| `RecordArray` / `NumericRecordArray` → `RecordBatch` / `NumericRecordBatch`; batch-axis operator semantics | [#235](https://github.com/TARPS-group/prob-pipe/issues/235) Chapter 2 |

The docstrings in this PR are written to the design reference's vocabulary (field = named leaf, key vs. path, canonical order) so they stay accurate as the follow-ups land.

## What changes

**Leaf-keyed mapping.** `record["x"]` reaches a leaf and raises on an interior node — use `record.at_path("x")` for a leaf or subtree, and `record.children` for the one-level view. `keys()` / `values()` / `items()` / iteration enumerate every leaf by full path in canonical first-appearance order; `in` and `[]` are leaf-only and coincide.

**Construction.** Path-keyed unflatten (`/`-string keys, field-xor-prefix collision rule, malformed-key errors, canonical first-appearance order), template carry-down to subtrees, and `from_nested_dict`.

**Transforms & edits.** `map` / `map_with_keys` (structure-preserving, per-leaf spec re-inference) and `replace` / `merge` / `without` (template-threaded) live on `_NamedTree`; `select` stays a projection.

**Removed / renamed API (breaking):**
- `EventTemplate.leaf_paths` → `keys()`
- `to_leaf_list` → `list(record.values())`
- `from_leaf_list` → `from_field_values`
- `map_with_names` → `map_with_keys`
- internal storage `_fields` / `_specs` → `_tree` (dropped `_field_map`)

**Kept as temporary aliases:** `Record.fields` (→ `children`) and `Record.to_dict` (→ `to_nested_dict`). These names are pervasive across the distribution/model surface (11 classes define their own `.fields`; `to_dict` is overloaded with `Provenance` / pandas / views), so renaming them belongs with the distribution value-model work in [#328](https://github.com/TARPS-group/prob-pipe/issues/328).

**Batch types.** `RecordArray` / `NumericRecordArray` gain the field-navigation surface (`at_path` / `children` / `is_field`, leaf-only string `[]`). Their batch-axis mapping (`len` / iteration / integer `[]`, and the meaning of `keys` / `values` / `items` / `in`) stays top-level for now, and `replace` / `merge` / `without` raise `NotImplementedError` on a batch — reconciled with the batch-axis rework in [#235](https://github.com/TARPS-group/prob-pipe/issues/235) Chapter 2.

**Nested-aware diagnostics & broadcasting.** The ArviZ / xarray export (`extract_draws`, `to_named_posterior_dataset`), internal log-likelihood reconstruction (`_loo`), and vmapped-output stacking now export and rebuild leaves by full path, so a *nested* posterior / sample round-trips. (Previously these assumed a flat top-level structure.)

## Docs & changelog

User Guide notebook 02 (and the nested-access cells in notebook 04) are updated to the leaf-keyed idiom; STYLE_GUIDE §1.10 and a `CHANGELOG` `[Unreleased]` entry describe the new surface and the removals. Module- and class-level docstrings for `Record`, `EventTemplate`, and `NumericRecord` have been rewritten to the design reference's vocabulary.

## Follow-ups (draft PRs opened to sequence the refactor)

- **[#336](https://github.com/TARPS-group/prob-pipe/pull/336)** — `Tracked` / `Annotated` identity & metadata mixins (design II.2).
- **[#337](https://github.com/TARPS-group/prob-pipe/pull/337)** — unify leaf specs under a `ValueSpec` base with `is_valid` (design III.1).
- **[#338](https://github.com/TARPS-group/prob-pipe/pull/338)** — align `NamedTree` / `EventTemplate` / `Record` / `NumericRecord` with the reference (II.1, III.1/III.3); depends on #336 and #337.

Sequencing: this PR → #336 and #337 (independent of each other) → #338 (needs both). The distribution-layer work ([#328](https://github.com/TARPS-group/prob-pipe/issues/328)) and the batch-axis rework ([#235](https://github.com/TARPS-group/prob-pipe/issues/235) Chapter 2) follow.

## Known issues carried forward

- **`ArraySpec` eq/hash quirk:** `ArraySpec((), dtype=None) == ArraySpec((), dtype=float64)` is `True` (numpy treats `None` as the default float dtype) yet the two hash differently. This PR no longer relies on the quirk (subtree re-templating is unconditional); the underlying eq/hash contract is flagged for the `ValueSpec` work (#337).
- **JAX pytree aux:** `_record_unflatten` still re-infers the template (a schema/class-lossy round-trip), as in #317 — revisit whether the template should live in the aux.


jhuggins:
@arob5 — heads-up: I've pushed three commits addressing the consolidated review findings on this PR (`009022b` fix, `651e596` tests, `318d389` docs). Summary below so you can re-review the deltas rather than the whole diff.

> 🤖 **Fixes authored by Claude** (Anthropic, via Claude Code), reviewed and pushed by @jhuggins. Every regression below was reproduced before fixing and now has a pinned test. Verified: ~1,270 targeted tests pass (core records/templates, product, broadcast, transition, empirical, GLM, sweep, diagnostics, inference), `ruff format` + `ruff check` clean repo-wide, notebooks re-executed.

## Regressions fixed (all reproduced vs `main` first)

**1. Leaf-only `template[name]` crashes on nested inputs.** Two callers still indexed an interior subtree: `_sample_nested` (depth-2 batched `ProductDistribution` sampling raised `KeyError: 'sub' is a subtree…`) and `_RecordArrayView.__init__` (`view` / `select` / `select_all` on any nested batch — `select_all` is the workflow-sweep splat path). Both now use `template.children[name]`, matching the sites you had already migrated. Tests: depth-2 batched product sampling; `view`/`select_all` on a nested batch.

**2. Structural edits rebuilt through `type(self)` — two failure modes.**
- *Template side:* `Record(x=1.0, y=2.0).replace(x="hello")` raised `TypeError` (the auto-promoted `NumericEventTemplate` rejected the opaque spec), contradicting `replace`'s own docstring. Template edits now rebuild through a `_rebuild_class()` hook that returns the base `EventTemplate`, so `__new__` re-decides the promotion in both directions (demote on opaque, re-promote when the last opaque leaf goes).
- *Value side:* the flatten-and-rebuild strategy demoted untouched nested `NumericRecord` children to the outer class (breaking subtree equality, `.to_vector`, and template↔value class agreement). I first tried dispatching the child class off the sub-template's numericness, but that provably conflicts with your pinned round-trip law `type(r)(dict(r), event_template=r.event_template) == r` — a plain `Record` may legitimately hold an all-numeric subtree, so the template cannot determine the value class. `Record.without/merge/replace` are instead reworked as **recursive child surgery**: untouched children are reused *verbatim* (class, `name`, provenance, backend aux all preserved — they are the same objects), only touched subtrees are rebuilt recursively, and the edited `event_template` is assembled from the children's own templates so the lock-step invariant holds by construction. All 449 pre-existing collection/record/template tests still pass, including the round-trip law.

**3. Nested-batch round-trips compared unequal.** `RecordArray.__eq__` fell back to identity for children `jnp.array_equal` can't handle, so `tpl.from_vector(ra.to_vector()) == ra` and pickle round-trips were `False` for nested batches — violating `from_vector`'s documented law. It now recurses into `Record`/`RecordArray` children.

## Hardening

- **Batch guards completed:** `map` / `map_with_keys` / `from_nested_dict` on a batch now raise the same clean `NotImplementedError` as `replace`/`merge`/`without` instead of a cryptic constructor `TypeError`.
- **Overlapping `replace` paths raise** (`{"a": …, "a/b": …}` silently dropped the descendant update, order-dependently) — on both `Record` and `EventTemplate`, either listing order.
- **Metadata preservation:** path-keyed `NumericRecord({"a/b": xr.DataArray(…)})` now materialises nested children *before* leaf coercion, so backend aux is captured by the owning record and `to_native()` restores the `DataArray` (it was orphaned under the `/`-key). Supplying a template whose subtree **is** the child's own template object stores the child verbatim (identity check — deliberately not `==`, which the `ArraySpec` dtype quirk (#337) makes unsafe).
- **Caller sweep:** the remaining `for f in x.fields: x[f]` sites (transition resampling, broadcast marginals, `_empirical` helpers/accessors, GLM coercion) now iterate leaf-keyed, so nested values flow through — or, where nested reconstruction isn't defined yet (broadcast batch-stacking), raise a clear `NotImplementedError` mirroring your `_make_stack` guards.
- **API/docs accuracy:** dropped the redundant `is_leaf` (shipped alongside `is_field` with no legacy callers; kept `is_field`); `keys()`/`values()`/`items()` return materialised tuples per the dict-like contract; `children` no longer copies the storage dict per access; removed the `Tracked`/`Annotated`/`annotations` claims (that's #336's surface — docstrings now describe `name`/`source`); fixed `to_vector`/`from_vector` crossrefs to `NumericEventTemplate` (mkdocstrings links); CONTRACTS.md's vocabulary example no longer mandates the removed `to_leaf_list`/`leaf_paths` API; STYLE_GUIDE §1.10 scopes the leaf-keyed protocol to `Record`/`EventTemplate` and documents *both* exceptions (batch types → #235, distributions → #328).

## Tests (the suite was green partly because it missed the buggy paths)

- Fixed the **unfalsifiable assertion** in the nested-export test (`shape[-1] == 30 or ndim >= 1` — the right arm was always true, and the actual shape is `(60,)`): all three export tests now check **exact leaf↔column values against the seeded chains**, and the `_loo` fallback test checks against an independent NumPy baseline (`atol=1e-5`; observed deviation ~7e-7 vs float32 storage) — a swapped intercept↔slope mapping now fails.
- Added **flat-posterior compat tests** for `extract_draws` / `to_named_posterior_dataset` / `_add_log_likelihood` on the `event_template` branch every real posterior takes — pinning that flat posteriors keep their un-prefixed ArviZ variable names.
- Unrolled `test_edits_unsupported_on_batch` (the `op is not nra.merge` bound-method comparison was always true, leaving a dead branch and zero-arg calls that passed only because the overrides take `*args`).
- Renamed the issue-suffixed files to behavior names (`test_named_collection.py`, `test_nested_diagnostics_export.py`) and stripped issue-number narration from comments, per repo convention.
- New regression tests for every fix above, plus: reversed-order field-vs-prefix collision (a distinct `_unflatten_paths` branch), nested pickle round-trips, `map` subclass preservation, deep-merge regrouping order, batch `children`/`is_field`, and the **transitional batch split** (`keys`/`len`/`in` top-level vs leaf-only `[]`) pinned in both directions so #235 inherits an explicit spec.
- Notebooks 02/04: teaching cells use `keys()` instead of the deprecated `fields` alias, and both are re-executed so demo cells render output on the docs site (`mkdocs-jupyter` uses stored outputs).

## Left for the scheduled follow-ups (flagged, not fixed here)

- Batch `keys`/`len`/`in` staying top-level while `[]` is leaf-keyed — documented + pinned; the real reconciliation belongs to the #235 batch-axis rework.
- The two coexisting nested-batch storage forms (pre-canonical `NumericRecord` child vs canonical `NumericRecordArray`, distinct treedefs) — now documented on `NumericRecordArray` with the canonical form named; normalisation left to #235.
- `ArraySpec` eq/hash dtype quirk → #337, as you noted in the PR body.
- netCDF rejecting `/` in variable names for nested exports — documented on `to_named_posterior_dataset` and `_leaf_keys`; a separator decision can wait until the surface stabilises.

Happy to adjust anything — the three commits are independent, so individual pieces are easy to drop or amend.

jhuggins (approved):

jhuggins (approved):

Issue #333, opened 2026-07-01
Standardized cross-method inference validation harness (canonical models + exact/approximate contract)


jhuggins:
## Summary

`probpipe.validation` (shipped in #301) gives us a method-agnostic metric engine — `Reference`, `score_posterior`, `simulation_based_calibration`, `interval_coverage` — but **only `blackjax_nuts` and `blackjax_sgld` actually use it** (via `tests/inference/test_validation_suite.py`). The other ~11 registered methods each hand-roll a bespoke correctness check, and the same reference models are re-implemented across many files.

This issue proposes a **single canonical validation suite** that every inference method runs through, so that (i) one shared set of models + reference posteriors validates everything, (ii) adding a new method carries a clear, mechanical validation obligation, and (iii) the suite applies the right standard to **exact** methods (must reproduce a trusted reference) versus **approximate** ones (SGLD, VI, ABC, amortized SBI — validated by calibration + looser bounds, not exact reproduction).

Every test here is **fast**: closed-form references where they exist, a *saved* long-NUTS draw set otherwise. The genuinely slow tier (big data, BNNs, high dimension, full SBC sweeps) stays in the separate `probpipe-benchmark` repo (#302).

## Motivation — current state

A review of every inference-method correctness test found:

- **Six different "correctness" idioms**, one per author: inline analytic-conjugate recovery (NUTS `tests/inference/test_blackjax_mcmc.py:109`, ESS `test_blackjax_ess.py:413`, nutpie `test_nutpie.py:275`, pyabc `test_pyabc.py:100`), data-generating-parameter recovery (SGLD/SGHMC `test_blackjax_sgmcmc.py:246`), calibration bands (all BayesFlow), cross-method-vs-another-fit (`test_bayesflow_likelihoods.py:357`), component-vs-scipy (`test_glm.py:51`), and pure smoke (`cmdstan` `test_cmdstan_method.py:42`, `pymc_advi` `test_pymc_model.py:381`, tfp). "Correct" means something different in each file.
[...]

PR #334, opened 2026-07-01
ci: per-file test selection for the inference package

jhuggins:
## What

Refine the change-based CI test-selection so a single inference-**method** source change runs only its own test file plus the registry/dispatch integration tests, instead of the whole `tests/inference/` folder.

First step of the standardized validation harness (#333). As `tests/inference/` grows to cover many heavy, independent optional backends, folder-granular selection reruns far more than necessary on a one-backend change.

## The problem

Change-based selection maps a changed source module to its **test folder** by subpackage (`probpipe/inference/X.py → tests/inference/`). The reverse-dependency import graph expands a leaf method module to `{itself, inference/__init__.py}` — and that `inference/__init__.py` registration facade is pulled in by *every* method change (it imports each method to register it), so it always mapped back to the whole folder, defeating any per-method granularity.

```
$ python3 scripts/ci/import_graph.py expand probpipe/inference/_blackjax_sgmcmc.py
probpipe/inference/__init__.py      # ← pulled in by the registration facade
probpipe/inference/_blackjax_sgmcmc.py
```

## The fix

Map the affected-but-not-directly-changed `inference/__init__.py` to the **registry/dispatch integration tests** (`test_inference.py`, `test_inference_registry.py`) rather than the whole folder, and map each method module to its `test_<name>.py` by filename convention. Shared-core inference modules without a matching test file (`_registry.py`, `_approximate_distribution.py`, …) still fall back to the whole folder — correct, since they affect every method. Non-inference subpackages keep the existing folder-granular behavior.

| Changed source | Before | After |
|---|---|---|
| `probpipe/inference/_blackjax_sgmcmc.py` | whole `tests/inference/` | `test_blackjax_sgmcmc.py` + `test_inference.py` + `test_inference_registry.py` |
| `probpipe/inference/_pyabc.py` | whole `tests/inference/` | `test_pyabc.py` + the two integration tests |
| `probpipe/inference/_registry.py` (shared core) | whole `tests/inference/` | whole `tests/inference/` (unchanged) |
| `probpipe/inference/__init__.py` (direct edit to the facade) | whole `tests/inference/` | whole `tests/inference/` (a facade edit can make/break any method's registration) |
| `probpipe/core/*.py` | `tests/core/` | `tests/core/` (unchanged) |

## How

The source→test mapping moves out of inline `ci.yml` bash into a unit-tested `resolve_test_targets()` in `scripts/ci/import_graph.py` (new `test-targets` CLI subcommand) — that module is already "the single source of truth for the logic ci.yml uses." `ci.yml` now calls it in one line; the changed-test-file passthrough, dedup, foundational run-all triggers, and push-to-`main` full-run backstop are all unchanged.

## Testing

- **9 new unit tests** in `tests/ci/test_import_graph.py` (`TestResolveTestTargets`: method→own-test+integration, direct-`__init__`→folder, shared-core→folder, no-matching-test→folder, nested→folder, root→root-tests, re-export→root-trigger, other-subpackage→folder, CLI) — full file: **17 passed**.
- `ruff check` + `ruff format --check` clean; `ci.yml` validated as parseable YAML.
- Sanity-checked the resolver against the real repo tree (outputs match the table above).

## Scope / follow-up

This is the **test-selection** half of the harness plan's CI step. The **dependency-isolation** half — extending the existing `bayesflow`/`stan` gated-leg pattern to `pymc`/`nutpie`/`pyabc` so the main job stops installing them — is a separate follow-up PR: it restructures jobs + coverage flags and has a `nutpie`×`pymc`×`stan` install entanglement (nutpie compiles pymc *or* stan models) that warrants its own focused review.

Because selection only ever *narrows* what runs on a PR (never what installs), and the push-to-`main` full run remains the backstop, the blast radius here is contained.


scripts/ci/import_graph.py
```diff
@@ -242,6 +247,86 @@ def _cmd_expand(args: argparse.Namespace) -> int:
...
+    (``skip_init=False``, as the test job requires), then each affected module
+    is mapped to a target: a root-level ``probpipe/X.py`` selects the root
+    ``tests/test_*.py`` files; ``probpipe/inference/...`` uses the
+    filename-convention granularity of :func:`_inference_test_targets`; any
+    other ``probpipe/<subpkg>/...`` selects the ``tests/<subpkg>/`` folder. A
+    folder target supersedes individual file targets beneath it. Returns a
+    sorted list.
+    """
+    rdeps = reverse_dependency_map()
+    changed_sources = [f for f in changed_files if f.startswith(PACKAGE)]
```
SoymilkWinsAgain, line 309: Modifying `scripts/ci/import_graph.py` alone will not trigger any tests or notebook legs.

`resolve_test_targets()` only processes probpipe* sources, and `scripts/ci/import_graph.py` is not included in the workflow's foundational list. It is recommended to add `scripts/ci/import_graph.py` to at least the `RUN_ALL` foundational set of the test job, as it also controls notebook selection; ideally, it should also be added to the `RUN_ALL_NB` set of the notebook job.

scripts/ci/import_graph.py
```diff
@@ -242,6 +247,86 @@ def _cmd_expand(args: argparse.Namespace) -> int:
...
+    * ``__init__.py`` -- the registration facade -- maps to the registry/dispatch
+      integration tests (:data:`_INFERENCE_INTEGRATION_TESTS`).
+    * A flat per-method module ``_<name>.py`` maps to
+      ``tests/inference/test_<name>.py`` by filename convention.
+    * Anything else -- a module with no matching ``test_<name>.py`` (shared core
+      such as ``_registry.py``), or a nested subpackage module (``sub/_foo.py``,
+      which has no flat filename convention) -- falls back to the whole
+      ``tests/inference/`` folder.
+    """
+    if rel_path == "__init__.py":
```
SoymilkWinsAgain, line 283: Directly modifying `probpipe/inference/__init__.py` now only runs `test_inference.py` and `test_inference_registry.py`, which is too narrow compared to the original `tests/inference/`.

The PR body suggests mapping "affected-but-not-directly-changed inference/__init__.py" to `registry/dispatch` integration tests. However, the implementation doesn't distinguish between direct seeds and transitive affected modules. Directly modifying the registration facade might delete or corrupt the registration of an optional method. It's recommended that `direct probpipe/inference/__init__.py` still be mapped to `tests/inference`, and that two integration tests be used only for `__init__` methods derived from method module back dependencies, along with corresponding regression tests.

SoymilkWinsAgain (changes requested):
Also, the PR body still says "...13 tests passed", but there are in fact 16 tests.

scripts/ci/import_graph.py
jhuggins: Done in 289a337. Added `scripts/ci/import_graph.py` to the test job's `RUN_ALL` foundational set and the notebooks job's `RUN_ALL_NB` set, so a change to the resolver now forces the full suite — which includes its own `tests/ci/test_import_graph.py`.

I left the `bayesflow`/`stan` legs out on purpose: they gate on their own file globs and never call the resolver, so a change to `import_graph.py` can't affect their selection.

scripts/ci/import_graph.py
jhuggins: Good catch — fixed in 289a337. `resolve_test_targets` now computes the seed set (the directly-changed modules) and passes `is_direct` into `_inference_test_targets`:

- a **directly-changed** `probpipe/inference/__init__.py` maps to the whole `tests/inference/` folder (a facade edit can make or break any method's registration, including optional backends);
- only the **transitively-affected** `__init__` (pulled in as a reverse-dependency of a method change) narrows to the two integration tests — so per-method granularity is preserved.

Added a regression test, `test_direct_init_change_maps_to_whole_folder`, alongside the existing transitive-`__init__` case.

jhuggins:
Thanks for the review, @SoymilkWinsAgain — all three points addressed in 289a337 (see the inline replies):

1. **CI trigger** — `scripts/ci/import_graph.py` added to the `RUN_ALL` (test) and `RUN_ALL_NB` (notebooks) foundational sets, so a change to the selection logic now runs the full suite.
2. **Direct vs transitive `inference/__init__.py`** — a direct facade edit now maps to the whole `tests/inference/` folder; only the transitive edge (a method change) narrows to the integration tests. Regression test added.
3. **Test count** — corrected the PR body (13 → 17) and added a direct-`__init__` row to the behavior table.

CI is re-running on the new head. Re-requesting review.

SoymilkWinsAgain (approved):
I updated the branch, everything looks fine

PR #332, opened 2026-07-01
docs(design): add consolidated design reference

jhuggins:
## Summary

Adds a new top-level `design/` directory holding the **ProbPipe Consolidated Design Reference** — a target-state design document that coordinates the active threads reshaping the core value, distribution, and collection abstractions. It fixes the shared design *principles* and the precise contracts for the core abstractions, workflow functions, and operations, so future work has one reference to build from and less re-litigation.

Split into a `README.md` (overview + conventions) and one file per part:

- `01-design-principles.md` — Part I: the core (C-series) and derived (D-series) principles.
- `02-shared-abstractions.md` — Part II: `NamedTree`, identity/metadata, `Batch`, and the generic dispatch registries.
- `03-values-and-distributions.md` — Part III: `EventTemplate`, `Record`, `LinOp`, `Distribution`, capabilities, `ConditionalDistribution`, batches, factored distributions, composition, the hierarchy, plus the conversion and constraint-reparameterization registries.
- `04-workflow-functions.md` — Part IV: how an ordinary function is lifted (broadcast, sweep, dispatch, orchestration, provenance).
- `05-operations.md` — Part V: `sample`, `log_prob`, moments, `condition_on`, composition, `pushforward`, `marginal`/`factor`, and batched operations.
- `06-distribution-catalog.md` — Part VI: the concrete families — parametric, empirical, mixtures, pushforward results, the Gaussian algebra, random functions/measures, approximate, and the conditional families (including GLM likelihoods).

A seventh part — a higher-level agentic interface — is noted as planned.

`design/` is a **contributor-facing architecture reference** and is deliberately kept out of the mkdocs user site. It is intended as the durable home for the design contracts; the transient `CONTRACTS.md` (the #235 discipline) can point here and be retired when #235 closes.

## Linked issue(s)

Refs #235, #255, #284, #328 (value / distribution / composition model), #184, #309 (multiple dispatch), and #281 (agentic interface). This PR consolidates their design intent and does not close them.

## Test plan

Documentation only; no code changes. The files are standard GitHub-flavored Markdown and render on GitHub. `design/` is intentionally outside the mkdocs site, so `mkdocs serve` is unaffected.

## Breaking changes

None.

## Documentation

- [x] N/A — no public API changed, so no docstrings to update
- [x] N/A — `design/` is a contributor reference, not the User Guide / tutorials
- [x] N/A — no CHANGELOG entry (a design document, not a released change)

## Checklist

[...]

PR #335, opened 2026-07-04
fix(core): harden container edge cases

SoymilkWinsAgain:
## Summary

Fixes two core container bugs with a deliberately small diff:

- `DistributionArray` integer indexing now follows NumPy/JAX-style bounds semantics: valid negative indices wrap per axis, while positive overflow and excessive negative indices raise `IndexError` with axis/size context. A 0-d `DistributionArray` accepts `da[()]` and rejects axis indexing.
- `NumericRecordArray.mean()` and `.var()` now recurse through nested numeric record fields instead of passing nested record containers directly to JAX reductions. This covers both canonical nested `NumericRecordArray` fields and accepted plain `Record` nested fields whose leaves carry the batch shape.

## Root cause

The `DistributionArray.__getitem__` integer fast path used modulo-style wrapping, so out-of-bounds integer keys could silently select a valid component. `NumericRecordArray._reduce()` treated nested record containers as ordinary array leaves, which broke reductions for nested numeric records.

## User impact

Indexing errors are now surfaced instead of silently wrapping to the wrong distribution component, and nested numeric record batches can be reduced structurally with `mean()` / `var()`.

## Linked issue(s)

N/A — small standalone bug fix.

## Test plan

- `uv run --frozen pytest -s -p no:xdist -o "addopts=" tests/core/test_distribution_array.py tests/core/test_record_array.py tests/core/test_event_template.py -x -v` — 278 passed.
- `uv run --frozen ruff check probpipe/core/_distribution_array.py probpipe/core/_record_array.py tests/core/test_distribution_array.py tests/core/test_record_array.py tests/core/test_event_template.py` — passed.
- `uv run --frozen ruff format --check probpipe/core/_distribution_array.py probpipe/core/_record_array.py tests/core/test_distribution_array.py tests/core/test_record_array.py tests/core/test_event_template.py` — passed.

## Breaking changes

None.

## Documentation

- [x] CHANGELOG updated for the fixed behavior.

## Checklist

- [x] Linked to an issue above, or marked N/A for a small standalone fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -265,6 +265,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Core container indexing and nested reductions.** `DistributionArray`
+  integer indexing now raises `IndexError` for positive overflow and negatives
+  past the axis bounds, while 0-d arrays accept only empty-tuple indexing.
+  `NumericRecordArray.mean()` and `.var()` now recurse through nested numeric
+  record fields instead of treating nested records as arrays.
+
 - **Linear-algebra and Gaussian-conditioning edge cases on the algebra bug-fix
   branch.** `RootLinOp.diag()` now squares diagonal roots; `CholeskyLinOp`
   keeps lower-root (`L @ L.T`) and upper-root (`U.T @ U`) representations
--- a/probpipe/core/_distribution_array.py
+++ b/probpipe/core/_distribution_array.py
@@ -447,6 +447,12 @@ class DistributionArray[T](Distribution[T]):
         ``shape=batch_shape`` object array.
         """
         bshape = self._batch_shape
+        if not bshape:
+            if key == ():
+                if self._backend is not None:
+                    return self._backend.cell(0)
+                return self._components[0]
+            raise IndexError("too many indices for 0-d DistributionArray")
         # Normalise the key to a tuple, one entry per leading axis.
         if isinstance(key, tuple):
             if len(key) > len(bshape):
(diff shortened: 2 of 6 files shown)


arob5:
Note that the design contracts for container classes are refactored as part of #332

jhuggins:
Thanks @SoymilkWinsAgain — these are correct, well-tested fixes. The wrinkle is that the record-array changes overlap the in-flight records redesign in #327 (`dev/record-named-collection`), which reworks these same containers. To keep #335 from blocking #327, let's **land #327 first and bring the still-needed pieces back as a follow-up on top of it.** Here's the per-change plan (diffed against #327's current head):

**`RecordArray.__eq__` (nested recursion) — drop it.** #327's `__eq__` already recurses on nested records (`if isinstance(a, Record) or isinstance(b, Record): ... a != b`), and a bit more thoroughly — it also compares `fields`/`template` and the child types up front. Redundant once #327 lands.

**`NumericRecordArray._reduce` / `mean` / `var` (nested recursion) — keep, as a follow-up on #327.** This one is *not* covered by #327: its `_reduce` still applies the reduction directly to every field (`{name: fn(self._tree[name], axis) for name in self._tree}`) with no recursion into nested `NumericRecordArray` fields, so multi-field nested `.mean()` / `.var()` is still wrong there — and untested. Please carry this fix and its nested-reduction tests forward, re-applied against #327's `_reduce`. (Equivalently it could go straight into #327 — coordinate with that author.)

**`RecordArray.stack` (nested support) — defer.** #327 deliberately raises `"RecordArray.stack does not yet support nested templates"`, whereas #335 implements it. That's a design decision, not a bug, so let's hold this until we settle whether nested stacking belongs in the redesigned contract. Leave it out of the follow-up for now.

**`DistributionArray.__getitem__` (bounds + 0-d) — independent.** #327 doesn't touch `_distribution_array.py`, so this fix is orthogonal and won't conflict either way. It can ride along in the follow-up or land on its own — no coordination needed.

Two notes for #327 while we're here: (1) the `_reduce` nested-recursion gap above (no nested mean/var test there), and (2) `test_event_template.py::test_nested_multi_axis_batch_shape` still carries the old *"RecordArray.__eq__ is broken for a nested \*Array … Tracked in #235"* workaround even though #327's `__eq__` is fixed — worth confirming that case passes and cleaning up the comment.

Net: after #327 merges, rebase the follow-up on it — keep `_reduce` (+ tests) and the `DistributionArray` fix, drop `__eq__` (redundant), defer `stack` pending the design call.

jhuggins (changes requested):
See my posted comment from earlier

SoymilkWinsAgain:
Pushed a narrowed bugfix-only update for this PR.

Current scope is now intentionally limited to:

- `DistributionArray` integer indexing bounds checks and 0-d empty-tuple indexing.
- `NumericRecordArray.mean()` / `.var()` recursive reductions through nested numeric record fields, including canonical nested `NumericRecordArray` fields and accepted plain `Record` nested fields with batch-shaped leaves.
- Focused regression tests and the matching changelog entry.

The previous nested `RecordArray.stack` and `RecordArray.__eq__` changes are no longer part of this PR diff.

Validation run locally:

- `uv run --frozen pytest -s -p no:xdist -o "addopts=" tests/core/test_distribution_array.py tests/core/test_record_array.py tests/core/test_event_template.py -x -v` — 278 passed.
- `uv run --frozen ruff check probpipe/core/_distribution_array.py probpipe/core/_record_array.py tests/core/test_distribution_array.py tests/core/test_record_array.py tests/core/test_event_template.py` — passed.
- `uv run --frozen ruff format --check probpipe/core/_distribution_array.py probpipe/core/_record_array.py tests/core/test_distribution_array.py tests/core/test_record_array.py tests/core/test_event_template.py` — passed.

jhuggins (approved):

Issue #328, opened 2026-06-26
Distribution value model: event_template on the base, removing RecordDistribution, and factorization as a capability (SupportsFactors)


arob5:
# Distribution value model: `event_template` on the base, removing `RecordDistribution`, and factorization as a capability (`SupportsFactors`)

**Status.** A design proposal, surfaced while implementing #326 (the `Record` / `EventTemplate` leaf-keyed collection redesign) and developed across several design sessions since. It works out the distribution-layer questions that #326 deliberately left out of scope. It is written to be read by a contributor who was not part of the original discussion: the vocabulary is defined before it is used, the recommendations are stated as precise contracts, and the genuinely undecided points are collected as open questions at the end. Nothing here is implemented yet, and nothing is locked — but most of it has been worked through far enough to state as a recommendation rather than a question.

**Relationship to the master plan.** #235 is the master value-model plan and remains the living "final plan." The process is: settle a large question *here* first, then **fold the resolution into #235** and reconcile it with the adjacent proposals. This issue should not be implemented piecemeal ahead of that. It coordinates closely with #284 (the composition algebra `*` → `ProductDistribution` / `ChainDistribution` / `ConditionalProduct` / `ConditionalChain`), #255 (`ConditionalDistribution` — probability kernels), and the deferred batch work (#235 Chapter 2). Those are adjacent and dependent, not absorbed here.

---

## 1. The problem

Today every distribution in ProbPipe is a `RecordDistribution`. `TFPDistribution` — the base for `Normal`, `Gamma`, and the other TFP-backed scalar distributions — inherits from `NumericRecordDistribution`, which inherits from `RecordDistribution`. So even a scalar `Normal` is "Record-shaped": it carries a single-field `EventTemplate`, and one draw is wrapped as a single-field `Record`. Three consequences follow, and this issue argues each is a symptom of conflating things that should be separate:

1. **`event_template` lives on `RecordDistribution`, not on the base `Distribution`.** The base has no notion of the structure of its own draws; the entire named-collection surface (`fields`, item access, `select`, the per-field views) is defined further down the hierarchy. As a result a `fields` / component vocabulary is present on nearly every distribution, including scalar ones where it means nothing.

2. **A distribution whose draw is genuinely multi-valued is the same class as a scalar one.** `ProductDistribution` (an independent joint) is a `NumericRecordDistribution` holding a `dict` of components, distinguished from a `Normal` only by how many entries the `dict` has.

[...]

Issue #112, opened 2026-04-05
Formalize diagnostics: DiagnosticsStore on Distribution


jhuggins:
### Context

`MCMCApproximateDistribution` currently stores an `inference_data: Any | None` field alongside its frozen JAX chain arrays. Every backend extracts JAX arrays into `_chains` *and* produces an `InferenceData` object, creating redundant copies of the posterior. There is no enforced relationship between them.

More broadly, any distribution can carry diagnostics — MCMC chains have acceptance rates and divergences, importance-weighted samples have ESS, variational approximations have ELBO history, and even exact distributions can have statistical diagnostics (e.g., sensitivity to prior hyperparameters, moments, tail behavior). Currently there is no unified place for this metadata.

---

### Proposed change

Add a `DiagnosticsStore` at the `Distribution` level (alongside `_validation_results` and `_source`). It is `None` by default and populated when diagnostics are available. Replace `_inference_data` and `_algorithm` on `MCMCApproximateDistribution` with the shared mechanism.

`az.InferenceData` is assembled on demand from authoritative sources:

```
az.InferenceData ← frozen distribution data (samples, chains, weights)
                 + DiagnosticsStore (sample_stats, algorithm, computed metrics)
```

---

### Frozen/mutable alignment

| Concern | Owner | Mutability | Level |
|---|---|---|---|
| Distribution data | `_samples`, `_chains`, etc. | frozen | subclass |
| Provenance | `_source` | write-once | `Distribution` |
| Diagnostics | `_diagnostics` → `DiagnosticsStore` | see below | `Distribution` |
| Validation results | `_validation_results` | mutable accumulator | `Distribution` |

Within `DiagnosticsStore`:

| Field | Mutability | Examples |
|---|---|---|
| `algorithm` | frozen at construction | `"tfp_nuts"`, `"rwmh"`, `"importance_sampling"`, `None` for exact |
| `sample_stats` | frozen at construction | acceptance rate, step size, divergences, ESS of weights |
| `computed` | mutable accumulator | R-hat, bulk/tail ESS, LOO-CV, MCSE, sensitivity metrics |
| `_backend_trace` | frozen at construction | native PyMC/nutpie `InferenceData`, `CmdStanMCMC` fit |

---

### Which distributions carry diagnostics

Diagnostics can be computational (about the algorithm that produced the distribution) or statistical (about the distribution itself).

| Distribution type | `algorithm` | Diagnostic examples |
|---|---|---|
[...]

PR #336, opened 2026-07-05
feat(core): Tracked & Annotated identity/metadata mixins (design II.2)

arob5:
## Summary

This PR introduces the two identity-and-metadata mixins from **Part II.2** of the [consolidated design reference (#332)](https://github.com/TARPS-group/prob-pipe/pull/332) — `Tracked` (name + provenance) and `Annotated` (annotations) — and migrates the objects that previously each re-implemented naming, provenance, and metadata (`Distribution`, `Record`/`NumericRecord`, and the batch types) onto them.

Per the design reference, `Tracked` and `Annotated` are **mixins that carry state and concrete methods**, not `typing.Protocol`s — a tracked object *is* one. The structural `Supports*` protocols are a separate mechanism and are out of scope here.

## What landed

1. **`probpipe/core/tracked.py`** — the new module holding both mixins, with full contract docstrings (per `CONTRACTS.md`, these are the canonical contract location):
   - `Tracked`: `name`, `name_is_auto`, write-once `provenance` via `with_provenance(...)` (`None` is a no-op so `Provenance.create()` results pass through unguarded), and `with_name(name)` — a shallow copy with `name_is_auto = False` and rename provenance (`operation="with_name"`, parent = the original). The mixin is `__slots__ = ()` and writes through `object.__setattr__`, so it works on the immutable slotted `Record` family and on plain-attribute `Distribution`s alike.
   - `Annotated`: `annotations: Mapping[str, Any] | None`, widening the old `xarray.DataTree | None` slot to a general mapping (a `DataTree` remains a valid value; the `diagnostics` accessors are unchanged).
2. **Migrations.** `Distribution(Tracked, Annotated, ABC)` and `Record(_NamedTree, Tracked, Annotated)` drop their duplicated `name`/`source`/`with_source`/`renamed`/`auxiliary` code. The batch types (`RecordArray`, `NumericRecordArray`, `DistributionArray`) inherit the `Tracked` attributes and methods through their bases; they are contractually `Tracked` only — the `annotations` slot they currently inherit is an interim artifact of subclassing `Record`/`Distribution`, noted in the code, and goes away with the `Batch` rework (#235 Ch. 2).
3. **Renames (hard cut, no aliases):** `source → provenance`, `with_source → with_provenance`, `renamed → with_name`, `auxiliary`/`_auxiliary` → `annotations`/`_annotations`, and `make_posterior(auxiliary=)` → `annotations=`. Applied across `probpipe/`, `tests/`, `example_scripts/`, the API docs, and the user-guide notebooks. The *backend* aux registry (`NumericRecord._aux`, `register_aux`, docs "auxiliary-metadata registry") is unrelated and untouched, as planned.
4. **`name_is_auto` wired everywhere a name is auto-derived:** `Record`/`RecordArray` default names, `DistributionArray`, the broadcast/marginal distributions, empirical/bootstrap families, `RandomMeasure`, `ProductDistribution`, `SequentialJointDistribution`, `JointGaussian`/`JointEmpirical`, `TransformedDistribution`, KDE, and `MinibatchedDistribution`. `Distribution.__init__` gains a `name_is_auto: bool = False` keyword; user-supplied names stay `False`. `Record.__reduce__` round-trips the flag.
5. **`Provenance`/`ParentInfo` reconciliation:** `ParentInfo.obj → parent` and `ParentInfo.source → provenance` (the serialized `to_dict` key follows: `"source"` → `"provenance"`).
6. **One construction-time guarantee.** The non-empty-name check lives on `Tracked`'s own metaclass (`_TrackedMeta`), covering every tracked term — the `Record` family included — and the `Distribution`-only metaclass is gone (`_RecordDistributionMeta` rebases onto `_TrackedMeta` for its `event_template` check). Shared helpers keep the idioms in one place: `auto_name(name, default)` resolves optional names at every auto-name constructor, and `Tracked._restore_identity` is the single post-unpickle identity restore (used by the record, record-array, and product reducers, which all round-trip `name_is_auto` and provenance).
7. **Identity propagation hardened.** `with_name` works on every host, including classes whose `__new__` takes required arguments (`_shallow_copy` allocates with `object.__new__`; the class is already resolved), and it copies the annotations *container* so post-rename annotation writes don't show through on the original. Operations that rebuild an object from a parent keep `name_is_auto` consistent with where the name came from: product conditioning and `DistributionArray` slices mirror the parent's flag; sweep/broadcast stacks, per-cell batched-params names, `FullFactorialDesign`'s derived default, the minibatch draw/log-prob wrappers, and transition resampling mark derived names auto; `RandomMeasure`/`RandomFunction` preserve a caller-supplied flag. Classes that bypass `Distribution.__init__` (view classes, modeling wrappers, `BayesFlowModel`) initialize identity via `_init_tracked`.
8. **Convention docs updated in-PR.** CONTRIBUTING.md design principles 1 and 6 and the key-abstractions table now describe the `Annotated` annotations store and the `Tracked` name guarantee (with a `Tracked`/`Annotated` row); STYLE_GUIDE.md §9.2 is rewritten around `_annotations` and `tracked.py` joins the public core-module list; the provenance docs use the op form (`sample(anc.parent, ...)`).

## Decisions on the open questions

1. **`Provenance.metadata` is kept**, documented in its docstring as scalar/string operation metadata (rename old/new names, algorithm labels, ESS, …). It is widely load-bearing at provenance-assembly sites and orthogonal to object-level `annotations`; folding or dropping it would lose recorded information. Flagged here as a deliberate extension beyond the reference's minimal `Provenance(operation, parents)`.
2. **`ParentInfo.obj → parent` was renamed now**, together with `ParentInfo.source → provenance` — this is the identity/provenance PR, and deferring would leave the vocabulary split across two PRs.
3. **No temporary aliases** — a single hard cut, as the pre-stable default. The migration was mechanical; the full test suite passes without any compat shim.

## Notable side effects (called out for review)

- `BootstrapReplicateDistribution`'s internal source-distribution attribute was also named `_source`, silently colliding with the provenance slot that `Distribution.source` read (and being skipped by content fingerprints, so replicates of *different* source distributions could fingerprint equal). It is renamed `_source_dist` and now participates in fingerprints. Fingerprints for that class change; no test relied on the old digests.
- `with_name` provenance records `operation="with_name"` (was `"renamed"`).
- The fingerprint generic-fallback skip-list is now `{_name, _name_is_auto, _provenance, _annotations, _sampling_cost}` — identity metadata only.

## Out of scope (unchanged from the plan)

- `Record.__init__` reshaping to the required-first-positional `name` contract — the NamedTree/Record alignment PR (needs this PR's `name_is_auto` machinery).
- `fields → children` and `RecordDistribution` removal — #328.
- Batch-axis rework and `*Array → *Batch` — #235 Chapter 2.

- [x] Contract of every touched abstraction was clear before coding (ambiguities resolved & noted — see the three decisions above).
- [x] New/changed contracts fully documented in docstrings (types, orderings, raises, invariants — `tracked.py`, `provenance.py`).
- [x] Docstring openings lead with the API/contract; rationale deferred to Notes sections.
- [x] Code verified to obey the documented contract; `tests/core/test_tracked.py` asserts it (auto-flag semantics, copy semantics, write-once, error cases, pickle round-trips). Full suite: 3667 passed, 54 skipped.
- [x] Docstrings describe the plan's target contract; the batch types' interim `Annotated`-by-inheritance is labeled as an interim implementation detail.
- [x] Variable names consistent with the canonical table (`name`, `provenance`).
- [x] `CONTRACTS.md` index updated (mixin contract location now points at `probpipe/core/tracked.py`).


probpipe/core/tracked.py
```diff
@@ -0,0 +1,357 @@
...
+        lineage.
+        """
+        object.__setattr__(self, "_name_is_auto", bool(name_is_auto))
+        if provenance is not None:
+            object.__setattr__(self, "_provenance", provenance)
+        return self
+
+    # -- copying -------------------------------------------------------------
+
+    def _shallow_copy(self) -> Self:
```
SoymilkWinsAgain, line 292: `Tracked._shallow_copy()` directly calls `cls.__new__(cls)`, therefore `with_name()` cannot handle classes with `__new__` that require parameters. For example, `TransformedDistribution.__new__` requires `base` and `bijector`, and in practice, `transformed.with_name("y")` throws a `TypeError`. Multiple dynamic view classes also have the same risk. This violates the core contract that "all Tracked objects support `with_name`".

probpipe/core/_random_measures.py
```diff
@@ -114,8 +115,9 @@ class RandomMeasure[T](Distribution[Distribution[T]]):
     components like any other.
     """
 
-    def __init__(self, *, name: str | None = None):
-        super().__init__(name=name or type(self).__name__)
+    def __init__(self, *, name: str | None = None, name_is_auto: bool = False):
```
SoymilkWinsAgain, line 117: The `auto_name` function is called again for the already parsed non-empty name, overriding the `True` passed to `MinibatchedDistribution` in `_minibatch.py`.

probpipe/distributions/_product.py
SoymilkWinsAgain, line 419: It returns the automatic name as a normal `name=` to the constructor, and the result flag becomes `False`.

probpipe/core/_distribution_array.py
SoymilkWinsAgain, line 513: The automatic name was mistakenly labeled as the user name here.

docs/api/provenance.md
```diff
@@ -57,8 +84,8 @@ probpipe.provenance_config.mode = ProvenanceMode.FULL
 
 posterior = wf(prior)
 anc = provenance_ancestors(posterior)[0]
-anc.obj                       # the live Normal distribution
-anc.obj.sample(key, (100,))   # sample from it
+anc.parent                       # the live Normal distribution
+anc.parent.sample(key, (100,))   # sample from it
```
SoymilkWinsAgain: ProbPipe explicitly uses ops-not-methods, so it should be changed to  sth like `sample(anc.parent, key=..., sample_shape=...)`.

SoymilkWinsAgain (changes requested):
Also:

The PR head still specifies Distribution._auxiliary in STYLE_GUIDE.md. CONTRIBUTING.md also still uses _auxiliary/auxiliary/"source tracking", and its abstract table still describes the old API.

The new test covers basic mixin behavior but does not cover the key boundaries mentioned above. At least the following should be added:

1. `with_name()` for `TransformedDistribution` and dynamic views;
2. Auto-named `MinibatchedDistribution`;
3. Auto-named flags after Product conditionalization;
4. Auto-named flags after DistributionArray slices.

The PR body should add `kind:breaking-change` label.

probpipe/core/tracked.py
jhuggins: Good catch — confirmed and fixed in 0dab1f9: `_shallow_copy` now allocates with `object.__new__(cls)` (the class is already resolved, so the class-selection `__new__` overrides must not run again). Regression tests cover `with_name` on `TransformedDistribution`, `FlattenedDistributionView`, `_RecordDistributionView`, and the empirical router.

probpipe/core/_random_measures.py
jhuggins: Confirmed and fixed in 0dab1f9: `RandomMeasure.__init__` (and `RandomFunction.__init__`, which had the same shape) now applies the default only when no name was supplied, so `MinibatchedDistribution`'s `name_is_auto=True` survives. Test added.

probpipe/distributions/_product.py
jhuggins: Confirmed and fixed in 0dab1f9: the conditioned product mirrors the parent's `name_is_auto` (set directly post-construction — the `**components` signature leaves no room for a keyword). The same sweep also fixed the analogous sites: sweep/broadcast stacks, per-cell batched-params names (eager and backend paths), `FullFactorialDesign`, the minibatch wrappers, and transition resampling. Tests added.

probpipe/core/_distribution_array.py
jhuggins: Confirmed and fixed in 0dab1f9: `_make_distribution_array` gained a `name_is_auto` passthrough and the slice path mirrors the parent's flag. Test added covering both directions (user-named stays False, auto stays True).

docs/api/provenance.md
jhuggins: Fixed in 97b27f9 — both snippets now use the op form (`sample(anc.parent, key=key, sample_shape=(100,))`), with `sample` added to the snippet import.

jhuggins:
Thanks for the careful review — every point was verified and addressed:

- **Inline comments 1–4** (with_name on custom-`__new__` hosts; `RandomMeasure` clobbering the flag; product conditioning and `DistributionArray` slices flipping auto→user): all confirmed and fixed in 0dab1f9, each with a regression test. The `with_name` breakage reproduced exactly as you described. Beyond the four cited sites, the same commit sweeps the whole repo for the shared bug shape — every operation that rebuilds an object from a parent now keeps `name_is_auto` consistent with where the name came from (sweep/broadcast stacks, per-cell batched-params names on both the eager and backend paths, `FullFactorialDesign`, the minibatch draw/log-prob wrappers, transition resampling).
- **Convention docs**: 97b27f9 updates CONTRIBUTING.md (design principles 1 and 6, key-abstractions table with a `Tracked`/`Annotated` row) and STYLE_GUIDE.md (§9.2 rewritten around the annotations store; `tracked.py` added to the public core-module list), plus the ops-form fix in `docs/api/provenance.md`.
- **Tests**: all four requested boundaries are covered in `tests/core/test_tracked.py` (`TestWithNameOnCustomNewHosts`, `TestFlagPropagation`), and 50ed31b also renames the old-vocabulary test symbols (`test_source_*` → `test_provenance_*`, `TestRenamed*` → `TestWithName*`, …).
- **Label**: `kind:breaking-change` added.

Full suite: 3667 passed, 54 skipped; ruff clean. Re-requesting review.

arob5:
## Review — Tracked & Annotated mixins

Reviewed the full diff, ran the touched test modules (all green), confirmed `ruff` clean, `kind:breaking-change` set, and that the `source`→`provenance` / `auxiliary`→`annotations` / `obj`→`parent` rename sweep is complete across source, tests, docs, and notebooks. The design is sound and coverage of the new mixins is strong. No critical issues — the gaps below are additive (no shipped-behavior changes required).

### Recommended

1. **Add a regression test for the `_source_dist` fingerprint fix.** The PR calls out a genuine latent bug — sampleable-source `BootstrapReplicateDistribution`s stored the source in `_source`, which the fingerprint skip-list silently dropped, so replicates of *different* source distributions fingerprinted equal. I verified the fix works:

   ```python
   fp(B(Normal(0, 1, name="x"), replicate_size=10)) != fp(B(Normal(5, 1, name="x"), replicate_size=10))  # True
   ```

   but nothing locks it in. Suggest a test in `tests/core/test_fingerprint.py`: two bootstraps over different `SupportsSampling` sources (same `replicate_size`) fingerprint distinctly, and a same-source pair still matches.

2. **`with_name` is untested on the batch types.** `RecordArray` / `NumericRecordArray` / `DistributionArray` are positioned as first-class `Tracked` terms and inherit `with_name` (which relies on `_shallow_copy` walking their slots/`__dict__`), but `tests/core/test_tracked.py` only covers `Record`, `Normal`, views, `ProductDistribution`, and `EmpiricalDistribution`. Suggest a `with_name` case on a `RecordArray`/`NumericRecordArray` asserting copy identity, shared field data, `name_is_auto=False`, and original unchanged.

### Minor

3. **PR body opens with process narration** — "**Implemented** (this replaces the placeholder)." reads as review-round history; CONTRIBUTING asks the body to describe the final state only. The rest of the body is excellent.

4. **Redundant `auto_name` call in `RandomMeasure.__init__`** (`probpipe/core/_random_measures.py`): inside `if name is None:` it calls `auto_name(name, ...)` with `name` known to be `None`, i.e. just `(default, True)`. The sibling `RandomFunction.__init__` writes this directly. Picking one idiom across the paired constructors reads cleaner.

5. **Duplicated `__all__` section comment** in `probpipe/__init__.py`: the `# Identity & metadata mixins` header appears twice (before `Annotated`, again before `Tracked`) due to alphabetical placement. Cosmetic.

### Verified correct
- `_DistributionMeta` fully removed; `_RecordDistributionMeta` rebased onto `_TrackedMeta`; no dangling references.
- `_auxiliary` → `_annotations` sweep complete (no residual references); diagnostics read/write paths both use `_annotations`.
- Public `.auxiliary` removal and `make_posterior(annotations=)` rename are documented breaking changes with the label applied.
- Auto/user name-flag propagation (conditioning, slices, per-cell backends, sweeps, `FullFactorialDesign`) is consistent and tested.
- `provenance_config` reset by an autouse fixture, so the `ProvenanceMode.OFF` test doesn't leak under xdist.
- Convention docs (STYLE_GUIDE §9.2, CONTRIBUTING principles 1 & 6 + table, CONTRACTS.md) updated in-PR.

jhuggins:
@arob5 All five items addressed in edda740:

1. **Fingerprint regression test** — `TestBootstrapSourceFingerprint` in `tests/core/test_fingerprint.py`: different sampleable sources fingerprint distinctly, same-source pairs match.
2. **`with_name` on the batch types** — `TestWithNameOnBatchTypes` covers `RecordArray`, `NumericRecordArray`, and `DistributionArray` (copy identity, shared field data, `name_is_auto=False`, original unchanged).
3. **PR body opener** — the process-narration sentence is gone; the body now describes the final state only.
4. **`RandomMeasure.__init__`** — writes the default-name branch directly, matching `RandomFunction`; the redundant `auto_name` call (and its now-unused import) removed.
5. **`__all__` comment duplication** — both stray section comments dropped.

Core suite green (1600 passed), ruff clean.

YonghoLim (approved):
Reviewed from the diagnostics side (owner of #193/#324), since this PR does a breaking rename of the storage mechanism that subsystem depends on.

- Ran the full `tests/diagnostics/` suite on this branch: 316/316 pass.
- Checked the whole repo for leftover `_auxiliary` references: none.
- Read every non-rename line touching `probpipe/diagnostics/`, `tests/diagnostics/`, and `docs/api/diagnostics.md`: the diagnostics-side change here is a purely mechanical `_auxiliary` → `_annotations` rename (one comment reword aside) — no logic changes.
- No stale `auxiliary=` kwarg or public `.auxiliary` references left anywhere in diagnostics code/tests/docs.

Also: this resolves the open architectural question from #112 (where diagnostic/algorithm-produced metadata should live) — it lands cleanly on `Annotated.annotations` per design/06-distribution-catalog.md §VI.7. Left a comment on #112 noting that.

LGTM from the diagnostics side.

Issue #255, opened 2026-06-11
Proposal: ConditionalDistribution[S, T] — first-class probability kernels


jhuggins:
# Proposal: `ConditionalDistribution[S, T]` — first-class probability kernels

## Summary

ProbPipe has distributions but no formal notion of a *conditional* distribution — what
probability theory calls a **probability kernel** (Markov kernel): a measurable family
`K : S → P(T)` assigning to each conditioning value `s` a distribution `K(s, ·)` over `T`.
The concept is already load-bearing in at least five places (likelihoods, amortized SBI
posteriors, predictive checks, regression-type models, random-function evaluation), each with
its own ad hoc calling convention. This proposal introduces:

1. **`ConditionalDistribution[S, T]`** — an abstract base (a sibling of `Distribution[T]`,
   *not* a subclass), generic in the conditioning type `S` and the event type `T`. Required
   primitive: `_condition_on(given) -> Distribution[T]`. Structural metadata: a
   `given_template` (schema of `S`) alongside the usual `event_template` (schema of `T`).
2. **A capability-protocol family mirroring the unconditional one** —
   `SupportsConditionalLogProb`, `SupportsConditionalSampling`,
   `SupportsConditionalExpectation`, `SupportsConditionalMean`,
   `SupportsConditionalVariance`, and `SupportsConditionalCovariance`, with methods
   `_conditional_log_prob(given, value)`, `_conditional_sample(given, key, sample_shape)`,
   `_conditional_mean(given)`, etc. Each gives a fused path and a static capability
   declaration for dispatch, and each defaults to materialize-then-evaluate via
   `_condition_on`. The design rule throughout: **`Distribution` and
   `ConditionalDistribution` behave as similarly as possible.**
3. **Op-layer integration** — users interact through the *existing* ops: `condition_on(K, s)`
   evaluates the kernel (it already dispatches through `SupportsConditioning` with no change);
   `log_prob(K, y, given=s)` / `sample(K, given=s, key=...)` / `mean(K, given=s)` expose the
   fused paths; and `condition_on(K, θ_posterior)` composes with the existing
   workflow-broadcast machinery to produce posterior predictive distributions, with a
   `predictive(K, dist, ...)` convenience op as a committed follow-up PR.

With these pieces, `Likelihood` *is* a `ConditionalDistribution` supporting conditional
log-prob, `GenerativeLikelihood` *is* one supporting conditional sampling, and the
[...]

PR #337, opened 2026-07-05
feat(core): unify leaf specs under a ValueSpec base with is_valid (design III.1)

arob5:
# Summary

This PR brings the leaf "spec" classes into line with **Part III.1** of the [consolidated design reference (#332)](https://github.com/TARPS-group/prob-pipe/pull/332): a common `ValueSpec` base with an `is_valid` check, plus the shape/naming corrections the concrete specs needed.

## Why `ValueSpec`, not `FieldSpec`

A spec describes the structure of **a value** — a shape-and-dtype for an array, an inner template for a distribution, and so on. It does **not** describe a *field*: a field (in the `NamedTree` sense) is a *named* leaf, and the name comes from the tree, not from the spec. A `ValueSpec` carries no name and knows nothing about paths; it becomes a field only once an `EventTemplate` gives it one. The old `LeafSpec` alias named the union after its position in the tree rather than what it is, so it is replaced by the `ValueSpec` base.

## What this PR does

1. **`ValueSpec(ABC)`** with abstract `is_valid(value) -> bool`, documented per `CONTRACTS.md` as the base contract for the specs. The docstring states the subclass obligation explicitly (frozen, hashable, by-value dataclass — the union is open and templates hash their specs). `ArraySpec` / `OpaqueSpec` / `DistributionSpec` / `FunctionSpec` subclass it and implement `is_valid`:
   - `ArraySpec.is_valid` — the value is a numeric array-like (numeric Python scalar, or an object with a numeric `dtype` and `shape`) whose shape equals `shape` exactly; its dtype equals `dtype` when set; every element satisfies `support.check` when set. Never raises — an invalid value returns `False`. Two documented boundaries: the `support` check is not `jit`-traceable (Python `bool` reduction), and complex inputs follow JAX's lexicographic ordering under real-ordering constraints (set `dtype` to exclude them; the constraint-layer fix is #344).
   - `OpaqueSpec.is_valid` — the value is a non-array, non-mapping Python object (per the design reference: a numeric array-like belongs under an `ArraySpec`, and a mapping denotes tree structure, never a leaf). `meta` is spec metadata and is not checked.
   - `DistributionSpec.is_valid` — the value is a `Distribution` whose `event_template` equals the spec's, unconditionally: a distribution always carries the schema of its draws, so one that exposes no template — or whose template property fails to run, whatever it raises — does not satisfy any `DistributionSpec`, and `is_valid` returns `False` rather than propagating.
   - `FunctionSpec.is_valid` — the value is callable; the input/output templates document the intended signature but cannot be checked against a bare callable.
2. **`LeafSpec` removed** (hard cut, no alias): the type alias is gone from `probpipe.core.event_template` and the top-level `probpipe` exports; `ValueSpec` is exported instead. Internal isinstance dispatch uses the base class, and docstrings/error messages do not enumerate the concrete subclasses (the set will grow, e.g. `ConditionalDistributionSpec`) — the module docstring keeps the one catalog.
3. **`DistributionSpec.inner_template` → `event_template`**, matching III.6 and the canonical-name table in `CONTRACTS.md` ("a distribution's structural schema" → `event_template`).
4. **`FunctionSpec` accepts a bare `ValueSpec`** on either side and wraps it in a single-field template, so `FunctionSpec(ArraySpec(...), ArraySpec(...))` types a scalar function by the same convention that presents a scalar value as a single field. The auto-derived field names are **`input`** and **`output`** (a bare spec carries no name to key by, and the distinct names keep the two sides' field names disjoint — the same rule III.8 applies to given/event templates). The fields are typed `EventTemplate` (the stored, post-normalisation form, guaranteed after construction); the hand-written `__init__` accepts `EventTemplate | ValueSpec` and raises `TypeError` on anything else. `DistributionSpec` likewise validates that `event_template` is an `EventTemplate`; a bare `ValueSpec` is deliberately *not* auto-wrapped there.

## Design decisions

1. **`is_valid` does the deep check.** For `ArraySpec` that is shape, and dtype/support when set — rather than a structural-only check. The III.1 open point ("validation scope") concerns *when* validation runs, not what `is_valid` means; this PR only provides the predicate and does not decide when construction runs it. The tree-alignment PR can still choose structural-only construction checks by calling something weaker than `is_valid`.
2. **The `ArraySpec` eq/hash quirk is fixed here.** `numpy` treats `np.dtype(None)` as the default dtype, so `ArraySpec((), dtype=None) == ArraySpec((), dtype=float64)` was `True` while the two hashed apart — an eq/hash contract violation. Now: `dtype` is normalised to `numpy.dtype` at construction (so `"float32"`, `jnp.float32`, and `np.dtype("float32")` produce equal, equal-hashing specs; the field is annotated `npt.DTypeLike | None` accordingly), and a custom `__eq__` treats an unset dtype as equal only to unset. `record.py`'s child-template adoption comment is updated accordingly. Behavior change: `repr` of a dtype-carrying spec shows the normalised form (`dtype=dtype('float32')`).
3. **`DistributionSpec.is_valid` is strict.** The design-target contract is that every distribution carries its template, so a missing/`None`/raising template is simply invalid — there is no leniency for distributions that do not yet expose one.

## Notable behavior for review

- `OpaqueSpec.is_valid` rejects `Mapping` values, per the design reference ("mappings are never leaves"). The current `Record` constructor still accepts a plain-`dict` leaf, so the spec `infer_from` derives for a dict leaf does not validate that leaf — this divergence is documented in the docstring as interim and pinned by a deliberate test; the tree-alignment PR resolves it in the design's direction.
- `DistributionSpec.is_valid` compares the *full* template, so a spec keyed `x` matches `Normal(name="x", ...)` (whose single-field template is keyed by the distribution's name) but not `Normal(name="y", ...)` — consistent with V.2's draw-wrapping convention.
- Two known limitations are tracked separately because their fixes span other layers: bfloat16/ml_dtypes arrays classify as non-numeric (#343 — the numeric-kind gates in the template and record layers must move in lockstep; fixed by the follow-up PR stacked on this one), and ordered `Constraint`s do not reject complex values (#344 — touches every ordered constraint class).

## Out of scope / blocked (per #332's III.1 scope)

- **`ConditionalDistributionSpec`** — needs `ConditionalDistribution` (#255) first.
- The substrate-owned validation wiring (`NamedTree._leaf_type()` + construction-time `is_valid` checks) — the NamedTree/Record alignment PR, which builds on this one.
- Symbolic dimensions in `ArraySpec.shape` (`tuple[int | str, ...]`) — a separate III.1 deliverable; shapes here remain concrete ints.

## Dependencies & sequencing

- Independent of the `Tracked`/`Annotated` PR (#336): no semantic overlap, only two one-hunk textual conflicts (adjacent `CHANGELOG.md` entries, and one comment block in `record.py` both PRs touch) for whichever lands second.
- Prerequisite for the NamedTree/EventTemplate/Record/NumericRecord alignment PR.

- [x] Contract of every touched abstraction was clear before coding (ambiguities resolved & noted — the design decisions and the pinned `input`/`output` field names above).
- [x] New/changed contracts fully documented in docstrings (types, checks, raises — `ValueSpec` and each concrete spec's `is_valid`).
- [x] Docstring openings lead with the API/contract; rationale deferred.
- [x] Code verified to obey the documented contract; `tests/core/test_event_template.py` asserts it (per-spec `is_valid` acceptance/rejection, bare-spec wrapping, eq/hash consistency across dtype spellings, pickle round-trips, `TypeError` cases, never-raises regressions). Suites run locally: `tests/core`, `tests/distributions`, `tests/converters`, `tests/inference`, `tests/modeling`, `tests/ci`, `tests/diagnostics`, `tests/validation` — all green.
- [x] Docstrings describe the target contract; the one labeled interim detail is the record layer's plain-`dict` leaf divergence noted on `OpaqueSpec.is_valid`.
- [x] Variable names consistent with the canonical table (`event_template`, `value`).
- [x] `CONTRACTS.md` index updated (`ValueSpec` added to the spec row).
- [x] `ruff format` and `ruff check` clean (verified with the CI-pinned ruff 0.15.16).


probpipe/core/event_template.py
```diff
@@ -673,46 +718,185 @@ def __post_init__(self) -> None:
...
+    event_template: EventTemplate
+
+    def __post_init__(self) -> None:
+        if not isinstance(self.event_template, EventTemplate):
+            raise TypeError(
+                f"DistributionSpec.event_template must be an EventTemplate, "
+                f"got {type(self.event_template).__name__}"
+            )
+
+    def is_valid(self, value: Any) -> bool:
```
SoymilkWinsAgain, line 839: Here, the `event_template` property that throws `RuntimeError` will directly propagate the exception, but the new test only covers `TypeError`. The public contract and PR description both promise not to throw exceptions, but the implementation only catches `AttributeError`/`TypeError`.

probpipe/core/event_template.py
SoymilkWinsAgain, line 988: A standard JAX bfloat16 array is being treated as non-numeric. `ArraySpec((2,), dtype=jnp.bfloat16).is_valid(jnp.ones(..., dtype=jnp.bfloat16))` returns `False`, and an `OpaqueSpec` is generated during inference. Although the PR has deferred this to #343, this directly violates the new API's "numeric array" contract. I suggest finding a workaround.

probpipe/core/event_template.py
```diff
@@ -673,46 +718,185 @@ def __post_init__(self) -> None:
...
+            # No template attribute, or an auto-deriving template property
+            # that cannot run — either way the schema is unknown, so the
+            # value cannot be certified against this spec.
+            return False
+        return template == self.event_template
 
 
 @dataclass(frozen=True)
-class FunctionSpec:
+class FunctionSpec(ValueSpec):
```
SoymilkWinsAgain, line 870: The stored property type of FunctionSpec remains the constructor input union type. At runtime, both properties are guaranteed to be normalized to EventTemplate, as is the design document, but the annotation is retained as EventTemplate | ValueSpec.

SoymilkWinsAgain (changes requested):
In addition, the PR description contains explicit review/process narration. Phrases such as "Implemented (replaces placeholder)," "Review round," and "maintainer call" violate the final-state requirement of the baseline CONTRIBUTING.

probpipe/core/event_template.py
jhuggins: Fixed in 94a53bf. The guard is now `except Exception` around the `event_template` read — a validity probe is total, so any failure to read the value's schema (whatever the property raises) means the value cannot be certified, and `is_valid` returns `False`. The regression test is parametrized over `TypeError`, `RuntimeError`, and `ValueError` properties.

probpipe/core/event_template.py
jhuggins: Fixed in #345, stacked on this PR (base `dev/value-spec-classes`). The fix stays out of this PR's scope deliberately: the dtype gate is duplicated across the template layer, both `NumericRecord`/`NumericRecordArray` leaf checks, and the `Design` marginals probe, and the sites must move in lockstep — widening only the gate here would make `infer_from` call a bfloat16 record numeric that `NumericRecord` construction rejects. #345 replaces the duplicated kind-sets with one shared `_is_numeric_dtype` predicate (numpy kinds, plus kind-`"V"` dtypes exactly when `jnp.issubdtype(dtype, jnp.number)` holds), so your repro — `ArraySpec((2,), dtype=jnp.bfloat16).is_valid(jnp.ones(2, dtype=jnp.bfloat16))` — passes there, `infer_from` infers `ArraySpec`, and structured dtypes stay non-numeric.

probpipe/core/event_template.py
jhuggins: Fixed in 94a53bf. `FunctionSpec` is now `@dataclass(frozen=True, init=False)` with the fields annotated `EventTemplate` — the stored, guaranteed form — and a hand-written `__init__` typed `EventTemplate | ValueSpec` for the documented bare-spec convenience. Constructor input and stored type no longer share one annotation.

jhuggins:
All four review items are addressed:

- **Never-raises contract** — `DistributionSpec.is_valid` now guards the `event_template` read with `except Exception` (94a53bf), with regression tests over `TypeError` / `RuntimeError` / `ValueError` properties.
- **`FunctionSpec` stored type** — fields are annotated `EventTemplate` (the guaranteed stored form); a hand-written `__init__` (`dataclass(init=False)`) carries the `EventTemplate | ValueSpec` constructor union (94a53bf).
- **bfloat16** — fixed in #345, stacked on this branch, which unifies all four numeric-dtype gates behind one ml_dtypes-aware predicate; kept out of this PR because the gates span the template and record layers and must move together.
- **Description** — rewritten to describe the final state per CONTRIBUTING (no process narration).

probpipe/core/event_template.py
```diff
@@ -673,46 +718,197 @@ def __post_init__(self) -> None:
...
+        if self.support is not None:
+            return bool(jnp.all(self.support.check(value)))
+        return True
 
 
 @dataclass(frozen=True)
-class OpaqueSpec:
-    """A non-array Python-object leaf (str, DataFrame, ...).
+class OpaqueSpec(ValueSpec):
+    """A non-array, non-mapping Python-object leaf (str, DataFrame, ...).
```
arob5: My original conceptualization of `OpaqueSpec` was just as a fallback when the value couldn't be described by any of the other specs. I had imagined `is_valid` would simply return `True` for any value. It seems like explicitly barring certain types might be unnecessary here.

probpipe/core/event_template.py
```diff
@@ -673,46 +718,197 @@ def __post_init__(self) -> None:
...
+            # template property that fails to run (whatever it raises) means
+            # the schema is unknown, so the value cannot be certified
+            # against this spec.
+            return False
+        return template == self.event_template
+
+
+@dataclass(frozen=True, init=False)
+class FunctionSpec(ValueSpec):
     """A leaf whose value is a callable.
```
arob5: Using "leaf" here contradicts the base class docstring explanation of the name `ValueSpec`. 

Also, should `input_template`/`output_template` be typed as `EventTemplate | None`? When both are absent, this spec would simply indicate "this value is a function" (this would apply when the template is inferred, as well as for the case that the function has variable input or output types). The use of `input_template`/`output_template` are is then reserved for cases where there is a known or required input/output structure for the function.

arob5:
## Review

A clean, well-tested, well-documented refactor. No correctness bugs and nothing critical. Conventions are followed throughout: the rename sweep is complete (no dangling `LeafSpec`/`inner_template` anywhere in `probpipe`/`tests`/`docs`), the existing `_full_array_shape_or_none` helper is reused rather than reimplemented, `__all__`/exports/`CONTRACTS.md` index are all updated, and the CHANGELOG entry is present under `[Unreleased] → Changed`, marked breaking, with `kind:breaking-change` on the PR.

### Recommended (should fix)

1. **Documented `jax.jit`-traceability of `is_valid` is untested.** `ArraySpec.is_valid`'s Notes make a specific, falsifiable claim: the shape and dtype checks are traceable under `jax.jit`, but a `support`-carrying spec is not (because `bool(jnp.all(...))` forces concretization). Per `CONTRACTS.md` ("tests *assert the contract*, not just happy-path values") and `STYLE_GUIDE` §2.3 on jit paths, this boundary deserves a guarding test — e.g. `jax.jit(lambda v: ArraySpec((3,), dtype=jnp.float32).is_valid(v))(jnp.ones(3, jnp.float32))` returning a static `True`, plus asserting the `support` path concretizes/raises under trace. Without it, a change that accidentally concretizes the shape/dtype path passes CI silently.

### Minor (nice to have)

2. **`ArraySpec.dtype` attribute type is looser than its stored value.** The field is annotated `npt.DTypeLike | None`, which correctly types the *constructor input*, but `__post_init__` normalises it so the stored value is always `np.dtype | None`. A reader doing `spec.dtype.kind` type-checks against `DTypeLike`'s `str`/`None`/type members. Deliberate tradeoff (the CHANGELOG calls out the `DTypeLike` annotation) and it mirrors the `FunctionSpec` field-vs-init pattern — but there the *field* keeps the narrow stored type and only `__init__` widens. A one-line note at the field, or matching the `FunctionSpec` approach, would make the asymmetry read as intentional.

3. **`ArraySpec.is_valid`'s "never raises" is narrower than `DistributionSpec`'s.** `DistributionSpec.is_valid` wraps the template probe in `try/except Exception` for totality; `ArraySpec.is_valid` relies on `hasattr` (AttributeError only) and on `support.check` never raising on an already-validated array. Sound for all normal values — the base docstring scopes the guarantee to "a mismatched value" — so this is only a note: the two specs word the same guarantee differently. Consider aligning the wording.

4. **PR body leans on design-reference section shorthand** ("III.1", "III.6", "V.2", "III.8"). The link to #332 mitigates it, but the bare roman-numeral refs read as insider shorthand for an outside reader (per `CONTRIBUTING.md` on body readability). Expanding them to short phrases would help.

### Verified correct (not findings)

- The `LeafSpec` hard cut and `inner_template → event_template` rename leave **zero** dangling references in `probpipe`, `tests`, or `docs`. §1.12 "rename sweeps are complete" is satisfied.
- `_to_spec` collapses the four-way `isinstance` tuple to `(ValueSpec, EventTemplate)`; the remaining `isinstance(spec, ArraySpec)` sites elsewhere genuinely need `ArraySpec` specifically (to read `.shape`/`.dtype`), so they correctly stay concrete.
- The existing `match="spec must be"` test still matches the reworded `_to_spec` error — not stale.
- The `ArraySpec` eq/hash fix is correct: `eq=False` + hand-written `__eq__`/`__hash__`, comparing `dtype` set-ness explicitly to dodge `np.dtype(None) == float64`. Pickle round-trip and cross-spelling equality are pinned.
- No issue numbers or AI-artifact comments leaked into the code; the `record.py` comment that previously cited `#337` was *removed* (an improvement).

All findings are polish — nothing blocks merge on correctness.

probpipe/core/event_template.py
jhuggins: Agreed on the fallback intent — softened in 353f896 with one refinement. `OpaqueSpec.is_valid` now accepts any value **except** a `Mapping`: a mapping denotes tree structure (a subtree), never a leaf, so it's the one thing an opaque leaf shouldn't accept. Everything else is valid, including a numeric array or scalar (an explicitly-opaque field admits it, even though `infer_from` would route a numeric to `ArraySpec`).

probpipe/core/event_template.py
jhuggins: Both fixed in 353f896.

- The "leaf" framing is gone from all four spec docstring openings (not just `FunctionSpec`) — they now describe the value the spec is for, consistent with the `ValueSpec` naming rationale.
- `input_template` / `output_template` are now optional, defaulting to `None`. A bare `FunctionSpec()` means "any callable, structure unspecified" — the inference / variable-signature case you raised. The templates are kept as `EventTemplate`s (typed `EventTemplate | None`) rather than a flat tuple of types, since an `EventTemplate`'s named fields already express a multi-argument signature: `FunctionSpec(EventTemplate(x=(), y=(3,)), EventTemplate(out=()))` types `f(x, y) -> out`, and the bare-`ValueSpec` sugar covers the single-input case. So `None` was the only missing piece.

probpipe/core/event_template.py
jhuggins: Revisiting this (d2b49d3): we reversed the earlier `except Exception` fix. Catching everything masks exactly the bugs worth surfacing, so on Jonathan's call we narrowed the contract instead of broadening the catch — which is closer to your original framing (align the contract to the implementation, not the other way around).

`is_valid` now catches only `AttributeError` and `TypeError` — the two conditions that genuinely mean a distribution has no derivable schema (no `event_template` attribute; the documented auto-derive `TypeError`). Any *other* error from reading `event_template` (your `RuntimeError`, or a `ValueError`/`KeyError`) is treated as a malfunctioning distribution and **propagates** rather than being silently read as "invalid". The base `ValueSpec.is_valid` and `DistributionSpec` docstrings now state this — a non-match returns `False`, but an unexpected error is not suppressed — and a test pins that `RuntimeError`/`ValueError`/`KeyError` propagate while `TypeError` is a non-match. So the contract and the catch match, without hiding real failures.

PR #338, opened 2026-07-05
feat(core): align NamedTree/EventTemplate/Record/NumericRecord with the design reference (II.1, III.1/III.3)

### LINKED ISSUE #353: Backend-metadata (aux) policy: fix transform/pickle/conversion loss; pytree loss stays by design
## Summary

`NumericRecord` keeps backend metadata (xarray/pandas dims, coords, index, dtype)
in a private `_aux` map so that `to_native()` can reconstruct the original backend
object from the bare `jax.Array` leaves. That aux is currently lost across several
boundaries. **Status: the structural-transform fix is decided and in progress**
(implementation underway on top of the #338 substrate). An audit of the adjacent
surfaces expanded this issue into the single aux-policy record: what survives
where, what is being fixed, and what stays lossy **by design** — so a future
implementer neither re-litigates the fixed parts nor "fixes" the design-mandated
one.

## Background: how aux works

- `NumericRecord` stores, alongside its bare-array leaves, a private `_aux` dict
  **keyed by top-level field name** → `(hooks, blob)`, where `blob` is the
  captured backend metadata. It is populated at coercion time in
  `NumericRecord._coerce_fields` (`probpipe/core/_numeric_record.py`): a field
  whose raw value is a registered backend type (see
  `probpipe/core/_array_backend.py`, `aux_for`) gets an entry, while a nested
  `NumericRecord` child is stored *by reference* and carries its own `_aux`.
- `NumericRecord.to_native()` walks one level, calling `hooks.restore(array, blob)`
  for each aux entry and recursing into nested `NumericRecord` children. It
  returns a permissive `Record`.
- Structural transforms rebuild the root through `Record._rebuild_root`
  (`probpipe/core/record.py`) from the already-extracted bare arrays and thread
  **no** aux.
- The pytree registration's static aux is the triple
  `(event_template, name, name_is_auto)` (`_record_flatten` in `record.py`);
  `_aux` is not part of it.

## The four surfaces (decision table)

| surface | behavior today (verified on `dev/named-tree-record-alignment` @ a5aab3e) | decision |
|---|---|---|
| structural transforms (`without` / `merge` / `replace` / `with_path_names`) | top-level aux lost; nested preserved only via by-reference child reuse | **fix — in progress**: thread a filtered `_aux` through the rebuild (rules below) |
| `to_numeric()` on an already-numeric record | **not idempotent**: returns a rebuilt copy and severs the nested children's by-reference aux | **fix alongside** the transform work (same construction path); also repairs the nested example below |
| pickle / `__reduce__` | aux lost at **every** level (no by-reference escape); `reloaded == original` stays `True`, so the loss is silent | **thread `_aux` through `__reduce__`** — no technical obstacle (blobs are picklable); sequence with/before #352 (see below) |
| JAX pytree round-trip (`tree_map` / `jit` / `vmap`) | aux lost | **keep as-is — design-mandated**; no change proposed (see below) |

## Corrected examples

```python
import xarray as xr
from probpipe import NumericRecord, Record

da = xr.DataArray(...)  # carries dims / coords

# Top-level backend leaf: aux LOST after a transform (the in-progress fix)
nr = Record("nr", temps=da, extra=1.0).to_numeric()
nr.to_native()["temps"]                          # -> DataArray (restored)
nr.without("extra").to_native()["temps"]         # -> bare jax.Array; dims/coords gone

# Nested backend leaf: survives construction AND transforms via by-reference reuse,
# but only under direct construction / auto-promotion:
inner = NumericRecord("grp", temps=da)
outer = NumericRecord("outer", grp=inner, extra=1.0)
outer.without("extra").to_native().at_path("grp/temps")   # -> DataArray restored

# CORRECTION to the earlier version of this issue: appending a redundant
# .to_numeric() severs the nested aux BEFORE any transform, because
# to_numeric() on an already-numeric record rebuilds rather than returning self:
outer2 = Record("outer", grp=inner, extra=1.0).to_numeric()
outer2.at_path("grp") is inner                            # -> False
outer2.to_native().at_path("grp/temps")                   # -> bare jax.Array (no edit at all)
```

The earlier nested example claimed restoration through the `.to_numeric()` form;
verbatim execution shows it does not hold — the asymmetry is real but the
redundant conversion breaks it. Hence the `to_numeric()` idempotence row above.

## Transform aux rules (the in-progress fix)

| transform | aux rule |
|---|---|
| `without` | keep entries for surviving top-level fields |
| `replace` | drop entries for replaced fields (the new value supplies its own hook or none); keep untouched |
| `merge` | union by field name |
| `with_path_names` | rename keys to match renamed top-level fields |

Implementation friction, for the record: `_rebuild_root` lives on the base
`Record` (which has no `_aux`), so this needs a `NumericRecord`-level hook plus a
construction path that accepts carried aux (`NumericRecord.__init__` already takes
an `aux` argument; the public/promotion path does not currently expose it).

## Why the pickle surface is in scope (previously unlisted)

Unlike the transform surface — where every current `to_native()` call site is the
immediate construct-then-export shape — pickle round-trips of records are already
routine and about to become more so:

- **Today:** under `prefect_task` / `prefect_flow` with a Ray/Dask task runner,
  per-row inputs and results cross process boundaries via cloudpickle
  (`tests/core/test_record_serialization.py` exists for exactly this reason). A
  distributed run therefore silently degrades `to_native()` fidelity relative to
  the same pipeline run in-process.
- **With #352 (Prefect result caching):** `persist_result=True` pickles results to
  storage, so a **cache hit would return an observably different object than a
  cold run** (bare arrays vs restored backend types) — caching stops being
  transparent. #352's acceptance criteria currently document this as a reload
  caveat; once the `__reduce__` threading lands here, that criterion flips to
  "restores".

Equality cannot catch the loss (`__eq__` compares the coerced arrays), which is
why the regression tests below must assert the restored **native type**, not `==`.

## Why the pytree half stays as-is (design-mandated)

`design/03-values-and-distributions.md` (Pytrees note) fixes the registration:
"its static aux data is the template and name only, so provenance and annotations
do not survive a JAX transform boundary; lineage instead rides on the computation
layer." Backend `_aux` is an implementation artifact of the array-backend
registry, not a design-reference concept — but the same closure applies
technically: the static aux participates in treedef equality (the `jit` cache
key), and aux blobs contain unhashable coords/indexes, so carrying them there
would break `jit`; treating coords as traced leaves instead is semantically wrong
(they are not part of the differentiable value). Note the design constraint is
scoped to the **JAX boundary only** — it does not mandate loss across transforms,
conversion, or pickle, which is why the other three rows are fixable.

## Acceptance criteria

Transforms + conversion (in progress):
- [ ] `to_native()` after `without` / `merge` / `replace` / `with_path_names`
  restores top-level backend leaves untouched by the edit, matching the
  nested-leaf behavior.
- [ ] `to_numeric()` on an already-`NumericRecord` is identity (or at minimum
  preserves nested child references and aux); the corrected nested example above
  restores after an edit.
- [ ] Regression tests covering each transform → `to_native()`, for both top-level
  and nested backend leaves, and for the `replace` / `merge` aux semantics —
  asserting restored native types, not `==`.
- [ ] Transform docstrings in `record.py` updated to state that backend aux for
  untouched leaves survives.

Pickle (sequence with/before #352):
- [ ] `__reduce__` carries `_aux` at every nesting level; pickle and cloudpickle
  round-trips restore `to_native()` for top-level and nested backend leaves.
- [ ] A test asserts the restored native type across a dump/load cycle (the `==`
  check alone passes even when aux is lost).
- [ ] #352's reload-fidelity acceptance criterion updated from "documented as
  lost" to "restores".

Unchanged by design:
- [ ] The JAX-round-trip docstrings (`_record_flatten`, the `NumericRecord`
  pytree note) remain "does not survive", citing the design constraint.
- [ ] `design/02` / `design/03` updated only if the structural-transform contract
  wording there needs to reflect aux survival (the pytree text is already
  correct).

## Sequencing

The transform/conversion work targets the #338 substrate (`_rebuild_root`, the
promotion construction path) and lands after #338. The pickle row should land
before or with #352 so result caching never exhibits hit-dependent fidelity.

arob5:
## Summary

Brings the four core value/structure classes — `NamedTree` (the shared tree substrate), `EventTemplate`, `Record`, and `NumericRecord` — into line with **Parts II.1 and III.1/III.3** of the consolidated design reference (`design/` on `main`, [#332](https://github.com/TARPS-group/prob-pipe/pull/332), as amended by [#346](https://github.com/TARPS-group/prob-pipe/pull/346)). This is the "make the collection substrate match the contract" step, after [#327](https://github.com/TARPS-group/prob-pipe/pull/327) established the leaf-keyed collection interface and on top of [#336](https://github.com/TARPS-group/prob-pipe/pull/336) (`Tracked`/`Annotated`) and [#337](https://github.com/TARPS-group/prob-pipe/pull/337) (`ValueSpec`).

The single placeholder commit is superseded by the implementation below.

## Breaking changes

- **`Record` / `NumericRecord` are constructed name-first.** The signature is now `Record(name, fields=None, /, *, event_template=None, name_is_auto=False, **kw_fields)`: the record's name is a required first positional argument. The old `name=` keyword and the nameless forms are removed — no deprecation shim. Operations that build records internally derive a `record(<field-keys>)` name and mark it `name_is_auto=True` through a private `_auto_record` helper, so the previous auto-naming behavior is preserved for operation-produced records.
- **All-numeric records auto-promote to `NumericRecord`.** `Record(...)` whose fields are all coercion-free numeric leaves (bare jax/numpy arrays and numeric scalars) returns a `NumericRecord`; this reverses the previous explicit "not auto-promoted" behavior and mirrors `EventTemplate → NumericEventTemplate`.
- **`EventTemplate.from_field_values` is removed**, replaced by `Record.from_field_values(name, template, values)`. `NumericRecord.from_vector(name, template, vec)` is added as the value-level inverse of `to_vector`.
- **A `Mapping` is never a leaf.** Construction materializes a mapping-valued field into a nested subtree, exactly as `from_nested_dict` reads every mapping as a subtree. Diagnostics payloads that were carried as Records with dict-valued fields are now plain dicts.

## What landed

**`NamedTree` (II.1)** — relocated to a new public module `probpipe/core/named_tree.py` (was the private `_NamedTree` in `event_template.py`). It owns the leaf-keyed mapping interface, nested-dict (de)construction, the structural edits `merge` / `without` / `replace`, and gains `with_path_names(mapping=None, /, **kwargs)` — rename leaves or whole subtrees by path, or by bare name when unique. Each family declares its `_leaf_type()` (`ValueSpec` for templates, arbitrary values for records), and construction checks every leaf against that declared type (a trace-safe `isinstance` check) and validates each leaf against its spec's `is_valid` — structure only (shape + dtype via `numpy.can_cast` same-kind, so a cross-kind dtype raises); an `ArraySpec`'s `support` is descriptive metadata, not checked by `is_valid` (a data-dependent check that is not `jax.jit`-traceable); `is_multi_field` moved down from `EventTemplate`; `is_leaf` removed in favor of `is_field`.

**`EventTemplate` / `Record` / `NumericRecord` (III.1 / III.3)** — auto-promotion with the numeric axis re-derived under every structural transform (a `replace`/`without`/`merge` that drops the last non-numeric leaf promotes; one that introduces a non-numeric leaf demotes). The pytree registration carries the event template and identity `(template, name, name_is_auto)` in the treedef aux, so `tree_map`/`jit`/`vmap` preserve a record's template, name, and auto flag; provenance and annotations do not cross the transform boundary.

## Design decisions (the placeholder's four open questions, resolved)

1. **`from_field_values` placement** — moved to `Record.from_field_values(name, template, values)`; the `EventTemplate` form is removed outright (hard cut, no shim), since value construction belongs on the value type and the template argument is now explicit.
2. **`NamedTree` module** — relocated to its own public module `probpipe/core/named_tree.py` in this PR, matching the reference's framing of it as standalone Part II infrastructure.
3. **Auto-promotion vs. an explicit `event_template`** — promotion fires only when **both** the template permits it (no template, or a `NumericEventTemplate`) **and** every field value is coercion-free numeric. An explicit non-numeric template therefore wins over numeric data (stays a plain `Record`), and — the sharper edge — a plain `Record` holding a verbatim backend leaf (xarray/pandas) stays plain even when its own numeric template is threaded back through an edit or `from_field_values`, because coercing that leaf is lossy. The numeric axis stays a function of the current leaves, re-derived under every transform.
4. **`with_path_names` bare-name ambiguity** — confirmed: a bare name resolves to the unique node so named and raises `ValueError` if the tree contains it more than once; full paths are always unambiguous.

## Deviation from the reference (called out)

The pytree aux is `(template, name, name_is_auto)` — the reference lists "template and name". The extra `name_is_auto` flag is required so a reconstructed record re-derives an auto-name but preserves a user-given one; without it, the transform-identity rule cannot be applied after a JAX round-trip.

## Correctness notes

- **Promotion requires coercion-free values, not just a permitting template.** A plain `Record` holding an xarray/pandas leaf carries a numeric template (inference reads the leaf's shape/dtype) but is excluded from `NumericRecord` promotion, so re-threading that template does not silently coerce the leaf: `from_field_values(r.name, r.event_template, r.values()) == r` holds.
- **Pytree flatten emits children in template order** (matching unflatten), so `tree_map` over a record built with an explicitly-ordered template never transposes field values.
- **One `_rebuild_root` helper owns the transform-identity rule.** Every structural edit (`without` / `merge` / `replace` / `with_path_names`) re-derives an auto-derived name and preserves a user-given one; `Record(...)` with the name omitted reports `Record`, not the promoted `NumericRecord`.

## Out of scope

- `fields` / `to_dict` temporary aliases (coupled to the distribution-layer `fields → children` rename, [#328](https://github.com/TARPS-group/prob-pipe/issues/328) §6).
- `RecordArray` / `NumericRecordArray` → batch renames and the batch-axis contract ([#235](https://github.com/TARPS-group/prob-pipe/issues/235) Chapter 2, [#328](https://github.com/TARPS-group/prob-pipe/issues/328) Q8).
- Distribution-layer `event_template` / `event_shape` moves ([#328](https://github.com/TARPS-group/prob-pipe/issues/328)).

## Verification

- Full suite green (`pytest tests/`), plus `ruff check` / `ruff format` clean across `probpipe`, `tests`, and `example_scripts`.
- All four edited docs notebooks re-execute end-to-end under the User-Guide and tutorial (nutpie + bayesflow + pymc + pyabc) dependency sets, matching CI.
- `CHANGELOG.md`, `CONTRACTS.md`, `STYLE_GUIDE.md`, and `CONTRIBUTING.md` updated for the new substrate and name-first constructors.

## References

- Design reference **Part II.1** (`NamedTree`) and **Part III.1 / III.3** (`EventTemplate`, `Record`, `NumericRecord`) — `design/` on `main` ([#332](https://github.com/TARPS-group/prob-pipe/pull/332)), amended by [#346](https://github.com/TARPS-group/prob-pipe/pull/346).
- Leaf-keyed collection design — [#326](https://github.com/TARPS-group/prob-pipe/issues/326), implemented in [#327](https://github.com/TARPS-group/prob-pipe/pull/327).
- Master value-model plan — [#235](https://github.com/TARPS-group/prob-pipe/issues/235) (predates the reference; where they differ, the reference wins).

- Backend-metadata round-trip for `NumericRecord` — structural transforms (`without`/`merge`/`replace`/`with_path_names`) and pickle/cloudpickle now preserve `xarray`/`pandas` aux for `to_native`, consistent with the already-preserved nested case. Closes #353.

- Follow-up: [#355](https://github.com/TARPS-group/prob-pipe/issues/355) — store native-form leaves in `NumericRecord` and convert lazily at the compute boundary; would dissolve the backend-aux threading shipped here and the promotion exclusion for backend-typed leaves.


probpipe/core/record.py
```diff
@@ -311,29 +405,63 @@ def __init__(
...
+        the field re-derives it.
+        """
+        if child._name == field_name:
+            return child
+        renamed = child._shallow_copy()
+        object.__setattr__(renamed, "_name", field_name)
+        object.__setattr__(renamed, "_name_is_auto", True)
+        return renamed
+
     def _validate_event_template(self, event_template: EventTemplate) -> None:
```
SoymilkWinsAgain: Only the structure is checked here, but the design requires all values ​​to satisfy `ValueSpec.is_valid`, and specifies that `from_field_values` throws an error if the `shape` does not match. 

The current test treats binding the same float value to an int32 template as a valid scenario, which should be modified.

probpipe/core/_numeric_record.py
```diff
@@ -377,6 +522,19 @@ def to_native(self) -> Record:
     # ---------------------------------------------------------------------
 
     def __reduce__(self):
```
SoymilkWinsAgain, line 516: This method does not serialize `event_template`, and the recovery function will re-infer the template from the native value. This causes Ray/cloudpickle transports to silently change schema and equality. Existing serialization tests only check the aux, not the template.

probpipe/core/named_tree.py
```diff
@@ -0,0 +1,807 @@
...
+                )
+            groups[head] = value
+            is_prefix[head] = False
+    result: dict[str, Any] = {}
+    for head, val in groups.items():
+        result[head] = _unflatten_paths(val) if is_prefix[head] else val
+    return result
+
+
+class NamedTree:
```
SoymilkWinsAgain, line 94: The implementation is a regular `NamedTree` class, with accessors and transformations frequently returning `Any`; `NamedTree[int]` will directly throw a `TypeError`. This contradicts the design of `NamedTree[L]` and diminishes the value of the `py.typed` package for downstream users.

CHANGELOG.md
```diff
@@ -169,9 +208,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
     (best-effort, lossy inference). The value upcast is consolidated to
     `Record.to_numeric()`.
-  - **Moved** `to_vector` / `from_vector` and `leaf_shapes` onto
-    `NumericEventTemplate`; `from_vector`'s `non_numeric` argument is dropped.
-    `numeric_leaf_shapes` is consolidated into `leaf_shapes`.
+  - **Moved** `leaf_shapes` onto `NumericEventTemplate`; `numeric_leaf_shapes`
+    is consolidated into `leaf_shapes`. (`to_vector` / `from_vector` are now
+    value-level methods on `NumericRecord` / `NumericRecordArray` — see the
+    value-model entry above — not template methods.)
   - **Added** leaf-keyed (de)composition: the mapping protocol
```
SoymilkWinsAgain, line 222: The new entry says EventTemplate.from_field_values ​​has been removed, but this entry still says it's responsible for rebuilding values. It should be removed.

SoymilkWinsAgain (changes requested):
Also, I think the PR body isn't precise. It now says that mapping-valued fields will be rejected, but the implementation and design actually materialize them into subtrees. Suggest revising

arob5:
One thing to keep in mind is the logical split between module-level and class/function-level docstrings, given how they will display in the API documentation. For example, there is currently redundancy between the module-level and class-level docstrings for `NamedTree`. I typically think of the logical split to go something like this:
- Module: brief introduction to the abstraction implemented in the module. Then information on where it fits within ProbPipe more broadly and how it might interact with other parts of the package. 
- Class: all of the specifics regarding the abstraction itself go here

probpipe/core/record.py
jhuggins: Done in f391184. Construction now enforces per-leaf shape and dtype: shape must match, and dtype follows `numpy.can_cast` — an exact or value-preserving (safe) cast is silently accepted, a within-kind narrowing warns, and a cross-kind (unsafe) cast raises, so binding a float value to an int-dtype spec is now rejected. `from_field_values` inherits this (it builds through construction), and the equality test that bound a float to an int32 template is reworked into an error assertion.

The data-dependent `support` check is intentionally *not* run at construction: it reduces to a Python `bool`, so it can't run under `jax.jit` tracing, and construction happens inside traces (pytree unflatten). Shape + dtype are trace-safe and are enforced. This resolves the III.1 validation-scope open point (design updated accordingly).

probpipe/core/_numeric_record.py
jhuggins: Fixed in f92febf. `event_template` is now serialized in `Record.__reduce__` and both `NumericRecord.__reduce__` branches (the aux/native path included) and threaded back through the unpicklers via `event_template=`, so the schema — and therefore equality — survives a `pickle` / cloudpickle (Ray) round-trip instead of being re-inferred. Base `Record` had the same gap; `RecordArray` already serialized its template.

Regression coverage: `TestPicklePreservesTemplate` in `test_record_serialization.py`, using a leaf `support` / `OpaqueSpec.meta` (values `infer_from` cannot recover) for plain `Record`, `NumericRecord`, cloudpickle, and the native/aux path.

probpipe/core/named_tree.py
jhuggins: Agreed. Scoped out of this PR to keep it behavioral, and tracked in #356 — making `NamedTree` generic via PEP 695 (`class NamedTree[L]`), typing the leaf-trafficking methods so leaf types reach downstream users, and the related gap that there's no CI type-checker enforcing the shipped `py.typed` today.

CHANGELOG.md
jhuggins: Fixed in 8458bcb — the entry no longer credits the removed `EventTemplate.from_field_values`; reconstruction from a leaf list is now `Record.from_field_values`.

jhuggins:
Addressed the review on `dev/named-tree-record-alignment`:

- **PR body corrected** — mapping-valued fields materialize into subtrees, not rejected.
- **`__reduce__`** now serializes `event_template`, so schema/equality survive pickle & cloudpickle (f92febf).
- **Construction enforces per-leaf shape + dtype** (f391184): dtype via `numpy.can_cast` tiers — exact/safe silent, within-kind narrowing warns, cross-kind raises. The `support` check stays descriptive since it isn't `jax.jit`-traceable (construction runs inside traces).
- **CHANGELOG** `from_field_values` entry fixed (8458bcb).
- **Generic `NamedTree[L]`** tracked separately in #356.

Full `tests/core` green (1740 passed).

jhuggins:
Thanks — adopted that split in 2bb3def. Trimmed the `NamedTree` and `NumericRecord` **module** docstrings to a brief intro + where each fits in ProbPipe (its interactions with `Record` / `EventTemplate` and the compute paths) + a pointer to the class, leaving the specifics in the class docstrings that already held them. The `NamedTree` redundancy you flagged — the field/path terminology and the mappings-never-leaves rule — is gone.

Two notes on the rest:
- `record.py`'s module docstring already follows this split (placement-focused: the non-random counterpart to `Distribution`, the boundary wrapper for `WorkflowFunction`s, the `Record` family), so I left it as-is.
- `event_template.py`'s module docstring is more specifics-heavy, but it's currently serving as the overview/catalog for a multi-class module — and `ValueSpec`'s class docstring points back to it for the spec catalog. Happy to redistribute those specifics into the `EventTemplate` / `ValueSpec` / concrete-spec class docstrings if you'd like the same treatment there; flagging it since it'd cascade to that cross-reference.

probpipe/core/record.py
jhuggins: Follow-up / refinement (cc31579): in the end we removed the `support` check from `is_valid` altogether rather than special-casing it at construction. `ArraySpec.is_valid` now validates structure only — shape + dtype (via `numpy.can_cast` same-kind) — and is fully `jax.jit`-traceable; `support` stays a descriptive field on the spec, not checked by `is_valid`. Construction calls `spec.is_valid(value)` directly (the separate validation helper is gone), so there's no duplicated shape/dtype logic.

Net effect on the original point is unchanged: a value that violates its field's shape/dtype — e.g. a float bound to an int-dtype spec (a cross-kind cast) — is rejected at construction. Two refinements vs my earlier note: a same-kind cast (incl. a within-kind narrowing) now passes silently (no warning), and `support` conformance is descriptive rather than enforced (it isn't `jax.jit`-traceable, and `is_valid` runs at construction, which happens under trace).

arob5:
## Review

### Summary
A large, high-quality refactor (+3782/−2132, 84 files) that extracts the shared tree substrate into a new public `NamedTree`, makes `Record`/`NumericRecord` name-first with numeric auto-promotion, and moves `to_vector`/`from_vector`/`from_field_values` onto the value types. The core modules are exceptionally well-documented, the migration is applied consistently across consumers and docs, and no lingering references to removed APIs survive anywhere in the tree. I found **one real naming bug** and some **test-coverage gaps**; no value-level correctness or backward-compat problems. Conventions (naming, imports, types, docstrings, dependency graph, `__all__`, CHANGELOG, `kind:breaking-change` label) are all satisfied.

### Recommended (should fix)

**1. Name-shadowing bug in `_product.py` `_map_components` (line ~564).** The PR added a `name` parameter to thread `self.name` through, but the loop `for name, comp in components.items():` (line ~567) rebinds it, so `return Record(name, fields, name_is_auto=True)` (line ~572) uses the **last top-level component key** rather than the passed-in name. `_mean()` / `_variance()` call `_map_components(self.name, …)` and return the record directly, so `mean(product)` / `variance(product)` come back named after the last component (e.g. `"y"`) instead of the distribution's name. Values and structure are correct — this is an identity/auto-name regression that directly defeats the PR's own stated goal ("an operation that produces a record supplies a meaningful name … the producing distribution's name"). On `main` this returned a derived `record(x,y)`-style name, so the new behavior is actually *worse* than before. Fix: rename the loop variable (e.g. `for field_key, comp …`).

**2. Test gap: backend aux (xarray/pandas) preservation across structural edits is untested.** The PR body prominently claims `xarray`/`pandas` aux now survives `without`/`merge`/`replace`/`with_path_names` (closes #353), and `_numeric_record.py` carries ~90 lines of machinery (`_carry_aux`, `_backend_aux_by_leaf_id`, `_reattach_backend_aux`) to make it work. Aux fidelity is tested only for pickle/cloudpickle and construction — never through a structural edit + `to_native()`. This is the single in-scope behavior with dedicated new code and no regression test. A small test class (e.g. `NumericRecord("nr", temps=xr_da, extra=jnp.array(1.0)).without("extra").to_native()["temps"]` asserting coords/attrs survive, plus the `merge`/`replace`/`with_path_names` analogues) would close it.

### Minor (nice to have)

**3. Dead/misleading `name` param in `_product.py` `_sample_nested` (line ~533).** Same shadowing (`for subkey, name in zip(keys, names)`); the `name` parameter is never used and `return Record(name, …)` uses the last field key. Harmless today because the result is always stored under a field key where `_named_by_key` renames it correctly — but it's dead code and should be fixed alongside #1.

**4. Shape-only assertions where a value check is cheap (STYLE_GUIDE §8.6 "real value over shape"):**
- `test_named_tree.py::TestPytreeAuxSplit.test_numeric_record_jit_and_vmap` asserts only `out.shape == (4,)`; inputs are known, so a per-row value check would catch a field-transpose that preserves shape.
- `test_event_template.py::TestToVector.test_multi_field_value` asserts only `vec.shape`; the exact concatenation is known and would also verify canonical ordering.
- `test_named_collection.py::test_map_preserves_class_and_reinfers_template` maps `lambda x: 2*x` over `jnp.zeros(3)` — since `2*0==0`, an identity/no-op map passes identically; use a non-zero input.

**5. A few untested error paths:** `with_path_names` field-vs-prefix re-wrap branch (`named_tree.py` ~735-738, distinct message); `replace` "both positional mapping and keyword arguments" `ValueError` and the no-op `return self` early return; `at_path` non-string-segment `TypeError`.

**6. PR body hygiene (CONTRIBUTING.md "no review-round narration").** The body is otherwise excellent and describes the final state, but the "Post-implementation review round" and "A second review pass swept the stale constructor examples …" sections are review-process narration an outside reader doesn't need. Consider folding those fixes into the relevant "What landed" bullets. (Non-blocking.)

### Verified clean
- Diagnostics dict-valued-field migration (`_mcmc.py`, `_ppc_spc.py`): pure ops correctly converted from records-with-dict-fields to plain payload dicts, with all internal consumers updated in lockstep — exactly the case the "mappings are never leaves" rule would otherwise have silently exploded into subtrees.
- No remaining `Record(name=…)` keyword form, `EventTemplate.from_field_values`, template-level `.from_vector`, `_record_cls`, or `_NamedTree` references anywhere in `probpipe/`.
- The field-order-transpose bug the PR fixed has solid regression coverage (`test_roundtrip_with_out_of_order_template`, `test_tree_map_realigns_values_by_template_order`).
- Docs fully migrated: CHANGELOG, CONTRACTS.md, STYLE_GUIDE.md, CONTRIBUTING.md, `design/`, `docs/api/records.md`, example scripts, and `__init__.py`/`__all__` (`NamedTree` exported).

### Suggested changes
1. In `_product.py`: rename the loop variable in `_map_components` so `return Record(name, …)` uses the threaded `name`; do the same in `_sample_nested`. **[fixes #1, #3]**
2. Add a test class exercising xarray/pandas aux fidelity through `to_native()` after `without`, `merge`, `replace`, and `with_path_names`. **[fixes #2]**
3. Tighten the three shape-only assertions in #4 to concrete-value checks. **[fixes #4]**
4. Optionally add the error-path tests in #5 and trim the review-round narration from the PR body (#6).

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

probpipe/core/_numeric_record.py
```diff
@@ -437,15 +599,119 @@ def ndim(self) -> int:
         return int(getattr(leaf, "ndim", 0))
 
 
+# ---------------------------------------------------------------------------
+# 1-D vector reconstruction (value-level; the template supplies only layout)
+# ---------------------------------------------------------------------------
+
+
+def _value_treedef(
```
SoymilkWinsAgain, line 607: This function consistently uses a float32 placeholder to construct a Record carrying the actual template. The results are:

1. For templates with dtype=int32 or bool, even if the input vector's dtype matches perfectly, NumericRecord.from_vector will throw a ValueError before processing the actual vector.

2. Batch paths of NumericRecordArray.from_vector may successfully return an object that is actually float32, with the template declared as int32 and is_valid=False.

3. Even with tree_unflatten validation disabled, shape-changing jax.tree.map can still return a stale template: the actual scalar, and the template is still (2,).

Furthermore, the current tests do not cover dtype-pinned single-value, batch, or mixed dtype templates.

probpipe/core/named_tree.py
```diff
@@ -0,0 +1,807 @@
...
+        Construction validates every leaf against this type via
+        :meth:`_check_leaf`, so a malformed tree fails at construction rather
+        than at first navigation. A family whose leaves are arbitrary values
+        declares ``object``, making the isinstance check vacuous; the
+        mappings-are-never-leaves rule still applies.
+        """
+        return object
+
+    @classmethod
+    def _check_leaf(cls, path: str, value: Any) -> None:
```
arob5, line 206: It seems like this method should incorporate the "Mapping not allowed as leaf" invariant, which is mentioned in docstrings but never actually codified in `NamedTree`. This prevents having to duplicate logic ensuring the invariant is satisfied in downstream classes. 

The complementary question of how constructors handle mappings (e.g., converting them to the internal node type) would remain within the purview of inheriting classes, given that `NamedTree` does not implement construction logic.

probpipe/core/event_template.py
```diff
@@ -775,10 +202,11 @@ def is_valid(self, value: Any) -> bool:
             actual = getattr(value, "dtype", None)
             if actual is None:
                 actual = np.asarray(value).dtype
-            if np.dtype(actual) != self.dtype:
+            # A same-kind cast (a widening promotion or a within-kind
+            # narrowing) matches; a cross-kind cast (e.g. float where an int
+            # dtype is declared) does not.
+            if not np.can_cast(np.dtype(actual), self.dtype, casting="same_kind"):
```
SoymilkWinsAgain, line 208: The record did not actually perform the conversion. I had the AI ​​run a test, and both int32 -> float32 and float64 -> float32 failed. Furthermore, `support=positive` returns `True` for values ​​containing negative numbers. "Being able to convert" does not mean the current value conforms to the authoritative schema. This also violates the ValueSpec's own contract and the base design's provision that is_valid overrides dtype/support.

probpipe/core/named_tree.py
```diff
@@ -0,0 +1,796 @@
...
+    structure (see :meth:`_check_leaf` and :meth:`from_nested_dict`).
+    Navigation yields views into the same storage; the structure-preserving
+    transforms (:meth:`map`, :meth:`replace`, :meth:`merge`, :meth:`without`,
+    :meth:`with_path_names`) return a tree of the same family, rebuilt through
+    :meth:`_rebuild_class`.
+
+    This substrate provides no constructor logic; each family's constructor
+    owns storage (the ``_tree`` attribute) and validation policy.
+    """
+
```
SoymilkWinsAgain, line 132: `NamedTree` still does not implement the generic contract in its design. The implementation is still `class NamedTree`, and leaf access and transformations return a large number of `Any` values. `NamedTree[int]` will throw a `TypeError`.

CHANGELOG.md
SoymilkWinsAgain: This seems conflict with above... Line 150 says " `is_valid` — a data-dependent check that is not `jax.jit`-traceable." Not sure which one is correct. I suggest we confirm this first before fixing the `is_valid` issue I just commented.

SoymilkWinsAgain (changes requested):
The PR document still retains the process description, which contradicts repo regulations.

arob5:
A couple comments on the `ValueSpec` hierarchy:
1. Asymmetry between `DistributionSpec` and `FunctionSpec` constructors: the former requires user pass inner `EventTemplate` directly, while latter provides the convenience path that wraps `ValueSpec`s as single-field templates.
2. I imagine the single-field wrapping logic will come up in other places (e.g., when `ConditionalDistribution` is implemented). It seems worthwhile to pull out the core logic into a helper function.

arob5:
In `NumericEventTemplate`'s `_post_validate`, the error raised for `OpaqueSpec` and the error raised for `FunctionSpec`/`DistributionSpec` should probably be consolidated into a single "only array specs allowed" error message.

jhuggins:
Good points — and there's a wrinkle that makes this a cross-cutting convention question rather than a local fix.

The single-field wrap `FunctionSpec` offers is (per `design/03` III.1 and III.5) the same convention `LinOp` uses ("built from a bare matrix defaults to single-field templates"). `DistributionSpec` deliberately omits it because the two specs *use* their templates differently: `FunctionSpec.is_valid` is `callable(value)` — the input/output templates are documentary and never validated — whereas `DistributionSpec.is_valid` is a template **equality** check. So a fixed-name single-field wrap (e.g. `"value"`) would make `DistributionSpec` *reject* a real scalar distribution, whose draw is keyed by the distribution's own name, rather than accept it.

Rather than resolve that locally, I've filed #357 to settle the convention — and your helper-extraction (which `ConditionalDistribution` will also want) — consistently across `FunctionSpec` / `DistributionSpec` / `LinOp`; it also captures your `FunctionSpec.is_valid` under-validation observation. PR #338 makes no spec-constructor change.

probpipe/core/_numeric_record.py
jhuggins: Fixed in 375f893. `_reconstruct_from_vector` now casts each reconstructed block back to its field's declared dtype, so a dtype-pinned (int32 / bool) template round-trips through `to_vector` (which promotes to a common float across mixed-dtype fields). The throwaway `_value_treedef` skeleton is now built with `_validate_leaves=False`, so its float32 fill no longer trips construction validation against a dtype-pinned template — that was the `ValueError`-before-processing you flagged. Regression coverage: `test_dtype_pinned_vector_round_trip` (a mixed int32/float32 template).

On the shape-changing `jax.tree.map` case (stale template vs the actual scalar): that's the inherent JAX-pytree caveat — unflatten must reproduce the flattened treedef, so a shape-changing map yields a record whose carried template no longer matches its leaves. It's documented as unsupported rather than validated (validating under a transform isn't possible), and the unflatten path deliberately skips leaf validation for exactly that reason.

probpipe/core/event_template.py
jhuggins: You're right that `is_valid` no longer checks `support` and treats a same-kind dtype as valid — that's a deliberate decision (cc31579), now stated in the design (`design/03` III.1 §89 and the III.3 construction-validation note §178), not an oversight.

The reason is trace-safety. `is_valid` is the check construction runs, and `Record` construction happens *inside* `jax.jit` / `vmap` — the pytree unflatten path reconstructs a value within the trace. `support` reduces to a data-dependent `bool(jnp.all(constraint.check(value)))`, which cannot execute under tracing. So `is_valid` is scoped to the trace-safe structural checks (shape + dtype-kind), and `support` is descriptive metadata rather than an enforced constraint. On dtype specifically, we use `numpy.can_cast(..., "same_kind")` so a template written `float64` accepts a JAX-default `float32` (and vice-versa) instead of spuriously rejecting it; a cross-kind cast (e.g. a float value against an int spec) still fails.

So the contract is: a value satisfying `is_valid` conforms *structurally*. Enforcing `support` / exact-dtype conformance would need a separate, non-trace-safe validation pass — deliberately out of scope for this PR. If you see a trace-safe way to fold `support` in, I'm glad to reconsider.

probpipe/core/named_tree.py
jhuggins: Acknowledged — `NamedTree` is still a plain (non-generic) class here. We've deliberately scoped the generic `NamedTree[L]` work out of this PR to keep it behavioral, and tracked it in #356: PEP 695 `class NamedTree[L]`, typing the leaf-trafficking methods so leaf types reach downstream users, plus the related gap that there's no CI type-checker enforcing the shipped `py.typed` today. Not addressed in PR #338 by design.

CHANGELOG.md
jhuggins: Good catch — reconciled in 375f893. The two entries now agree: `is_valid` is a structural check (shape + dtype), and an `ArraySpec`'s `support` is descriptive metadata that `is_valid` does **not** check (it isn't `jax.jit`-traceable — see the `is_valid` thread for the rationale). The `#337` entry no longer lists `support` among `is_valid`'s checks.

probpipe/core/named_tree.py
jhuggins: Done in e89b285. `NamedTree._check_leaf` now enforces the "a mapping is never a leaf" invariant directly — it raises if a mapping reaches leaf validation — so the rule is codified in the substrate rather than docstring-only. As you noted, how a constructor *materialises* a mapping into the node type stays the family's concern: `Record` / `EventTemplate` materialise before `_check_leaf`, so for them this is a backstop; any family that routes leaf validation through `_check_leaf` now gets the invariant for free.

jhuggins:
Thanks for the thorough review — all points addressed:

- **#1 / #3 — `_product.py` name-shadowing:** fixed in 375f893. Renamed the loop variable in `_map_components` (and `_sample_nested`), so `mean(product)` / `variance(product)` are named after the distribution again rather than the last component. Regression: `test_mean_variance_named_after_distribution`.
- **#2 — aux across structural edits:** this is already covered — `TestBackendAuxSurvivesStructuralTransforms` in `tests/core/test_array_backend.py` exercises `without` (top-level + nested), `merge`, `replace`, and `with_path_names`, each followed by `to_native()` asserting coords/dtype survive (plus a source-unchanged check). Looks like the review ran against a commit before that test landed.
- **#4 — shape-only assertions:** tightened all three to concrete-value checks (e89b285) — the vmap output (catches a field transpose), the `to_vector` concatenation + field ordering, and a non-zero `map` input.
- **#5 — error-path tests:** added in e89b285 — `at_path` non-string segment, `replace` both-positional-and-keyword + the no-op return-self, and the `with_path_names` field-versus-prefix collision.
- **#6 — PR-body narration:** the "Post-implementation review round" section is reframed as factual "Correctness notes".

Separately, from your later comments: the `NumericEventTemplate` non-array-leaf error is now a single consolidated "only ArraySpec leaves allowed" message (41e123b), the mappings-never-leaves invariant is codified in `NamedTree._check_leaf` (e89b285), and the `ValueSpec` single-field-wrap convention is tracked in #357.

Issue #340, opened 2026-07-07
Nested empirical / nested-prior inference broken: NumericRecordArray not leaf-keyed


jhuggins:
## Summary

Constructing an `EmpiricalDistribution` / `RecordEmpiricalDistribution` over a
**nested** record raises `KeyError: "'…' is a subtree, not a field; use
at_path() to navigate to it"`. This breaks **nested-prior inference end to end**
(e.g. NPE over a prior with `outer={r, m}`), and is currently failing CI on
`main` (the `bayesflow` job — see the companion fix PR, which `xfail`s the test
that exercises it).

## Root cause

The leaf-keyed `Record`/`EventTemplate` redesign (#326 / #327) migrated the
**single** record types but **not the batch types**. Batch-type leaf-keying was
explicitly deferred in #326 (the `*Array → *Batch` phase of #235). As a result
the mapping protocol is **inconsistent** between single and batched records:

```python
tpl = EventTemplate(outer=EventTemplate(a=(), b=()), m=())

nr = tpl.from_vector(jnp.arange(3.0))          # NumericRecord (single)
nr.keys()     # ['outer/a', 'outer/b', 'm']    ← leaf-keyed ✓
nr["outer/a"] # works ✓

b = tpl.from_vector(jnp.arange(9.0).reshape(3, 3))   # NumericRecordArray (batched)
b.keys()      # ['outer', 'm']                 ← still TOP-LEVEL ✗
b["outer"]    # KeyError: "'outer' is a subtree…"
```

`_empirical.py` stores its samples as a batched record (`_record_data`, a
`NumericRecordArray`) and iterates `self._record_data.fields` + indexes
`self._record_data[f]` in ~15 places (`_validate_record_samples`,
`event_shapes`, `obs_shapes`, `dtypes`, `real_support`, `mean`/reductions,
`__repr__`, …). For a nested record those top-level `fields` include subtree
names, so `_record_data["outer"]` raises.

## Repro

```python
import jax.numpy as jnp
from probpipe.core.event_template import EventTemplate
from probpipe.core._empirical import RecordEmpiricalDistribution

tpl = EventTemplate(outer=EventTemplate(a=(), b=()), m=())
draws = tpl.from_vector(jnp.arange(15.0).reshape(5, 3))   # nested batched record
RecordEmpiricalDistribution(draws, name="post")            # KeyError at construction
```

## Why it wasn't caught pre-merge

[...]

PR #341, opened 2026-07-07
fix(tests): unbreak main CI — leaf-keyed nested access + xfail nested-prior NPE

jhuggins:
## Summary

Unbreaks `main`'s CI. The `bayesflow` job went red after #327 merged; every other job (the `test` matrix across 3.12/3.13/3.14, `stan`, `notebooks`, `lint & format`) is green.

## Root cause

`tests/inference/test_bayesflow_posteriors.py::TestBayesFlowMethods::test_nested_prior_end_to_end` fails with `KeyError: "'outer' is a subtree, not a field; use at_path() to navigate to it"`. Two layers:

1. **Test bug.** `_NestedLikelihood.generate_data` read the per-draw record by *subtree* (`params["outer"]["r"]`). The leaf-keyed `Record` redesign (#326/#327) now rejects subtree bracket access — you index a **leaf path** (`params["outer/r"]`) or navigate with `at_path()`. → fixed here.

2. **Deferred core migration (the real blocker).** Even with (1) fixed, NPE builds a posterior over a nested **`NumericRecordArray`**, and the *batch* record types were **not** migrated to leaf-keyed — this was explicitly deferred in #326 (the `*Array → *Batch` phase of #235). So `RecordEmpiricalDistribution` over a nested record raises at construction. Repro + analysis in **#340**.

Verified `NumericRecordArray` is still top-level-keyed while single records are leaf-keyed:
```python
nr = tpl.from_vector(jnp.arange(3.0))            # NumericRecord (single)
nr.keys()   # ['outer/a','outer/b','m']  ← leaf-keyed
b = tpl.from_vector(jnp.arange(9.0).reshape(3,3)) # NumericRecordArray (batched)
b.keys()    # ['outer','m']              ← still top-level; b["outer"] raises
```

## What this does

- Fixes `_NestedLikelihood.generate_data` to leaf-path access (correct regardless of xfail).
- `xfail`s `test_nested_prior_end_to_end` (`strict=False`) with a reason linking **#340**, so `main` goes green while the deferred batch-type migration is completed.

**Only `tests/inference/test_bayesflow_posteriors.py` changes — no core/runtime code.**

## Why xfail rather than a full fix

The genuine fix requires completing the deferred leaf-keyed migration of the batch record types and then `_empirical.py`'s `.fields` iteration — core redesign work (owner: the #326/#235 effort) that can't be done or verified safely in a CI-unblock PR (BayesFlow isn't in the local env either). #340 tracks it and the un-xfail.

## Verification

`ruff check` + `ruff format --check` clean; the file parses. The end-to-end xfail behavior is verified by this PR's own `bayesflow` CI job.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -247,7 +247,18 @@ jobs:
         # would always fail here. The floor is enforced on the full-suite
         # run below; keep partial runs exempt with --cov-fail-under=0.
         run: |
-          uv run pytest ${{ steps.changed.outputs.test_files }} --cov-fail-under=0 --cov-report=xml --cov-report=html
+          # A changed test file can be entirely skipped in this job's
+          # environment (e.g. a bayesflow/stan test when that extra isn't
+          # installed here — the dedicated extra job runs it). pytest then
+          # collects 0 tests and exits 5; treat that as a pass rather than
+          # failing the targeted run.
+          ec=0
+          uv run pytest ${{ steps.changed.outputs.test_files }} --cov-fail-under=0 --cov-report=xml --cov-report=html || ec=$?
+          if [ "$ec" = 5 ]; then
+            echo "No tests collected (all skipped in this environment); treating as pass."
+            ec=0
+          fi
+          exit "$ec"
 
       - name: Run tests (full suite)
         if: steps.changed.outputs.skip != 'true' && steps.changed.outputs.run_all == 'true'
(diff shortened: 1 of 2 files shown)


jhuggins:
Added a second commit for a CI-harness failure that surfaced here, unrelated to the test fix.

The `test (3.12/3.13/3.14)` jobs were failing (fast — ~1 min). This PR changes only `tests/inference/test_bayesflow_posteriors.py`, so the **"changed files only"** run in the `test` job targets just that file — but BayesFlow isn't installed in that job, so `importorskip` skips the whole module, pytest collects **0 tests and exits 5**, and the step fails. (The dedicated `bayesflow` job *does* install the extra and runs it — it's green here now that the nested test is `xfail`ed.)

The `ci.yml` change makes the targeted run treat **exit code 5 (no tests collected) as a pass**, so any PR touching only an extras-gated test file (bayesflow / stan) no longer red-fails the main `test` job. The full-suite run (coverage floor) and the dedicated extra jobs are unaffected. This is a pre-existing latent issue, not specific to this change.

PR #309, opened 2026-06-22
feat(core): add RegistryCatalog + SupportsRegistryCataloging adapters (Stage 2)

tcikezu:
## Summary

Stage 2 of the multiple-dispatch plan from #184; builds on #204 (Stage 1).

Lands a **discoverability layer** over every dispatch surface in the
process: a global `probpipe.registry_catalog` singleton, a
`SupportsRegistryCataloging` protocol that any registry can satisfy
(conforming or not), and small adapters wiring the existing
`ConverterRegistry` and the new `bijector_registry` facade into the
catalog. No dispatch behaviour changes; existing tests and the
tutorial notebooks that treat `list_methods() -> list[str]` as a public
surface keep working unchanged.

After this PR:

```python
>>> import probpipe
>>> print(probpipe.registry_catalog)
RegistryCatalog (
  bijectors   factory     12 method(s)  — Constraint -> Bijector factory dispatch ...
  converters  converter    4 method(s)  — Cross-type distribution converters ...
  inference   dispatch    12 method(s)  — Inference-method dispatch for condition_on.
)
>>> print(probpipe.registry_catalog.describe("inference"))
inference  (dispatch) — Inference-method dispatch for condition_on.

  Auto-dispatched (by priority):
     88  nutpie_nuts  (probpipe.inference._nutpie)
     85  blackjax_nuts  (probpipe.inference._blackjax_mcmc)
     ...
  Opt-in only (reachable via method=...):
      0  tfp_nuts  (probpipe.inference._tfp_mcmc)
     ...
```

## Migration approach — fully back-compat (Option B)

During planning we considered a breaking change to
`BaseDispatchRegistry.__init__` (require `name`) and to `list_methods()`
(return `list[MethodSummary]`). Implementation-time grep showed the
actual blast radius was wider than the original plan claimed: ~60 bare
`UnaryDispatchRegistry()` / `BinaryDispatchRegistry()` calls across
~10 test files, plus 3 user-facing tutorial notebooks that treat
`name in inference_method_registry.list_methods()` as a public membership
check — silently broken under the new shape.

Final approach: **`list_methods()` keeps returning `list[str]`** (used
by tests + tutorials + the `method=` named override). A new
`method_summaries() -> list[MethodSummary]` carries the rich
introspection the catalog needs. The constructor auto-derives
`register_in_catalog` from whether a `name` was supplied. **Zero
existing test or notebook edits required.**

## Design notes

- **Two checks, not one** — *satisfying the protocol* vs *being in the
[...]

PR #346, opened 2026-07-10
docs(design): resolve the judgement-call findings from the second design-reference review

jhuggins:
## Summary

This PR resolves the judgement-call findings from the second review of the design reference, then carries a set of follow-on design decisions that grew out of the review discussion. Every change is design prose in `design/` (regenerated into the vault copy); no code or public API changes, so no CHANGELOG entry.

The changes land as focused commits, grouped below by theme. Reviewers can veto per theme.

## Capability algebra

- Joint moment capabilities are decided **at construction**: an edge-free joint derives moments componentwise, jointly Gaussian factors promote to `FactoredMultivariateGaussian`, and any other dependent joint carries no moment capability (the Monte Carlo fallback answers).
- `d[path]` is contractual on **every** distribution via a `FieldView(Distribution)` class and `__getitem__` on the base, with a per-capability derivation table (III.7): moments project from the parent, only the density routes through per-path `_marginal`, and only `_sample` needs a sampling parent.
- The Monte Carlo moment fallback is defined per event kind (arrays average; functions → a lazy mean function; measures → a mixture mean measure); `cov`/`quantile` keep numeric-only fallbacks.
- `SupportsExpectation` is restricted to laws that integrate an arbitrary `f` exactly (finite support). The conditional capability vocabulary is closed by one mirroring rule.

## Lifting semantics

- **Layering:** the computation lift *only samples* and never consults a registry; `pushforward` builds the registry on top, and its sampling fallback *is* the lift, so `pushforward(f, d)` with no matching rule equals the lifted `f(d)`. This resolves the three previously inconsistent return contracts.
- `fixed_args=` binds a multi-parameter map's side arguments **as data**, keeping type-keyed dispatch on the map itself (rather than wrapping in `functools.partial`, which would lose its registered identity).
- Trigger: unannotated parameters expect values; `Distribution`/`Supports*` annotations pass through. Co-sampling is grouped by **root ancestor**; unrelated groups take the product law.
- Output wrapping has no error case: a mapping return becomes fields; anything else becomes one field keyed by the function's name (`OpaqueSpec` when non-array, tuples included); `output_template=` declares structure inference cannot recover.
- `include_inputs=True` nests each sampled input under its parameter name (plain arguments go to provenance); `condition_on` always binds the value, and the predictive mixture is requested only via `predictive`.

## Batch algebra

- `Batch` gains **`axis_groups`**: axes partition into ordered levels; `batch_shape` stays the flat concatenation; nesting is **views over grouped columnar storage**, no dedicated nested classes. `sample` over a `DistributionBatch` nests laws over draws; multi-level `quantile` likewise.
- Each level carries a **name** (`level_names`), auto-derived and repinned with `with_level_names` (the batch-level counterpart of `with_path_names`). **Level names are unique within a batch**: an auto-minted collision disambiguates by the smallest free integer suffix (`draw`, `draw2`, …, so nested/posterior-predictive sampling just works), while a user rename onto an existing name raises.
- Batched operations **align levels by name** (not by position); mismatched-but-corresponding levels are lined up by renaming one first, exactly as `joint()` realigns fields. `select(**levels)` indexes by name — xarray's `isel`, with `level_names` as the dimension names (no label-based `sel`, since levels carry no coordinate labels).
- `LinOp`/`LinOpBatch` apply elementwise (`matmat`-over-batches deferred); `LinearPushforwardDistribution` (VI.4) is the generic linear rule's lazy exact-moment carrier.

## Reproducibility and tracking

- `Provenance` gains `controls` (resolved key, sample count, selected method) and `inputs` (plain arguments by name, fingerprinted in tiers with `fingerprint_is_weak`).
- **Randomness rule:** no ambient RNG state. Only draw-producers (`sample`) require an explicit key — without ambient state a defaulted key would silently repeat a draw. Deterministic-contract ops (moments, marginal, inference) default a key derived **structurally** from `seed` — split along the computation graph and by index per broadcast/batch axis. That preserves common random numbers (nearby-input comparisons and reparameterization gradients stay low-variance), and because each unit's key is fixed by index rather than a shared counter, the result is identical across `jax`/`sequential`/`thread` and parallel execution races on no mutable RNG state. Operation signatures are the control surface.

## Registries and types

- `ConverterRegistry(BinaryDispatchRegistry)` keyed on `(type(source), target)` through the base's abstract key extraction; `Converter(BinaryDispatchMethod)`; `ConversionInfo(MethodInfo)` carries fidelity, realized **through priority tiers** (so priority stays the sole selection order and `min_fidelity` is a `check`-time floor). The bijector facade stays non-dispatch but is cataloged.
- Catalog vocabulary aligned with PR #309's committed rename (`EntrySummary`, `entry_count`, `entry_summaries()`/`describe_entry()`, `register_method` removed).
- Smaller clarifications: names are human labels (not identity); MRO-distance dispatch tie-break; numeric-axis re-derivation under transforms; the pytree children/aux split; the `marginal` fallback records its route; empirical atoms in native batch form.
- **`with_names` → `with_path_names`** (de-confused from `with_name`).

## Naming

The lifted-function abstraction is renamed **`WorkflowFunction` → `Computation`** (`@computation`, Part IV retitled *Computations*, file `04-computations.md`). "Workflow function" over-indexed on orchestration (off by default) and never connected to *lifting*. Kept: the C6 principle "reproducible **workflows**" and the pipeline sense — computations are the nodes, a workflow is what they compose into.

This is design-leading-code: the rename is **decided** for the target state. Sweeping the implementation and the convention docs (`STYLE_GUIDE.md`, `CONTRIBUTING.md`, the public API, tests, notebooks — ~540 token occurrences) is deliberately a separate implementation PR, sequenced after the in-flight core PRs merge to avoid mass conflicts. The README already states the standing convention that the design describes the target and contributor docs describe the code until implementation catches up.


jhuggins:
**Cluster A - the capability doctrine** (2b6e88b). Joint moment capabilities are decided at construction (edge-free joints componentwise, jointly Gaussian promoted to `FactoredMultivariateGaussian`, otherwise the Monte Carlo fallback). `d[path]` is contractual on every distribution via `FieldView`, with a per-capability derivation table in III.7: moments project from the parent, only the density needs `SupportsMarginals`, only sampling needs a sampling parent. The moment fallback is defined per event kind, `SupportsExpectation` is restricted to finite support, and the conditional capability vocabulary is closed by one mirroring rule.

jhuggins:
**Cluster B - lifting semantics** (af8fc24). The lift can only sample and never consults a registry; `pushforward` layers the registry over the lift and falls back to it exactly, with `fixed_args=` binding a map's side arguments as data so registered identity survives type-keyed dispatch. Output wrapping has no error case (a mapping return becomes fields; anything else, tuples included, becomes one function-named field, opaque when not an array), and `include_inputs=True` nests each sampled input under its parameter name, with plain arguments recorded in provenance instead.

jhuggins:
**Cluster C - batch algebra** (d5c6767). Batches gain `axis_groups` (ordered levels over flat columnar storage, presented as views), so `sample` over a `DistributionBatch` nests laws over draws and the laws/draws distinction is a type fact. Alignment is level-wise broadcast-zip with outer products by explicit reshaping. `LinOp` and `LinOpBatch` apply elementwise, with `matmat`-over-batches deferred as an open point, and VI.4 adds `LinearPushforwardDistribution` as the generic linear rule's lazy exact-moment carrier.

jhuggins:
**Cluster D - reproducibility** (3507a9b). `Provenance` now records the resolved controls and the plain inputs (fingerprinted in tiers, with `fingerprint_is_weak`); operations require explicit PRNG keys and never draw from ambient state; operation signatures are the control surface, with the options namespace reserved for wrapped user functions; a workflow-function `seed` derives the `Key` that invoked operations consume.

jhuggins:
**Registry and type clarifications** (b9b4974). Names are human labels, not identity; dispatch ties break by MRO distance within a priority, then registration order; the numeric template axis re-derives under every transform; the pytree aux is template and name only, so provenance does not cross JAX boundaries; the `marginal` fallback records its route and sample count; empirical atoms are stored in the event type's native batch form.

jhuggins:
**Converters and the catalog** (1e554d2). The converter registry is now a `BinaryDispatchRegistry` subclass keyed on `(type(source), target)` through the key extraction the base leaves abstract, with fidelity riding `ConversionInfo(MethodInfo)` and realized through priority tiers, so priority stays the sole selection order and `min_fidelity` is a `check`-time feasibility floor. The bijector facade stays outside type dispatch and is cataloged via `SupportsRegistryCataloging`. The same commit aligns II.4 with the entry rename committed on #309 (`EntrySummary`, `entry_count`, `entry_summaries()` / `describe_entry()`, `register_method` removed) - cc @tcikezu: the II.4 update you flagged there is covered here, so it needn't be done twice.

jhuggins:
**`with_path_names`** (fb2b3a8). The field-renaming method is renamed so it no longer reads as an accidental sibling of `with_name`: the `with_` prefix is kept, since the functional semantics match (a shallow copy differing only in names), and the middle term says what is renamed - nodes at paths, with bare names resolving to unique paths.

jhuggins:
Post-merge-of-decisions audit against the locked plan surfaced three fixes (391ef9e), all cross-consistency rather than new decisions:

- **III.8 / B6.** The conditioning contract still said "conditioning on a *distribution* over the given yields the predictive mixture" — the auto-detection B6 removed. V.4 was updated in af8fc24 but III.8 carried the same claim; it now matches (a value is always bound; the predictive mixture comes from `predictive()`).
- **II.4 forward reference.** The `entry` definition named converters, bijector factories, and inference methods, all defined in later sections; it now describes the term by role.
- **II.3 / C2.** Added the "nesting is unbounded in principle, two levels in practice" note from the C2 resolution.

Every other decision in the seven cluster commits verified against the plan: the FieldView derivation table and gate (A2), the per-event-kind moment fallback (A3), `SupportsExpectation` finite support (A4), the capability mirror rule (A5), the lift/pushforward layering and `fixed_args=` (B1), `include_inputs` layout (B4), output wrapping (B5), `axis_groups` and nested sampling (C2), elementwise operator application (C3/C5), provenance controls and explicit keys (D1/D2), the converter dispatch subclass and fidelity-through-priority (E1), and the `with_path_names` sweep (E8) are all present and consistent.

jhuggins:
**Named batch axis groups** (e42a731). Following the C2 nested-batch resolution, each batch level now carries a **name**:

- `Batch.level_names` gives one name per level, aligned with `axis_groups`, auto-derived by the operation that produces the level and repinned with `with_level_names` — the batch-level counterpart of `with_path_names` (a shallow copy that changes only the names, never the shapes or elements). Level names are independent of the field names within each element, so a `RecordBatch` can have a `draw` level over records with fields `mu`/`sigma` with no clash.
- Batched operations now **align levels by name** rather than by position (V.8): a level in both operands must be broadcast-compatible, a level in one broadcasts across the other, and two levels meant to correspond under different names are lined up by renaming one with `with_level_names` first — exactly the role `with_path_names` plays for `joint()`'s field realignment. Level *ordering* stays outermost-first (layout, `to_vector`), which is orthogonal to by-name matching.
- Canonical level names: `sample`'s draw level is `draw` (V.2, V.8), `quantile`'s inner level is `quantile`, and swept outer levels keep their names.

This is the follow-on to the "should the axis groups be named?" question — resolved yes, fully named, with by-name alignment as the payoff.

jhuggins:
**By-name level selection** (6af7bcd). Adds `Batch.select(**levels)`, the read-side dual of the by-name alignment — it indexes a batch along its named levels and returns a view, parallel to xarray's `.isel()` with `level_names` playing the role of xarray dimension names.

- An indexer is an int, slice, `None`, or a tuple of these over the level's axes: an int drops its axis, a slice or `None` keeps it, `None` meaning the whole axis (`:`).
- A multi-axis level takes one indexer per axis; a shorter indexer fills the leading axes and leaves the rest whole, so `draw=i` on a two-axis `draw` level is `draw=(i, None)`.
- Selecting an entire single-axis level by an integer removes it, yielding the inner batch or element exactly as positional `[]` and iteration do; unnamed levels are kept whole.
- No label-based `.sel()` counterpart: batch levels carry no coordinate labels (a `draws` level is anonymous draws, a `laws` level anonymous laws), and adding coordinates would be xarray's index machinery — out of scope.

jhuggins:
Two design refinements from a review-side discussion, both landing here since #346 has become the post-review design-refinement branch.

**Rename: `WorkflowFunction` → `Computation`** (3b854d3). "Workflow function" over-indexed on orchestration (which is off by default), read oddly for a single function, and never connected to *lifting* — the flagship it enables. Renamed to `Computation`: `@computation` decorator, class `Computation`, Part IV retitled **Computations**, file `04-workflow-functions.md` → `04-computations.md`, and the compounds (computation lift / sweep / layer). Kept deliberately: the C6 principle name "reproducible **workflows**" and the pipeline sense — computations are the nodes, a *workflow* is what they compose into.

**Randomness rule reconciled** (7e1a425). The V.0 "an operation that needs randomness raises without a key" overshot: an operation that only falls back to Monte Carlo *under the hood* (a moment, a marginal, an inference result) is deterministic by contract, so requiring a key there burdened the caller and leaked the closed-form-vs-MC distinction into the signature (against C3/D3).

- **V.0:** only draw-producers (`sample`) require an explicit key — without ambient state a defaulted key would silently *repeat* a draw. Deterministic-contract ops take the key as an optional control defaulting to the `seed`-derived one; the resolved key is always recorded in `provenance`. No ambient state (the real principle) is kept.
- **IV.3:** the `seed` derives keys **structurally** — split along the computation graph and by index per broadcast/batch axis, **never folding in input values**. That preserves common random numbers: perturbing an input reuses the same keys, so nearby-input comparisons and reparameterization gradients stay low-variance instead of being swamped by independent sampling noise. Vary the `seed` for an independent stream. (This corrects an earlier idea of hashing inputs into the key, which would have broken exactly this.)
- **IV.4:** because each unit's key is fixed by its index rather than a shared counter, the result is identical across `jax`/`sequential`/`thread` and parallel execution contends for no mutable RNG state — nothing locked, no key reused. That's the parallelization-safety guarantee.
- **V.2:** `sample` splits its key by draw index; inside a `Computation` the `seed` supplies it by the same split, so keys are never hand-threaded.

Roadmap W2.4/W7.5/W7.6 and the D-cluster row updated to match.

arob5:
## Design review

A well-executed set of decisions — the capability doctrine, the lift/pushforward layering, and converters-as-dispatch-subclasses are all internally coherent and clearly argued, and the mechanical renames (`with_names`→`with_path_names`, workflow-function→computation) swept cleanly through `design/`. The main issues are in the **newly specified batch-level machinery** (one reachable self-contradiction) and in **drift between this PR's vocabulary and everything outside `design/`** — the PR body and the convention docs.

### Critical — spec is undefined in a reachable case

**1. Level names must be unique, but nothing says so, and nested sampling collides.** Two new features key on level names: alignment "matches the operands' levels **by name** … there is no positional fallback" (`05-operations.md` V.8) and `select(**levels)` targets levels by name via kwargs (`02-shared-abstractions.md` II.3). Both are ill-defined if two levels share a name — and `select`'s `**levels` *literally cannot address* a duplicate (Python kwargs forbid repeated keys). Yet, unlike field names ("Factor names are unique across the list, and a duplicate is an error") and sibling field keys ("Sibling uniqueness applies…"), **level names carry no uniqueness rule**. And the design produces a collision on a normal path:

- `sample(d, sample_shape=(N,))` → batch with a level named `draw` (V.2)
- feed those draws through `given=` → `DistributionBatch` carrying that `draw` level (V.8)
- `sample(that_batch, sample_shape=(S,))` → appends **another** `draw` level (`level_names == (*d_batch.level_names, "draw")`, V.8) → two `draw` levels, unaddressable and ambiguous under by-name alignment.

Needs (a) an explicit uniqueness invariant on `level_names`, and (b) a rule for when an operation would mint a duplicate (auto-suffix? raise? require rename first?).

### Recommended

**2. The PR body is stale — it predates ~half the branch.** The body documents 7 commits; the branch has 13. The 6 later commits include the single biggest change in the diff. The body:
- never mentions the **`WorkflowFunction`→`Computation` rename** (a file delete+add, plus README and Part V rewrites) — it still says "workflow-function lift" throughout §Lifting semantics;
- under §Batch algebra mentions only `axis_groups`, omitting `level_names`, `with_level_names`, and `select()`;
- describes V.8 alignment as "level-wise broadcast-zip, outermost level first" — but the final design aligns **by name** with no positional fallback (commit `e42a731` superseded this).

A reviewer reading the body reviews a design that no longer exists; it should be regenerated against the final diff.

**3. The `Computation` rename creates a vocabulary split with everything outside `design/`.** `STYLE_GUIDE.md` (§1.4 "Standalone workflow functions", `@workflow_function`, the `WorkflowFunction` public-class rule) and `CONTRIBUTING.md` (orchestration section, output-wrap principle, object dictionary) still say `WorkflowFunction`/`@workflow_function` throughout — as does the public API. Design leading code is normal here, so this isn't "stale STYLE_GUIDE." But the PR should state intent: is the rename decided (→ tracking issue to sweep code + convention docs on implementation) or provisional? Unmarked, a user hits `Computation` in the design and `WorkflowFunction` in the style guide and API with nothing connecting them.

**4. The default-seed source for a standalone Monte Carlo fallback is undefined.** V.0 says `sample` "requires an explicit PRNG `key` and raises without one," but a moment/marginal falling back to MC "takes the key as an optional control defaulting to one derived from the `seed`." Inside a `Computation`, `seed` is defined — fine. But a bare `mean(d)` on a sample-only distribution *outside* any computation has no defined `seed`. Either there's an implicit constant default (then say so, and note repeated calls return the identical estimate — the same "silent repeat" the `sample` rule prevents, here tolerated because the contract is deterministic) or it's unspecified. One sentence resolves it.

**5. The doctrine's central claim — "`isinstance` stays honest" — has no stated mechanism for per-instance capabilities.** The doctrine rests on "support decidable at construction is expressed as structure, so `isinstance` stays honest." But `FieldView` is one class whose supported capabilities vary with its parent instance and path (III.7 table), and an edge-free `FactoredDistribution` advertises moment capabilities a dependent one doesn't — same class, different instances. With `@runtime_checkable` protocols, `isinstance` checks *method presence*, which is per-class, so a single class either always or never satisfies `SupportsMean`. The doc asserts honesty without saying how one class presents per-instance capability sets (dynamic class synthesis? per-instance protocol registration?). May already be solved in the implementation, but the doctrine commit is the right place to name the mechanism in a sentence.

### Minor

**6. `pushforward` + `fixed_args`: the "exactly one" remaining parameter isn't enforced in prose.** What happens when `fixed_args` leaves *two* parameters unbound (which does `d` map to, or is it an error)? One clause.

**7. Edge-free joint moments should state the intersection requirement.** "derives its moments componentwise from its factors" is only available when *each* component supports that moment, the way sampling/log-prob explicitly state "the intersection of the factors'." Making it explicit parallels the sampling rule.

**8. PR body process jargon** ("second review of #332," "judgement calls," "@tcikezu flagged this on #309") — fine for this audience, low priority.

### Solid, no action
- The `with_name` / `with_path_names` / `with_level_names` trio is coherent; the `fb2b3a8` rationale for de-confusing `with_names`↔`with_name` is sound.
- The view capability-derivation table is mathematically correct row by row (projection for mean, `PΣPᵀ` sub-block for cov, `f∘π` composition for expectation, `condition_on(s)[p∖s]` commuting conditioning with marginalization).
- The lift/`pushforward` layering (V.6: "the sampling fallback *is* the lift") cleanly resolves the three previously inconsistent return contracts; `fixed_args` avoiding `functools.partial` to preserve type-keyed dispatch is right.
- `ConverterRegistry(BinaryDispatchRegistry)` keying on `(type(source), target-value)` is consistent with II.4's "arity subclasses supply key extraction."
- No CHANGELOG / `kind:breaking-change` needed — design prose, not code or public API.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

jhuggins:
Thanks — thorough review. Resolutions below.

**1. Level-name uniqueness + nested-`draw` collision (critical) — fixed (`0132674`).** Confirmed reachable on the posterior-predictive path (sample → `given=` → sample). Added the invariant that **level names are unique within a batch**, resolved by the same auto-vs-user split the rest of the design uses:
- an operation that would mint a colliding level disambiguates deterministically with the smallest free integer suffix (`draw`, then `draw2`), so nested/posterior-predictive sampling just works and stays reproducible;
- a `with_level_names` rename onto a name already present raises, as field and factor renames do.

Chose auto-suffix over raise-and-rename precisely because the collision is on a *common* path — raising there would be hostile, and `draw`/`draw2` is interpretable (outer = posterior draws, inner = predictive draws). xarray errors on duplicate dims, but its dims are user-supplied; ours are auto-minted, so self-disambiguation is the right divergence. II.3 and V.8 updated.

**2. Stale PR body — fixed.** Regenerated against all commits: it now covers the `Computation` rename, `level_names`/`with_level_names`/`select`, **by-name** alignment (the positional description was superseded by `e42a731`), the randomness reconciliation, and the uniqueness rule.

**6. `fixed_args` arity — fixed (`2160704`).** V.6 now states `fixed_args` must leave exactly one parameter free (the one `d` maps to); leaving two unbound is an error.

**7. Edge-free joint moment intersection — fixed (`2160704`).** III.10 now says an edge-free joint has a given moment exactly when every factor does — the intersection rule sampling and log-prob already state.

**3. `Computation` vocabulary split — no design change; tracked separately.** This is design-leading-code, which the README already frames as standing convention (the design describes the target; contributor docs describe the code until implementation catches up). The rename is **decided**, and the code + `STYLE_GUIDE`/`CONTRIBUTING`/API/tests/notebooks sweep (~540 occurrences) is deliberately its own implementation PR, sequenced after the in-flight core PRs so it doesn't mass-conflict. I added a paragraph to the PR body stating this intent.

**4. Default seed for a standalone MC fallback — not a gap.** There is no "outside a computation": every operation *is* a computation (Part V), so `mean(d)` carries the `seed` control with its default like any other, and the reconciled rule applies uniformly. Repeated calls returning the identical estimate is the intended deterministic-contract behavior, already implied by the reconciliation. No separate case to specify.

**5. Per-instance capability mechanism — left to implementation.** The doctrine states a behavioral guarantee (capabilities decided at construction, so `isinstance` reflects true support); *how* a class presents per-instance capability sets — most-specific-class selection, which the factored path already invokes — is an implementation concern that C3 keeps out of the design prose. Noting it here so the intent is on record.

**8. Process jargon — subsumed** by the body regeneration in #2.

PR #347, opened 2026-07-10
feat(discrepancies): kl_divergence via a binary dispatch registry (Stage 4)

tcikezu:
## Stage 4 — `kl_divergence` via a binary dispatch registry

Draft. **Stacked PR** — base is `dev/registry-catalog` (Stage 2 + #309), not `main`. It cannot merge until that base lands; opening it now so there's a concrete, runnable example of a `BinaryDispatchRegistry` in the repo to look at alongside the freshly-merged design reference (#332).

### What this adds
- `probpipe/discrepancies/` — a new subpackage for distribution-to-distribution discrepancies.
- `kl_registry`, a `BinaryDispatchRegistry` keyed on `(type(p), type(q))` with three methods:
  | Prio | Method | Applies when |
  |---|---|---|
  | 90 | `kl_normal_normal` | both `Normal` (exact, closed-form) |
  | 70 | `kl_tfp` | both wrap a TFP dist with a registered closed-form KL |
  | 5 | `kl_mc` | `p` samples + has `log_prob`, `q` has `log_prob` (MC fallback) |
- `kl_divergence(p, q, *, method=None, **kwargs)` public op in `probpipe.core.ops`.

### Notes for reviewers
- Independent of Stage 3 (dispatch-conventions guide): the docstring reference to `docs/contributor/dispatch-conventions.md` was removed so this branch is self-contained.
- Tests: `tests/discrepancies/test_kl_divergence.py` — 11 passed locally.


tcikezu:
Discrepancy is a binary operation on pairs of distributions. I am unsure if _kl_registry.py ought to belong in a discrepancies module. Also would like some thought on where/whether to place discrepancy in the design docs @arob5

Issue #352, opened 2026-07-12
feat(core): wire fingerprints into Prefect cache_key_fn for cross-run task caching


erozercan:
### Summary

Wire ProbPipe's `fingerprint()` utility (PR2, #318) into Prefect task caching so that a workflow step is "skipped and its result reused" when the exact same computation: same user-function body and same inputs: has already run, including across separate processes. This is PR3 of the provenance/caching series (PR1 #244 `ProvenanceMode`, PR2 #318 `fingerprint()`). 

### Motivation

Today Prefect is used only as a parallel executor: `@task` carries a `name` and nothing else — no caching, no result persistence. So re-running a pipeline recomputes everything from scratch, and a mid-pipeline failure loses all progress. The original design discussion asked for three behaviors Prefect can provide but we don't use:

- **Re-run with same inputs** → skip unchanged steps.
- **Failure mid-pipeline** → resume from the failed step (cached upstream results).
- **Change only the final step** → re-run only that step.

PR2 built the stable content hash these need. PR3 consumes it.

## Current architecture (where the work lands)

- The Prefect `@task` is created in `probpipe/core/_workflow_execution.py::map_task()` — **not** in `node.py`. Every Prefect path (`prefect_task`, `prefect_flow`) funnels through `map_task`, so it is the single chokepoint for `cache_key_fn` + `persist_result`.
- `node.py::_make_execution_config()` builds the `WorkflowExecutionConfig` handed down to `map_task`, and has access to the `WorkflowFunction` itself.
- `Provenance.create()` (`provenance.py`) currently computes `fingerprint()` for every parent on every call, even though nothing consumes it yet.

## Two design decisions taken up front

1. **Caching is opt-in**, via explicit global config on `PrefectConfig` (mirrors `WorkflowKind`), default OFF.
2. **Fingerprinting is gated** — `Provenance.create()` only computes fingerprints when caching is active, resolving the open question deferred from PR2.

## Key constraints

- `fingerprint()` cannot stably hash a *raw* Python function (it falls to a `repr()` that embeds a memory address → process-dependent). The function fingerprint must be computed at the `WorkflowFunction` level in `node.py` (where the correct code-hashing path lives) and threaded down into the config — computed once, not per row.
- Prefect's built-in `cache_policy=INPUTS` is unsuitable: it hashes the identical `run_func` wrapper source (missing user-body changes) and pickles JAX arrays unreliably. A **custom `cache_key_fn`** is required.
[...]

PR #359, opened 2026-07-13
feat(core): store native-form leaves in NumericRecord; convert lazily at the compute boundary

### LINKED ISSUE #355: Store native-form leaves in NumericRecord; convert lazily at the compute boundary
## Summary

`NumericRecord` currently coerces every leaf to a `jax.Array` **eagerly at
construction**, capturing the backend metadata (`xarray` dims/coords/attrs,
`pandas` index/columns/dtypes) in a private aux side-channel so `to_native()`
can reconstruct the original object. This issue proposes inverting that:
**store each leaf in its native form** and convert to `jax.Array` **lazily, at
the compute boundary** (pytree flatten for `jit`/`vmap`/`grad`, `to_vector`,
the single-field scalar shim), with a set-once per-leaf cache.

One access semantic is already decided: `record["x"]` (and every other
navigation view — `children`, `at_path`) returns the **native leaf verbatim,
with no conversion** — views stay views (`D7 – Single source of truth`);
conversion belongs to the operations that need arrays.

## Motivation

- **`C3` / `D7`.** The design principles say representations are computational
  detail, converted as needed, and every view is derived on demand from one
  authoritative store (`design/01-design-principles.md`). Eager coercion does
  the opposite: it converts up front and keeps the authoritative (native) form
  only as a reconstruction recipe. `design/03` III.5's `LinOp` already applies
  the lazy philosophy to matrices ("without materializing them"); this extends
  it to data.
- **Disk-backed / lazy data.** Today `Record("data", X=<dask-backed
  DataArray>).to_numeric()` forces the whole array into device memory at
  construction. Native leaves defer materialization to first compute use —
  data that lives on disk or is lazily loaded stays there until an operation
  actually needs the values.
- **The aux machinery is a carrying cost.** Because coercion separates data
  from metadata, the metadata must be captured, stored, and threaded through
  every path that rebuilds a record. PR #338 had to add exactly that threading
  twice (issue #353): identity-keyed aux re-attachment across the structural
  transforms (`without` / `merge` / `replace` / `with_path_names`) and
  pickling through the native form (the aux hooks are closures, so the blob
  cannot go in the pickle payload). Native-form leaves dissolve the whole
  problem: metadata is never separated from data, so there is nothing to
  capture, restore, or lose — `probpipe/core/_array_backend.py` (the
  `AuxHooks` registry), `NumericRecord._aux`, the aux-carrying transform
  overrides, and the aux-aware `__reduce__` all retire.
- **The promotion-exclusion wart dissolves.** Today a record holding an
  all-numeric `xarray` leaf must stay a *plain* `Record` (the promotion probe
  rejects backend-typed leaves) purely so eager coercion does not silently
  destroy the native form (`design/03` III.3, `record.py` `__new__`). Under
  native storage, numeric data in a native container can simply *be* a
  `NumericRecord` — the more faithful typing (`D1`).

## Proposed design

- `NumericRecord` validates leaves as numeric **without converting** (the
  existing probe already reads `.dtype` / `.shape` attributes only) and stores
  them verbatim.
- Conversion to `jax.Array` happens at the compute boundary: pytree flatten,
  `to_vector`, the scalar shim, and any numeric kernel entry point. A
  set-once per-leaf cache (the functional `object.__setattr__` pattern already
  used elsewhere) prevents repeated conversion; memory then holds native +
  cached array only after first compute touch — the mirror of today's
  array + metadata blob, with the native side authoritative.
- `ArraySpec` and the template layer stay data-free: the template describes
  structure and never depends on the value layer. `ArraySpec.is_valid` can
  already validate a native leaf via its exposed shape/dtype.
- A JAX round-trip still returns bare `jax.Array` leaves — unflatten cannot
  rebuild native containers (coords are non-hashable and cannot live in the
  static treedef aux without breaking `jit`). This matches the current
  documented boundary behavior (`design/03` III.3 Notes).
- `to_native()` becomes trivial (leaves already native); `to_numeric()`
  validates rather than coerces.

## What retires

- `probpipe/core/_array_backend.py` (`AuxHooks` / `register_aux` / `aux_for`).
- `NumericRecord._aux`, `aux`, the aux-carrying overrides of
  `without` / `merge` / `replace` / `with_path_names`, and the native-form
  `__reduce__` branch.
- The backend-leaf exclusion clause in the `Record.__new__` promotion probe
  (and its sentence in `design/03` III.3).

The regression tests added with the aux work
(`TestBackendAuxSurvivesStructuralTransforms` in
`tests/core/test_array_backend.py`, `TestNumericRecordAuxPickle` in
`tests/core/test_record_serialization.py`) pin **observable** contracts —
metadata survives structural transforms and pickling — that the native model
satisfies trivially, so they transfer as the migration's safety net (adapted
away from `.aux` internals where they touch them).

## Open decisions

- **Aliasing / mutability.** Native objects are mutable, and eager
  `jnp.asarray` currently snapshots values at construction. Verbatim storage
  means an external in-place mutation of a passed-in DataFrame reaches the
  record — a plain `Record` already accepts this for opaque leaves, but it
  weakens the spirit of `C2` for numeric data. Options: document (match plain
  `Record`), or shallow-copy where cheap (xarray/dask metadata copies are
  cheap; a pandas deep copy is not).
- **Equality / hashing / fingerprints** on disk-backed leaves force
  materialization when computed — presumably acceptable (on demand), but
  worth stating.
- **Batches.** Native columns raise stacking/alignment questions
  (`RecordArray` today rejects backend leaves per element). Defer to the
  `*Array` → `Batch` rework; batches can stay eagerly coerced initially.

## Scope

A rewrite of `probpipe/core/_numeric_record.py` internals plus the
retirements above — deliberately **after** PR #338 merges. The aux-threading
fixes shipped there are the correct bridge under the current architecture and
keep the observable contract this change preserves. `design/03` III.3 then
gains the two-sentence native-form contract in place of a metadata-survival
note ("leaves are stored in native form; conversion to `jax.Array` happens at
the compute boundary, and a JAX round-trip returns bare arrays").

## Acceptance criteria

- [ ] `NumericRecord` stores native leaves verbatim; `record["x"]` /
  `children` / `at_path` return them unconverted.
- [ ] `jit` / `vmap` / `grad`, `to_vector`, and the scalar shim work
  unchanged, converting via the cached boundary.
- [ ] A dask-backed / lazily-loaded leaf is not materialized at construction
  (pinned by a test that fails if construction touches values).
- [ ] All-numeric records holding native leaves auto-promote to
  `NumericRecord` (exclusion clause removed), with the equality/hashing
  semantics stated.
- [ ] The existing transform/pickle metadata-survival tests pass under the
  new storage model.
- [ ] `design/02` / `design/03` updated (promotion rule, backend note);
  CHANGELOG entry for the behavior changes.

## References

- Current contract and its costs: `design/03` III.1/III.3, PR #338
  (NamedTree/Record alignment; documents the promotion exclusion and the
  pytree-boundary rule), issue #353 (the aux-survival fixes this proposal
  dissolves).
- Design principles: `C2`, `C3`, `D1`, `D7` in `design/01`; `LinOp` laziness
  precedent in `design/03` III.5.

jhuggins:
# Summary

Closes #355. `NumericRecord` now stores each leaf **in its native form** — a `jax` / `numpy` array, an `xarray.DataArray`, a `pandas` object, or any registered array backend — and converts to `jax.Array` **lazily at the compute boundary**, with a set-once per-leaf cache. Builds on #338 (now merged; base is `main`), whose record substrate this rewrites.

## The storage inversion

Construction **validates without converting**: leaves are checked from container metadata only (shape / dtype / registered-backend hooks) and stored verbatim, so a lazy or disk-backed value is not materialised at construction. A bare Python scalar — which carries no metadata — normalises to a 0-d `jax.Array`. Navigation (`record["x"]` / `children` / `at_path`) returns native leaves unconverted; conversion happens at the compute boundary — the JAX pytree flatten that `jit` / `vmap` / `grad` traverse, `to_vector`, and the single-field scalar shim — through a set-once cache, so each leaf materialises at most once.

Consequences that fall out of data and metadata never separating:

- **Metadata survives everything except a JAX transform.** Structural transforms (`without` / `merge` / `replace` / `with_path_names`) and pickling reuse the native leaves verbatim at every nesting level. A JAX round-trip still returns bare-array leaves — unflatten cannot rebuild native containers (unhashable coords cannot ride the static treedef aux without breaking `jit`), matching the design reference's pytree note.
- **`NumericRecord.to_native()` is removed** — under native storage every invocation was a no-op (leaves are already native; navigation is the export), so the method reified a conversion that no longer exists. `to_numeric()` stays: it does real work on a plain `Record` (validation) and is the identity on a `NumericRecord` (idempotence).
- **The promotion exclusion dissolves.** All-numeric records holding native containers auto-promote to `NumericRecord`, `EventTemplate.infer_from` infers `ArraySpec` for them, and `ArraySpec.is_valid` accepts them — the record's class, its template's numericness, and per-leaf validation all agree.

## The `ArrayBackend` registry (replacing the aux capture/restore registry)

`AuxHooks` bundled two roles: metadata **capture/restore** (dissolved — nothing is separated, so nothing needs restoring) and backend **recognition/conversion** (still needed, relocated). The second survives as a slim, type-keyed registry in the same module:

```python
ArrayBackend(event_shape=..., numpy_dtype=..., is_numeric=..., to_jax=..., to_numpy=...)
register_array_backend(leaf_type, backend)
array_backend_for(obj)
```

Resolution is **registry-first, duck-typing fallback**: numpy-protocol containers with numpy dtypes (xarray, `pandas.Series`, zarr, dask arrays) need no registration. Registration exists for containers the duck path cannot see or convert — non-numpy dtypes (a torch-style tensor), device tensors wanting custom conversion, frames. The built-in registration is `pandas.DataFrame` (per-column `.dtypes`, no single `.dtype`). One registration makes a type recognised, validated, promoted, converted — including at the eager batch-stacking boundary — and content-fingerprinted everywhere at once; the end-to-end test registers a fake tensor whose dtype the duck path cannot read and asserts all of it.

`fingerprint()` gains a native-container tier (concrete type + materialised shape/dtype/bytes via `to_numpy`) replacing the process-dependent `repr` fallback — cross-process cache keys over native-leaf records stay stable. Container metadata (coords / index) is documented as outside the digest.

Batch hooks (a per-backend native `stack` / `element`) deliberately await the `*Array` → `Batch` rework; `ArrayBackend`'s keyword-only fields keep that addition non-breaking for existing registrations. Until then the batch layer stays eagerly coerced, now routing per-element conversion through `to_jax` so registered backends batch correctly under the eager model.

## Design decisions

1. **Aliasing / mutability: documented, no defensive copies.** Native leaves are stored by reference, matching plain `Record` opaque leaves; mutating a passed-in container mid-pipeline is user error. The class docstring also names the corollary: mutation after a leaf has crossed a compute boundary desynchronises the cached conversion from the native view.
2. **Equality / hashing / fingerprints materialise on demand** — computing them on a record with lazy leaves forces materialisation; stated on the class and in the fingerprint tier list.
3. **Batches defer native columns** to the batch rework (the hook semantics — xarray coord alignment, pandas index reordering, element views — are that rework's open questions, and III.4's "pytree of arrays" definition of `NumericRecordBatch` would need amending); the registry's conversion is adopted by the eager batch boundary now.

## Breaking changes

- `AuxHooks` / `register_aux` / `aux_for` and `NumericRecord.aux` are **removed**; `ArrayBackend` / `register_array_backend` / `array_backend_for` replace them (a one-line migration for registrants).
- `record["x"]` on a `NumericRecord` returns the **native leaf** (previously always a `jax.Array`).
- All-numeric records holding backend containers now **auto-promote** (previously they stayed plain `Record` by documented exclusion), and `infer_from` classes them numeric.
- `NumericRecord.to_native()` is **removed** (previously rebuilt a plain `Record`; under native storage it had no non-identity case). Read fields directly for native containers.

## Tests

The metadata-survival contracts pinned by the previous aux suites carry over, re-expressed against native storage (`tests/core/test_array_backend.py`, rewritten): transform/pickle survival asserts native **types**, not `==` (equality passes even when fidelity is lost). New pins: **no materialisation** at construction/navigation/inference/transforms (an instrumented lazy leaf whose `__array__` counts); the **set-once cache** (one conversion across repeated `to_vector` and flatten); registry lookup/MRO/overwrite with a snapshot-restore fixture; DataFrame recognition (homogeneous vs mixed dtypes, per-instance numericness); the **end-to-end registration** test; and the eager batch boundary through a registered converter. Existing suites updated where they pinned the old coercion contract (leaves now verbatim; backend records promote; identity `map` preserves natives — an upgrade the old suite asserted the absence of).

## Verification

- Full local suite green: `tests/core` (1740), plus distributions / record / converters / inference / modeling / diagnostics / validation / ci / linalg.
- `ruff format` / `ruff check` clean at the CI-pinned version.
- `design/03` III.3 rewritten to the native-form contract; API docs (`records.md`, `extending.md`, `index.md`, `converters.md`), `CONTRIBUTING.md` (key-abstractions row + registry section), and the records user-guide notebook updated; CHANGELOG entry under Unreleased → Changed.

- [x] Contract of every touched abstraction was clear before coding (the three design decisions above were settled up front; the registry-first resolution order and the frames coherence rule are stated where they live).
- [x] New/changed contracts fully documented in docstrings (`NumericRecord` storage/boundary/aliasing/equality sections, `ArrayBackend` hooks, `to_jax_array`, the fingerprint tier).
- [x] Docstring openings lead with the API/contract; rationale deferred.
- [x] Code verified to obey the documented contract; tests assert it (native verbatim storage, no-materialisation, cache-once, boundary conversion, transform/pickle survival by type, promotion coherence, registration end-to-end).
- [x] Variable names consistent with the canonical table.
- [x] `ruff format` and `ruff check` clean (CI-pinned 0.15.16).


probpipe/core/_fingerprint.py
```diff
@@ -163,6 +167,17 @@ def _update(h: hashlib._Hash, obj: Any, depth: int, max_array_bytes: int | None)
             _update(h, v, depth + 1, max_array_bytes)
     elif _is_tfp_object(obj):
         _update_tfp_object(h, obj, depth, max_array_bytes)
+    elif (content := _numeric_container_to_numpy(obj)) is not None:
```
SoymilkWinsAgain, line 171: This method only hashes the container type and the converted shape/dtype/bytes, explicitly ignoring coords/index, and also effectively ignoring dims, attrs, name, columns, etc. Since this metadata is now observable input to the Record, and the fingerprint is defined as a complete content identifier and future cache key, this will cause different inputs to have the same traceability identity, leading to incorrect reuse of results after caching is enabled. 

(AI report: In actual testing, two `xarray.DataArray`s with identical values ​​and coordinates `[0, 1]` and `[10, 20]` respectively yielded the same Record fingerprint, but when users calculated the coordinates using the native leaf array, they obtained `1` and `30` respectively.)

probpipe/core/record.py
SoymilkWinsAgain: This line directly calls `jnp.asarray`. For types that register `ArrayBackend` precisely because they don't support the NumPy protocol, a fallback to the object's own `!=` value will occur if the conversion fails. 

(The AI ​​report tested two registered `Box` leaf objects with identical content; their `to_vector()` values ​​were both [1, 2], but the corresponding `NumericRecord` comparison returned `False`.)

SoymilkWinsAgain, line 922: As below, the registration backend is bypassed. `np.asarray` is executed directly on the leaves. The same registered Box example returns a 0-dimensional object array containing Box objects, instead of the `[1, 2]` provided by the backend `to_numpy`. Registered numerical leaves should also use the registry-first path in this public transformation interface.

probpipe/core/_array_backend.py
```diff
@@ -1,246 +1,231 @@
...
-_register_pandas()
+    def _is_numeric(df: pd.DataFrame) -> bool:
+        # Lazy import: this module is imported by event_template, which owns
+        # the shared dtype predicate — resolve it at call time, post-load.
+        from .event_template import _is_numeric_dtype
+
+        dtypes = df.dtypes
+        return len(dtypes) > 0 and all(_is_numeric_dtype(d) for d in dtypes)
+
+    def _numpy_dtype(df: pd.DataFrame) -> np.dtype | None:
```
SoymilkWinsAgain: This function has a bug: calling `np.dtype(...)` on `Int64Dtype()`/`Float64Dtype()` will throw a `TypeError`. These DataFrames are first accepted by `is_numeric` and automatically promoted to `NumericRecord`, and then accessing `record.dtype` or using dtype-pinned `ArraySpec.is_valid` fails.

probpipe/core/_numeric_record.py
```diff
@@ -217,45 +234,45 @@ def __init__(
...
-                total += int(val.size)
+                backend = array_backend_for(val)
+                shape = backend.event_shape(val) if backend is not None else val.shape
+                total += int(prod(shape))
         object.__setattr__(self, "_vector_size", total)
-        # Aux is ``None`` if no field had a registered hook — keeps the
-        # common all-jax case allocation-free and lets ``to_native``
-        # short-circuit.
-        object.__setattr__(self, "_aux", aux if aux else None)
+        # The set-once converted-leaf cache backing the compute boundary.
```
SoymilkWinsAgain, line 259: This violates the promise of "materializing each leaf at most once": a new cache is created with each construction. In practice, the materialization count after `to_vector()` on the original record is 1, but when `record.without(...).to_vector()` is used again to materialize the same original leaf, the count becomes 2. For disk or remote backends, this negates the main benefits of set-once caching. Either propagate the cache to unmodified leaves, or explicitly narrow the public promise to "once per Record instance".

probpipe/core/record.py
SoymilkWinsAgain: I think this need to update

probpipe/core/_numeric_record.py
```diff
@@ -41,91 +42,107 @@
...
     ----------------------------------
-    Coercing a field to a bare ``jax.Array`` discards any metadata its original
-    backend carried — an ``xarray`` array's dimensions, coordinates, and
-    attributes, or a ``pandas`` object's index, columns, and dtypes. That
-    metadata is captured at construction, stored on the record keyed by field,
-    and readable through :attr:`aux`. :meth:`to_native` reverses the conversion,
-    restoring each field to its original backend type. Because restored ``xarray``
-    or ``pandas`` leaves are no longer ``jax.Array``\\ s, :meth:`to_native`
-    returns a plain :class:`Record` rather than a ``NumericRecord``.
+    :meth:`~Record.__eq__`, :meth:`~Record.__hash__`, and content
```
SoymilkWinsAgain: In practice, a record hash is a structured hash.

SoymilkWinsAgain (changes requested):

probpipe/core/_fingerprint.py
jhuggins: Fixed in 97bd213, and this drove the central decision of the round. You are right: under native storage the metadata is observable input, so a content identifier that ignores it gives different inputs the same identity — a cache-collision risk once #352 lands. `ArrayBackend` gains a `metadata` hook (coords / dims / attrs / name for xarray; index / columns / dtypes for pandas), xarray/pandas are now registered built-ins that supply it, and the `container:` fingerprint tier folds it in. Your exact repro now differs: two `DataArray`s with coords `[0,1]` vs `[10,20]` fingerprint differently and compare unequal. `Record.__eq__` was made metadata-aware in the same commit for consistency.

probpipe/core/record.py
jhuggins: Fixed in 97bd213. `__eq__` now compares numeric leaves through `_to_jax_array` (registry-first, then `jnp.asarray`), so a registered non-numpy backend is compared by its converted values, not object `!=`. Your `Box` example — two leaves with `to_vector() == [1,2]` — now compares equal. It also compares the native-container metadata (per the fingerprint thread above), so equal-value/different-coords records are unequal. Regression test: `TestRegisteredBackendIdentity::test_eq_routes_through_registry`.

jhuggins: Fixed in 97bd213. `to_numpy()` now resolves `array_backend_for(val).to_numpy(val)` first for a registered leaf, falling back to `np.asarray` only for the duck path. The `Box` leaf comes back as `[1, 2]` (its backend `to_numpy`), not a 0-d object array. Test: `test_to_numpy_routes_through_registry`.

probpipe/core/_array_backend.py
jhuggins: Fixed in 97bd213 at the gate, not the symptom. `np.dtype(Int64Dtype())` raising is the tell that a pandas nullable/masked dtype is not a dense numpy dtype — so `_is_numeric_dtype` now returns `False` for anything not `np.dtype`-coercible, and such a frame no longer promotes to `NumericRecord` (it stays a plain `Record`). That closes the crash on `record.dtype` / dtype-pinned `is_valid` / `to_vector` at once, and is the right semantics (NA/masking is not a dense numeric array). `numpy_dtype` is also made defensive (`_safe_numpy_dtype`). Tests: `TestNullableDtypeNotNumeric`.

probpipe/core/_numeric_record.py
jhuggins: Addressed in 97bd213 by narrowing the promise (your first option). The docstring now reads "each leaf materialises at most once **per record instance**"; a structural transform rebuilds through `__init__` with a fresh cache. Propagating the cache across transforms would mean reintroducing exactly the per-transform `NumericRecord`-level overrides this PR deleted (the old `_carry_aux` machinery), so it is deliberately left as a possible future perf item rather than done here — the honest scope is per-instance.

probpipe/core/record.py
jhuggins: Fixed in 97bd213 — the module-docstring one-liner now reads "all fields are numeric, stored in native form (converted to JAX arrays lazily, at the compute boundary)" rather than "all fields are JAX arrays".

probpipe/core/_numeric_record.py
jhuggins: Right — fixed in 97bd213. `Record.__hash__` no longer calls `jnp.asarray`; it is a coarse structural hash over shape + canonicalized dtype (read from container metadata), so hashing a lazy / disk-backed leaf does not materialise it. The dtype is canonicalized the way a boundary conversion would be, so it stays consistent with the (now metadata-aware) `__eq__`; records differing only in values or coords collide in hash, which is legal. The "Equality, hashing, and lazy leaves" docstring is corrected to say only equality and fingerprint materialise.

jhuggins:
Thanks for the thorough review — all seven addressed in 97bd213, each with a regression test; replies inline. The unifying fix: the registry-first resolution used at construction / `to_vector` / the fingerprint value tier was **not** applied to `__eq__`, `__hash__`, and `to_numpy`, which still assumed the numpy protocol. Those three now route through the array-backend registry.

**Correctness (bugs):**
- `__eq__` and `to_numpy()` route through the registry — a registered non-numpy backend (your `Box`) now compares equal and materialises through its own `to_numpy`.
- pandas nullable/masked dtypes (`Int64Dtype`) are no longer classed numeric (not `np.dtype`-coercible), so such frames stay plain `Record`s instead of promoting then crashing — fixed at the shared gate.

**Native metadata is now part of identity (the design call):** `ArrayBackend` gains a `metadata` hook; xarray/pandas are registered built-ins that supply coords / index / dims / attrs; `__eq__` and `fingerprint()` fold it in. Two records with equal values but different coords are now unequal and fingerprint differently. The two fingerprint tests that previously asserted coords-are-excluded are flipped to assert inclusion.

**Contract/docs:** `__hash__` is now a coarse structural hash (shape + dtype, no values or metadata) that does not materialise a lazy leaf, consistent with the coord-sensitive `__eq__` via legal collisions; the cache promise is scoped to "per record instance"; the stale `NumericRecord` one-liner and hashing note are corrected.

Full `tests/core` (1755) + distributions / record / converters / inference / modeling / diagnostics / validation green; ruff clean at the pinned version.

probpipe/core/_numeric_record.py
```diff
@@ -41,91 +42,112 @@
...
 
     Single-field records as scalars
     --------------------------------
-    A ``NumericRecord`` with exactly one field behaves like a thin wrapper around
-    that field's value: ``float(r)``, ``int(r)``, ``bool(r)``, ``np.asarray(r)``,
-    ``jnp.asarray(r)``, and the ``r.shape`` / ``r.dtype`` / ``r.ndim`` attributes
-    all forward to the sole leaf. This lets a workflow function that returns a
-    single-field record be used in ordinary numeric expressions without
-    unwrapping the field by hand. A record with more than one field, or whose one
-    field is itself a nested record, raises ``TypeError`` from these conversions,
```
arob5, line 126: small thing, but this seems to contradict our definition of field. If the object being referred to is a nested record, then it is an internal node, not a field.

probpipe/core/_numeric_record.py
```diff
@@ -217,45 +239,45 @@ def __init__(
...
+        (it carries no metadata to preserve); every other numeric leaf is
+        stored verbatim.
 
         Raises ``TypeError`` on non-numeric input with a message that
         names the offending field and its type.
         """
         cls_name = cls.__name__
         out: dict[str, Any] = {}
-        aux: dict[str, Any] = {}
         for field_name, raw in raw_fields.items():
```
arob5: similar naming nit here - these are not the field names under the stated definition of "field"

arob5:
Generally speaking, these updates are not consistently following the definitions for things like "field", "path", etc. specified in `NamedTree`. They should be updated here, and we should also ensure that LLMs in the future reference such conventions when writing code.

probpipe/core/_numeric_record.py
```diff
@@ -272,11 +294,36 @@ def _validate_and_coerce(
...
+        if isinstance(val, jnp.ndarray):
+            return val
+        cache = self._jax_cache
+        arr = cache.get(field_name)
+        if arr is None:
+            arr = _to_jax_array(val)
+            cache[field_name] = arr
+        return arr
+
+    def _leaf_as_jax(self, path: str) -> jnp.ndarray:
```
arob5: docstring should make clear that paths to interior nodes are not valid input here. This internal helper operates on a field (a leaf). A more precise API: `_leaf_as_jax(self, key: str)`. And more precise variable naming in the first line of the function: `child_name, sep, descendant_key = key.partition(_PATH_SEP)`. If the above method is renamed as suggested, then to be consistent this method should be renamed `_field_as_jax(self, key: str)`.

probpipe/core/_numeric_record.py
```diff
@@ -272,11 +294,36 @@ def _validate_and_coerce(
...
-            if hooks is not None:
-                aux[field_name] = (hooks, hooks.capture(raw))
-            out[field_name] = raw if isinstance(raw, jnp.ndarray) else jnp.asarray(raw)
-        return out, aux
+            out[field_name] = jnp.asarray(raw) if isinstance(raw, _NUMERIC_SCALARS) else raw
+        return out
+
+    # -- The compute boundary: lazy, cached conversion -----------------------
+
+    def _as_jax(self, field_name: str) -> jnp.ndarray:
```
arob5: I think the function name and docstring should make more clear that this operates on a child field (i.e., a leaf that is one-level down). A name selecting an interior node is not valid input. A more useful name would be something like `_child_field_as_jax(self, field_name: str)`.

probpipe/core/_numeric_record.py
```diff
@@ -347,149 +399,9 @@ def from_vector(cls, name: str, template: NumericEventTemplate, vec: Array) -> N
...
-                if entry is not None:
-                    carried[field_name] = entry
-        current = self._aux or {}
-        if carried and not (set(carried) <= set(current)):
-            object.__setattr__(self, "_aux", {**carried, **current})
+    def to_numeric(self) -> NumericRecord:
+        """Return ``self`` — a ``NumericRecord`` is already numeric (identity)."""
+        return self
 
     # -- Single-field scalar-like coercion ---------------------------------
```
arob5, line 413: This large comment is probably not necessary. All of this information should be in the class docstring already.

probpipe/core/_numeric_record.py
jhuggins: Agreed. Fixed in b4bcf04: the docstring no longer calls a nested record a field/leaf. The numeric-leaf-invariant section now reads *"Every field (a leaf) must be numeric — …; interior nodes are themselves `NumericRecord` instances"*, and the single-field-scalar section says *"whose single child is a nested record — an interior node, not a field"*.

probpipe/core/_numeric_record.py
jhuggins: Fixed in b4bcf04: the loop variable is now `name` (it iterates child names, each mapping to a leaf *or* an interior node). The error messages follow the convention too — a bad nested child reads `child {name!r} … must be NumericRecord`, while a bad leaf keeps `field {name!r} must be a numeric …`.

probpipe/core/_numeric_record.py
jhuggins: Done in b4bcf04: renamed to `_field_as_jax(self, key)` with `child_name, sep, descendant_key = key.partition(_PATH_SEP)`, and the docstring now states *key* must resolve to a field (a leaf) — a path naming an interior node is not valid input. Went with `_field_as_jax` (over `_leaf_as_jax`) for consistency with `_child_field_as_jax` below, as you suggested.

probpipe/core/_numeric_record.py
jhuggins: Renamed to `_child_field_as_jax(self, field_name)` in b4bcf04; the docstring now says *field_name* must name a field one level down (a leaf), not an interior node.

probpipe/core/_numeric_record.py
jhuggins: Removed in b4bcf04 — replaced with a one-line pointer to the class docstring's "Single-field records as scalars" section, which already carries this.

PR #360, opened 2026-07-13
refactor(core): make NamedTree generic over its leaf type (NamedTree[L])

### LINKED ISSUE #356: Make NamedTree generic (NamedTree[L]) so leaf types reach py.typed consumers
## Summary

The design sketches `NamedTree` as generic over its leaf type — `class NamedTree[L]`
(`design/02-shared-abstractions.md` II.1) — but the implementation
(`probpipe/core/named_tree.py`) is a plain, non-generic class: `NamedTree[int]`
raises `TypeError` (it is not subscriptable), and the leaf-trafficking accessors
and transforms (`values`, `[]`, `items`, `map`, ...) are annotated `Any`.

Because ProbPipe ships `probpipe/py.typed`, downstream users are told the package
is typed, yet they receive `Any` exactly where the design promises the leaf type
`L`. This weakens the value of the typed marker and contradicts the II.1 contract.

## Proposed change (typing only, no behavior change)

- Make `NamedTree` generic with PEP 695 syntax: `class NamedTree[L]:` (the project
  is `requires-python >=3.12`, and PEP 695 `type` aliases are already used in
  `event_template.py`).
- Type the leaf-trafficking surface with `L`: `values() -> tuple[L, ...]`,
  `__getitem__(...) -> L`, `items() -> tuple[tuple[str, L], ...]`,
  `map(Callable[[L], L], ...) -> Self`, `map_with_keys(...) -> Self`. Navigation
  that legitimately returns a leaf **or** a subtree (`at_path`, `children`) stays
  `L | Self`.
- Bind `L` in the subclasses: `class EventTemplate(NamedTree[ValueSpec])` and
  `class Record(NamedTree[Any], Tracked, Annotated)` (records hold arbitrary leaf
  values, so `Any` is the honest bound; `NumericRecord` narrows informally to
  arrays but the static leaf type stays `Any` via `Record`).
- The runtime `_node_type()` / `_leaf_type()` hooks and the node/leaf partition
  are unaffected — this is annotations only.

Verify by a spot check that `NamedTree[int]` subscripts without error and that a
downstream `reveal_type(EventTemplate(...)["x"])` is `ValueSpec` (and
`Record(...)["x"]` is `Any`).

## Open sub-question — enforcement

There is **no CI type-checker** configured today (no mypy / pyright / pyre in
`pyproject.toml` or `.pre-commit-config.yaml`), so the `py.typed` contract is not
actually enforced and generic annotations could silently rot. Adding a type-check
gate (pyright or mypy over `probpipe/`) is a separate, larger decision worth
settling alongside this — otherwise the generic surface is advisory only.

## Scope

A typing-only sweep of `named_tree.py` plus the two subclass base declarations.
Deliberately deferred out of PR #338 (per SoymilkWinsAgain's review) to keep that
PR focused on behavior; tracked here per that decision.

## References

- Design: `design/02-shared-abstractions.md` II.1 — the `NamedTree[L]` class
  sketch and the `_leaf_type()` / `_node_type()` hooks.
- Shipped typed marker: `probpipe/py.typed`.
- Origin: SoymilkWinsAgain's review on PR #338 (`named_tree.py` thread).

jhuggins:
## Summary

Makes `NamedTree` generic over its leaf type — `class NamedTree[L]` — so the
`py.typed` leaf type reaches downstream consumers. Closes #356 (part 1; the
enforcement sub-question is a separate effort — see Notes).

ProbPipe ships `probpipe/py.typed`, but the shared `NamedTree` substrate was a
plain non-generic class: `NamedTree[int]` raised `TypeError`, and the
leaf-trafficking accessors were annotated `Any`. Downstream typed users were
therefore handed `Any` exactly where `design/02-shared-abstractions.md` (II.1)
promises the leaf type `L`, weakening the shipped typed marker.

This is an **annotations-only** change — no runtime behavior changes.

## What changed

- `NamedTree` is generic via PEP 695: `class NamedTree[L]:` (the project is
  `requires-python >=3.12`; PEP 695 `type` aliases are already used in
  `event_template.py`).
- `L` is threaded through the leaf-trafficking surface: `values() ->
  tuple[L, ...]`, `__getitem__ -> L`, `items() -> tuple[tuple[str, L], ...]`,
  `_walk_leaves -> Iterator[tuple[str, L]]`, and `at_path` / `children` ->
  `L | Self` (they legitimately return a leaf **or** a subtree).
- The structure-preserving transforms (`map` / `map_with_keys` / `without` /
  `merge` / `replace` / `with_path_names`) and `from_nested_dict` return
  `Self`.
- Leaf type bound in the two direct subclasses: `EventTemplate(NamedTree[ValueSpec])`
  and `Record(NamedTree[Any])` (records hold arbitrary leaf values, so `Any`
  is the honest bound). `NumericRecord`, the record-array types, and the
  distribution views inherit the `Any` binding through `Record`; `_leaf_type()`
  continues to carry the runtime narrowing.
- Follow-up commit: `__getitem__` (annotated `-> L`) delegates to `at_path`
  (typed `L | Self`), which pyright cannot narrow via the runtime
  `_node_type()` guard, so it reported `reportReturnType`. Narrowed with a
  post-guard `cast(L, node)`. This keeps the change **net-neutral on type
  debt** — the advisory pyright job reports the same pre-existing
  `named_tree.py` errors as before this PR (the subclass-owned `_tree`
  accesses) and no new ones.

## Verification

- **`reveal_type` spot check (pyright):** `EventTemplate(...)["x"]` →
  `ValueSpec`; `EventTemplate.values()` → `tuple[ValueSpec, ...]`;
  `EventTemplate.without("x")` → `EventTemplate | NumericEventTemplate` (the
  `Self` return, with the documented auto-promotion surfaced); `Record(...)["x"]`
  → `Any`. 0 errors.
- **Runtime:** `NamedTree[int]` subscripts; `NamedTree.__type_params__` is
  `(L,)`; the abstract-instantiation guard still trips; `jax` pytree
  flatten/unflatten and `vmap` over a `NumericRecord` round-trip unchanged.
- **Tests:** the core `NamedTree` / `Record` / `EventTemplate` /
  `NumericRecord` / array-backend / serialization / record-array suites pass
  (570 passed).
- `ruff format` / `ruff check` clean.

## Notes

- Stacked on #338 (`dev/named-tree-record-alignment`), so the diff here is only
  the generic-typing change; GitHub will retarget to `main` once #338 lands.
  Independent of #359 (native-form leaves) — that PR touches `_numeric_record`
  / the array backend, this one touches only annotations and the two class
  declarations, so they do not overlap.
- **Enforcement is out of scope here.** The advisory `typecheck` CI job
  (`pyrightconfig.json`, `basic` mode, `continue-on-error`) already reports on
  this PR; this change is net-neutral on that count. Moving the checker from
  advisory to gating is a separate burn-down tracked as the enforcement
  sub-question on #356.


arob5:
**No automated regression guard for the added static types.**

Every observable effect of this PR is at type-check time (`values() -> tuple[L, ...]`, `EventTemplate(...)["x"]` → `ValueSpec`, `Record(...)["x"]` → `Any`, etc.). The verification here was a manual `reveal_type` spot check; nothing in `tests/` pins it. Because the `typecheck` CI job is advisory (`basic` mode, `continue-on-error`), a future edit could silently regress `values() -> tuple[L, ...]` back to `Any` and CI would stay green — losing exactly what this PR buys.

Suggestion: add a small typing test using `typing.assert_type` on the key accessors, runnable under the existing `pyrightconfig.json`:

```python
from typing import Any, assert_type

from probpipe.core.event_template import EventTemplate, ValueSpec

assert_type(EventTemplate(...)["x"], ValueSpec)
assert_type(EventTemplate(...).values(), tuple[ValueSpec, ...])
assert_type(Record(...)["x"], Any)
```

Caveat: the repo has no precedent for `assert_type`/`reveal_type` tests today, so this would be a new pattern — reasonable to defer, but if so it's worth one sentence in the PR body stating the static-type guard is intentionally out of scope (the same way enforcement already is).

arob5:
**`-> Self` on the mutating transforms is mildly unsound for the auto-promoting `EventTemplate` family.**

`NumericEventTemplate._rebuild_class()` redirects rebuilds to the base `EventTemplate` so an edit re-decides promotion. For `replace` / `map` / `merge`, that edit can **demote** an all-numeric template to a mixed base `EventTemplate` (e.g. replacing an array spec with an opaque one). A receiver statically typed as `NumericEventTemplate` (reachable via `numeric_subset()` or an explicit annotation) then has the result mis-typed as `Self` = `NumericEventTemplate`.

This is still strictly better than the prior `-> Any`, and the common cases are sound:

- values typed as the `EventTemplate | NumericEventTemplate` constructor union,
- `Record` (never demotes),
- `without` (removal can't introduce opacity, so an all-numeric template stays numeric).

Only `replace` / `map` / `merge` on a bare-`NumericEventTemplate` static receiver, in the demoting direction, is affected.

Options: leave `-> Self` as the pragmatic base contract and just acknowledge the narrow imprecision, or have `EventTemplate` override the *annotations* of the promoting transforms to `-> EventTemplate | NumericEventTemplate`. The latter is arguably scope creep for an otherwise base-only change — flagging for a call rather than requesting it.

arob5:
**Is `from_nested_dict` still earning its keep?** (out of scope for this PR — raising it here for a follow-up)

`from_nested_dict` was added under an earlier design where the constructor accepted only the node type or leaf types, so a nested `dict` needed a dedicated reader. Under the current **"a mapping is never a leaf"** invariant the constructor materialises any mapping value into a subtree itself, so a nested dict passes straight through `Record(name, data)`. On this branch `from_nested_dict` reduces to:

```python
flat = cls._flatten_paths(data)
return cls(name, flat) if event_template is None else cls(name, flat, event_template=event_template)
```

`_flatten_paths` recurses into every mapping (no veto), then hands the result back to the constructor — which would have re-materialised the same structure anyway. There is no `recurse_into` predicate and no opaque-dict-leaf escape hatch anywhere on the branch, so the method supplies nothing the constructor lacks beyond passing `name` and forwarding `event_template` (both already constructor parameters).

Verified against this branch:

| Check | Result |
|---|---|
| `Record("r", nested) == Record.from_nested_dict("r", nested)` | `True` (same keys, same `NumericRecord` promotion) |
| Same, with `event_template=` forwarded | `True` |
| Constructor keeping a dict as an opaque leaf? | Not possible — `Record("r", {"meta": {...}})["meta"]` → *"'meta' is a subtree"* |

One behavioural divergence, and it's a footgun rather than a feature — for a malformed input mixing a `/`-path key with a nested-dict value under the same prefix:

```python
Record("r", {"y/a": 1.0, "y": {"b": 2.0}})             # ValueError: 'y' is used both as a field and as a path prefix
Record.from_nested_dict("r", {"y/a": 1.0, "y": {"b": 2.0}})  # silently succeeds → keys ['y/a', 'y/b']
```

`from_nested_dict` accepts what the constructor (correctly) rejects, because `_flatten_paths` and `_unflatten_paths` resolve the collision differently. So removing it would *tighten* validation, not loosen it.

Proposal for a follow-up (not this PR): drop `NamedTree.from_nested_dict`, the `Record` / `EventTemplate` overrides, the `_record_array.py` unsupported-stub, and `_flatten_paths`. Callers to migrate: the `test_from_nested_dict_*` round-trip tests in `tests/core/test_named_collection.py`, whose story becomes `Record(name, r.to_nested_dict())`. Since it's public API, external/notebook callers want a sweep first. Flagging here so the redundancy is captured while the context is fresh; happy to open the issue.

Issue #357, opened 2026-07-13
Decide the single-field-wrap convention across FunctionSpec / DistributionSpec / LinOp


jhuggins:
## Summary

The `ValueSpec` hierarchy is inconsistent about one convenience: **wrapping a bare
`ValueSpec` into a single-field `EventTemplate`**. `FunctionSpec` offers it;
`DistributionSpec` does not. The design (`design/03` III.1 and III.5) frames this
wrap as a *cross-cutting convention* — "the same convention that presents a scalar
draw as a single-field value" — that also governs `LinOp`. This issue is to decide
the convention **consistently** across the specs, and, if kept, to DRY the wrap
logic into a shared helper (per @arob5's PR #338 review).

## Current state

- **`FunctionSpec`** (`probpipe/core/event_template.py`): its hand-written
  `__init__` accepts `EventTemplate | ValueSpec | None` for `input_template` /
  `output_template` and wraps a bare `ValueSpec` into a single-field template under
  the field name `"input"` / `"output"`. Its `is_valid` is just `callable(value)`
  — the templates are **documentary only**, never validated, so the wrap name is
  cosmetic and can't cause a false rejection.
- **`DistributionSpec`**: `__post_init__` requires an `EventTemplate` and raises
  otherwise — no convenience path. Its `is_valid` compares
  `value.event_template == self.event_template` — an **equality** check. So a
  fixed-name single-field wrap (e.g. `"value"`) would make it *reject* a real
  scalar distribution, whose draw is keyed by the distribution's own name
  (design III.6), not the wrap convention.
- **`LinOp`** (design III.5, not yet implemented): "built from a bare matrix
  defaults to single-field templates, by the same convention."

So the constructor asymmetry @arob5 flagged is real, but the specs *use* their
templates differently (doc-only vs equality-checked), which is why "just make them
symmetric" is not a free win.

## The decision to make

Which direction, applied consistently across `FunctionSpec` / `DistributionSpec` /
`LinOp`?

- **Keep + DRY + extend.** Keep the convenience; extract the
  `bare ValueSpec → single-field EventTemplate` wrap into a shared helper (it
  recurs — `ConditionalDistribution` will want it too); and decide whether/how
  `DistributionSpec` adopts it. Adoption would require its equality `is_valid` to
  be name-tolerant (compare structurally modulo the top-level field name), or a
  documented wrap-name convention.
- **Remove + require meaningful names.** Drop the convenience from `FunctionSpec`
[...]

Issue #252, opened 2026-06-10
Tracking: estimator integration through fit and first-class Function


SoymilkWinsAgain:
## Goal

Integrate estimator training into `ProbPipe` without introducing a parallel estimator object model:

```text
training data
-> fit
-> reusable `Function` or `ConditionalDistribution`
-> apply / condition / sample / summarize
```

For deterministic point estimators, `fit` returns an ordinary first-class `Function`. For estimators that define an explicit predictive law, it returns `ConditionalDistribution`.

This issue is now the map and status tracker. Detailed design and implementation decisions live in the sub-issues below.

The existing single mapping base is named `Function` throughout this plan. This is the `Computation` → `Function` rename tracked in #377, not a second callable hierarchy.

## Established design boundary

- `Function` remains the reusable deterministic mapping abstraction and becomes a first-class tracked term.
- `fit` uses a controlled producer contract to publish trustworthy, fitted Functions. Consumers do not dispatch on fitted origin.
- `FunctionSpec`, `FunctionBatch`, and `RandomFunction` remain callable-generic. Fit-produced instances may carry stronger per-result guarantees without narrowing those stable abstractions.
- Result kind and canonical schema are planned before lifted execution.
- Universal `TermId`, executable persistence, random-function realization sharing, and advanced estimator semantics are separate decisions, not hidden prerequisites of the minimum point-`fit` slice.

## Dependency map

```text
#377 rename Computation to Function
  └─> #368 first-class tracked Function
  └─> #369 closed result planning
      └─> #370 controlled fitted-mapping contract
           └─> #371 minimum point fit and adapters
                 └─> #372 lifted fit
                 └─> #375 staged semantic extensions

#373 identity scope -------- independent, non-blocking
#374 persistence/replay ---- independent, non-blocking

#255 ConditionalDistribution design
  └─> probabilistic branches in #375
```

### Foundation

- [x] #377 — rename `Computation` to `Function` and `@computation` to `@function`
- [ ] #368 — first-class, tracked `Function`
- [ ] #369 — closed result planning
- [ ] #370 — controlled fitted-mapping producer contract

### Minimum estimator slice

- [ ] #371 — `fit` core and deterministic estimator adapters

### Lifted results

- [ ] #372 — reuse `FunctionBatch` and `RandomFunction` without narrowing them

### Independent architecture decisions

- [ ] #373 — distinguish operation identity, artifact identity, and `TermId`
[...]

PR #361, opened 2026-07-14
refactor(core)!: require explicit EventTemplates in FunctionSpec (#357)

jhuggins:
## Summary

Resolves the `FunctionSpec` / `DistributionSpec` single-field-wrap asymmetry
(#357, from @arob5's PR #338 review) by **tightening** — requiring explicit,
meaningfully-named templates — rather than extending the convenience. Also
aligns the (not-yet-implemented) `LinOp` design in the same spirit, with one
deliberate naming difference that is **offered for review** (see below).

Addresses #357.

## What changed — `FunctionSpec` (code, settled)

- `FunctionSpec` no longer wraps a bare `ValueSpec` into a single-field template
  keyed `"input"` / `"output"`. Each side must be an explicit `EventTemplate`
  (or `None`); a single-field signature is written out:
  `FunctionSpec(EventTemplate(x=()), EventTemplate(out=()))`.
- Dropping the wrap lets `FunctionSpec` become a plain `@dataclass(frozen=True)`
  with `__post_init__` validation — the same shape as `DistributionSpec` —
  removing the hand-written `__init__`. This *is* the resolution of the
  constructor asymmetry, and it drops @arob5's "DRY the wrap into a helper"
  sub-ask (there is no wrap left to share).
- **Breaking**: `FunctionSpec` is a public export. Migration: replace
  `FunctionSpec(ArraySpec(...), ArraySpec(...))` with explicit templates.

Why tighten rather than keep the convenience: once a callable is represented by
a value that exposes authoritative input/output templates (the first-class
`Function` of #252), `is_valid` can check a signature *structurally*, the way
`DistributionSpec` already checks `value.event_template == self.event_template`.
At that point a fixed wrap name (`"input"`/`"output"`) would cause **false
rejections** against a real function's caller-chosen names — the same failure
mode #357 already documents for `DistributionSpec`. Tightening now makes that
future validation sound.

## What changed — `LinOp` design (for review)

`design/03` §III.5 is updated so a `LinOp`'s input/output `NumericEventTemplate`s
name the numeric spaces it maps between (a bare matrix supplies those names
explicitly, instead of defaulting to a single-field placeholder).

**The one point for reviewer feedback:** a `LinOp`'s
two sides **may share a name**, in deliberate contrast to
`ConditionalDistribution`, whose given/event fields must stay *disjoint*
(design/03 §III: a kernel's conditioning value and its draw are distinct random
variables — `K(x, dy)`, not `K(x, dx)`). The rationale for the difference:

- a `ConditionalDistribution` is a **kernel between two draws** → disjoint names
  keep the distinct realizations distinguishable;
- a `LinOp` is an **operator on a space** → a self-adjoint operator (covariance,
  precision, Hessian) is an endomorphism on *one* space, so its input and output
  templates are *equal*, and that shared name is exactly what lets the operator
  algebra confirm that a composition lines up or that two operands act on the
  same space (transpose swaps the sides, so `A == Aᵀ` requires them equal).

Does that distinction hold, or should `LinOp` and `ConditionalDistribution`
share one naming rule? No `LinOp` code exists yet, so this is a design decision
only.

## Verification

- `tests/core/test_event_template.py`: 172 passed (the bare-spec test class is
  rewritten to assert rejection; explicit-template sites unchanged).
- `ruff format` / `ruff check` clean.
- pyright: net-neutral — `event_template.py` keeps its 2 pre-existing errors
  (both outside `FunctionSpec`) and gains none.

## Stacking / conflicts

Targets `main` (#338 has merged; its tip is an ancestor of `main`, so this PR's
diff is just the `FunctionSpec` commit). No overlap with #359 or #360 (also now
→ `main`): neither touches `FunctionSpec` or `test_event_template.py`, and their
`event_template.py` / `design/03` hunks are in different regions (only
`CHANGELOG.md` is shared — additive).


arob5:
A few points to tighten the language:
- Can have a square operator that still maps between two different spaces. Phrasing in the design is a little loose on this point.
- It seems like the default behavior is to allow input/output templates to be the same or distinct. This is the case for `FunctionSpec`. `ConditionalDistribution` is thus the exception to this behavior. Current design reads as if `LinOp` is the exception.
- Need to formalize what is load-bearing here: for example, is symmetry defined with respect to a tag or is it the result of a check `input_template == output_template`?
- Naming rule makes sense to me. `FunctionSpec`/`LinOp` names correspond to domain/codomain names, which can be the same. For `ConditionalDistribution` the input and output spaces could be the same (the latter representing a measure over the input space) but the names need to signify the given vs. event distinction (e.g., current state vs. next state).

jhuggins:
Thanks — tightened the design language in 36df65e. Point by point:

1. **Square ≠ same space.** Agreed, and the phrasing no longer ties shape to space at all. The two sides now *coincide exactly when the operator maps a space to itself (an endomorphism)* — shape-agnostic. A square operator that maps between two distinct spaces (e.g. the Jacobian of an `ℝⁿ → ℝⁿ` map) just names its sides distinctly, and the paragraph no longer implies "non-square ⇒ distinct".

2. **Default vs. exception.** Right. The independent-templates default now lives on `FunctionSpec` — "the two templates are independent … a callable may map a space to itself or between two different spaces" — and the `LinOp` paragraph drops the `ConditionalDistribution` contrast entirely, so neither is framed as the exception. Framing note: with the planned `Function` hierarchy (`LinOp` a special case), a `ConditionalDistribution` is a *separate mathematical object* — a probability kernel — not a `FunctionSpec` variant, so it carries its own rules rather than deviating from the operator ones.

3. **What's load-bearing — tag or check.** The check: `input_template == output_template` is the structural fact the operator algebra reads for "same space / composes". Symmetry / self-adjointness is a *separate* declared property carried in `flags` (`"symmetric"`), not implied by template equality — the old text conflated the two ("self-adjoint … templates equal"). Now stated explicitly.

4. **Naming rule.** Kept, and grounded `ConditionalDistribution`'s disjoint-names rule in the given/event *role* distinction (conditioning value vs. produced law): the names stay disjoint even when `S = T` (`state → next_state`), independent of the domain/codomain naming for functions/operators.

PR #362, opened 2026-07-14
feat(core): pandas nullable/masked numeric columns are numeric leaves (NA -> NaN)

jhuggins:
## Summary

Makes pandas **nullable / masked numeric** columns (`Int64`, `Float64`, `boolean`, pyarrow numerics) first-class numeric leaves, so records carrying missing data promote to `NumericRecord` and flow through the numeric machinery. Missing values are encoded as `NaN` at the compute boundary — the only missing-value representation `jax` offers.

Base: **`main`** (originally stacked on #359, which has since merged).

### Motivation

Missing data is a common case for real datasets, and we want ProbPipe to handle it rather than force users to pre-clean or lose the numeric path. With the native-form-leaves work (now in `main`), a frame with an `Int64`/`Float64`/`boolean` column stayed a plain `Record`: the generic numeric gate `_is_numeric_dtype` rejects masked dtypes because they are not dense-numpy-coercible. That is the right default for a *bare* value on the duck-typing path, but the pandas backend can do better for its own types.

### Design

The change lives mainly in the pandas `ArrayBackend` hooks (`is_numeric`, `numpy_dtype`, `to_jax`, `to_numpy`) in `_array_backend.py`. It also **touches `event_template.py`**: a one-paragraph clarification to the `_is_numeric_dtype` docstring stating that the generic (duck-typing) gate stays strict while a registered backend may still recognise its own masked dtypes. Because every numeric gate resolves **registry-first**, promotion, `ArraySpec` inference, the compute boundary, and `fingerprint()` all pick the change up at once.

- **Recognition:** a column is nullable-numeric when `pd.api.types.is_extension_array_dtype(dtype) and is_numeric_dtype(dtype)` — admits `Int64`/`Float64`/`boolean`/pyarrow numerics, excludes categorical / string / datetime.
- **Conversion & dtype:** the container densifies to its columns' **common dense dtype** (`_dense_dtype`). A masked numeric column contributes a float — a nullable float keeps its own width (`Float32 → float32`), a nullable integer / boolean promotes to `float64` — and every other column contributes its own dtype; the result is their `np.result_type` promotion. So a mixed frame promotes correctly and a **complex column stays `complex128`** rather than being truncated to float. `numpy_dtype` reports exactly the dtype the conversion produces, and each NA is encoded as `NaN`.

The generic `_is_numeric_dtype` gate stays strict: a conservative duck path over a backend that knows its own types.

### Why this fits native storage

The nullable frame is **stored verbatim**, so its validity mask is intact at rest — `record["x"]` returns the pandas object with real `NA`s. `NA → NaN` happens *only* when converting to `jax`, where a mask cannot survive anyway. No information is lost until you actually compute.

### Semantic notes (inherent, not bugs)

- **Integer / boolean columns promote to float** — there is no integer `NaN`.
- **NA and a genuine `NaN` are indistinguishable once in `jax`** — at rest the mask is preserved; after conversion both read as `NaN`.

### Tests

`TestNullableNumericMissingData` — promotion; `NA → NaN` at the boundary; mask preserved at rest; `Series`/`DataFrame`; `boolean`/`Float64`; mixed nullable+plain columns; `numpy_dtype`; categorical/string stay opaque; and the **nullable+complex** cases (imaginary parts preserved, `numpy_dtype == complex128`, own-width nullable floats, and equality/fingerprint distinguishing imaginary parts while treating NA as equal). `TestGenericGateRejectsExtensionDtype` — the generic gate stays strict (incl. the `np.dtype` `ValueError` regression pin).

Full `tests/core` plus distributions / converters / modeling / inference / diagnostics / validation green.

### Also

Corrects a stale line in the `_array_backend` module docstring (`xarray.DataArray` / `pandas.Series` are registered, carrying `metadata` hooks). Includes a documentation-prose pass (parentheticals, sentence length, en/em-dashes) over the module docstring, the `_is_numeric_dtype` docstring, and the CHANGELOG entry.

### DIFF
--- a/probpipe/core/_array_backend.py
+++ b/probpipe/core/_array_backend.py
@@ -35,8 +35,8 @@ for two reasons. The first is to expose their identity-bearing metadata (an
 columns) through the ``metadata`` hook. The second, for pandas, is to handle
 nullable / masked numeric columns (``Int64``, ``Float64``, ``boolean``). The
 generic duck path cannot see those columns as numeric, but the pandas backend
-accepts them and converts through ``float64``, encoding each NA as ``NaN``. Bare
-``numpy`` / ``jax`` arrays stay unregistered — the duck path covers them and
+accepts them, encoding each NA as ``NaN`` in the columns' common dense dtype.
+Bare ``numpy`` / ``jax`` arrays stay unregistered — the duck path covers them and
 they carry no metadata beyond their values.
 
 Batch stacking hooks (a per-backend native ``stack`` / ``element``) are
@@ -296,23 +296,45 @@ def _register_pandas() -> None:
         # numeric dtype (generic gate), or a nullable numeric one (NA -> NaN).
         return _is_numeric_dtype(dtype) or _is_nullable_numeric(dtype)
 
+    def _dense_dtype(dtypes: Any) -> np.dtype | None:
+        # The single dense numpy dtype a Series / DataFrame of these column
+        # dtypes densifies to (via to_numpy / to_jax), or None when a column has
+        # no dense numpy dtype. A masked numeric column has no dense numpy view
+        # and must hold NA as NaN, so it contributes a float: a nullable float
+        # keeps its own width (Float32 -> float32), a nullable integer / boolean
+        # promotes to float64. Every other column contributes its own dtype. The
+        # result is their common promotion, so a complex column yields complex128
+        # rather than being truncated to float.
+        contribs: list[np.dtype] = []
+        for d in dtypes:
+            if _is_nullable_numeric(d):
+                nd = getattr(d, "numpy_dtype", None)
+                if nd is not None and np.issubdtype(nd, np.floating):
+                    contribs.append(np.dtype(nd))
+                else:
+                    contribs.append(np.dtype("float64"))
+            else:
+                nd = _safe_numpy_dtype(d)
+                if nd is None:
+                    return None
+                contribs.append(nd)
+        return np.result_type(*contribs) if contribs else None
+
     def _dense(obj: Any, dtypes: Any) -> np.ndarray:
-        # Any masked column forces the whole result through float64, encoding
-        # each NA as NaN; otherwise the native numpy view is used unchanged.
+        # A masked column has no dense numpy view; converting with an explicit
+        # target dtype encodes each NA as NaN and keeps complex columns complex
+        # (see _dense_dtype). Otherwise the native numpy view is used unchanged.
         if any(_is_nullable_numeric(d) for d in dtypes):
-            return obj.to_numpy(dtype="float64", na_value=np.nan)
+            return obj.to_numpy(dtype=_dense_dtype(dtypes), na_value=np.nan)
         return obj.to_numpy()
 
     def _series_metadata(s: pd.Series) -> Any:
         return {"index": np.asarray(s.index), "name": s.name}
 
     def _series_numpy_dtype(s: pd.Series) -> np.dtype | None:
-        # A nullable column converts through float64 (see _dense), so float64
-        # is the leaf's dtype — not the masked dtype, which has no dense numpy
-        # form. This depends on the dtype, not on whether NAs are present.
-        if _is_nullable_numeric(s.dtype):
-            return np.dtype("float64")
-        return _safe_numpy_dtype(s.dtype)
+        # Report the single dense dtype the series densifies to (see
+        # _dense_dtype), so numpy_dtype matches what to_numpy / to_jax produce.
+        return _dense_dtype((s.dtype,))
 
     register_array_backend(
         pd.Series,
@@ -331,19 +353,9 @@ def _register_pandas() -> None:
         return len(dtypes) > 0 and all(_column_numeric(d) for d in dtypes)
 
     def _frame_numpy_dtype(df: pd.DataFrame) -> np.dtype | None:
-        # A masked / nullable numeric column has no dense numpy view, so the
-        # whole frame converts through float64 (NA -> NaN). Otherwise a numeric
-        # frame densifies (to_numpy / to_jax) to one dense array whose dtype is
-        # its columns' common promotion — e.g. int64 + float64 -> float64 — so
-        # report that rather than None. ``None`` only when a column has no dense
-        # numpy dtype (a non-numpy, non-nullable extension column).
-        dtypes = df.dtypes
-        if any(_is_nullable_numeric(d) for d in dtypes):
-            return np.dtype("float64")
-        numpy_dtypes = [_safe_numpy_dtype(d) for d in dtypes]
-        if not numpy_dtypes or any(nd is None for nd in numpy_dtypes):
-            return None
-        return np.result_type(*numpy_dtypes)
+        # Report the single dense dtype the frame densifies to (see
+        # _dense_dtype), so numpy_dtype matches what to_numpy / to_jax produce.
+        return _dense_dtype(df.dtypes)
 
     def _frame_metadata(df: pd.DataFrame) -> Any:
         return {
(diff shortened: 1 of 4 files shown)


probpipe/core/_array_backend.py
```diff
@@ -268,32 +284,63 @@ def _register_pandas() -> None:
...
     from .event_template import _is_numeric_dtype
 
+    def _is_nullable_numeric(dtype: Any) -> bool:
+        # A masked / extension *numeric* dtype: numeric, but with a validity
+        # mask and no dense numpy view. Excludes categorical / string /
+        # datetime, which are extension dtypes yet not numeric.
+        is_ext = pd.api.types.is_extension_array_dtype(dtype)
+        return is_ext and pd.api.types.is_numeric_dtype(dtype)
+
+    def _column_numeric(dtype: Any) -> bool:
```
SoymilkWinsAgain, line 297: `_column_numeric()` accepts `complex128`, but `_dense()` will cast the entire DataFrame to float64 upon finding any nullable/extension numeric column.

Suggestions:

1. Calculate the common dense dtype for the entire container: nullable columns should contribute at least `float64`, other columns should participate in `np.result_type`, thus selecting `complex128` when complex columns exist.

2. Make `_frame_numpy_dtype()` report the same common dtype as the actual transformation.

3. Add tests for nullable integers + complex DataFrames, while also checking for imaginary parts, NAs, and fingerprint/equality.

SoymilkWinsAgain (changes requested):
Additionally, the PR body needs updating: the current base is now "main," no longer "stacked on #359". The description also claims no changes to `event_template.py`, but the final diff includes documentation updates for that file. `CONTRIBUTING.md` explicitly treats expired PR descriptions as review blockers.

probpipe/core/_array_backend.py
jhuggins: Fixed in 00a1616. You're right — `_dense` hard-coded a `float64` cast whenever any nullable column was present, truncating a sibling complex column, and `numpy_dtype` reported `float64` rather than what the conversion produced.

New `_dense_dtype(dtypes)` computes the container's single dense dtype (suggestions 1 & 2): a masked numeric column contributes a float — a nullable *float* keeps its own width (`Float32 -> float32`), a nullable integer / boolean promotes to `float64` — and every other column contributes its own dtype, combined via `np.result_type`. So `Int64 + complex128 -> complex128` (imaginary parts preserved, NA -> `nan+0j`), and both `_dense` and the `numpy_dtype` hooks now go through it, so the reported dtype matches the actual transformation.

Tests added (suggestion 3): a nullable `Int64` + `complex128` frame keeps its imaginary parts with `numpy_dtype == complex128` and NA -> nan; a nullable `Float32` keeps `float32`; and equality/fingerprint distinguish records differing only in an imaginary part while treating the NA position as equal.

(One deviation from "at least float64": per a maintainer call, a nullable *float* column keeps its own width rather than being floored to float64 — lossless, and float32 already holds NaN.)

jhuggins:
Thanks — updated the PR description: the base is now **`main`** (no longer "stacked on #359", which merged), and it now states that the diff touches `event_template.py` (the `_is_numeric_dtype` docstring clarification) plus the CHANGELOG and a documentation-prose pass, rather than claiming no changes there. The complex-truncation fix is in 00a1616 (see the inline thread).

probpipe/core/_array_backend.py
```diff
@@ -268,35 +284,78 @@ def _register_pandas() -> None:
...
+        # result is their common promotion, so a complex column yields complex128
+        # rather than being truncated to float.
+        contribs: list[np.dtype] = []
+        for d in dtypes:
+            if _is_nullable_numeric(d):
+                nd = getattr(d, "numpy_dtype", None)
+                if nd is not None and np.issubdtype(nd, np.floating):
+                    contribs.append(np.dtype(nd))
+                else:
+                    contribs.append(np.dtype("float64"))
```
SoymilkWinsAgain: [P1] Preserve complex extension dtypes here. _is_nullable_numeric() accepts every pandas extension dtype that pandas considers numeric, including Sparse[complex128]. SparseDtype exposes its dense type as .subtype, not .numpy_dtype, so this branch falls back to float64; Series([1+2j, 3+4j], dtype=Sparse[complex128]) then reports numpy_dtype == float64 and both to_numpy() and NumericRecord.to_vector() silently return [1, 3]. Please derive the contribution from the extension dtype's underlying dense dtype (for example, handling .subtype and complex/inexact kinds) and add a Sparse-complex regression test.

probpipe/core/_array_backend.py
```diff
@@ -259,7 +265,17 @@ def _register_pandas() -> None:
     A ``Series`` speaks the numpy protocol but carries an index and name; a
     ``DataFrame`` additionally has no single ``.dtype`` (per-column ``.dtypes``
     the duck path cannot read). Both answer shape / dtype / numericness from
-    metadata without materialising values. No-op without pandas.
+    metadata without materialising values.
+
+    Nullable / masked numeric columns (``Int64``, ``Float64``, ``boolean``,
+    pyarrow numerics) are accepted as numeric even though the generic gate
+    rejects their dtype: such a column has no dense numpy view, so the whole
+    container converts through ``float64`` with each NA encoded as ``NaN`` —
```
SoymilkWinsAgain: [P2] Update the dtype documentation to match the new common-dtype behavior. This still says the whole container converts through float64, but the implementation now preserves nullable Float32 and promotes mixed complex frames to complex128. The CHANGELOG entry and TestNullableNumericMissingData class docstring make the same stale float64-only claim.

SoymilkWinsAgain (changes requested):
The latest commit fixes the nullable-plus-plain-complex DataFrame case and the PR description is now aligned with the current base. However, one extension-complex path still silently loses data, so I am requesting changes. I reproduced it on pandas 3.0.3: a Series with dtype Sparse[complex128, nan] is accepted as numeric, but numpy_dtype, to_numpy, and NumericRecord.to_vector() all become real-valued and discard the imaginary components. The focused backend suite passes (64 tests), which shows this extension-dtype case is not currently covered. I also noted stale float64-only documentation inline.

probpipe/core/_array_backend.py
jhuggins: Fixed in d18a8b9. Right on the diagnosis — `SparseDtype` exposes its dense type as `.subtype`, not `.numpy_dtype`, and my check only preserved the floating kind, so `Sparse[complex128]` fell to `float64`. New `_extension_dense_dtype(d)` reads the underlying dtype via `.numpy_dtype` then `.subtype`, and `_dense_dtype` keeps it when it already holds NaN (float **or** complex, at its own width), promoting only integer / boolean to `float64`. So `Sparse[complex128]` densifies to `complex128` (imaginary parts kept, NA -> `nan`). Added a Sparse-complex regression test: `numpy_dtype == complex128`, `to_numpy` / `to_vector` keep the imaginary parts, NA -> nan, and equality distinguishes imaginary parts.

probpipe/core/_array_backend.py
jhuggins: Fixed in d18a8b9 — synced the stale float64-only wording in all three spots: the `_register_pandas` docstring, the CHANGELOG native-leaf entry, and the `TestNullableNumericMissingData` class docstring. They now describe the columns' common dense dtype (a nullable float keeps its own width, a complex column stays complex, integer / boolean promote to `float64`).

jhuggins:
Both points from the latest review are addressed in d18a8b9 (details in the inline threads): the `Sparse[complex128]` extension path now preserves imaginary parts via a subtype-aware, complex-preserving `_extension_dense_dtype`, and the stale float64-only docs are synced across the docstrings and CHANGELOG. Full `tests/core` plus the dependent suites are green.

PR #364, opened 2026-07-15
feat(core): wire fingerprints into Prefect cache_key_fn for cross-run task caching

erozercan:
> **Draft / WIP.** Steps 1–4 (below) are implemented, tested, and verified
> end-to-end. Step 5 (serialization audit + docs) and Step 6 (CHANGELOG + docs
> page) are still to come — opening now for early review of the core design.

## Summary

PR3 of the provenance/caching series (PR1 #244 `ProvenanceMode`, PR2 #318
`fingerprint()`). Adds **opt-in, cross-run Prefect task caching** keyed by
PR2's content fingerprints: a workflow step is skipped and its result reused
when the same computation — same user-function body, same inputs, same
environment — has already run, including across processes and machines.

Caching is **off by default**; nothing changes for existing users unless they
set `prefect_config.cache_mode = CacheMode.FINGERPRINT`.

Refs #117.

## Motivation

Today Prefect is used only as a parallel executor — the `@task` carries a
`name` and nothing else, so every re-run recomputes from scratch and a
mid-pipeline failure loses all progress. This PR turns on Prefect's result
caching, using a custom cache key that actually understands ProbPipe's types
(JAX arrays, `Record`s, distributions) and, critically, the user-function body
— which a source-hashing policy can't see through the generic `run_func`
wrapper.

## Design decisions (settled in plan review)

- **Opt-in, default OFF**, configured on `PrefectConfig` (caching is
  Prefect-only), mirroring `WorkflowKind`.
- **`CacheMode` enum**: `OFF` / `FINGERPRINT`.
- **No Record pickle fix** — already done: `Record`/`NumericRecord` define
  `__reduce__` (#336), pinned by `tests/core/test_record_serialization.py`.
  The serialization step is an audit, not new code.
- **No fingerprint gating** — fingerprints stay always-on in
  LIGHTWEIGHT/FULL as today, so there is **no breaking change** and no
  provenance-content↔caching-config coupling.
- **`cache_result_storage`** — one config field wired to **both** the task's
  `result_storage` and the cache policy's `key_storage` (a shared result store
  with local key records still misses on other workers). Default = Prefect
  local.
- **No TTL**; instead an **environment salt** (probpipe + jax versions +
  `jax_enable_x64`) is folded into the key, so env drift is a miss, not a
  stale hit.

## How the cache key is built

```
cache_key = sha256(func_fingerprint | environment_salt | fingerprint(inputs))
```

- `func_fingerprint` — computed once at the `WorkflowFunction` level (the only
  place `fingerprint()` can stably hash the user body) and threaded down via
  `WorkflowExecutionConfig`. Detects a changed function body.
- `environment_salt` — probpipe/jax versions + x64 flag. Detects env drift.
- `fingerprint(inputs)` — content hash of the per-row inputs. Empirically
  verified that Prefect binds a `**kwargs` task's args as `{"kwargs": {...}}`
  with **unordered** keys, so the whole mapping is fingerprinted and
  `fingerprint()`'s key-sorting makes the result order-independent.

## Changes

**`probpipe/core/config.py`**
- `CacheMode` enum (`OFF` / `FINGERPRINT`), next to `WorkflowKind`.
- `PrefectConfig.cache_mode` (type-checked setter), `caching_enabled`,
  `cache_result_storage` + `resolve_cache_storage()`, and a
  `PROBPIPE_CACHE_MODE` env var with an `_initial_cache_mode()` resolver
  (env-var parity with `PROBPIPE_WORKFLOW_KIND`).
- `CacheMode` added to `__all__`.

**`probpipe/core/_fingerprint.py`**
- `environment_salt()` — stable `probpipe=…;jax=…;x64=…` string.

**`probpipe/core/node.py`**
- `_func_fingerprint_value()` (memoized `fingerprint(self)`).
- `_make_execution_config` attaches `func_fingerprint` and
  `cache_result_storage` to the config **only** on the Prefect paths and
  **only** when caching is enabled; otherwise both stay `None` and the config
  is byte-identical to before.

**`probpipe/core/_workflow_execution.py`**
- `WorkflowExecutionConfig` gains `func_fingerprint` and
  `cache_result_storage`.
- `_cache_task_options()` builds the `cache_key_fn`, `persist_result=True`, and
  storage; `map_task` applies them. Returns `{}` when caching is off, leaving
  the task decorator unchanged.

**`probpipe/__init__.py`** — export `CacheMode`.

**`tests/conftest.py`** — the autouse config-reset now also resets
`prefect_config` (prevents the new `cache_mode` global leaking between tests
under xdist).

## Test plan

New `tests/core/test_caching.py` plus additions to `test_prefect_config.py`
and `test_fingerprint.py`:

- **Config** — `CacheMode` enum, defaults, type-checked setter,
  `caching_enabled` both ways, storage store/resolve, full
  `PROBPIPE_CACHE_MODE` env-var parity (incl. bad-value-crashes-at-import
  subprocess test).
- **Environment salt** — components present, reflects real jax version, and
  **toggling `jax_enable_x64` changes the salt** (restored in `finally`).
- **Threading** — `func_fingerprint`/storage set only on Prefect paths when
  caching on; `None` otherwise; memoized; two different function bodies →
  different fingerprints.
- **Cache key** — deterministic, order-independent, changes with inputs / func
  fingerprint / env salt, identical across a **subprocess**, storage wired to
  both surfaces.
- **End-to-end** (marked `prefect`, uses `prefect_test_harness`) — a real
  Prefect run confirms a re-run **hits** the cache (user function not
  re-executed) while a changed input **misses**.

Verified manually as well: caching OFF → 4 executions, ON → 2, over the same
call sequence.

Pre-existing, unrelated to this PR (confirmed identical on `main` in a clean
worktree): the `test_output_values_correct` JAX-tracer failures and the
NumericRecord/vmap broadcast failures under the local env.

## Still to do (before marking ready)

- [ ] **Step 5** — serialization audit: confirm persisted result types
  (`EmpiricalDistribution`, `Record`/`NumericRecord` incl. aux round-trip,
  `RecordArray`, TFP families) reload correctly; document JAX-x64 /
  Python-version key caveats.
- [ ] **Step 6** — CHANGELOG entry (Added: `CacheMode`, `cache_mode`,
  `cache_result_storage`, `PROBPIPE_CACHE_MODE`, env-salted keys) + a caching
  docs page.

## Note for reviewers

During review I found commit `15679eb` (Jul 13, "preserve backend aux across a
NumericRecord pickle round-trip") appears to **already fix** the `NumericRecord`
aux-loss caveat raised in plan review — so the "aux is dropped on reload"
acceptance criterion looks stale. Step 5's audit will confirm aux round-trips
faithfully; flagging in case that changes anyone's expectations.

## Breaking changes

None. Caching is opt-in and off by default; fingerprinting behavior is
unchanged (gating was dropped), so no public field changes.

## Checklist

- [x] Opt-in, default-OFF; OFF path byte-identical to before
- [ ] CHANGELOG entry (Step 6)
- [ ] Docs page (Step 6)
- [x] `area:core` label
- [x] Linked to tracking issue

### DIFF
--- a/probpipe/__init__.py
+++ b/probpipe/__init__.py
@@ -40,7 +40,13 @@ from probpipe.core._array_backend import (
 from probpipe.core._distribution_array import DistributionArray
 from probpipe.core._numeric_record import NumericRecord
 from probpipe.core._record_array import NumericRecordArray, RecordArray
-from probpipe.core.config import ProvenanceMode, WorkflowKind, prefect_config, provenance_config
+from probpipe.core.config import (
+    CacheMode,
+    ProvenanceMode,
+    WorkflowKind,
+    prefect_config,
+    provenance_config,
+)
 from probpipe.core.constraints import (
     Constraint,
     boolean,
@@ -212,6 +218,7 @@ __all__ = [
     "BootstrapDistribution",
     "BootstrapReplicateDistribution",
     "BroadcastDistribution",
+    "CacheMode",
     "Categorical",
     "Cauchy",
     "ConditionallyIndependentLikelihood",
--- a/probpipe/core/_fingerprint.py
+++ b/probpipe/core/_fingerprint.py
@@ -42,6 +42,7 @@ a 3.13 one for the same source.
 from __future__ import annotations
 
 import hashlib
+import importlib.metadata as _metadata
 import math
 import struct
 from typing import Any
@@ -49,7 +50,7 @@ from typing import Any
 import jax as _jax
 import numpy as _np
 
-__all__ = ["fingerprint"]
+__all__ = ["environment_salt", "fingerprint"]
 
 # JAX is a hard dependency of probpipe, so ``jax.Array`` is always importable.
 _JAX_ARRAY_TYPE = _jax.Array
@@ -92,6 +93,39 @@ def fingerprint(obj: Any, *, max_array_bytes: int | None = _DEFAULT_MAX_ARRAY_BY
     h = hashlib.sha256()
     _update(h, obj, depth=0, max_array_bytes=max_array_bytes)
     return h.hexdigest()[:16]
+
+
+def environment_salt() -> str:
+    """Return a stable string identifying the fingerprint-relevant environment.
+
+    :func:`fingerprint` hashes user bytecode and input *content*, but not the
+    surrounding runtime.  Folding this salt into a cache key ensures that an
+    environment change is a cache **miss** rather than a stale **hit** on an
+    otherwise-identical content digest.  It captures:
+
+    - the ProbPipe version,
+    - the JAX version, and
+    - the ``jax_enable_x64`` flag — it changes array dtypes, and therefore the
+      dtype recorded in every array fingerprint, so the same values under
+      ``x64=True`` and ``x64=False`` must key differently.
+
+    The Python minor version is not included explicitly: it changes the
+    ``WorkflowFunction`` bytecode digest (already part of the cache key), so a
+    Python upgrade already invalidates keys for the function component.
+
+    Returns
+    -------
+    str
+        A human-readable, deterministic salt such as
+        ``"probpipe=0.1.0;jax=0.4.38;x64=False"``.
+    """
+    try:
+        probpipe_version = _metadata.version("probpipe")
+    except _metadata.PackageNotFoundError:
+        probpipe_version = "unknown"
+    return (
+        f"probpipe={probpipe_version};jax={_jax.__version__};x64={bool(_jax.config.jax_enable_x64)}"
+    )
 
 
 # ---------------------------------------------------------------------------
--- a/probpipe/core/_workflow_execution.py
+++ b/probpipe/core/_workflow_execution.py
@@ -7,15 +7,20 @@ assemble inputs and interpret outputs outside this module.
 
 from __future__ import annotations
 
+import hashlib
 from collections.abc import Callable
 from concurrent.futures import ThreadPoolExecutor
 from dataclasses import dataclass
 from typing import Any, Literal
 
+from ._fingerprint import environment_salt, fingerprint
+
 try:
     from prefect import flow, task
+    from prefect.cache_policies import CacheKeyFnPolicy
 except ImportError:
     task = flow = None
+    CacheKeyFnPolicy = None
 
 WorkflowExecutionMode = Literal[
     "sequential",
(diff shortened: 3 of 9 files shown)


erozercan:
@jhuggins picking PR3 (Prefect fingerprint caching) back up after being away, and the execution-layer refactor that landed since means Step 4's design no longer fits. I'd like your ask it before I build it, since our original PR3 plan was about the new managed-execution / RNG-replay subsystem.

**Quick recap of PR3's plan** (third in the provenance/caching series after `ProvenanceMode` and `fingerprint()`): opt-in cross-run task caching via a custom `cache_key_fn` = `sha256(func_fingerprint | environment_salt | fingerprint(inputs))`, plus `persist_result` and shared result/key storage, wired at the single `@task` in `_workflow_execution.py::map_task`.

**What still ports cleanly:** the config layer (`CacheMode` on `PrefectConfig`, `PROBPIPE_CACHE_MODE`), the environment salt, and threading the function fingerprint down — modulo the `WorkflowFunction`→`Function` rename and the immutable `set_attribute` change.

**Where it breaks — and why I don't want to just force it in:**

1. **The cache key inputs moved.** `map_task` now runs `run_func(payload)` where `payload` is a `ManagedPrefectPayload` (RNG state, broker handle, `attempt` counter), so what Prefect hands `cache_key_fn` is the managed wrapper, not the user inputs. Fingerprinting it would ignore the real inputs and change on every retry. The key would have to reach into the work item for the actual arguments.

2. **Caching conflicts with the managed-outcome contract.** A cache hit skips `run_func` and returns a saved result — but `_run_prefect_payloads` / `_validate_prefect_outcome_for_payload` validate every outcome's frame/attempt/RNG report against its payload, and `_coordinate_prefect_randomness` runs afterward. A cached result has no fresh managed outcome, so it would fail validation or bypass the reproducibility machinery.

**My questions:**

- Where should caching hook now — still the `@task` in `map_task`, or a different layer that produces a valid managed outcome on a hit?
- How should a cache hit satisfy the managed-outcome/RNG-coordination contract (cache the whole validated outcome? synthesize one?)?
- Does the replay system change the caching goal — is durable cross-run *result* reuse still the target, or does replay already cover part of what PR3 intended?
- Would you rather caching be designed jointly with the replay model, and is now a good time, or should PR3 wait?

Happy to write this up as a short design proposal once I know the direction. Branch `dev/prefect-cache-key` still has the pre-refactor Steps 1–4 if useful as a reference; I can share the diff.

SoymilkWinsAgain:
Hi @erozercan, first answer your question:

Your diagnosis is right. The caching boundary should no longer be the raw return value of the managed `run_func` task.
1. The input of `task` now contains the payload of RNG authority, attempt, etc. So, I would suggest moving the hook into the parent-side managed orchestration, after a fresh payload/attempt has been reserved but around the actual Prefect submission. A hit should skip submission; a miss should use the existing execution, coordination, and retry path. Only a successfully validated and accepted result should be published to the cache.
2. We should not cache a complete `ManagedExecutionOutcome`. Its index, frame, attempt, and report ownership are operational state belonging to one submission. Instead, cache an attempt-independent semantic artifact and synthesize a fresh outcome for the current payload on a hit.
3. Replay does not replace durable result reuse. Replay just re-executes current code and inputs while validating RNG behavior, while caching skips execution. But I believe replay can inform future stochastic-cache compatibility rules and techniques. Therefore, I would suggest you take a closer look at the relevant sections.
4. The RNG system has been stabilized. And we are now working in the orthogonal direction. I would say it is now a good time for PR3.

I'm sorry I need you to redo it. I genuinely didn't know at the time that this had such a large overlap with your PR. I will write new design docs soon (currently waiting for #444). I would suggest you read #400 for details about how new RNG system works. Here's a very simple model:

The implementation is divided into four layers:

```text
RNG identity and ownership
    ↓
Immutable stochastic planning and jointness
    ↓
Managed execution and transport
    ↓
Provenance and replay
```

You can think of it as a tree of ownership and a DAG of random source.

```text
                   random source DAG
                   source 0  source 1
ownership tree
occurrence 0        key 00    key 01
occurrence 1        key 10    key 11
```

SoymilkWinsAgain (changes requested):
See my previous commit.

PR #365, opened 2026-07-15
refactor(core): extract shared registry-resolution helpers

jhuggins:
## Summary

Extracts the **registry-first, duck-typing-fallback** resolution into one shared helper per numeric question, so each fallback lives in a single place and can't drift between call sites. Only `to_jax` had such a helper (`_to_jax_array`); the rest were open-coded at ~12 sites.

Stacked on #362 (base `dev/native-form-missing-data`).

### Motivation

Every numeric gate asks one of a few questions about a leaf — *is it numeric? what's its shape / dtype? materialise it to numpy / metadata* — and each resolved the same way inline: the registered `ArrayBackend` hook if the type is registered, else the numpy-protocol duck path. Because the duck fallback was copied per site, they had **already drifted**:

- the numeric predicate was implemented **twice** — `_is_numeric_leaf` and, inline, inside `_full_array_shape_or_none`;
- the dtype fallback existed in **three forms** — `NumericRecord.dtype` used a bare `getattr(leaf, "dtype", None)`, while `_canonical_dtype_str` and `ArraySpec.is_valid` added an `np.asarray(...).dtype` fallback.

Both were harmless today (the leaf invariant guarantees a stored leaf has a real dtype), but they're latent bugs and make any change to the resolution contract a ~12-site edit kept in sync by hand.

### What this does

Adds, in `_array_backend.py` alongside `_to_jax_array`:

| Resolver | Question |
|---|---|
| `_is_numeric_leaf(v)` | is this a numeric leaf? |
| `_event_shape_of(v)` | its event shape (`()` for a scalar) |
| `_numpy_dtype_of(v)` | its single numpy dtype, or `None` |
| `_to_numpy_array(v)` | materialise to numpy (numpy twin of `_to_jax_array`) |
| `_metadata_of(v)` | its identity-bearing container metadata, or `None` |

`_NUMERIC_SCALARS` moves here too, as the single scalar-set definition. Call sites in `_numeric_record`, `event_template`, `record`, and `_fingerprint` now delegate.

### Two behaviours corrected (not just consolidated)

- `_full_array_shape_or_none` is now `_event_shape_of(v) if _is_numeric_leaf(v) else None` — the duplicated numeric decision is gone, so it and promotion can't disagree on what counts as numeric.
- `NumericRecord.dtype` shares `_numpy_dtype_of` with the other dtype sites, removing the bare-`getattr` drift.

Verified equivalent on scalars, registered numeric/non-numeric containers, duck arrays, and strings.

### Deliberately not touched

- **`Record.to_numpy()`** — its fallback returns non-numeric leaves *as-is* (avoids boxing an opaque object), a genuinely different contract from the numeric `_to_numpy_array`.
- **`_fingerprint._numeric_container_to_numpy`** — a tiered gate (returns `None` for bare arrays/scalars handled in earlier fingerprint tiers); only its metadata line moved to `_metadata_of`.
- **`__hash__` double lookup** — left as two resolver calls; `array_backend_for` is a dict probe + short MRO walk, so a special combined shape+dtype resolver wasn't worth bloating the one-per-hook API.

`_is_numeric_leaf` keeps its historical import path (re-exported from `_numeric_record`).

### Tests

Pure refactor; no behaviour change on valid leaves. Full `tests/core` (1772) plus distributions / converters / modeling / inference / diagnostics / validation (1784, 54 skipped) green. The shared-predicate regression test now asserts **both** shared predicates (the dtype-level `_is_numeric_dtype` and the leaf-level `_is_numeric_leaf`).

### DIFF
--- a/probpipe/core/_array_backend.py
+++ b/probpipe/core/_array_backend.py
@@ -216,6 +216,98 @@ def _to_jax_array(value: Any) -> jax.Array:
 
 
 # ---------------------------------------------------------------------------
+# Leaf resolvers: registry-first, duck-typing fallback
+# ---------------------------------------------------------------------------
+#
+# Every numeric gate asks one of a few questions about a leaf — is it numeric?
+# what is its shape / dtype? materialise it to numpy — and each resolves the
+# same way: the registered backend's hook if the leaf's type is registered,
+# else the numpy-protocol duck path. These helpers hold that one resolution
+# (mirroring ``_to_jax_array`` for ``to_jax``) so each question's duck fallback
+# lives in one place and cannot drift between call sites.
+
+# Numeric Python scalars. A bare scalar leaf is normalised to a 0-d
+# ``jax.Array`` at ``NumericRecord`` construction, but it still counts as
+# numeric everywhere the raw value is inspected.
+_NUMERIC_SCALARS = (bool, int, float, complex, np.integer, np.floating, np.bool_)
+
+
+def _is_numeric_leaf(value: Any) -> bool:
+    """Whether *value* is a numeric leaf — the shared numeric predicate.
+
+    Registry-first: a registered container answers through its ``is_numeric``
+    hook; a numeric Python scalar counts; everything else duck-types on a
+    numeric ``dtype`` / ``shape``. Strings, object arrays, and opaque types are
+    not numeric. Every "is this leaf numeric?" site shares this one decision,
+    so promotion, ``_full_array_shape_or_none``, and spec validation cannot
+    drift apart on what counts.
+    """
+    if isinstance(value, (str, bytes)):
+        return False
+    if isinstance(value, _NUMERIC_SCALARS):
+        return True
+    backend = array_backend_for(value)
+    if backend is not None:
+        return backend.is_numeric(value)
+    from .event_template import _is_numeric_dtype
+
+    return hasattr(value, "dtype") and hasattr(value, "shape") and _is_numeric_dtype(value.dtype)
+
+
+def _event_shape_of(value: Any) -> tuple[int, ...]:
+    """The numeric leaf's event shape — registry ``event_shape``, else ``.shape``.
+
+    Reads container metadata only, never the values. A scalar reports ``()``.
+    Assumes *value* is numeric (see :func:`_is_numeric_leaf`).
+    """
+    backend = array_backend_for(value)
+    if backend is not None:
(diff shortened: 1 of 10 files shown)


arob5:
**Upward dependency inversion + per-call lazy import on a hot path**

`_is_numeric_leaf` (`probpipe/core/_array_backend.py`) does `from .event_template import _is_numeric_dtype` *inside the function*, on the duck path. This is forced today: `event_template` imports `_array_backend` at module top (`event_template.py:71`), so `_array_backend` can't import `event_template` at top without a cycle.

But `_is_numeric_leaf` sits on genuine hot paths — the `NumericRecord` construction leaf gate (`_numeric_record.py:268`), workflow-result coercion (`_workflow_result.py:71`), and template inference (`_full_array_shape_or_none`) — and bare numpy/jax arrays (the common case) always take that duck path. Previously the `_is_numeric_dtype` import was module-level (zero per-call cost) in `_numeric_record`.

Suggested fix: move `_is_numeric_dtype` (and the trivial `_NUMERIC_KINDS = frozenset("biufc")`) **down into `_array_backend.py`**, co-locating the dtype-level predicate with the leaf-level resolver that wraps it. That:

- removes the per-call lazy import,
- removes the `_array_backend → event_template` upward edge, and
- is arguably the correct home now that the leaf resolvers live in `_array_backend`.

It's a wider sweep — four modules import `event_template._is_numeric_dtype` (`_record_array`, `_broadcast_distributions`, `event_template`, `record/design`) and `test_numeric_record.py::test_numeric_dtype_predicate_shared` pins those paths — so it's a fair call whether it belongs here or in a follow-up. If deferring, a one-line comment at the lazy import noting the cycle it avoids would help.

jhuggins:
@arob5 Done in 66212c4 — took the move now rather than deferring. `_is_numeric_dtype` and `_NUMERIC_KINDS` now live in `_array_backend`, co-located with the leaf resolvers that wrap the predicate. That removes the per-call lazy import (three sites, including both `_register_*` hooks) and the `_array_backend -> event_template` upward edge, so the DAG is clean (`_array_backend <- event_template <- record <- _numeric_record`). `event_template` re-exports `_is_numeric_dtype` (an explicit `as` re-export) so `_record_array` / `_broadcast_distributions` / `record.design` and `test_numeric_dtype_predicate_shared` need no changes; the test also pins the new home. Full `tests/core` + dependent suites green.

PR #366, opened 2026-07-16
refactor(core)!: drop from_nested_dict and _flatten_paths

jhuggins:
## Summary

Removes `Record.from_nested_dict` / `NamedTree.from_nested_dict` and the private
`NamedTree._flatten_paths`. Under the current **"a mapping is never a leaf"**
invariant the constructor already materialises every nested mapping value into a
subtree, so `Record(name, data)` reads a nested dict directly and these helpers
earned nothing extra.

Follow-up to #360, proposed by @arob5 in a review comment there. Public API is
removed with **no deprecation** (per discussion).

## Why

`from_nested_dict` predates the mappings-are-never-a-leaf invariant. On current
`main` it reduced to `cls(name, _flatten_paths(data), ...)`, and `_flatten_paths`
recurses into every mapping before handing the result back to the constructor —
which re-materialises the same structure anyway. Verified equivalence:
`Record(name, data) == Record.from_nested_dict(name, data)` for any nested dict.

Removing it also **tightens validation**. `_flatten_paths` and the constructor's
`_unflatten_paths` resolve a malformed `/`-path-vs-nested-dict collision
differently, so `from_nested_dict` silently accepted input the constructor
correctly rejects:

```python
Record("r", {"y/a": 1.0, "y": {"b": 2.0}})               # ValueError (correct)
Record.from_nested_dict("r", {"y/a": 1.0, "y": {"b": 2.0}})  # silently -> ['y/a', 'y/b']
```

## What changed

- **Migrated the two `_flatten_paths` callers** — `Record.ensure` and the
  workflow-output wrap in `_workflow_result` — to pass the dict straight to the
  constructor (behaviour-identical apart from the collision now raising).
- **Removed** `NamedTree.from_nested_dict`, the `Record` override, the
  `RecordArray` `NotImplementedError` stub, and `NamedTree._flatten_paths`.
- **Tests**: round-trips migrated to `Record(name, r.to_nested_dict())`; the
  `from_nested_dict`-named cases became constructor tests; added a test that the
  `/`-collision now raises.
- Swept docstrings, `design/02`, `STYLE_GUIDE.md`, `CONTRIBUTING.md`,
  `docs/api/records.md`, and `CHANGELOG.md`.

Net **-69 LOC** in `probpipe/core`.

## Migration (public API)

- Nested-dict construction: `Record.from_nested_dict(name, data)` ->
  `Record(name, data)`.
- Round-trip: `Record.from_nested_dict(name, r.to_nested_dict())` ->
  `Record(name, r.to_nested_dict())`.

## Verification

- Full `tests/core` passes (1777) on this branch rebased onto current `main`.
- `ruff format --check` and `ruff check` clean.


arob5:
**Stale convention-doc entry the sweep missed.** `CONTRIBUTING.md:537` — the "Key abstractions" table still describes `NamedTree` as providing *"nested-dict (de)construction."* The parenthesized *(de)construction* reads as a paired reader/writer API, i.e. exactly the `from_nested_dict`/`to_nested_dict` pair this PR splits up.

This PR swept the same phrasing out of `STYLE_GUIDE.md` and `design/02-shared-abstractions.md` but didn't update `CONTRIBUTING.md` (the PR body's "What changed" also lists the swept docs and omits it).

Suggested fix — mirror the wording already landed in `design/02`:

> …structural edits (`merge` / `without` / `replace` / `with_path_names`), and nested-dict export (`to_nested_dict`) that the constructor reads back.

PR #367, opened 2026-07-16
docs(design): target package structure

jhuggins:
## Summary

Adds **`design/package-structure.md`**, a companion to the design reference's numbered parts: the target package and module layout that realizes the reference as an import architecture. Design prose only, no code changes.

## What it fixes

- **One package per layer**, in the reference's dependency order: `core/` (Part II) → `values/` (III.1–III.4, all leaf value kinds) → `linalg/` (III.5) → `distributions/` (III.6–III.14) → `functions/` (Part IV) → `operations/` (Part V) → `families/` (Part VI) → `inference/` / `diagnostics/` / `validation/`.
- **Imports point strictly downward; registration flows upward.** Lower layers define registries; higher layers populate them at import time (families register evaluation rules and converters, inference methods register themselves). Defining the inference-method registry with `condition_on` in `operations/` also fixes today's one inverted edge (`core/ops.py` importing the registry from `inference/`).
- **A spec lives with the type it admits**: `ArraySpec`/`OpaqueSpec`/`FunctionSpec` in `values/`, `DistributionSpec`/`ConditionalDistributionSpec` with the distribution classes — and the same rule places each type's batch form.
- **Modules private, packages public, one curated top-level namespace** as the only import a user needs; tests mirror the tree.

## Grounded in the implementation's seams

The module boundaries follow seams the code already has, so the layout is reachable by moves rather than rewrites — a correspondence table in the doc maps each load-bearing move:

- **`functions/` is the Function engine, kept whole.** Today's `node.py` plus the seven `_workflow_*` modules are one ~3,400-line machine (call classification, broadcast planning and grouping, the sampling lift, the batch sweep, execution dispatch, result wrapping) whose stages change together; each target module maps 1:1 onto an existing `_workflow_*` file. It sits **above** `distributions/` because that is what its imports already say: the lift samples distributions, materializes `EmpiricalDistribution` results, and consults the converter registry.
- **`operations/` is thin, matching what `ops.py` actually is**: capability-dispatched definitions wrapped by the decorator, one module per Part V section, plus the registries the operations own.
- **`EmpiricalDistribution` lives in `distributions/`**, matching its current `core/` placement: it is the closure family the lift and every Monte Carlo fallback construct, so it must sit below the machinery that uses it — the single deliberate exception to part-per-package (its Part VI entry is unchanged).

## The `Function` unification

The document is written against the decided direction that the computation abstraction becomes **`Function`**, the universal function representation: a tracked term with input and output templates, what a `FunctionSpec` admits and a `FunctionBatch` holds, with `LinOp` its linear subtype and invertibility a capability rather than a `Bijector` class.

The import graph forces a representation/engine split: the function value kind's **base** (what `LinOp` subclasses and the spec references) lives in `values/` with the other leaf value kinds, below `distributions/`; the **engine** lives in `functions/` above them. The detailed `Function` contract (decorator, call behavior, the base's exact surface, how the engine attaches) is being drafted separately and is flagged as an open point — this document constrains only placement, so the two efforts don't block each other.

## `pushforward` → `evaluate`

This PR also renames the pushforward operation to **`evaluate`**, swept across `design/` and this package layout (`operations/_evaluate.py`, evaluation rules, VI.4 retitled *Evaluation results*). Two reasons:

- **The op is broader than a pushforward, and simpler to define.** `evaluate(f, v)` is the call `f(v)` with the rule registry layered on top: the built-in call already covers every operand kind (plain call, sweep, sampling lift), and `evaluate` consults the registry first, falling back to exactly that call — so the two differ only where a registered rule improves on sampling, and the lift/registry layering is unchanged.
- **It follows the reference's own naming convention** (the interface name in prose, the mathematical concept in math statements, exactly as *kernel* is reserved for `ConditionalDistribution`'s definition). "Pushforward" survives as mathematics — `f♯μ` in the operation table and V.6, C4's descriptor, the transformed-family label, the affine-pushforward closure statement — and in the `LinearPushforwardDistribution` class name.

cc @arob5 @SoymilkWinsAgain


arob5:
**"one module per Part V section" slightly overstates the mapping.**

The `operations/` layer bullet and the tree realize V.1–V.7 (seven modules: `_moments`, `_sample`, `_density`, `_condition`, `_joint`, `_pushforward`, `_marginal`). But V.8 ("Batched operations") has no `operations/` module — batching is the `functions/_sweep.py` engine — and V.0 ("The operation model") is the decorator itself, not a module. Consider narrowing the phrasing to "one module per operation section (V.1–V.7)" so it matches the actual layout.

arob5:
**Two `_function.py` modules across packages may invite confusion.**

The layout puts the function kind's base in `values/_function.py` and the engine (holding `Function`) in `functions/_function.py` — same basename in two packages. Since the base class's name is itself flagged as an Open point, a distinguishing name for the value-layer module (e.g. `_function_kind.py`) may age better and make imports/greps unambiguous. Worth revisiting once the forthcoming `Function` design fixes the base's name.

jhuggins:
**`pushforward` → `evaluate`** (e01afc2). The operation is renamed and **broadened**: `evaluate(f, v)` is total over the operand kinds — a native value returns `f(v)`, a batch the elementwise sweep, and a distribution the law of `f(X)` for `X ~ v`, the pushforward `f♯v`. The contract pins the relation to the lift precisely: for a value or a batch, `evaluate(f, v)` and the lifted call `f(v)` agree exactly, and for a distribution the registry fallback *is* the lift, so the two differ only where a registered rule improves on sampling. The B-cluster layering (the lift never consults a registry) is deliberately unchanged.

Naming rationale, now stated in V.6's Rationale: the operation is named for what it does across every operand kind, with *pushforward* reserved for mathematical statements — the same convention that reserves *kernel* for `ConditionalDistribution`'s definition. So `f♯μ` stays in the math, C4 keeps its descriptor, III.12's family label stays "Transformed (pushforward)", and `LinearPushforwardDistribution` keeps its name.

Swept: V.6 (retitled, widened contract, "evaluation rules"), V.8's map-operand exemption, III.7, VI.0/VI.2/VI.4 (retitled *Evaluation results*)/VI.5/VI.6, the README operation table (the evaluation and pushforward rows merge), and this PR's package layout (`operations/_evaluate.py`).

jhuggins:
Both taken:

- **Two `_function.py` modules** — agreed; the value-layer module is now **`values/_function_base.py`** (f1914db). The `functions/` hub keeps `_function.py` on the package-hub pattern (`linalg/_linop.py`, `distributions/_distribution.py`), while `_function_base` follows the existing `_distribution_base.py` naming and sorts next to `_function_batch.py`. It also stays correct whatever the forthcoming `Function` design names the base class, which remains that design's call.
- **"One module per Part V section"** — right, it overstated the mapping. Narrowed to "one module per operation section (V.1–V.7)" with the two exclusions now stated: V.0's operation model is the wrapping itself, and V.8's batching is the engine's sweep (latest commit). The correspondence table row for `ops.py` says the same.

SoymilkWinsAgain:
Uh, I have a problem. I've been working on how to avoid adding a `Function` class yesterday. The solution is almost there...

Issue #370, opened 2026-07-17
Proposal: define a controlled fitted-mapping producer contract


SoymilkWinsAgain:
## Parent

Tracking issue: #252

## What to establish

Define a controlled producer contract for deterministic mappings created by `fit`.

The contract answers a narrow question: when may `fit` publish an ordinary `Function` as a reusable fitted mapping?

Core should validate checkable invariants:

- authoritative, immutable, schema-complete input and output templates;
- total correspondence between the Python signature and top-level input fields;
- an event-valued application result plan;
- all free output dimensions declared by input dimensions;
- private ownership of the implementation carrier;
- conformance to any plan-owned common schema.

Adapters should attest to properties Python cannot prove in general:

- deterministic versus probabilistic application semantics;
- reentrancy or required serialization;
- schema-resolution regime;
- implementation ownership and absence of externally reachable mutable aliases.

A private immutable certificate may preserve these attestations for diagnostics, but ordinary consumers must not dispatch on it.

## Why this is necessary

A nominal new subclass would not prove determinism, schema stability, ownership, or thread safety. `fit` nevertheless needs stronger admission rules before fitted atoms enter `FunctionBatch` or `RandomFunction`. Putting the guarantee at the producer boundary gives `fit` a checkable contract without narrowing every `Function` or every callable-valued abstraction.

## Construction boundary

Core should own a controlled factory equivalent to:

```python
_make_fitted_function(
    implementation,
    *,
    declaration_ref,
    signature,
    input_template,
    output_template,
    plan,
    producer_attestation,
) -> Function
```

Ordinary declarations use public `@function`; fitted mappings use this controlled factory.

The factory validates, installs private state, constructs an ordinary tracked `Function`, and attaches training provenance.

## Acceptance criteria

- [ ] Fitted point mappings are ordinary `Function` instances.
- [ ] Unsafe atoms fail before publication.
- [ ] Structural checks and producer attestations have an explicit boundary.
- [ ] Uncontrolled application randomness cannot be published as point `Function`.
- [ ] No stable callable abstraction is narrowed.
- [ ] No consumer-visible behavior depends on construction origin.
- [ ] Attestations remain inspectable without becoming public ontology.

## Dependencies

Depends on first-class `Function` (#368) and closed result planning (#369).

Issue #368, opened 2026-07-17
Make `Function` a first-class tracked term


SoymilkWinsAgain:
## Parent

Tracking issue: #252

## What to establish

Make `Function` a first-class, immutable, tracked `ProbPipe` term that may be declared statically with `@function` or produced dynamically by another operation.

This is the minimum core change needed for a deterministic `fit` result to remain an ordinary `ProbPipe` mapping.

The public contract should cover:

- reusable deterministic mapping semantics;
- `Tracked` name and provenance;
- authoritative signature, input template, and output template;
- a private `FunctionImplementation` that owns execution state;
- one invocation planner for decorated and dynamically produced Functions;
- call-local binding and reentrant execution.

For fitted estimators, the mutable backend object remains private implementation state. The resulting public object is still an ordinary `Function`.

## Why this is necessary

A point estimator must return a reusable mapping with training lineage. If dynamically produced Functions are not first-class and tracked, `fit` either needs a parallel public ontology or loses provenance when its result is later applied.

This change also prevents estimator integration from becoming a second callable hierarchy.

## Compatibility constraints

- Plain callable wrapping remains valid.
- Existing callable-valued abstractions remain general.
- `FunctionSpec`, `FunctionBatch`, and `RandomFunction` are not narrowed to `Function`.
- Universal artifact identity and portable serialization are not prerequisites.
- No consumer may branch on whether a Function came from `fit`.

## Acceptance criteria

- [x] Static and dynamically produced `Functions` use the same call planner.
- [x] `Function` is a `Tracked` first-class term.
- [x] A produced Function can carry training provenance and remain a parent of later results.
- [x] Backend state is reachable only through the private implementation carrier.
- [x] Signature and authoritative templates have one coherent contract.
- [x] Call-local symbolic binding is reentrant.
- [x] No second callable hierarchy, estimator-specific wrapper, or hidden fitted subtype is required.
- [x] Existing plain-callable use cases remain valid.

## Dependencies

Blocked by the base rename in #377. This is the foundation for the remaining sub-issues under #252.

Issue #350, opened 2026-07-10
Design-alignment roadmap: implementation waves toward the design reference


jhuggins:
# Design-alignment roadmap — implementation waves (tracking)

> **📌 Standing directive.** As with #235: read **`CONTRACTS.md`** at the repo root before implementing any item here, and fully document every touched contract in docstrings. Where #235's older phase plan or any pre-design issue disagrees with the design reference, **the design reference (`design/`) wins.**

> **Last updated 2026-08-19 — status refresh + two corrections.** Three checklist items have merged since the last snapshot and were still shown open: **W1.3** (#397), **W3.1** (#394), and **W3.2** (#399 / #401 / #402 / #404, which removed `RecordArray` and `NumericRecordArray` outright). **W4.1 is partly delivered** — see its item for what is done and by which PR. Two corrections: W3.2's "adjacent hole" no longer exists, and the #369 line in the dependency backbone reflects a framing that has been superseded; both are marked in place. One prerequisite added: #441 blocks W2.1. **2026-08-21:** #441 closed as superseded by #443 (core-simplification proposal); W2.1's added prerequisite is now #443.
>
> **Last updated 2026-07-22 — wave reorder + whole-number renumber.** Two changes this revision. **(1) Reorder:** the `Distribution`-base consolidation and Record-tower dissolution move **ahead of** the concrete batch types. Implementing #368 (PR #380) surfaced the schema-ownership debt directly: `event_template`, `dtypes`, and `supports` are overlapping — potentially contradictory — schema sources living on the `RecordDistribution` tower, against D7's single-source rule. The hierarchy items need only the `Batch` **ABC** (per their own dependency annotations), so doing them first kills that debt at its root: the tower's scattered accessors die with the tower instead of being consolidated onto classes slated for deletion, and the concrete batch types then share the *authoritative* `event_template` from day one. **(2) Renumber:** the fractional waves (0.5, and the reorder would have forced a 1.5) are renamed to whole numbers. Old→new id map at the bottom of this note; the `W<wave>.<n>` ids below are the new scheme, with *(was …)* tags on every renumbered item.
>
> | old id | new id | item |
> |---|---|---|
> | W0.5.1 | **W1.1** | `Function` base (#380, merged) |
> | W1.1 | **W1.2** | `Batch[E]` ABC (promoted keystone) |
> | W0.5.2 | **W7.6** | `SupportsDifferentiation` (delayed; first consumers are in the map layer) |
[...]

PR #376, opened 2026-07-17
docs(design): the Function base and call engine, capability claims, and the operation model

jhuggins:
Aligns the design documents around the function kind and the operation model, building on the `evaluate` rename of PR #367. Design docs only — no code changes. Each decision is its own commit.

## The `Function` kind

**The base type is named `Function`** (was `Computation`), for the mathematical object it denotes — a function — and consistent with the types already named for it (`FunctionSpec`, `FunctionBatch`, `RandomFunction`). Part IV is retitled and its file renamed (`04-computations.md` → `04-functions.md`); the decorator becomes `@function`; the wrapped object is the `Function` and the raw thing it wraps a "plain callable."

**The base lives in the value layer** (§III.2, beside the spec that admits it): a tracked term wrapping exactly one callable, carrying `name`, `provenance`, and the input/output templates, all fixed at construction. It also carries its **controls** — the execution defaults, set at construction and revised functionally by `with_options` (`C2`) — which are opaque to the base and read by the engine at call time. Its own call is plain evaluation; the call path is the base's one extension point (`install_call_engine`, filled once at import).

**Part IV is the engine**: the call semantics beyond plain evaluation — argument classification, broadcast planning and co-sampling groups, the lift and the sweep, key splitting, execution dispatch, orchestration, result wrapping, provenance — installed on the base at import by the same upward registration the design uses everywhere. The engine contract states its obligations: on concrete values it agrees with plain evaluation, adding only the wrap and provenance, and it consults the evaluation-rule registry at each single-distribution application. §IV.0 carries a landing map giving each concern one home and its extension status.

**`LinOp` is the linear `Function` subtype** (§III.5): the map it wraps is the operator's action, and the operator algebra and structured queries are what linearity adds. With the base in Part III, the subclassing is a backward reference and imports point strictly downward. Two open points close as a consequence: §V.6's transport map (the first-class transport map is `Function` itself) and `package-structure.md`'s `Function` contract (the base class is `Function` in `values/_function_base.py`; the engine in `functions/` attaches at import).

## Capabilities are declared claims

**Invertibility**: the `SupportsInverse` capability replaces the `Bijector` class — a `Function` provides the inverse and the forward log-det-Jacobian, its `apply` the forward map — with *bijector* kept as a mathematical term, as *kernel* and *pushforward* are. Bijector-valued signatures annotate the protocol, the `Function` carrier understood; an invertible `LinOp` claims the same capability, its `logdet` supplying the Jacobian term. Constraint reparameterization, change-of-variables, and GLM links are retyped accordingly.

**Differentiation**: `D6` is restated as "Differentiability as a capability," and the `SupportsDifferentiation` protocol (the `Supports*` family names actions) is defined beside the base. The claim is declared at construction — the decorator's `differentiable` flag, `False` by default, asserts the callable differentiates end-to-end — and is consumed wherever gradients are required. It is a declaration, not a control, and is distinct from traceability: `auto` execution dispatch probes whether the call is array-traceable, which is what `jit` and `vmap` need.

## `Distribution` and `ConditionalDistribution`

A `Distribution` is the special case of a kernel with an empty conditioning value — nothing to condition on — so the marginal law exists and the unconditional operations apply (§III.8). The two stay distinct types, neither inheriting from the other: which unconditional operations an object carries turns on whether its given is empty, and each unconditional capability is mirrored by a conditional twin that prepends the given.

## The operation model

**`evaluate` leads with its denotation** (§V.6): applied to a value it returns `f(v)`; to a distribution, the pushforward `f♯v`; to a batch, the elementwise result.

**Every operation states its fallback status** (§V.0): a fallback with its domain (Monte Carlo on the samplers), total by definition (`sample`), or capability-only (`log_prob`, `factor`). The status is registry data, not contract — a numerical default registered later moves an operation without changing its signature.

**The evaluation-rule registry is consulted by the call itself.** The registry is defined beside the engine (`functions/_rules.py`) and populated upward by the families; every single-distribution application resolves through it, so the direct call `f(v)` and `evaluate(f, v)` take the same route and agree — `evaluate` is the operation form, adding `method=` selection and the operation's provenance record. Rules need not be exact: approximate schemes (quadrature, unscented transforms, quasi-Monte Carlo) register at recorded fidelity, and **the sampling lift is the rule registered at the generic pair** — the registry's floor and the route every plain callable takes, since rules key on the map's type. Grouped, multi-distribution lifts always take the sampling path, which co-sampling requires.

## Scope

Design docs only. The corresponding code moves are separate follow-ups: `WorkflowFunction` → `Function`, the `*Array` batch types → `*Batch`, and relocating the rule registry beside the engine. A writing pass keeps the added text in the documents' plain register, with rationale grounded in the design principles.


design/package-structure.md
```diff
@@ -2,7 +2,7 @@
...
 
-This document uses the **`Function`** vocabulary. A `Function` is the universal function representation: a tracked term carrying an input and an output template. `LinOp` is its linear subtype, invertibility is a capability, a `FunctionSpec` admits it, and a `FunctionBatch` holds them. Its detailed contract is unsettled (see Open points); this document fixes only where its pieces live.
+This document uses the **`Function`** vocabulary. A `Function` is the universal function representation: a tracked term carrying an input and an output template. `LinOp` is its linear subtype, invertibility is a capability, a `FunctionSpec` admits it, and a `FunctionBatch` holds them. The base class is `Function` itself, in the value layer, carrying templates, identity, and plain evaluation; the Part IV engine is installed on it at import, registration flowing upward. This document fixes where its pieces live.
```
SoymilkWinsAgain: This now seems to contradict issue #03 (which states that `FunctionSpec` accepts any `callable`). There are also many other unstable parts. I think you can refer to my new version #252 and its sub-issues, as well as the results we discussed and finalized earlier (`Computation` was renamed `Function`).

design/01-design-principles.md
```diff
@@ -28,6 +28,6 @@ ProbPipe's overarching aim is *simplification via abstraction*: complexity is ab
 
 **D5 — Explicit, carried structure.** The structure of a value — its named parts and their kinds — is represented explicitly and travels with the value and with the objects that produce and consume it. Structure is propagated forward from the producer that knows it.
 
-**D6 — Differentiability where possible.** An operation is differentiable end-to-end when the objects it acts on are backed by a differentiable, array-native representation, so gradients propagate through an entire workflow; when they are not, the operation remains correct but forgoes differentiability. Support for differentiation is encouraged but not required.
+**D6 — Differentiability as a capability.** Differentiability is a capability an object claims. Operations that require gradients dispatch on it, so support is settled at dispatch and a step that cannot differentiate is named there. Differentiability composes: a workflow is differentiable end-to-end exactly when every step claims it. Support for differentiation is encouraged but not required.
```
SoymilkWinsAgain, line 31: Part III's capability list lacks a corresponding protocol and doesn't explain:

1. Differentiability with respect to which variables;
2. How `Functions`, `Distributions`, and `LinOps` are declared;
3. How composition, views, and fallback propagate;
4. How gradient-requiring operations check it.

Furthermore, the changes are essentially just name replacements for D6 (the original driver assertions are retained without modification). For example, "array-backed operator/distribution is differentiable" and "lifted call is differentiable when using `jax` dispatch." This is precisely the kind of judgment method that the new D6 claims to no longer use.

design/03-values-and-distributions.md
```diff
@@ -401,7 +430,7 @@ Making each operation a *capability* rather than a base-class method follows `D3
...
+A `ConditionalDistribution[S, T]` is a *probability kernel* `K : S → P(T)` — a family of distributions p(· | s) indexed by a *conditioning value* `s : S`. Supply a value for what it conditions on and it yields an ordinary `Distribution` over what it produces. A `Distribution` is the special case of a kernel with an empty conditioning value — there is nothing to condition on — so the marginal law exists and `sample` / `log_prob` / `mean` apply unconditionally. A kernel with a non-empty given has no marginal law. The two are therefore distinct types, neither inheriting from the other: which unconditional operations an object carries turns on whether its given is empty, and the types mark the two cases. They are siblings sharing a capability vocabulary, each unconditional capability mirrored by a conditional twin that prepends the given.
```
SoymilkWinsAgain: `Distribution` is an empty-given corner that the empty-given kernel currently does not have a representative for. The design documentation does not specify:

1. What EventTemplate to use for the empty conditioning value;
2. Whether `ConditionalDistribution` should disable empty `given_template`;
3. Whether an empty-given `ConditionalDistribution` is automatically normalized to a `Distribution`;
4. Whether `ConditionalDistributionSpec` accepts `Distribution` in this corner.

The current `EventTemplate` rejects zero-field structures. Either change this setting or explicitly introduce a normalization rule here.

design/03-values-and-distributions.md
```diff
@@ -644,29 +673,26 @@ Conversion makes `C3 – Computational detail hidden by default, available on de
...
+Gradient-based and Hamiltonian inference work in an unconstrained space, so a constrained support must be reparameterized. The **constraint-to-bijector factory** maps a `Constraint` (the support an `ArraySpec` carries) to a *bijector*: a `Function` that takes `ℝⁿ` onto that support and claims `SupportsInverse`. Invertibility with a tractable Jacobian is a capability: a `Function` claims `SupportsInverse` by providing the inverse map and the forward log-determinant of its Jacobian, its own `apply` serving as the forward map. A `LinOp` claims it the same way, its inverse supplied by the operator algebra and its `logdet` the log-Jacobian, so one capability labels invertible maps, linear or not. Bijector-valued signatures annotate the protocol, with the `Function` carrier understood. `bijector_for(constraint)` returns the canonical such `Function`, and `register_bijector` plugs in a factory for a constraint type or for a specific constraint instance, with instance registrations taking precedence over type registrations.
```
SoymilkWinsAgain: This section mentions that `Function.apply` is a forward map, but Part IV does not define `apply`. Furthermore, the types of `canonical_link: Function` and `BijectorTransformedDistribution(..., bijector: Function)` themselves do not guarantee `SupportsInverse`.

It should be explicitly stated whether `__call__` or `apply` is used for forwarding, and runtime capability validation at construction time should be specified; if the capability is directed to all Functions, its input and output types should not be unambiguously fixed as Array -> Array.

SoymilkWinsAgain (changes requested):

design/03-values-and-distributions.md
```diff
@@ -687,14 +686,13 @@ Conversion makes `C3 – Computational detail hidden by default, available on de
...
...reparameterized. The **constraint-to-bijector factory** maps a `Constraint` (the support an `ArraySpec` carries) to a *bijector*: a `Function` that takes `ℝⁿ` onto that support and claims `SupportsInverse`. Invertibility with a tractable Jacobian is a capability: a `Function` claims `SupportsInverse` by providing the inverse map and the forward log-determinant of its Jacobian, its own `apply` (III.2) serving as the forward map. A `LinOp` claims it the same way, with its inverse from the operator algebra and its `logdet` the log-Jacobian, so one capability marks invertible maps, linear or not, over structured numeric values as well as bare arrays. A signature that expects a bijector annotates `SupportsInverse`, and a slot that requires it — `BijectorTransformedDistribution`'s `bijector`, a GLM `canonical_link` — checks the claim at construction and raises a capability error otherwise, the same `isinstance` check the distribution capabilities use. `bijector_for(constraint)` returns the canonical such `Function`, and `register_bijector` plugs in a factory for a constraint type or for a specific constraint instance, with instance registrations taking precedence over type registrations.
```
SoymilkWinsAgain: A small but significant issue remains: `SupportsInverse` does not include a forward mapping contract. The documentation directly labels parameters like `bijector` and `canonical_link` as `SupportsInverse` and verifies them using `isinstance`, but this protocol only requires `_inverse` and `_forward_log_det_jacobian`, and does not guarantee that the object has the subsequently required `apply`. Any non-Function object implementing these two methods will pass the check, but will fail during forward sampling or transformation.

design/03-values-and-distributions.md
SoymilkWinsAgain, line 134: I believe we should also check if `_differentiable` is `True`, not only if it exits.

Furthermore, `ClassVar` is inconsistent with the instance-specific `differentiable=` declaration. A dynamic typing/member mechanism that explicitly states "only the declarer owns the member" is needed, or an explicit predicate should be used instead of the current pseudocode implementation.

SoymilkWinsAgain (changes requested):
Also, some problems from AI:

1. The `evaluate` function makes unrealizable guarantees regarding direct calls to ordinary callables. The same passage states that `f` can be an ordinary Python callable, yet also that a direct call to `f(v)` will go through the evaluation registry. Unwrapped Python functions cannot intercept `__call__`; only Function/LinOp functions can. This equivalence relation should be restricted, specifying how `evaluate(plain_callable, v)` internally wraps or follows the generic rule.

2. Fallback status is referred to as registry data, but the documentation does not define a corresponding registry/metadata schema, nor does it declare the status of each operation item by item. It is recommended to add an `operation-status` table and a registration interface; otherwise, the "registering a numerical default to change the status" functionality cannot be implemented at present.

3. `functions/_rules.py` is defined as the location of the registry, but `operations/_evaluate.py` is still described as containing the rule registry.

design/04-functions.md
```diff
@@ -108,15 +126,15 @@ The `seed` fixes randomness deterministically and *structurally*. It is split al
...
+- **Dispatch — *how* the per-draw / per-element calls run.** `jax` vectorizes them (one `vmap`); `sequential` runs them one at a time; `thread` runs them on a thread pool; `auto` probes whether the call is array-traceable and picks `jax`, falling back to `sequential`. Under `jax`, a lifted call is traced end-to-end, and it differentiates end-to-end when the `Function` claims `SupportsDifferentiation`; dispatch never changes the result beyond floating-point effects of evaluation order. Because each unit's PRNG key is fixed by its index rather than taken from a shared counter, the result is identical across `jax`, `sequential`, and `thread`, and parallel execution contends for no mutable random state: nothing is locked, and no key is drawn twice.
 - **Orchestration — *whether* the call is traced.** Off by default. A computation can instead run as a traced task or flow, recording the computation graph for lineage and scheduling. Tracing never changes the result.
```
arob5: Stale rename: this still uses the old type name — "**A computation** can instead run as a traced task or flow". Should be "A `Function` can instead run…". (The "computation graph" later in the same sentence is fine — that's the dataflow-graph noun, not the renamed type.)

design/04-functions.md
```diff
@@ -13,33 +13,51 @@ Wrapping a function `f` layers four things on top of calling `f`, each defined i
...
 
 ## IV.2 — Lifting over distributions and batches
 
 ### Contract
 
-A computation compares each argument against the type its function expects, and lifts where they differ. Lifting always proceeds by **sampling**, whatever the function.
+A `Function` compares each argument against the type its function expects, and lifts where they differ. A single-distribution application resolves through the evaluation-rule registry (V.6), where the sampling lift is the rule registered at the generic pair: a plain callable therefore lifts by **sampling**, and a typed map takes its registered rule. Grouped, multi-distribution lifts always take the sampling path, which is what co-sampling requires.
 
 **The trigger.** A parameter that is unannotated, or annotated with a value type, expects a value, so a distribution passed in that position is lifted. A parameter annotated `Distribution`, `Distribution[...]`, or any `Supports*` protocol declares that the function consumes the distribution itself, which then passes through unlifted. Per draw, the function receives what `sample` returns, the draw's `Record`.
```
arob5: The lifting trigger keys on "any `Supports*` protocol" ⟹ "consumes the distribution itself". This PR broadens the `Supports*` family to also name *function* capabilities (`SupportsInverse`, `SupportsDifferentiation`), which annotate `Function`-typed slots rather than distributions. A parameter annotated `SupportsInverse` is a `Function`, so the blanket "any `Supports*` ⟹ consumes the distribution" rule now over-reaches. Suggest narrowing to the distribution capability protocols of §III.7 (e.g. "any distribution capability protocol").

design/03-values-and-distributions.md
```diff
@@ -94,10 +94,56 @@ As the *type layer*, an `EventTemplate` is the explicit structure that travels w
...
+
+A `Function` is authored with the `@function` decorator or produced by an operation; both use the same call path, and a produced `Function` carries its provenance like any other tracked term. Its capability protocols live beside the base: `SupportsInverse` (III.14) and `SupportsDifferentiation` are claims a `Function` carries, declared at construction and checked like the distribution capabilities of III.7. `SupportsDifferentiation` additionally declares *which* values differentiate, read through `is_differentiable`, as described below. The `Function` base is the tracked *wrapper*, not a restriction on what may be wrapped: the value-layer specs stay **callable-generic**. A `FunctionSpec` admits any callable — a plain lambda, a NumPy function, a `Function` — the `Function` being one such, not the required type, and a `FunctionBatch` holds a collection of them. No operation branches on whether a callable arrived bare or wrapped.
```
arob5: Minor tension: "Its capability protocols live beside the base: `SupportsInverse` (III.14) and `SupportsDifferentiation`…". Saying the protocols live *beside the base* (§III.2) while `SupportsInverse` is specified in §III.14 reads as self-contradictory. It's reconcilable — `package-structure.md` puts both in `values/_function_base.py`, so "beside the base" is the code location and III.14 the doc section — but a reader will trip. Consider "specified in III.14" or a half-clause distinguishing the two.

design/03-values-and-distributions.md
```diff
@@ -214,12 +264,12 @@ A `RecordBatch` makes `D1 – Mathematical fidelity` concrete on the value side:
...
 
 Its templates are always concrete, and construction from a template with unbound dimensions raises. A consumer whose sizes are not yet known holds the operator as a recipe, the operator class and its size-free parameters, and mints the instance once the sizes are bound. The base fixes the action and the square-only queries, and every query raises `LinAlgError` where it is undefined:
 
 ```python
-class LinOp(Tracked, ABC):
+class LinOp(Function, ABC):        # the linear subtype of the III.2 base
```
arob5, line 273: Minor: how `LinOp` fills the `Function` base's call/`apply` slot is left implicit. §III.2 defines `apply` as "evaluates the wrapped callable," but `LinOp` defines its action abstractly via `matvec`/`to_dense` with no wrapped callable. One clause on how the operator's action satisfies the base's `apply` / plain-evaluation path would close the gap — and the code follow-up will have to answer it regardless.

design/package-structure.md
jhuggins: Fixed in `9fdac05`, tightened in `027cb33`. You're right the wording read as narrowing the specs. III.2 now carries the full `Function`-base contract and states the callable-generic rule outright: a `FunctionSpec` admits **any** callable — a plain lambda, a NumPy function, a `Function` — the `Function` being one such, *not* the required type, and no operation branches on whether a callable arrived bare or wrapped. The base holds its wrapped callable and state privately (a stateful map exposes no backend object), and a `Function` can be authored via `@function` or produced by an operation — the general path your estimator work in #252/#368 fits into, without the reference hardcoding it.

I folded the contract into the design doc rather than only cross-referencing the cluster, on the principle that it should flow design → implementation: III.2 is now authoritative for the base and #368 conforms to it (its criteria already line up — first-class tracked term, private state, one call path, no separate `Function` category). `Computation → Function` confirmed throughout.

One ask: what does "issue #03" refer to? Want to be sure I've captured it.

design/01-design-principles.md
jhuggins: Good catch — the rename was skin-deep, and this PR had left a contradiction. Addressed across `155a4ba`, `79e9258`, `60e33cb`, and `b342969`, which answer your four questions directly:

1. **W.r.t. which variables:** the capability now declares it — `differentiable_template` is a non-empty `NumericEventTemplate` containing exactly the differentiable values (a sub-template of the numeric input template for a `Function` or `LinOp`, of the numeric event template for a distribution).
2. **How each type declares:** a `Function` through the decorator's `differentiable` argument, which takes the template itself — the claim is always explicit (no claim-everything shorthand, so it never silently widens), and omitting the argument declares nothing; a `LinOp` claims its whole input template; a distribution family claims the event values its sampling reparameterizes.
3. **Composition/views:** template algebra — a field view restricts its parent's template, a joint assembles its factors' under their field names, and a value is differentiable through a chain exactly when every step claims it.
4. **How gradient-requiring operations check:** `is_differentiable(x, values)` — every named value must lie in the template (no argument: the template covers every numeric value) — and the operation names the first step that fails, before any backend trace.

The implicit-judgment assertions you flagged (III.5/III.6/II.3/V.8 "array-backed ⇒ differentiable", IV.4's jax-dispatch coupling) are scrubbed to declared claims. Still deferred, as a stated Open point: whether a Monte Carlo fallback differentiates through its sampler's reparameterization, and the eventual `grad` operation the claims feed (a custom gradient method, then the AD route gated by the declared template, then finite differences at approximate fidelity).

design/03-values-and-distributions.md
jhuggins: Resolved in `d90b763`, reworded in `027cb33` after checking it against the formalization. A `Distribution` is *isomorphic* to a kernel with an empty given but not identical to one, so it is coherent for the code not to reify the empty corner even though the math has it. III.8 now states that, keeps the two as distinct tracked types, and answers your four questions:

1. No empty `EventTemplate` is introduced.
2. A `ConditionalDistribution` carries a non-empty `given_template` — the empty corner is `Distribution`, authored directly.
3. No normalization rule: nothing turns an empty-given kernel into a `Distribution`, because no empty-given kernel is constructible.
4. `ConditionalDistributionSpec` likewise requires a non-empty given; the empty corner is `DistributionSpec`'s.

So `EventTemplate`'s zero-field rejection stays as-is.

design/03-values-and-distributions.md
jhuggins: Addressed across `9fdac05` / `4f020fb` / `027cb33` / `37546fd`:

- **`apply` is defined** on the `Function` base (III.2): evaluation of the wrapped callable at a point, with no tracking or lifting — the raw map change-of-variables builds on; `__call__` is the tracked call that runs the Part IV engine. We kept the name: it is exactly point-application, `f(x)`.
- **`SupportsInverse` stays minimal** (`_inverse` + `_forward_log_det_jacobian`), typed over `Array | NumericRecord` rather than fixed to arrays, so a structured map or a `LinOp` claims it too.
- **Construction-time validation** is stated, and the typing hole you found in the follow-up is closed — see that thread: slots are typed by the `Function` carrier with the claim checked at construction, so the forward and the inverse are both guaranteed.

design/03-values-and-distributions.md
jhuggins: Right — typing the slots `SupportsInverse` traded one hole for another: any object with the two methods passed `isinstance`, then failed at the forward. Fixed in `37546fd` by keeping the protocol minimal and getting the forward from the **carrier**: a bijector-valued slot (`BijectorTransformedDistribution.bijector`, `canonical_link`, `glm_likelihood(link=)`) is typed `Function`, and construction checks that the argument claims `SupportsInverse` — two checks that together guarantee the full contract (the `Function` supplies `apply`; the claim supplies the inverse and the Jacobian). `bijector_for` returns, and `register_bijector` factories produce, such `Function`s — stated in their signatures.

design/03-values-and-distributions.md
jhuggins: Both points were right, and they pushed us to a better design than patching the flag (`60e33cb`, `b342969`). The `ClassVar` bool is gone: `SupportsDifferentiation` now provides a `differentiable_template` property — a non-empty `NumericEventTemplate` containing exactly the differentiable values — so `isinstance` is an ordinary structural check again (the property is a real member), and the explicit predicate exists: `is_differentiable(x, values)` checks the named values against the template, or, with no argument, that it covers every numeric value; it is `False` when the capability is undeclared. An object with nothing to claim does not declare the capability at all — the template is non-empty by construction — so there is no `False`-flag ambiguity and no existence-versus-value gap.

arob5:
Thinking more about some of the `LinOp` comments above -- as the plan currently stands, `LinOp` feels a bit forced into the `Function` hierarchy. I think this is because a `LinOp` is usually conceptually thought of as a lazy array, and the class is designed with this in mind (e.g., requiring a `to_dense` method). 

The `Function` constructor wraps a callable, but `LinOp` currently requires defining `to_dense` and `matvec`. It's currently unclear to me how this tension is resolved. It seems that `Function.apply` serves the same purpose as `LinOp.matvec`. `LinOp.matmat` exists for cases where there are more efficient routines for matrix-matrix operations, but this seems to overlap with `Function` batching logic.

Conceptually, it is also perhaps confusing to have a variable representing a covariance matrix carry around things like `n_broadcast_samples` and `seed`. 

If strict inheritance is maintained, we need to think more about these issues. Another route could be relaxing to protocol conformance, in which case a `LinOp` could satisfy the structure required of a `Function`.

arob5:
**Does a direct call record the selected rule's fidelity?**

The reframing moves the evaluation-rule registry down into the engine's call path, so `f(v)`, `A @ v`, and `evaluate(f, v)` take one route (IV.1 engine contract, §V.6). Combined with "rules need not be exact" — quadrature, unscented transforms, QMC "register at their recorded fidelity" — this means the **direct call** `f(d)` can now silently select an *approximate* closed-form rule, not just the sampling floor.

But the honesty guarantee is stated only on `evaluate`'s side:

- §V.6: *"The result records which rule produced it, mirroring the converter registry's recorded fidelity"* appears under `evaluate`'s contract.
- Both IV.1 and §V.6: *"`evaluate` is the operation form, adding `method=` selection and the operation's provenance record."*

So it's ambiguous whether a direct call stamps the selected rule's fidelity into the result's provenance, or whether only `evaluate` does. In the old model this couldn't bite, because the plain call *always sampled* (a single known fallback) and `evaluate` was the only path that could pick exact/approximate rules. Now that the engine itself selects among rules of varying fidelity on the direct-call path, a plain `f(d)` could return an approximate result with no recorded fidelity — which is exactly the mathematical-honesty property (D1) the converter-registry analogy is meant to guarantee.

Suggest making the plan explicit that the **rule records its fidelity on the result wherever it is invoked** (i.e., fidelity rides on the result via the rule, independent of whether the entry point was the direct call or `evaluate`), with `evaluate` adding only the extra operation-node provenance and `method=` selection on top. Otherwise the "same route, so they agree" claim holds for the *value* but not for the *provenance*, and the direct call becomes a way to get an unlabeled approximation.

arob5:
**Type-specificity vs. fidelity: which is the primary selection key?**

The evaluation registry is described two ways that can't both be the *primary* sort key once `LinOp` is a `Function` subtype and a given map matches multiple cells by subtyping:

- §V.6: *"the operation dispatches through a `BinaryDispatchRegistry` keyed on the map's and the distribution's types"* and *"rules key on the map's type"* — i.e., resolve by **type specificity** (MRO), as dispatch registries normally do.
- §V.6: *"an approximate scheme … registers at its recorded fidelity and **ranks between the exact rules and the sampling floor**"* — i.e., resolve by a **global fidelity order**.

These give opposite answers whenever a *specific-type approximate* rule competes with a *less-specific exact* rule.

**Concrete example.** Two rules are registered for pushing a Gaussian through a linear operator:

1. Exact, at `(LinOp, GaussianDistribution)` — the closed form the section itself cites, `N(A·mean, A Σ Aᵀ)`, fidelity `exact`.
2. Approximate, at `(SparseLinOp, GaussianDistribution)` for the sparse subclass — e.g. a randomized/sketched estimate of `A Σ Aᵀ` that is far cheaper at scale, registered at its recorded (approximate) fidelity.

Now evaluate a `SparseLinOp` on a Gaussian (`SparseLinOp <: LinOp`, so **both cells match**):

- **Type-specificity first** ("keys on the map's type", MRO): the `(SparseLinOp, …)` cell is most specific → the **approximate** rule wins, and the exact rule is silently skipped even though it applies to this operand.
- **Fidelity first** ("approximate ranks between exact and the sampling floor"): the **exact** `(LinOp, …)` rule wins, and the purpose-built specific rule is ignored.

Both are defensible, they disagree, and the plan doesn't say which. This matters more now that the same resolution runs on the direct-call path (see [prior comment](https://github.com/TARPS-group/prob-pipe/pull/376#issuecomment-5027597558)), so the choice determines whether `A @ d` silently downgrades to an approximation.

Suggest stating the composite ordering explicitly — e.g. lexicographic by `(fidelity, type-specificity)` with `method=` as the override, or the reverse — and giving the tie-break rule when both keys are equal.

arob5:
**`SupportsInverse` leaves no user-facing way to invert a bijector.**

Retiring the `Bijector` class drops its three public methods (`forward`, `inverse`, `forward_log_det_jacobian`). The replacement capability exposes:

```python
@runtime_checkable
class SupportsInverse(Protocol):
    def _inverse(self, y: Array | NumericRecord) -> Array | NumericRecord: ...
    def _forward_log_det_jacobian(self, x: Array) -> Array: ...
```

Both are **underscore-prefixed**, i.e. *implementer* methods per the V.0 convention (*"The implementer counterparts are `_`-prefixed … The user-facing names are the bare operations"*). The forward map is the public `apply` (III.2). So of a bijector's three operations, forward is public (`apply`) but inverse and log-det are private — and I grepped the design docs: there is **no user-facing `inverse` (or forward-log-det) operation defined anywhere**. Change-of-variables inside `evaluate` can reach `_inverse`, but a user cannot.

Inverting a bijector is a first-class user need, not an internal detail. Reparameterization exists precisely so inference can move *both* directions between constrained and unconstrained space.

**Concrete example.** A user runs HMC on a positive scale parameter, whose observed value lives in the constrained space:

```python
b = bijector_for(positive)          # a Function: R -> R_+, claims SupportsInverse

# forward is fine — it's `apply`:
theta = b.apply(u)                  # unconstrained draw u -> constrained scale

# but to initialize the sampler from a known constrained value, or to score a
# constrained datum in unconstrained space, the user needs the inverse:
u0 = b.inverse(theta_init)          # <-- no such public operation exists
ldj = b.forward_log_det_jacobian(u0)  # <-- likewise gone
```

With the old class both lines worked. Under this plan the only available entry points are `b._inverse(...)` / `b._forward_log_det_jacobian(...)` — private methods users aren't meant to call — so the capability is effectively half-usable from user code.

Suggest defining the user-facing operations that consume this capability in the same pass — e.g. an `inverse(f, y)` operation (and a forward-log-det one) that dispatch on `SupportsInverse`, mirroring how `evaluate` is the operation form over `apply`. Otherwise the capability has a producer (`bijector_for`, change-of-variables) but no public consumer for the direction that reparameterization is *for*.

Minor, same block: `_inverse` is typed `Array | NumericRecord` but `_forward_log_det_jacobian(x: Array)` is `Array`-only, which contradicts the block's own comment that the capability is *"typed over the numeric value carried, not fixed to arrays."*

arob5:
Following up on my [comment above](https://github.com/TARPS-group/prob-pipe/pull/376#issuecomment-5027576896) about `LinOp` feeling forced into the `Function` hierarchy — there's a concrete correctness consequence in how `LinOp` claims `SupportsInverse`.

**`LinOp` breaks the "capability = declared, structurally-checked claim" model.**

The whole capability machinery rests on *method absence ⟺ capability absence*: a distribution that can't score simply doesn't define `_log_prob`, so `isinstance(d, SupportsLogProb)` is honestly `False` (III.7). `SupportsInverse` is checked the same way — §III.2: *"declared at construction and checked like the distribution capabilities of III.7,"* and the carrier check added in 37546fd gates bijector-valued slots on `isinstance(arg, SupportsInverse)`.

`LinOp` doesn't fit that model. §III.14 says *"A `LinOp` claims it the same way, with its inverse from the operator algebra and its `logdet` the log-Jacobian"* — i.e. the base provides `_inverse` / `_forward_log_det_jacobian` for **all** instances, delegating to the operator algebra. But §III.5 also says a `LinOp` is a general map `A : ℝⁿ → ℝᵐ`, and the square-only queries *"rais[e] `LinAlgError` where undefined."* So every `LinOp` — rectangular or singular included — **structurally satisfies `SupportsInverse`**, and the `isinstance` check can't tell an invertible operator from one that will raise at call time.

That defeats the construction-time guard the carrier check was just added to provide:

```python
# a rectangular operator — a perfectly legal LinOp, not invertible:
A = DenseLinOp(M, input_template=..., output_template=...)   # shape (m, n), m != n

BijectorTransformedDistribution(name, base, bijector=A)
#   construction check: isinstance(A, SupportsInverse)  ->  True   (methods exist on the base)
#   ...so it constructs happily, then:
log_prob(transformed, y)
#   -> A._inverse(y) -> operator algebra -> LinAlgError, at runtime
```

For a distribution the equivalent bad object never passes the gate (the method is genuinely absent). For `LinOp` it always passes, so the guarantee *"the claim supplies the inverse and the Jacobian"* holds for hand-written bijectors but not for the linear subtype the PR just folded in.

The two idioms need reconciling. Either the invertibility claim for the linear case is *not* structural `isinstance` (e.g. gate on squareness / a `positive_definite`-style flag at construction), or `LinOp` shouldn't expose the `SupportsInverse` methods unconditionally on the base — which cuts against `LinOp`'s "one base, `LinAlgError` where undefined" design. This is another facet of the forcing you flagged: capability-by-method-presence and `LinOp`'s always-present-queries are structurally incompatible.

design/04-functions.md
jhuggins: Fixed in `4e0e0ce` — "A `Function` can instead run as a traced task or flow"; kept "computation graph" as the dataflow noun, as you note.

design/04-functions.md
jhuggins: Agreed — the blanket rule predates the function capabilities and now over-reaches exactly as you describe. Narrowed in `efe84a9`: pass-through keys on `Distribution`, `Distribution[...]`, or a distribution capability protocol of III.7, and the function capabilities (III.2/III.14) annotate `Function`-valued parameters, which are values, so the ordinary value rule governs them — a distribution passed in such a position lifts.

design/03-values-and-distributions.md
jhuggins: Reworded in `4e0e0ce` to split the two senses: `SupportsDifferentiation` is defined in III.2, and `SupportsInverse`'s contract is given in III.14 with the reparameterization machinery that motivates it, while the protocol itself sits beside the base in the layout (`values/_function_base.py`, per `package-structure.md`).

design/03-values-and-distributions.md
jhuggins: Closed in `c780fda`, and the question settled a real design point: a `LinOp` is a callable like any `Function`. Its action is the map the base carries — `apply` evaluates the operator at a value (an `Array` or a `NumericRecord` conforming to `input_template`), with the operator's parameters as the private state where a wrapped callable's would be — and calling it takes the same call path as any `Function`, so `A(d)` and the `A @ d` sugar agree through the evaluation rules. `matvec` stays as linear-algebra syntactic sugar for `apply`, and `matmat` is the raw spelling of the operator's registered batched rule (see the reply to your thread comment on the `LinOp`/`Function` tension). Agreed the `WorkflowFunction → Function` code move has to realize exactly this — a good acceptance check for that PR.

jhuggins:
> `LinOp` feels a bit forced into the `Function` hierarchy…

Thought about this more with the `apply`/`matvec` question in hand — the inheritance stands, but your `matmat` observation exposed something real that the design now incorporates (`23b347c`, `c780fda`). On the points:

1. **Wrapped callable vs `to_dense`/`matvec`.** The base's map slot is filled by the operator's *action*: `apply` evaluates it (over an `Array` or a `NumericRecord` conforming to `input_template`), with the operator's parameters as the private state where a wrapped callable's would be. `matvec` is syntactic sugar for `apply` in the linear-algebra spelling — so `Function.apply` and `LinOp.matvec` don't overlap; they are one map in two spellings, and a `LinOp` is called like any `Function`.
2. **Lazy array.** The mathematical object is the linear map — the `D1` reading the design commits to; the lazy array is its representation, and `to_dense` is a derived export among the queries, not the identity of the class.
3. **`matmat` vs the engine's batching — the real catch.** Resolved by making **batched application registry-resolved**, exactly like single-distribution application: the elementwise sweep is the rule registered at the generic pair (the batch counterpart of the sampling lift), and fused batched implementations register above it. `matmat` is then the raw spelling of the operator's registered batched rule rather than a bespoke method beside the engine's batching — and V.8's "batched-backend optimization" becomes an ordinary registered rule too.
4. **Controls on a covariance.** They are constructor defaults that ride every `Function` and are read only by a call that needs them (e.g. a sampling-route evaluation); on an operator used as a covariance they are inert. The cost of uniformity is a few dormant defaults; the benefit is one call path with no operator special case.
5. **Protocol conformance instead of inheritance.** Considered and not taken: the base contributes concrete *state* — name, provenance, the templates, the controls — which a protocol cannot supply, so conformance would mean re-implementing the tracked-term machinery per class. One tracked base type per kind is the design's spine, with subtypes storing structure and capabilities claiming behavior.

jhuggins:
> **Does a direct call record the selected rule's fidelity?**

Exactly right — pre-#376 this couldn't bite because the plain call always sampled; now the direct call selects among rules of varying fidelity, and the recording guarantee lived only on `evaluate`'s side. Fixed as you suggest in `23b347c`: **the selected rule records its name and fidelity in the result's provenance wherever it is invoked** — direct call, `A @ v`, or `evaluate` — with `evaluate` adding only `method=` selection and the operation's provenance node on top. Stated in both the IV.1 engine contract and §V.6, so "same route, so they agree" now covers value and record alike, and a plain `f(d)` cannot return an unlabeled approximation.

arob5:
A batch of smaller consistency / organization points, grouped since each is minor. (A sixth — the "protocols live beside the base" vs `SupportsInverse` in §III.14 tension — is already tracked in the inline thread on that line, so it's left out here.)

**1. `with_options` returns `Function`, not `Self` (§III.2).**
```python
def with_options(self, **controls) -> Function: ...
```
Every other functional updater in the design returns `Self` — `with_flag`, `with_names`, `with_dims` (§III.5), `with_name`, `with_provenance` (§II). As written, `A.with_options(...)` on a `LinOp` erases the static type down to `Function`, dropping the operator algebra. Should be `-> Self`.

**2. The naming-consistency argument is weakened by the callable-generic rule.**
The case for `Function` (PR description) leans on parallelism with `FunctionSpec` / `FunctionBatch`. But §III.2 now states outright that a `FunctionSpec` *"admits any callable … the `Function` being one such, not the required type."* So `FunctionSpec` is **not** "the spec of a `Function`," unlike `DistributionSpec` (spec of a `Distribution`) or the `Record`/`RecordBatch` pairing. Callable-generic is a good deliberate choice — just noting the parallel that motivates the name doesn't actually hold, so that part of the rationale is softer than stated. No change required; flagging so the naming rationale isn't over-relied on.

**3. The §III.0 layer map doesn't reflect the new capability protocols.**
The III.2 row lists Contents as ``` `Function`, `FunctionBatch` / `OpaqueBatch` ``` and its Role never mentions `SupportsDifferentiation`, though that protocol is now defined in §III.2. The III.14 row still reads "constraint reparameterization (`bijector_for`)" though the `SupportsInverse` protocol now sits there too (per §III.2: *"its contract is given … in III.14"*). Contrast the III.7 row, which does name "The `Supports*` protocols." The map rows drifted from the sections' new content.

**4. "Empty-given corner" vs the currying rule (§III.8).**
§III.8 says *"there is no … rule that turns an empty-given kernel into a `Distribution`."* But the section also keeps the rule that conditioning a kernel on **all** its given fields yields a `Distribution` (`condition_on(K, s)` → `Distribution`), and currying on a subset yields a smaller kernel. So fully conditioning *does* produce the empty-given corner law. These are reconcilable — a kernel can't be *constructed* with an empty `given_template`, whereas full conditioning *produces* the corner — but as worded the two statements read as in tension. Worth one clause distinguishing "no empty-given kernel is ever constructed/represented" from "conditioning can produce the corner `Distribution`."

**5. "Every operation is a `Function`" isn't reconciled with the operations' own two-level model (§V.0).**
§V.0 keeps operations' `_x`/`x` (implementer/user-facing) structure and capability dispatch via `isinstance(obj, SupportsX)`, while §III.2 gives every `Function` an `apply`/`__call__` structure over *exactly one wrapped callable*. If `sample` "is a `Function`," which method is its `apply` and which its `__call__`, and how does that square with an operation that **dispatches across many types** rather than wrapping one callable? Plausibly `apply`↔the raw dispatched `_op` and `__call__`↔the wrapping `op`, but that identification is nontrivial (a dispatcher is not "one callable") and is currently left implicit. One or two sentences mapping the operation model onto the `Function` base would close it.

SoymilkWinsAgain:
Before I start #368, could we confirm a few points here? Otherwise I would be guessing at the base contract:

- Should the no-lifting rule be limited to `Distribution[...]` and protocols that explicitly consume distributions, rather than every `Supports*` protocol?
- Is `Function.apply` the canonical point-evaluation hook, with a regular `Function` delegating to its internal implementation and `LinOp` delegating to `matvec`?
- Does “wraps exactly one Python callable” allow a private callable implementation carrier for fitted state?
- Should `Function` carry an authoritative `inspect.Signature`, with `input_template` fields tied to parameter binding?

I do not think I can safely begin #368 until these are settled, since each one affects the shape of the base implementation.

jhuggins:
> **Type-specificity vs. fidelity: which is the primary selection key?**

Both readings were live — thanks for the sharp example. Settled in `3abfa3d`, and it's the same resolution the converter registry already uses: **fidelity-first, realized through priority tiers**. Priorities are assigned from fidelity tiers (exact rules above approximate ones, the sampling lift the floor at the generic pair); **within a tier** the more specific type pair wins, by the same closest-in-method-resolution-order tie-break the dispatch registries use, then registration order; `method=` overrides the ranking entirely.

Your example under this rule: the exact `(LinOp, Gaussian)` rule beats the approximate `(SparseLinOp, Gaussian)` rule, so `A @ d` never silently downgrades — the purpose-built sparse approximation is one `method=` away, and an *exact* sparse rule would win within the exact tier by specificity. Exactness is never traded for type specificity silently, which is the honesty property your fidelity-recording comment already pinned down.

jhuggins:
> **`SupportsInverse` leaves no user-facing way to invert a bijector.**

Right — the capability had a producer but no public consumer for exactly the direction reparameterization exists for. Fixed in `feca2e1`, with one refinement beyond your suggestion: the **Jacobian determinant is split into its own capability**, since it's genuinely separate from invertibility — the determinant exists only for a differentiable map, and a map can be invertible without a tractable one. So:

- `SupportsInverse` stays minimal: `_inverse` only, the forward being the claiming `Function`'s `apply`.
- `SupportsLogDetJacobian` carries `_log_det_jacobian(x)` — input generalized to `Array | NumericRecord`, fixing the typing inconsistency you flagged.
- A *bijector* is a `Function` claiming both; change of variables requires exactly the pair, while a slot needing only inversion (a GLM link) checks invertibility alone.

And the user-facing operations land as a new **§V.7**:
- `inverse(f)` returns the inverse map **as a `Function`** — so your example reads `u0 = inverse(b)(theta_init)` — itself invertible with `f` as its inverse, carrying the Jacobian claim whenever `f` does.
- `log_det_jacobian(f, x)` returns the log-determinant at `x`; the reverse direction composes rather than needing a second operation: `log_det_jacobian(inverse(f), y) = −log_det_jacobian(f, inverse(f)(y))`.
- Both are capability-only, with the status as registry data (a numerical route can register later at its recorded fidelity).

(`marginal`/`factor` and batched operations renumber to V.8/V.9; cross-references swept.)

jhuggins:
> **`LinOp` breaks the "capability = declared, structurally-checked claim" model.**

Correct, and it's the third instance of one theme this review surfaced: method presence can't witness everything (your fidelity catch, the differentiability flag, now this). Fixed in `feca2e1`:

- **`LinOp` claims the invertibility capabilities only when they apply** — the claim is guarded per instance by squareness (`input_template == output_template`); a rectangular operator makes no claim, so your `DenseLinOp(M)` example now fails at construction, not at `log_prob` time.
- Consumers check **`is_invertible(f)`** rather than bare `isinstance`: the structural check plus the claim's instance guard (unconditional for a declared bijector, squareness for a linear operator) — the same predicate pattern `is_differentiable` established.
- **Singularity stays a runtime `LinAlgError`**, exactly as for `solve`: the guard catches what construction *can* decide (shape), and numerics keep their honest failure mode. This is the capability doctrine's per-instance feasibility branch, not a new mechanism.

So the reconciliation you asked for goes your first way — the linear case's claim is not bare structural `isinstance` — without giving up "one base, `LinAlgError` where undefined" for the queries.

jhuggins:
> A batch of smaller consistency / organization points

All five addressed in `1479475` (except 2, which needs no edit):

1. **`with_options` → `Self`** — fixed; `A.with_options(...)` keeps the operator algebra.
2. **Naming rationale** — fair flag. The doc itself doesn't lean on the spec parallel; the name stands on the mathematical object a `Function` denotes, and callable-generic specs are the deliberate choice you note. Acknowledged that the PR description's parallelism argument is softer than stated.
3. **III.0 map rows** — updated: the III.2 row now names `SupportsDifferentiation` and the differentiability claim; the III.14 row names `SupportsInverse` and `SupportsLogDetJacobian`.
4. **Empty-given corner vs currying** — reconciled with the clause you suggested: a kernel is never *in* the empty-given state, since binding its last given field returns the `Distribution` directly; so "conditioning produces the corner's representative" and "no empty-given kernel exists" are both true.
5. **Operations onto the `Function` base** — mapped in V.0: an operation is a `Function` whose wrapped callable is its implementer dispatch (select the capability method, invoke on the raw types); `apply` is that raw, untracked result and `__call__` adds the engine's lifting and the wrap — which is also why V.9's batched sweep needs no machinery of its own, and how `raw=True` relates to the plain path.

jhuggins:
@SoymilkWinsAgain
1. **Yes.** Pass-through is limited to `Distribution`, `Distribution[...]`, and the distribution capability protocols of §III.7. The function capabilities (`SupportsInverse`, `SupportsDifferentiation`, `SupportsLogDetJacobian`) annotate `Function`-valued parameters, which are values, so a distribution passed there lifts (fixed in `efe84a9`, per arob5's same catch).

2. **`apply` is the canonical point-evaluation hook** — one nuance on direction: `matvec` is syntactic sugar *for* `apply` (its linear-algebra name), so a plain `Function`'s `apply` invokes its private implementation and a `LinOp`'s `apply` computes the operator action that `matvec` names. Which one the structured routine physically lives in is your wiring to pick, as long as `apply` is the public hook and the two agree.

3. **Yes, that's the intent.** §III.2 has the wrapped callable and its state staying private, reached only through the `Function`'s methods — "one Python callable" names the map, and the state that map needs, fitted parameters or a backend handle, lives behind the private implementation.

4. **Yes, with `input_template` as the single source of truth:** its top-level fields are the function's parameters, so binding an argument is binding a field by name. The `Function` doesn't store a separate `inspect.Signature` — it derives one from `input_template` on demand (one parameter per top-level field, in canonical order), so the two can't drift. When a plain callable is wrapped without an explicit template, its parameters seed those fields once; from then on the template is authoritative. Added to §III.2 in `739bc46`.

SoymilkWinsAgain:
@jhuggins Thanks — points 1–3 settle those parts. However, I do not think point 4 is fully settled yet, because deriving `inspect.Signature` solely from `input_template` cannot preserve parameter kinds, defaults, or variadic parameters. An `EventTemplate` describes value structure, not the complete Python calling convention, so this would also weaken compatibility with ordinary wrapped callables.

Our existing plan instead keeps a frozen `inspect.Signature` authoritative for Python binding and the templates authoritative for schema, validating their total correspondence at construction. This is recorded in #368 and #370, and summarized in #350 under W0.5.1. I will implement #368 using that contract for now. If review reveals a problem with it, we can revise the implementation and the plan together.

jhuggins:
You're right on point 4 — I'll walk back the derive-lazily framing. An `EventTemplate` is a value schema, so a signature reconstructed from it can't carry parameter kinds, defaults, or variadic parameters, and leaning on it would weaken exactly the ordinary-callable compatibility the callable-generic rule is there to protect.

So the contract is your #368/#370 one, and I've updated §III.2 to match (`fb991b2`): a `Function` carries a **frozen `inspect.Signature`, authoritative for Python argument binding**, while the **`input_template` is authoritative for the value schema**, and construction validates their **total correspondence** (parameters ↔ top-level fields). Two authorities, each for what it's actually the authority on. That lines up with W0.5.1 in #350 — go ahead and implement #368 against it, and we revise doc and plan together if review turns anything up.

Issue #369, opened 2026-07-17
Proposal: add closed result planning for `Function`


SoymilkWinsAgain:
## Parent

Tracking issue: #252

## What to establish

Add a closed internal result-plan family for Functions that may return first-class `ProbPipe` terms:

```text
_EventResultPlan
_TermResultPlan
_ResolvedFitResultPlan
```

Event wrapping remains the default. A term result is returned directly only when an internal plan predeclares its term family. For lifted `fit`, preflight must freeze the atom family, aggregate family, canonical schemas, validator, and all-or-error collection policy before any cell executes.

The first successful cell must never choose the result kind or common schema.

## Why this is necessary

The existing event-return boundary cannot distinguish “a Function as data” from “a Function that is itself the operation result.” Runtime inspection is also too late for Batch/Distribution lifting because the aggregate type must be known before execution.

Result planning is a reusable Function-layer facility. Estimator integration is its first concrete consumer; the plan types remain private and closed.

## Fit result algebra

```text
point adapter
plain        -> Function
sweep        -> FunctionBatch
distribution -> RandomFunction

probabilistic adapter
plain        -> ConditionalDistribution
sweep        -> ConditionalDistributionBatch
distribution -> Distribution[ConditionalDistribution]
```

This algebra is invocation-specific and does not redefine the global contracts of `FunctionBatch`, `RandomFunction`, or `ConditionalDistribution`.

## Schema rules

- A lifted result has one plan-owned authoritative common schema.
- Polymorphic schemas compare by joint input/output alpha-equivalence while
  preserving the complete repeated-symbol relation.
- “First atom wins,” schema unions, and ragged aggregate schemas are forbidden.
- A post-fit learned schema is allowed for one concrete fit.
- A lifted fit requires a preflight-stable schema or an explicit common schema
  the adapter can enforce across all admitted cells or draws.

## Acceptance criteria

- [ ] Event wrapping remains the default.
- [ ] Term results require an explicit internal plan.
- [ ] Lifted aggregate family and common schema are fixed before execution.
- [ ] Canonical schema resolution is deterministic and order-independent.
- [ ] Learned-schema estimators fail before lifted work when a common schema cannot be guaranteed.
- [ ] Wrong atom family and schema mismatch have distinct failures.
- [ ] No runtime union, heterogeneous aggregate, or partial result escapes.

## Dependencies

Depends on the first-class `Function` contract in #368.

PR #380, opened 2026-07-21
Make Function a first-class tracked class

### LINKED ISSUE #368: Make `Function` a first-class tracked term

SoymilkWinsAgain:
## Summary

- Make `Function` an immutable, tracked, schema-aware computation with a captured Python signature, optional authoritative input/output templates, raw `apply()`, and a shared planner for decorated and dynamic implementations.
- Add call-local symbolic-dimension and RNG state, complete variadic lifting, Function-first provenance, backend-consistent nested output aggregation, and `DistributionArray.event_template`.
- Add `LinOp.apply(x)` as an exact alias of `matvec(x)`, plus documentation and regression coverage.

## Key contracts and scope

- A callable's signature is captured independently and is **not** derived from `input_template`, following the existing #368/#370 contract rather than the alternative discussed in #376.
- `apply()` performs signature/schema validation but returns the implementation's raw result without lifting, sweeping, wrapping, orchestration, or call provenance.
- `Function.__call__` shallow-copies directly returned tracked values into a new result term and replaces their provenance; `apply()` preserves the raw object's identity and provenance. This copy boundary also applies when provenance is disabled.
- Authoritative point outputs are normalized before aggregation so sequential, JAX, Prefect, broadcast, and sweep paths share the same structured representation and concrete template.
- Authoritative nested-output packing is private to Function aggregation; the public `RecordArray.stack` contract is unchanged.
- `DistributionArray.event_template` uses an explicitly supplied template when available, otherwise exposes a common component template only when all components agree.
- Variadic calls retain their `inspect.BoundArguments` structure. Each `*args` element and `**kwargs` entry is planned independently, receives the variadic parameter's annotation, and has a stable provenance/input label without leaking planner slot names into the callable.
- A plain callable parameter literally named `self` is an ordinary parameter. Bound methods rely on `inspect.signature(bound_method)` to remove their receiver naturally.
- Distribution pass-through remains an explicit capability allowlist; unrelated protocol annotations do not suppress lifting.
- Dynamic implementation fingerprints encode signature structure, stable defaults and annotations, templates, and implementation type. Opaque values contribute their type rather than process-specific identity or addresses.
- Provenance fingerprints use a conservative fidelity classification: structurally known values and closure-free Python functions receive portable content fingerprints, while closure-bearing functions, bound methods, partials, callable instances, classes, builtins, and unsupported objects use process-local identity and are marked `fingerprint_is_weak`. Weakness propagates through composite fingerprints so cross-run cache consumers can fail closed.
- `_from_implementation` is only a clean future factory seam. `OperationRef`, #369 result planning, and the #370 fitted-producer contract remain deferred.
- This PR does not change the `LinOp`/`Function` inheritance relationship.

Closes #368


SoymilkWinsAgain:
It has been confirmed that `LinOp` will be a subclass of `Function`. However, this is far beyond the scope of this PR, and a new PR will be initiated.

docs/api/workflows.md
```diff
@@ -4,6 +4,12 @@
 `@function`. `Module` is the stateful container with
 `@workflow_method` children.
```
jhuggins, line 5: `@workflow_method` needs to get renamed as well. Probably to `@module_method`, although there's a question as to whether Modules should exist at all...

docs/api/workflows.md
```diff
@@ -37,6 +43,84 @@ function whenever they can bind to that function. This keeps common names
...
+preserves that object's identity, annotations, and provenance. `__call__`
+instead creates a shallow independent result item: value data and templates are
+shared, the annotations container is copied, prior provenance is cleared, and
+the current Function and tracked inputs become the new direct parents. This
+copy is still made when provenance tracking is disabled.
+
+When a sweep returns distributions, its `DistributionArray.event_template`
+records the concrete authoritative output template. Ordinary
+`DistributionArray` construction exposes a common component template when all
+components agree, and otherwise returns `None`. Broadcast marginals and nested
```
jhuggins, line 147: An open question is whether Batches should enforce a uniform event template

probpipe/core/_workflow_call.py
```diff
@@ -44,19 +45,37 @@ class WorkflowCallOverrides:
...
     values: dict[str, Any]
     overrides: WorkflowCallOverrides
 
 
+@dataclass(frozen=True)
+class WorkflowInputRef:
+    """Reference to one planner-visible value in a resolved Python call."""
+
+    parameter_name: str
+    position: int | None = None
```
jhuggins: since only one of position or keyword can be None, I would suggest a single `position: int | str | None`

probpipe/core/event_template.py
```diff
@@ -147,9 +149,12 @@ class ArraySpec(ValueSpec):
 
     def __post_init__(self) -> None:
         shape = tuple(self.shape)
-        if not all(isinstance(d, int) and d >= 0 for d in shape):
+        if not all(
+            (isinstance(d, int) and d >= 0) or (isinstance(d, str) and bool(d)) for d in shape
```
jhuggins, line 161: the names also need to be unique (worth pinning in a regression test)

probpipe/core/event_template.py
```diff
@@ -908,6 +961,205 @@ def vector_size(self) -> int:
...
+) -> EventTemplate:
+    """Substitute all symbolic dimensions, failing if any remain unbound."""
+    missing = template.free_dims.difference(bindings)
+    if missing:
+        dimensions = ", ".join(sorted(missing))
+        raise ValueError(f"{context} has unbound symbolic dimensions: {dimensions}")
+    return _replace_template_dimensions(template, bindings)
+
+
+def _unify_template_node(
```
jhuggins, line 1030: It's worth including a brief comment for this and other private functions below, just as you did above

probpipe/core/_workflow_result.py
```diff
@@ -52,6 +59,12 @@ def _wrap_as_record(value: Any, field_name: str) -> Any:
     - Anything else (opaque Python object) → a single-field plain
       ``Record`` named and keyed by the function's own name.
     """
+    if output_template is not None:
+        return _wrap_declared_function_output(
+            value,
+            function_name=field_name,
+            output_template=output_template,
+        )
     if isinstance(value, (Distribution, Record)):
```
jhuggins, line 74: `_wrap_as_record` passes through only `Distribution`/`Record` here, and the same narrow guard appears twice more in `_function_contract.py` — `_wrap_declared_function_output` (~L301) and `_validate_function_output` (~L255, via `.event_template`). All three mean "this is an already-structured tracked value, keep it as-is," but they enumerate a *subset* of the tracked-term family. Any other `Tracked` value (today `Function` and `_ConditioningStep`) falls through and is buried as an opaque leaf in a single-field `Record`, losing its type and skipping the copy + provenance-replacement boundary. `_coerce_output`'s own `stack`/`nested` branch already uses `isinstance(value, Tracked)`, so the wrap path is inconsistent with the rest of this same function.

This gets sharper soon: `ConditionalDistribution` and the generic `Batch` types are tracked terms too (design/02; the `*Array → *Batch` rework), so with these enumerations they'd be silently mis-wrapped the moment they land.

Suggestion — generalize all three guards to `isinstance(x, Tracked)` (D4: the system is closed under operations, so every op returns a tracked term), and update the `Record | RecordArray | Distribution` contract wording in this module to "any tracked term." Two adjacent spots to keep in sync:
- `_fingerprint._update` enumerates the same families; a standalone `ConditionalDistribution`/`Batch` base (not subclassing `Record`/`Distribution`) would drop to the process-dependent `repr` tier — worth a generic `Tracked` fallback.
- the broadcast stacking dispatch (`_make_stack`/`_coerce_broadcast_output`) will need an explicit path when those types become broadcastable outputs.

To make this durable rather than vigilance-based: add a test that walks the concrete `Tracked` subclasses and asserts each is preserved (not buried) through the Function boundary and fingerprinted non-trivially — then a future `ConditionalDistribution`/`Batch` left unwired fails CI instead of being silently wrapped.

probpipe/core/_fingerprint.py
```diff
@@ -516,18 +518,31 @@ def _update_function(
...
+        h.update(implementation_type.__qualname__.encode())
+        h.update(b":signature=")
+        _update_signature_declaration(h, function.signature, max_array_bytes)
+        h.update(b":input_template=")
+        _update(h, function.input_template, 1, max_array_bytes)
+        h.update(b":output_template=")
+        _update(h, function.output_template, 1, max_array_bytes)
         return
 
+    func = implementation.callable
```
jhuggins, line 788: `_update_function` splits on implementation kind, and only the *dynamic* branch hashes `signature`/`input_template`/`output_template`; the plain-callable branch (here) hashes `__code__` + defaults + closure and `return`s, never reaching the templates. Since templates are declared independently of the callable and drive validation + output shape, two Functions sharing a callable but declaring different templates fingerprint identically — and because a Function is a provenance parent, their parent fingerprints collide (a wrong-cache-hit hazard for the `cache_key_fn` that `ParentInfo.fingerprint` is seeding).

Reproduced:
```python
g = lambda x: x
f1 = Function(func=g, output_template=EventTemplate(y=()))
f2 = Function(func=g, output_template=EventTemplate(y=(5,)))
fingerprint(f1) == fingerprint(f2)   # True — should differ
```
Fix by hashing the declaration (input/output templates) *unconditionally*, then appending implementation-specific content — i.e. hoist the template hashing out of the dynamic-only branch so both paths run it. Worth a callable-path template-sensitivity test too; the fingerprint tests currently exercise template sensitivity only on the dynamic path, which is why this slipped.

probpipe/core/node.py
```diff
@@ -444,38 +676,127 @@ def _call_with_options(
...
+                context=f"Function {self._name!r} output_template",
+            )
+            if self._output_template is not None
+            else None
+        )
+        provenance_parents: list[Tracked] = [self]
+        seen_parent_ids = {id(self)}
+        for ref in _workflow_call.iter_input_refs(self._signature_info, values):
+            value = _workflow_call.input_ref_value(values, ref)
+            if isinstance(value, Tracked) and id(value) not in seen_parent_ids:
```
jhuggins, line 722: `provenance_parents` is filtered to `isinstance(value, Tracked)`, so un-tracked inputs (plain arrays, scalars) are dropped — and since `Provenance` has no separate plain-arguments field (metadata here is just `{"func": name}`), they're recorded nowhere. `_make_parent(p)` already handles any object (`getattr(..., None)`, plus a content `fingerprint`), and `ParentInfo` documents `name: None for unnamed parents` — so the infrastructure is built to describe un-tracked parents; this filter discards them before `_make_parent` sees them.

Concretely, `condition_on(joint, x=jnp.array(0.0))` records parents `["condition_on", joint]` — the conditioned array (the data) is absent, and conditioning on `0.0` vs `5.0` yields identical parent fingerprints, so provenance can't distinguish what was conditioned on. This predates the PR (`main` filtered via `hasattr(v, "provenance")`, same effect), but the new `ParentInfo.fingerprint` / `cache_key_fn` foundation makes it consequential.

Design call: either include un-tracked inputs as fingerprinted `ParentInfo`s (loosening the `Sequence[Tracked]` typing), or add a distinct plain-arguments/controls field to `Provenance` (closer to the design/02 parents-vs-arguments-vs-controls split). Silently dropping them is the one option that clearly breaks reproducibility.

tests/core/test_first_class_function.py
```diff
@@ -0,0 +1,869 @@
...
+            dispatch=dispatch,
+        )
+
+        result = wrapped(rows)
+
+        assert result.event_template == EventTemplate(prediction=(2,))
+        assert result.batch_shape == (3,)
+
+    @pytest.mark.parametrize("dispatch", ["sequential", "jax"])
+    def test_distribution_broadcast_preserves_declared_output_template(self, dispatch):
```
jhuggins, line 720: This aggregation test asserts shape + declared template but not the aggregated *values*, and the `event_template ==` check elsewhere is near-tautological (the declared template is stamped onto the result, so the assertion re-reads its input). Here `func=lambda x: jnp.stack((x, x + 1))` checks only `samples["pair"].shape == (8, 2)` — a transposed/mis-ordered stack would pass; the load-bearing `pair[:, 1] == pair[:, 0] + 1` is never checked. Same pattern in the sweep variant (values `[[1,1],[2,2],[3,3]]` unchecked) and the `..._through_broadcast`/`..._through_sweep` cases (template echo only, no `num_atoms`/value). Add value assertions; the nested-mapping tests above already do this well — mirror them.

CHANGELOG.md
```diff
@@ -9,6 +9,32 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **First-class, tracked `Function` values (#368).** `Function` is now an
```
jhuggins, line 12: This changes observable provenance for existing ops — `condition_on`/`from_distribution` operation strings become `workflow.*`, parent counts grow by one (Function-first), a directly-returned tracked value is now a shallow copy rather than the identity object, and domain metadata (`metadata["conditioned"]`, `"ess"`) moves from the `__call__` result to the `apply()`/raw result (the call result's metadata is now just `{"func": name}`). It's intentional and well-documented, but it's filed under **### Added** — the provenance-semantics and identity-not-preserved parts belong under **Changed**, and this warrants `kind:breaking-change` (only `area:*` labels are set). Anyone asserting `result.provenance.operation == "condition_on"`, `len(parents) == 1`, or reading domain metadata off a `__call__` result will break — please confirm the `__call__`-metadata narrowing is intended.

jhuggins:
A few minor / polish items from Claude's review (non-blocking):

- `_bind_planned_function_inputs` (`_function_contract.py` ~L212-236) forwards a `None` `event_template` for a template-less `DistributionArray` under an authoritative `input_template`, surfacing "expected named fields, got NoneType" instead of the intended "no authoritative event_template for lifting."
- `_fingerprint.py` reads `cell.cell_contents` for every closure cell; an unbound cell raises `ValueError` out of `fingerprint()` — a sentinel / `suppress` keeps fingerprinting total (fix alongside the `_update_function` template comment, same file).
- `_make_stack` / `_make_marginal` (`_broadcast_distributions.py`) wrap a bare `ndarray` using a single leaf path as a flat field key with no `len(event_template) == 1` guard — unreachable from `__call__` today (invoke_point pre-wraps), so an explicit assert would make the dependency legible.
- `Function._from_implementation` is production-dead in this PR (tests only), and "seam" in its docstring / CHANGELOG is vague — consider deferring or renaming.

jhuggins (changes requested):

SoymilkWinsAgain:
> * `Function._from_implementation` is production-dead in this PR (tests only), and "seam" in its docstring / CHANGELOG is vague — consider deferring or renaming.

`_from_implementation` is indeed currently only called by tests. However, that's an interface reserved for the next step (#370 ), so I'll keep it. But the previous description was too vague.

The other 3 problems are solved.

probpipe/core/_function_contract.py
```diff
@@ -0,0 +1,313 @@
...
+    )
+
+    if isinstance(result, (Record, Distribution)):
+        try:
+            actual_template = cast(Any, result).event_template
+        except (AttributeError, TypeError) as error:
+            raise ValueError(
+                f"Function {function_name!r} output does not expose an authoritative event_template"
+            ) from error
+        if actual_template != concrete:
```
arob5: **Output-template validation is asymmetric — dtype/support-pinned templates are unsatisfiable by a `Record`/`Distribution` return.**

`_validate_function_output` validates the *same* declared `output_template` two different ways depending on what the function returns:

- **`Record`/`Distribution`** (this branch): strict structural equality, `actual_template != concrete`.
- **scalar/array** (below, ~L279): `ValueSpec.is_valid(result)`, which allows same-kind dtype casts and skips `support` entirely.

Because `Record.event_template` inference is lossy — it always yields `dtype=None` / `support=None` — any `output_template` that pins a dtype (or support) is **structurally unsatisfiable** by a `Record`/`Distribution` return, even when the underlying data conforms:

```python
@function(output_template=EventTemplate(y=ArraySpec((), dtype="float32")))
def f(a):
    return Record("r", {"y": jnp.array(3.0)})   # value IS float32
f.apply(1.0)
# ValueError: output event_template ...dtype=None... does not match declared ...dtype('float32')
```

The identical scalar `return 3.0` passes, so two returns carrying the same data get opposite verdicts based only on wrapper type. This is currently untested (no `output_template` test pins dtype/support).

Suggest routing this branch through conformance/unification with same-kind casting rather than `!=`. `_unify_event_templates` in `event_template.py` (expected-vs-authoritative-actual, `same_kind` dtype) looks like exactly the intended primitive — but it is currently unused anywhere in the tree, which reinforces that this path was meant to unify rather than compare for equality.

probpipe/core/_function_contract.py
```diff
@@ -0,0 +1,313 @@
...
+    if not isinstance(result, Mapping):
+        if len(concrete) != 1:
+            raise ValueError(
+                f"Function {function_name!r} returned a scalar/array for a multi-field "
+                f"output_template with fields {list(concrete.keys())}"
+            )
+        only_path = next(iter(concrete.keys()))
+        only_spec = concrete[only_path]
+        assert isinstance(only_spec, ValueSpec)
+        if not only_spec.is_valid(result):
```
arob5, line 331: **A `support` constraint declared on an `output_template` is never enforced.**

This scalar/array branch validates via `ValueSpec.is_valid`, and `ArraySpec.is_valid` deliberately skips `support` (it is a data-dependent, non-jit-safe check). So a `support` constraint on an `output_template` is silently unenforced on the common scalar/array path:

```python
@function(output_template=EventTemplate(y=ArraySpec((), support=positive)))
def f(a):
    return jnp.array(-3.0)   # negative
f.apply(1.0)   # passes — support never checked
```

Combined with the strict-equality issue on the `Record`/`Distribution` branch above (L262), `support` in an output template is at best a no-op (scalar/array) and at worst unsatisfiable (`Record`/`Distribution`). A user declaring, say, a positivity constraint on outputs gets neither enforcement nor a warning.

Suggest either enforcing it where feasible (outside jit traces), or rejecting/warning at construction on a support-bearing `output_template`, so the declaration doesn't imply a guarantee that isn't provided.

probpipe/core/_workflow_result.py
SoymilkWinsAgain: That's correct, but this is precisely the default event-result behavior explicitly preserved in #369. Ordinary `Functions` use an event-result contract by default. When returning another `Function`, the object should be wrapped in a `Record` as a `FunctionSpec`/opaque event leaf. Only the explicit `_TermResultPlan` in #369 can directly return the tracked term. #369 is a next step, so there's no need to worry about the issues mentioned here remaining.

In short, this is a deliberate legacy of this PR and is not within the scope of a fix.

probpipe/core/_fingerprint.py
SoymilkWinsAgain: This is correct and has been successfully reproduced.

Remediation: Adjust the `_update_function` to unify the encoding of frozen signature, input_template, and output_template first.

docs/api/workflows.md
SoymilkWinsAgain: I will write a new issue about this, because this isn't directly related to the theme of this PR. (This PR is already large enough, and it's connected to a series of subsequent projects. I don't want to complicate things.)

SoymilkWinsAgain:
Two issues will be addressed in separate issues, unrelated to this PR:

1. Make `LinOp` a subclass of `Function`.
2. `@workflow_method` needs to be renamed; one suggestion is `@module_method`. A more ontological discussion is whether Modules themselves are necessary (still open, no definitive conclusion).

probpipe/core/node.py
SoymilkWinsAgain: A new `Provenance.inputs` file has been added to record resolved regular parameters and their fingerprints. `parents` still only represents tracked lineage; DAG/ancestors do not traverse regular inputs.

probpipe/core/_function_contract.py
SoymilkWinsAgain: Solved. Now, the validation of Record/Distribution outputs has been changed from strict template equality to authoritative conformance.

probpipe/core/_function_contract.py
SoymilkWinsAgain: Solved. The root cause is that container branches perform strict template equality checks, while raw value branches intentionally ignore the `ArraySpec.is_valid` check for `support`. This design in the initial version was because `Constraint.check` is data-dependent, and outputs with `support` cannot safely enter the current JAX trace.

probpipe/core/node.py
jhuggins: With inputs and parents as separate channels, the *same* logical argument fingerprints differently depending on how it's wrapped. `f(x=np.array([2.]))` records a regular input `fp(array)`, while `f(x=NR(x=np.array([2.])))` records a tracked parent `fp(NR(x=array))` — even though, via the single-field shim, both mean "argument `x` = this array" and a coercing body can't tell them apart. 

I'd resolve it by normalizing inputs **up** to the tracked-term family rather than keeping a separate untracked channel: wrap a bare value bound to argument `x` as a **provenance-less** `NR(x=value)`, and pass an already-tracked input through unchanged. This is lossless (wrapping only adds structure; unwrapping a record would drop its name/provenance/backend metadata), and the equivalence then falls straight out of the existing structural fingerprint — which excludes `.name` and `.provenance`, so:

- an auto-wrapped `NR(x=arr)` fingerprints identically to a user's hand-written `NR(x=arr)` ✅
- `NR(y=arr)` stays distinct (field ≠ argument) ✅
- a piped value already wrapped as `NR(<source_op>=arr)` stays distinct from a literal — correct, it carries real lineage ✅

That also lets the regular-inputs channel fold back into `parents` as provenance-less *roots* (leaves in the DAG, no ancestry), so the model is one uniform family instead of two — up to you whether you still want to label literals separately for display.

I'd keep this **provenance-only**: the wrapping is for the fingerprint/lineage record, and bodies keep receiving native types (no change to what the function sees). That leaves exactly one semantic commitment, which I think we should state explicitly rather than leave implicit in the shim:

> **A single-field record named after its argument is interchangeable with the bare value; operations must not distinguish them.**

The single-field shim already implies this (`jnp.asarray(NR(x=arr)) == arr`), so we're just naming it.

SoymilkWinsAgain:
Found some inefficient code from earlier submissions. For example, instead of using the built-in `collections.deque`, `probpipe/core/provenance.py` uses a manually written queue. I believe this was introduced in the initial submission on March 19, 2026, and it was never noticed until this PR. Fixed together.

SoymilkWinsAgain:
Submitting a draft first to see if it can pass CI

probpipe/core/_function_contract.py
SoymilkWinsAgain: Unresolve this because it exposes deeper problems, shouldn't be patched up

SoymilkWinsAgain:
The previous authoritative-output fix over-corrected the Distribution case: it reconciled `event_template`, `dtypes`, and `supports` inside the Function boundary and then replaced the public result copy’s template. Those are currently separate legacy metadata surfaces, so this added another schema-ownership layer and could publish contradictory views.

The PR now uses a fail-closed transitional contract. Function validates an existing Distribution only against its own authoritative `event_template` and never rewrites that template. Metadata-bearing declarations therefore require a schema-complete, exactly matching Distribution template; current shape-only distributions reject that combination explicitly. Bare values, mappings, and Records retain their dtype/support validation. Consolidating Distribution schema ownership remains a separate follow-up.

tests/core/test_first_class_function.py
SoymilkWinsAgain: Complete matrix value assertions for sweep aggregations have been added. Assertions regarding `num_atoms` and column relationships for `broadcast` have been added. The number, order, location, and scale of components for `Distribution` broadcast/sweep have been validated.

probpipe/core/_workflow_call.py
SoymilkWinsAgain: combined

probpipe/core/event_template.py
SoymilkWinsAgain: This conflicts with the current symbolic-dimension contract. Names are in fact size variables, so repetition intentionally puts an equality constraint. (For example, ("n", "n") describes a square matrix.) This behavior has already been covered by a regression test. It is also required by the planned schema. Thus, enforcing uniqueness would remove that rather than tighten validation. Resolving it.

probpipe/core/event_template.py
SoymilkWinsAgain: Added brief docstring for 6 private helpers

probpipe/core/node.py
SoymilkWinsAgain: @jhuggins I believe this conflicts with the current value and provenance contracts. The single-field shim provides only narrow explicit coercions; it does not make a `NumericRecord(x=value)` interchangeable with the bare value, and a `Function` body can distinguish them. Provenance-only wrapping would therefore record a different input from the one actually consumed and could create unsafe cache equivalences. The existing design intentionally keeps tracked lineage in parents and plain resolved arguments in inputs

docs/api/workflows.md
SoymilkWinsAgain: I think it would be better to open a separate issue about this

SoymilkWinsAgain:
Going to open a separate issue talking about whether Batches should enforce a uniform event template

SoymilkWinsAgain:
The different fingerprint phenomena described in the review are true, but the proposed normalization scheme conflicts with the current contract, so it is temporarily patched. This will be discussed later.

jhuggins:
@SoymilkWinsAgain This looks good to me overall — the `Function` contract, the `apply` boundary, and the provenance split all read well, and the fingerprint/`Provenance.inputs` follow-ups addressed my earlier review comments.

Before merging I ran a high-effort Claude Code review over the diff (55 files, +4401/−545, vs `origin/main`). The full suite passes (3501 tests across `tests/core`, `distributions`, `modeling`, `diagnostics`, `validation`, `converters`, `linalg`, `record`), so none of these are CI-visible — each was reproduced with a standalone script and, where relevant, diffed against a `main` checkout. Posting them as-is; **please use your judgment about which to address here, which to push to a follow-up, and which to dismiss** — some may be intentional design choices, and a few clearly belong with the wrap-boundary work in #381/#369 rather than this PR.

Findings, most severe first:

- **`probpipe/core/_function_contract.py:420`** — `_validate_function_output_leaf_support` calls `bool(jnp.all(support.check(value)))` with no tracer guard, despite its docstring saying it runs "outside JAX tracing". Any Function declaring an `ArraySpec(support=...)` output raises `TracerBoolConversionError` inside `jax.vmap`: `dispatch="jax"` fails with the misleading "ensure the function is JAX-traceable", and `dispatch="auto"` silently falls back to `sequential` (verified: dropping `support=` restores `jax`).
- **`probpipe/core/_workflow_call.py:147`** — `input_ref_hint` applies the *variadic* parameter's annotation to every expanded element, so `**kw: Any` makes `build_broadcast_plan` skip sweeping at `_workflow_plan.py:66`. `Function(func=lambda **kw: kw["r"]["value"] * 2)(r=<3-row NumericRecordArray>)` returns a single `NumericRecord` here vs a `NumericRecordArray(batch_shape=(3,))` on `main`; the unannotated `**kw` still sweeps. ProbPipe's own ops all declare `**kwargs: Any`.
- **`probpipe/core/node.py:684`** — the PRNG key is re-seeded from `self._seed` per call, so repeated broadcast calls of one Function produce identical draws (`np.allclose(wf(n).samples, wf(n).samples)` is `True` here, `False` on `main`). Averaging over repeated calls no longer reduces Monte Carlo error, and per-step propagation in `iterate`/SMC loops reuses one noise realisation. (The PR body documents call-local RNG state as intentional, so this may be an expectations/docs question rather than a defect — but the MC-averaging consequence seems worth stating explicitly either way.)
- **`probpipe/core/_function_contract.py:298`** — `_validate_function_output` accepts a `RecordArray` at the template-to-template check then rejects it at the template-to-value check, so a declared `output_template` cannot describe a batched return: `ValueError: output/value has rank 1, expected rank 0 from shape ()` (both `__call__` and `apply`).
- **`probpipe/core/transition.py:189`** — `with_resampling`'s docstring still promises `dist.provenance.metadata["ess"]`, but the new result-copy semantics clear the implementation's provenance; the returned distribution carries only `{'func': ...}` and the `"resample"` node is not in its ancestry. The test was moved to `step.apply(...)`, the shipped docstring was not.
- **`probpipe/core/_workflow_result.py:159`** — `_copy_result_term` writes `_event_template`, but `RecordArray.event_template` reads `_template`, so the declared concrete template is silently not applied to batched results (contradicting the CHANGELOG claim).
- **`probpipe/core/provenance.py:228`** — every plain input is now content-hashed on every call (~14 ms per call for a 32 MB array, default LIGHTWEIGHT mode); callable/opaque inputs fall to `_update`'s `repr:` branch, embedding memory addresses and making fingerprints process-dependent — the exact problem `_update_declaration_value` was added to avoid for signatures.
- **`probpipe/core/node.py:970`** — the `_resolved_dispatch` per-instance cache was dropped, so `jax.make_jaxpr` re-probes on every `dispatch="auto"` call (1.95 ms vs 1.40 ms per trivial broadcast); the docstring still says it caches.
- **`probpipe/core/node.py:456`** — `_bind` as `MappingProxyType` makes `copy.deepcopy(fn)` raise `TypeError: cannot pickle 'mappingproxy' object` (worked on `main`).
- **`probpipe/core/_broadcast_distributions.py:804`** — the new bare-array + `event_template` branch reshapes to `batch_shape` before `_stack_declared_records`, which demands a leading axis of `prod(batch_shape)`; wrong for multi-dimensional batches, currently unreachable. The adjacent `assert len(event_template) == 1` guards (lines 430, 804) are validation expressed as `assert` and vanish under `-O`.

Happy to dig into any of these or open follow-up issues for the ones you'd rather not carry in this PR.

SoymilkWinsAgain:
>probpipe/core/node.py:684 — the PRNG key is re-seeded from self._seed per call, so repeated broadcast calls of one Function produce identical draws (np.allclose(wf(n).samples, wf(n).samples) is True here, False on main). Averaging over repeated calls no longer reduces Monte Carlo error, and per-step propagation in iterate/SMC loops reuses one noise realisation. (The PR body documents call-local RNG state as intentional, so this may be an expectations/docs question rather than a defect — but the MC-averaging consequence seems worth stating explicitly either way.)

The third impact point raised in the new review was not properly assessed previously. I didn't think about this issue much at the time. My initial thought was to avoid uncontrollable randomness related to the call order, but I forgot the underlying mathematical meanings. Currently, it appears to break the `Function` lifting mechanism.

Restoring the old `_key` requires only a small code modification, but it introduces a problem I tried to avoid: it introduces internal state of `Function`, call order dependencies, and concurrency issues. Another drawback is that this will cause `Function` to no longer be immutable, which in turn will affect fingerprint-related parts.

I didn't consider this issue at the time. The current situation is roughly as follows:

```python
wf = Function(func=f, seed=114514)
a = wf(distribution)
b = wf(distribution)
```

Previously, `a` and `b` were based on two different batches of samples; now they are based on the same batch. If `f` is deterministic, then the two results will be exactly the same. It doesn't affect other characteristics.

Restoring previous behavior is theoretically very simple. The only problem is that we must find a way to let the `ProbPipe` controller know how many calls we made.

The solution involves having the `ProbPipe` controller maintain a hidden global counter to automatically refresh the seed. Because this involves architectural changes and requires a reassessment of the impact, a separate issue will be created to explain the solution.

SoymilkWinsAgain:
> probpipe/core/provenance.py:228 — every plain input is now content-hashed on every call (~14 ms per call for a 32 MB array, default LIGHTWEIGHT mode); callable/opaque inputs fall to _update's repr: branch, embedding memory addresses and making fingerprints process-dependent — the exact problem _update_declaration_value was added to avoid for signatures.

This is a performance risk, but it does not currently cause any blockage. It will not be fixed until the behavior changes described above have been resolved.

SoymilkWinsAgain:
> probpipe/core/node.py:456 — _bind as MappingProxyType makes copy.deepcopy(fn) raise TypeError: cannot pickle 'mappingproxy' object (worked on main).

This is a bug. However, this wasn't introduced in this PR. Earlier, `Node._child_nodes` and `Node._inputs` were already `MappingProxyType`, so `deepcopy(Function)` would fail. This PR changed `Function._bind` to `MappingProxyType`. Now, `deepcopy` fails even earlier at `_bind`, but it's not the only root cause. Even if the freeze on `_bind` in #368 is lifted, `deepcopy` will still fail on the other two mapping proxies. A true fix requires an explicit implementation of the replication protocol. This is beyond the scope of this PR and will not be addressed at this time.

jhuggins (approved):

Issue #381, opened 2026-07-23
Proposal: reify term specs (TermSpec + RecordSpec) and make the wrap boundary kind-directed


jhuggins:
## Summary

Reify **term specs**: introduce a `TermSpec` taxonomy — `RecordSpec`, `DistributionSpec`, `ConditionalDistributionSpec`, `FunctionSpec`, one per kind — distinct from the `ValueSpec` leaf taxonomy, and make the wrap boundary **kind-directed and idempotent** (the identity on every tracked term). This resolves the asymmetry that currently makes term-valued results inexpressible in `output_template=`, and it settles, by one rule, the open questions about what `sample` returns and when a returned callable becomes a `Function`.

## The asymmetry (current state)

The spec taxonomy in `probpipe/core/event_template.py` has two universes that don't line up:

- `ValueSpec` **leaves**: `ArraySpec`, `OpaqueSpec`, `DistributionSpec`, `FunctionSpec`. Note that `DistributionSpec.is_valid` already validates a **tracked term** (`isinstance(value, Distribution)` + template equality) — it is a term spec in a leaf-only position.
- `EventTemplate = NamedTree[ValueSpec]` — the schema of a `Record` — sits **outside** the `ValueSpec` hierarchy, and it is the only spec admitted at top level: `output_template: EventTemplate | None`.

Consequences:

- There is no `RecordSpec`; the Rec kind's spec role is played by `EventTemplate` itself, unlike every other kind.
- "The result **is** a `Function`/`Distribution`/`ConditionalDistribution`" is inexpressible — `output_template` can only describe a `Record` with fields. This is the root of the event-vs-term tension in #369, surfaced concretely by #368/PR #380 (a fitted mapping is a term result, not an event), and of the open question of what `sample` returns for a `Distribution` whose draws are distributions or functions.

## Proposed taxonomy

Schemas stay pure; specs reify kind indices; the two are mutually recursive (leaves may be specs), and the recursion is well-founded (finite trees):

```python
# --- ValueSpec: the leaf taxonomy (base exists today) ---
class ValueSpec(ABC):
    """The structure of one leaf value; nameless until an EventTemplate
    stores it under a key. Frozen, hashable, compares by value."""
    @abstractmethod
    def is_valid(self, value: Any) -> bool: ...          # declared ONCE, here

@dataclass(frozen=True)
class ArraySpec(ValueSpec):                              # raw-value leaves: no kind
[...]

PR #382, opened 2026-07-23
docs(design): term-spec taxonomy and the kind-directed wrap boundary

jhuggins:
## Summary

Design-reference amendment for the term-spec taxonomy proposed in #381. This is **PR A** of a four-PR staging (A design docs · B the additive spec hierarchy · C the wrap boundary · D `sample`); it is docs-only and unblocked. The `design/` reference states target state, so this describes the whole taxonomy even though the code lands in slices.

Review moved the amendment past its original scope. Resolving *how a declaration stores* turned the spec into each term's single source of type, which put the slot on the tracked base and made the value-specification layer a shared abstraction. Part II therefore gains two sections and Part III loses one.

## The taxonomy

- **Two spec families.** Raw-value specs (`ArraySpec`, `OpaqueSpec`) name no kind. Term specs (`RecordSpec`, `DistributionSpec`, `ConditionalDistributionSpec`, `FunctionSpec`) each type one kind's tracked term. `TermSpec(ValueSpec)` is a marker whose concrete class *is* the kind, so `is_valid` stays declared once and a term spec is accepted anywhere a leaf is. `RecordSpec` completes the four corners.
- **Two roles, fixed by position.** At a named leaf, a term spec types a field holding a term (*as data*). As an output declaration, it states that the result *is* a term of that kind (*as result*).
- **Declarations store as specs.** A bare `EventTemplate` is accepted wherever a record declaration is meant and wraps to `RecordSpec(template)` at construction. After construction only the spec remains, so **the declared kind is the stored spec's class** — a structural test, not an inference. An *output* declaration is any `ValueSpec`, so a raw array or opaque result can be declared; an *event* declaration is record-valued, because `DistributionSpec.is_valid` checks it.
- **The spec is the single stored source of a term's type.** The slot is `TrackedTerm.spec`, declared once on the tracked base and narrowed by each kind. Convenience accessors expose that one object's properties and differ by kind: `event_template` on `Record`, `input_template` / `output_spec` on `Function`, `event_spec` on `Distribution`, `given_template` / `event_spec` on `ConditionalDistribution`. No accessor can disagree with the store or with another. A batch narrows it to the new **`BatchSpec`**, which types the *collection* — the element spec plus the named multiplicity — so `spec` is the batch's own type, `element_spec` a view on it, and a batch of values naming no kind still well typed.

## Structure

- **`Tracked` is renamed `TrackedTerm`** and carries four things: name, spec, provenance, annotations.
- **The spec layer moves to Part II** as `II.2 — Value specifications` (`ValueSpec`, `TermSpec`, the raw-value specs, `Constraint`, the two roles, the storage rule) and `II.3 — EventTemplate` (numeric promotion, symbolic dimensions, unification). `core/` houses the tracked base, so it cannot depend on the value layer: `ValueSpec`/`TermSpec`/`ArraySpec`/`OpaqueSpec`, `EventTemplate`, and `Constraint` move to `core/_specs.py`, `core/_event_template.py`, `core/_constraints.py`. Concrete term specs stay with their kinds (`RecordSpec` in `values/_record.py`, `FunctionSpec` in `values/_function_base.py`, `DistributionSpec` in `distributions/_distribution.py`, `ConditionalDistributionSpec` in `distributions/_conditional.py`).
- **`BatchSpec`** joins the term specs in II.2; II.5's level metadata becomes views on it, and every batch class carries `spec` plus an `element_spec` view.
- **Renumbering.** Old II.2–II.4 become II.4–II.6. The emptied III.1 is removed and Part III becomes III.1–III.13, so `Function` is III.1, `Record` III.2, `Distribution` III.5, `ConditionalDistribution` III.7, and the distribution batches III.8. `package-structure.md` follows.

## Per-kind changes

- **III.5 / III.7 — event declarations.** `Distribution` and `ConditionalDistribution` expose `event_spec`, a `TermSpec`, in place of an `event_template` property; a bare template still works at construction. A `DistributionSpec` as the event declaration is a random measure, whose draws are `Distribution`s; the same spec at a named leaf is a record law with a distribution-valued field. The two denote isomorphic spaces but declare different draw kinds, and the declaration decides — never `_sample`'s runtime return. `ConditionalDistribution`'s given side stays an `EventTemplate`.
- **III.5 / III.7 / III.8 — spec classes and batches.** `DistributionSpec` and `ConditionalDistributionSpec` are class-formatted like the other specs, with `event_spec: TermSpec` fields. The batch classes share their elements' declaration, so a batch of random measures declares term-valued draws as its elements do, and a batch's stored source is its element spec.
- **III.1 / III.2 — `Function` and `Record`.** `FunctionSpec.output_template` becomes `output_spec`, stored as `TermSpec | None`; the `@function` decorator parameter follows. `Record`'s constructor accepts an `EventTemplate` or a `RecordSpec`.
- **V.0 — the wrap boundary**, now five sub-bullets: a tracked term keeps its kind and is never re-wrapped; identity attaches *by derivation*, since terms are immutable and provenance is write-once, so the default result is a new term over the implementer's representation under fresh provenance while `raw=True` returns it untouched; a raw host wraps into its own kind; an output declaration is coerced and validated, wrong-kind and schema-mismatch raising distinct errors.
- **V.2 — `sample` returns the draw at the declared kind**, tracked: `RecordSpec` draws a `Record`, `DistributionSpec` a `Distribution`, `FunctionSpec` a `Function`. A non-empty `sample_shape` returns the batch type of that kind, not `RecordBatch` only. This settles what `sample` returns for a `Distribution[Distribution]` or `Distribution[Function]`.
- **III.5 — field views** exist on record laws; a term-drawing law has no fields to view. `LinOp` (III.4) keeps `input_template` / `output_template`, its `output_spec` being always a numeric `RecordSpec`.

## Notes

- No code changes. The additive hierarchy is PR B (#383); the wrap-boundary and `sample` behavior are PR C/D, gated on the first-class `Function` (#368 / #380). PR B's shipped `output_template: EventTemplate | TermSpec | None` becomes constructor sugar over a stored `output_spec`, and the base types grow the `spec` slot with accessors re-derived as views.
- Per #350's standing directive the design reference wins, so the clauses this amends — several of which state the opposite of the declaration-directed rule — should land before the C/D code is reviewed against settled text.
- The companion formalization tracks these decisions: outputs are value specifications on the `Dist`, `Cond`, and `Fun` sides, with the record output kept as sugar, and the base-type discussion states the spec-slot rule and the wrap at construction.

Part of #381.


design/03-values-and-distributions.md
```diff
@@ -27,13 +27,18 @@ The sections build in the order below, each depending only on those above it and
...
+
+A term spec is used in two positions, and position alone fixes its role. As a **leaf** of a template it types a field that holds a term — the *as-data* role, a `Distribution` stored inside a record, say. Heading an operation's **output declaration** it states that the operation's result *is* a term of that kind — the *as-result* role, how a fitted mapping declares that it returns a `Function`. An `EventTemplate` is not itself a spec: it is the *schema*, the structure that indexes a kind, and it is never one of its own leaves. A record output is declared with a bare `EventTemplate` directly, or as `RecordSpec(template)` where a uniform `TermSpec` is wanted — the two denote the same space, differing only in reading.
```
SoymilkWinsAgain: I suspect there's a minor bug: `Distribution` cannot represent the top-level kind of an event. The documentation specifies that when `TermSpec` is a leaf element of `EventTemplate`, it represents as-data, i.e., a field in the `Record`. However, random measures use the same leaf element of `DistributionSpec` to represent the sampling result itself as a top-level `Distribution`. Therefore, it's currently impossible to distinguish whether it represents a `Record` where field `x` stores a `Distribution`, or a random measure where its `draw` is directly a `Distribution`. `Distribution` still only carries `event_template: EventTemplate`, so the declaration doesn't store the "event root kind". We can only guess based on the runtime return value of `_sample()`, which is inconsistent with the goal of "declaration determining kind, wrong-kind reporting a separate error".

design/03-values-and-distributions.md
SoymilkWinsAgain: Should be `EventTemplate | TermSpec | None`. The same for below

design/05-operations.md
```diff
@@ -12,7 +12,7 @@ Every operation shares the following mechanics:
...
+- **The wrap boundary.** The boundary is kind-directed and wraps only the untracked. **A tracked term is never re-wrapped**: a `Record`, `Distribution`, `ConditionalDistribution`, or `Function` an operation returns is the result as-is. **A raw host wraps into its own kind**: a bare array or other raw non-mapping value becomes the single-field `Record` its `event_template` describes, a raw mapping becomes a structured `Record`, a raw callable becomes a `Function`, and a backend distribution enters through its converter. A concrete family such as a `Normal` can therefore be written in plain arrays, the user writing the family's natural constructor rather than supplying a template by hand. When the operation carries an output declaration (III.1), the result is coerced to the declared kind and validated, a wrong kind and a schema mismatch raising distinct errors; a raw callable is a `Fun` unless the declaration is a `ConditionalDistributionSpec`, the one distinction the value itself cannot reveal. A `ConditionalDistribution` adds the `given=` fused paths over its `_conditional_*` methods.
```
SoymilkWinsAgain: From CodeX:

"Returning tracked terms as is" is incompatible with fresh sample provenance.

V.0 specifies that tracked terms are "never re-wrapped" and returned as the same object; V.2 requires that the default sample appends a new sample provenance to an existing term. In the existing Tracked contract, provenance is write-once, and `with_provenance()` modifies and returns the original object.

If drawing an existing provenance, the new sample provenance will fail immediately; if not, the default sampling will modify the object returned by the implementer, and it will no longer be a true "pass-through as is". It needs to be clarified whether to clone the identity, wrap the lineage, or modify the provenance model.

design/05-operations.md
```diff
@@ -45,9 +45,9 @@ A mean is defined whenever draws can be averaged: coordinate-wise for arrays, po
 ### Contract
 
 `sample(d, key, sample_shape=(), raw=False)` draws from a distribution.
-- With `sample_shape=()` it returns a single draw, wrapped as the `Record` the `event_template` describes. A scalar law returns a single-field `Record` keyed by the distribution's name (the key is initialized from the name and thereafter independent of it). The wrap is unconditional: even a draw whose raw type carries identity, such as the `Distribution` a random measure yields, is returned as a `Record` with that draw as its leaf value.
+- With `sample_shape=()` it returns a single draw **at the event's own kind**, tracked. A draw over a record event is a `Record`; a scalar law returns a single-field `Record` keyed by the distribution's name (initialized from the name, thereafter independent of it). A draw whose event kind is itself a term is that term: a random measure's draw is a `Distribution`, a random function's draw is a `Function`. The record wrap applies only to a raw draw (V.0), so `sample` never buries a term as a `Record`'s leaf value.
```
SoymilkWinsAgain: If I understand correctly, this is still based on the Record-only model?

SoymilkWinsAgain (changes requested):
I couldn't find any instructions on which module the newly added `TermSpec` and `RecordSpec` should be placed in. Seems they were missing.

jhuggins:
@SoymilkWinsAgain Thanks, this motivated me to address some underlying issues. 

## Your comments

**1. `Distribution` cannot represent the top-level kind of an event.** Correct, and the gap was in the model as much as in the prose: the draw kind was being read off `_sample`'s runtime return, which contradicts declaration-determined kinds. The fix generalizes the output side. An event declaration is now a value specification, not a template, so `DistributionSpec` as the *declaration* is a random measure whose draws are `Distribution`s, while the same spec at a *named leaf* is a record law with a distribution-valued field. The two denote isomorphic spaces but declare different draw kinds. `Distribution` and `ConditionalDistribution` therefore expose `event_spec` rather than an `event_template` property, and a wrong-kind result raises the kind error, distinct from a schema mismatch (III.5, III.7, V.2).

**2. `EventTemplate | TermSpec | None` in the `Function` block.** Fixed, then superseded: see the storage rule below. The stored field is now `output_spec: TermSpec | None`, with the union surviving only at the constructor.

**3. CodeX on "returning tracked terms as is" vs. fresh `sample` provenance.** A genuine contradiction, and the resolution is that "never re-wrapped" was about **kind, not object identity**. Since terms are immutable and provenance is write-once, the default result is a *new* term over the implementer's representation under fresh name and provenance; the implementer's object is never modified, and `raw=True` returns it untouched. V.0's wrap boundary is now five sub-bullets, one of which states exactly this, and V.2's `raw` bullet matches.

**4. "Still based on the Record-only model?"** Yes, and fixed in two places. A non-empty `sample_shape` returns the batch type of the draw's kind (`DistributionBatch`, `FunctionBatch`, …), not `RecordBatch` only; and the batch classes share their elements' event declaration, so a batch of random measures declares term-valued draws as its elements do (III.8).

**5. Module placement for `TermSpec` and `RecordSpec`.** You were right that it was missing. It is now stated in the design text and in `package-structure.md` — with one revision forced by the layering, described next: `TermSpec` sits beside `ValueSpec`, but both now live in `core/`, not `values/`.

## What went beyond your comments

Resolving *how* a declaration is stored led to something we had already discussed but wasn't implemented: **a declaration is stored as a spec, and the spec is the single stored source of a term's type.** A bare `EventTemplate` is accepted wherever a record declaration is meant and wraps to `RecordSpec(template)` at construction, so after construction the declared kind is simply the stored spec's class — a structural test rather than an inference. That put the slot on the tracked base, which had three consequences:

- `Tracked` is renamed **`TrackedTerm`** and carries `spec: TermSpec` alongside name and provenance; each kind narrows it and exposes accessors for that one object's properties (`event_template` on `Record`, `input_template` / `output_spec` on `Function`, `event_spec` on `Distribution`, `given_template` / `event_spec` on `ConditionalDistribution`).
- Because `core/` houses the tracked base, it cannot depend on the value layer, so `ValueSpec`, `TermSpec`, `ArraySpec`, `OpaqueSpec`, `EventTemplate`, and `Constraint` move to `core/_specs.py`, `core/_event_template.py`, and `core/_constraints.py`. The concrete term specs stay with their kinds: `RecordSpec` in `values/_record.py`, `FunctionSpec` in `values/_function_base.py`, `DistributionSpec` in `distributions/_distribution.py`, `ConditionalDistributionSpec` in `distributions/_conditional.py`.
- The spec layer is a shared abstraction, so it moved into Part II as `II.2 — Value specifications` and `II.3 — EventTemplate`. Old II.2–II.4 shift to II.4–II.6, the emptied III.1 is gone, and Part III is renumbered III.1–III.13 (`Function` III.1, `Record` III.2, `Distribution` III.5, `ConditionalDistribution` III.7, the distribution batches III.8).

## What this means for #383

Nothing on the branch is wrong, but two things change shape:

- `FunctionSpec.output_template` becomes `output_spec`, and the union you flagged moves to the constructor: the stored field is `TermSpec | None`, with an `EventTemplate` wrapped to `RecordSpec(template)` on the way in.
- The base types grow the `spec` slot, with `event_template` and friends re-derived as views on it rather than stored independently. `Distribution`'s declaration widens to accept a `TermSpec` at the same time.

design/02-shared-abstractions.md
```diff
@@ -76,11 +78,142 @@ Using named paths is necessary to satisfy `C5 – Naming for unambiguous meaning
...
+
+---
+
+## II.4 — Identity, type & metadata: `TrackedTerm`, `Annotated`, `Provenance`
+
+### Contract
+
+Identity, type & metadata is the cross-cutting layer that lets any object carry, alongside its raw representation, four things: a **name** (what the object is called), a **spec** (the declaration of its type, II.2), a **provenance** (how it was produced), and free-form **annotations** (auxiliary information supplied by the user or an algorithm). The structure is provided by two mixins: `TrackedTerm` (name, spec, and provenance) and `Annotated` (annotations). Every first-class object, the kind an operation consumes and produces, must be a `TrackedTerm`, while structural helpers such as templates and specs are not. We call any such object a **tracked term**: a value, distribution, conditional distribution, linear operator, or batch that carries a name, its spec, and provenance.
```
SoymilkWinsAgain, line 223: `TrackedTerm.spec` is defined as a `TermSpec` and claims to be "the sole source of the term's own type"; however, `Batch` actually treats `.spec` as an element spec:

- `FunctionBatch.spec` -> `FunctionSpec`, but `FunctionBatch` itself is not callable.
- `DistributionBatch.spec` -> `DistributionSpec`, but it is not a `Distribution` itself.
- `OpaqueBatch.spec` -> `OpaqueSpec`, violating `TrackedTerm.spec: TermSpec`.
- `RecordBatch` contract doesn't declare the required `.spec`.

Therefore, whether `.spec` describes the batch itself or its elements is currently inconsistent, and the new single-source principle cannot be implemented.

design/04-functions.md
```diff
@@ -68,10 +68,10 @@ A `Function` compares each argument against the type its function expects, and l
...
...f — a `Function`, `Distribution`, or `Record` the callable produced is the result, not a field buried inside a fresh `Record`. A **raw callable** becomes a `Function`. A raw **mapping** becomes fields keyed by its keys, since mappings are never leaves. Any other raw return becomes one field keyed by the `Function`'s name, its spec inferred — an `ArraySpec` for an array and an `OpaqueSpec` otherwise, tuples included; an opaque output samples downstream but carries none of the numeric machinery, so a function whose output deserves structure returns a mapping or declares it. The optional `output_spec=` on the decorator declares what the producer knows and inference cannot recover. It accepts any `TermSpec` (II.2): at top level it declares the result's kind — a `FunctionSpec` for a fitted mapping, a `ConditionalDistributionSpec` for a predictive kernel — while a record schema (an `EventTemplate`, i.e. a `RecordSpec`) declares field structure such as a constrained support or a shared symbolic dimension. The declaration is bound per call by unification, which validates the output and, for a nested `RecordSpec`, keeps a tracked record field as a term-valued leaf rather than flattening it.
```
SoymilkWinsAgain: `Function.output_spec` is defined as both `ValueSpec` and `TermSpec` simultaneously.

Version II.2 and the latest intent allow arbitrary `ValueSpec`s such as `ArraySpec` and `OpaqueSpec`; the `Function` constructor also accepts `ValueSpec`. However, the property return type is still `TermSpec | None`, while Version IV.2 states that decorators only accept `TermSpec`.

Note that is not merely a difference in wording, because it determines if a user can declare raw array/opaque outputs. Considering the last few commits, the ultimate intent is clearly `ValueSpec`, therefore the property, Version IV.2, and the PR body are not yet synchronized.

design/04-functions.md
```diff
@@ -59,7 +59,7 @@ This makes `C1 – Uniform interface to distributions and values` and `C4 – Fu
...
 
-**The trigger.** A parameter that is unannotated, or annotated with a value type, expects a value, so a distribution passed in that position is lifted. A parameter annotated `Distribution`, `Distribution[...]`, or a distribution capability protocol of III.7 declares that the function consumes the distribution itself, which then passes through unlifted. The function capabilities of III.2 and III.14 annotate `Function`-valued parameters, which are values, so the value rule above governs them. Per draw, the function receives what `sample` returns, the draw's `Record`.
+**The trigger.** A parameter that is unannotated, or annotated with a value type, expects a value, so a distribution passed in that position is lifted. A parameter annotated `Distribution`, `Distribution[...]`, or a distribution capability protocol of III.6 declares that the function consumes the distribution itself, which then passes through unlifted. The function capabilities of III.1 and III.13 annotate `Function`-valued parameters, which are values, so the value rule above governs them. Per draw, the function receives what `sample` returns, the draw's `Record`.
```
SoymilkWinsAgain: V.2 correctly specifies that random functions/random measures return `Functions`/`Distributions` respectively, but other contracts still assume all draws are Records:

- `Function` lifting states that each call receives "the draw's Record".
- `include_inputs=True` unconditionally reads the `.event_template` of the parameters, while term-drawing Distributions only have `.event_spec`.
- Mean/variance still unconditionally declares that it returns a Record.
- `sample(DistributionBatch)` still states that the inner layer must be a RecordBatch.
- Random functions/measures are still described as FunctionSpec/DistributionSpec "leaf", contrary to the semantics of the new root declaration.
- One place still states that the field interface is available for "every distribution", even though the term-drawing law explicitly states there are no fields.

Therefore, the behavior of `Distribution[Function]` and `Distribution[Distribution]` after entering lifting, moments, or batched paths remains unclosed.

design/package-structure.md
```diff
@@ -24,38 +24,38 @@ probpipe/
 ├── __init__.py                # the curated public API
 ├── core/                      # Part II — shared abstractions
 │   ├── _named_tree.py         #   NamedTree (II.1)
-│   ├── _identity.py           #   Tracked, Annotated, Provenance, fingerprints (II.2)
-│   ├── _batch.py              #   Batch: axis groups, level names, select (II.3)
-│   ├── _dispatch.py           #   dispatch methods and registries (II.4)
-│   ├── _catalog.py            #   EntrySummary, RegistryCatalog (II.4)
+│   ├── _constraints.py        #   Constraint and the constraint factories (II.2)
```
SoymilkWinsAgain, line 27: The principle section states that `ArraySpec`/`OpaqueSpec` is located in values, but the directory tree and PR body place it in core; `_event_template.py` is labeled II.2, but it is actually II.3; the migration table moves `core/event_template.py` and `core/constraints.py` to values, which is the opposite of the target tree.

SoymilkWinsAgain (commented):
Also PR body needs to update

jhuggins:
@SoymilkWinsAgain All four land, and the first one was a genuine hole in the amendment rather than a wording slip. Addressed in 3bfa3f4c.

**A batch's `.spec` (02:216).** You are right that the principle could not be implemented as written: `spec` was declared to name the term's *own* type while every batch used it for the *element's*, so `OpaqueBatch.spec -> OpaqueSpec` was not even a `TermSpec`, `FunctionBatch` was specified as callable though it is not, and `RecordBatch` declared no spec at all.

New **`BatchSpec`** in II.2 resolves it as a term spec in its own right, carrying the element spec together with the named multiplicity. `spec` now means the same thing on a batch as on any other term, `element_spec` / `axis_groups` / `level_names` are views on it, `OpaqueBatch` is well typed, and `RecordBatch` gains the spec it was missing. It also matches the model, where a batch inhabits the *family* kind over its element's kind rather than the element kind itself — the companion formalization said "a batch's spec is its element's" and has been corrected the same way.

**`output_spec` (04:71).** Correct, and correct that it is not just wording. The widening reached the prose and the constructor but not the property, which still returned `TermSpec`, nor IV.2, which still said the decorator takes only a `TermSpec`. Both now read any value specification. Note the code in #383 already stores `ValueSpec | None`, so the design was the side that had drifted.

**Record-only assumptions (04:62).** All six generalized: lifting now receives the draw at the kind the law's event declaration names; `include_inputs` says what a term-drawing law contributes; `mean` / `variance` and `expectation` wrap at the declared kind, so a random function's mean is a `Function`; batched `sample` over a `DistributionBatch` yields the batch form of the declared kind; the field interface is offered only where fields are declared. Random functions and measures are also no longer described as a `FunctionSpec` / `DistributionSpec` **leaf**, in Part III and in the catalog — that phrasing predated the root-declaration reading and directly contradicted it.

Worth adding for scope: in #383 a term-valued *event* declaration is refused at construction, precisely because `is_valid` would otherwise have to report "not implemented" as "invalid". So `Distribution[Distribution]` is not constructible in code today, and these six were a design-side gap rather than a live behavioural hole. The `Distribution`-side support that makes such a declaration checkable is separate work.

**package-structure.md (pkg:27).** All three were mine and all three are fixed: the principle sentence now states the layering reason `ValueSpec`, `TermSpec`, `ArraySpec`, and `OpaqueSpec` sit in `core/` — they are what the tracked base stores, and `core/` cannot import the value layer — the event-template module is labelled II.3, and the migration table no longer sends `core/event_template.py` and `core/constraints.py` to `values/`, which was backwards.

PR body updated too.

One thing to flag rather than let you find it: II.5 still says `select_at`. That method is renamed to `at_levels` in #386, which also carries the II.3-versus-II.5 numbering for it, so the rename is not duplicated here — it would only conflict.

PR #383, opened 2026-07-23
feat(core): term-spec hierarchy and declarations stored as specs

jhuggins:
## Summary

The term-spec sub-hierarchy from #381, together with the declaration-storage rule it exists to support. This is **PR B** of the staging in #382 (A design docs · B this · C the wrap boundary · D `sample`).

The two were originally planned as separate PRs. They are folded here because the storage rule's normalisation *needs* `RecordSpec` — the class this PR introduces — so split apart, B would ship a `RecordSpec` with no consumer and B2 would immediately rename an attribute B had just widened. Landing them together also means one breaking change rather than two, and one coherent hierarchy per release rather than a half-applied convention.

## The hierarchy

`ValueSpec` splits into two families:

- **Raw-value specs** — `ArraySpec`, `OpaqueSpec` — type a plain value that names no ProbPipe kind.
- **Term specs** — one per kind, whose concrete class *is* the kind. `TermSpec(ValueSpec)` is the marker `isinstance` reads. New `RecordSpec` completes the four corners beside `DistributionSpec`, `FunctionSpec`, and the conditional spec still to come; `DistributionSpec` and `FunctionSpec` are reparented under `TermSpec`.

Because a tracked term is still one value that can occupy a leaf, a term spec *is* a value spec: `is_valid` stays declared once on `ValueSpec`, and anywhere a leaf is accepted a term spec is accepted unchanged.

## The storage rule

A **declaration** — of an event or of an output — is stored as a `ValueSpec`. A bare `EventTemplate` is accepted as construction sugar wherever a record declaration is meant, and is normalised to `RecordSpec(template)`, mirroring the existing `_FieldSpecInput` sugar in the same module. Of the stored specs only a `TermSpec` names a ProbPipe kind, and it names it by its class; a raw-value spec declares a raw value instead. After construction only the spec remains, so the declared kind is a **structural test** rather than an inference from a runtime value.

```python
DistributionSpec(EventTemplate(x=())).event_spec  # RecordSpec(...) — a record law
FunctionSpec(tau, tau).output_spec                # RecordSpec(...) — a record output
FunctionSpec(tau, FunctionSpec()).output_spec     # FunctionSpec    — returns a function
FunctionSpec(tau, DistributionSpec(tau))          # returns a Distribution
```

### Renames (breaking)

| Before | After | Stored as |
|---|---|---|
| `FunctionSpec.output_template` | **`output_spec`** | any `ValueSpec`, or `None` |
| `DistributionSpec.event_template` | **`event_spec`** | `RecordSpec` (see below) |

**Positional** construction is unchanged, and both constructors still accept an `EventTemplate` and wrap it, so `DistributionSpec(tau)` and `FunctionSpec(tau, tau)` keep working. **Keyword** construction moves to the new parameter name, as does every **read** of the old attribute:

```python
DistributionSpec(event_template=tau)    ->  DistributionSpec(event_spec=tau)
FunctionSpec(tau, output_template=tau)  ->  FunctionSpec(tau, output_spec=tau)
spec.event_template                     ->  spec.event_spec.event_template
```

There are no deprecation aliases: at `0.1.0` the only call sites outside the defining module are in its own test file, and an alias would preserve exactly the name the design rejects.

### Each field is declared at the type it stores

A field's annotation is the value it **holds**, not what the constructor accepts: `event_spec: RecordSpec` and `output_spec: ValueSpec | None`. The wider template spelling is carried by an explicit `__init__` (`dataclass(init=False)`, which keeps `eq`, `hash`, `repr`, and `dataclasses.replace`), so a type checker and the generated API reference both read the post-construction guarantee — and `is_valid` no longer casts its way back to it. `_to_record_declaration` is the narrow-in, narrow-out normaliser the record path needs, which retires the `cast` and its import rather than relocating it.

`ArraySpec` gets the same split, since it had the same mismatch: `dtype` is now declared as the `numpy.dtype` it stores rather than the `DTypeLike` spellings it accepts, and `shape` as the tuple rather than the iterable.

`RecordSpec.event_template` **keeps** its name. That is the naming rule, not an oversight: `*_template` is always a record schema, `*_spec` a declaration of any kind. Please don't apply the rename mechanically to it.

## What each declaration admits

`FunctionSpec.output_spec` accepts **any value specification**, matching the model's `Fun(σ, ρ)` with `ρ` a value specification and `E_Fun(σ,ρ) = Meas(E_σ, E_ρ)`: a callable may declare a term result of any kind, or a raw-value result that types the value the wrap boundary then places in a single-field `Record` keyed by the function's name. Nothing there is expressible-but-unsatisfiable, because the output side is unchecked — validity is callability alone.

An **event** declaration is narrower, and for a different reason: `DistributionSpec` takes an `EventTemplate` or a `RecordSpec` and **refuses a term-valued declaration**.

The reason is that a random measure could otherwise be *declared* but never *satisfied* — a `Distribution` exposes an `EventTemplate` and nothing that reports a term-valued draw kind — so `is_valid` would have to report "not implemented" as "invalid". That is exactly the conflation the two-partialities rule forbids (mathematically undefined versus computationally unavailable; design `III.5`'s error boundary). Refusing at construction removes the unsatisfiable state rather than choosing an answer about it, and puts the error where the declaration is written. Widening is non-breaking and belongs with the `Distribution`-side support that makes it checkable.

## `RecordSpec`'s leaf role is deferred

A `RecordSpec` is reached as a *declaration*, not as a template leaf: `Record` construction materialises a record-valued field into an interior template node, and `infer_from` does the same, so no in-tree route produces a `RecordSpec` leaf. The docstrings say so rather than describing the leaf role as live — `is_valid` is written for the leaf role a kind-directed wrap boundary will use, and that boundary is PR C.

## Fingerprinting

The stored declaration is a spec, not a template, so `_fingerprint.py` recurses into it and gains an explicit `RecordSpec` branch.

Chasing that surfaced a pre-existing defect worth its own note. The spec hasher records the object type and declaration fields, but the *generic* hasher had no `ValueSpec` branch, so a spec was routed there only as a template leaf. Reached any other way — bare, or inside a tuple, list, or mapping — it fell through to **identity** hashing, so two *equal* declarations produced different fingerprints and silently broke jit cache keys and provenance. `_update` now routes any `ValueSpec` to the spec hasher.

Deliberately **not** in this PR:

- `Tracked` → `TrackedTerm` and the `spec` slot on the tracked base, with the per-kind accessors becoming views. That touches nine files and has to absorb the scattered `self._event_template = …` assignments in `_empirical`, `_joint_empirical`, `_broadcast_distributions`, and `_record_distribution` — substrate work that should be revertable on its own.
- `Distribution` / `ConditionalDistribution` declarations accepting a `TermSpec`, and the batch classes. These depend on that slot.
- The wrap boundary and `sample` (PRs C/D), gated on the first-class `Function` (#368 / #380).

## Tests

`tests/core/test_event_template.py`: the renames throughout, plus tests for the storage rule — wrapping, idempotent normalisation, a term *output* declaration kept unwrapped, a term *event* declaration refused, the declared kind as the stored class, and `None` staying unspecified. Two tests that pinned the old store-as-given behaviour now assert the new contract. Two more pin the field split: that each declaration field's annotation resolves to the stored type (so a rewidening is caught), and that positional construction survives the rename while `event_template=` / `output_template=` raise.

`tests/core/test_fingerprint.py` gains `TestValueSpecFingerprints` (14 tests): every spec kind is reachable by the hasher, equal declarations agree while distinct declarations and distinct declared *kinds* do not, an unspecified output differs from a declared one, and a spec in each of the bare, in-tuple, in-list, and in-dict positions hashes strongly and stably. That class is what would have caught the two `AttributeError`s this PR briefly had after rebasing, and the identity-hashing defect above; both were verified to fail without their fixes.

**1020 pass** across every test module that touches these specs in `tests/core`, plus **114 pass** in `tests/inference` and `tests/modeling`. `ruff check` and `ruff format` clean at the pinned 0.15.16. `pyright` on `event_template.py` reports the same two pre-existing errors as before, both outside this diff's hunks — the change adds none and removes one `cast` suppression.

## Rebased

This branch is rebased onto `origin/main` (`f1987a4c`); it previously sat on the pre-`Function` merge base, which is why the two consumers above were invisible to it. Note one transient asymmetry until the follow-on lands: `FunctionSpec.output_spec` sits beside `Function.output_template`, which is a different attribute and is not renamed here.

Part of #381. Design reference: #382.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Value specs are fingerprinted by declaration, not identity (#381).** The
+  spec hasher now covers `RecordSpec` and recurses into a stored declaration
+  (`DistributionSpec.event_spec`, `FunctionSpec.output_spec`), which is a spec
+  rather than a template. The generic hasher also routes any `ValueSpec` to it,
+  so a spec reached other than as a template leaf — bare, or inside a tuple,
+  list, or mapping — records its type and declaration fields instead of falling
+  through to identity hashing. Previously such a spec hashed weakly, so two
+  *equal* declarations produced different fingerprints and silently broke jit
+  cache keys and provenance.
+
 ### Added
 
 - **`TermSpec` — the term-spec sub-hierarchy, and declarations stored as specs
@@ -28,8 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   declaration is record-valued for now, since a `Distribution` exposes an
   `EventTemplate` and nothing that reports a term-valued draw kind, so a
   random-measure declaration is refused at construction rather than accepted
-  and always reported invalid. Fingerprinting recurses into declarations and
-  covers `RecordSpec`, so equal declarations hash equally.
+  and always reported invalid.
 
 - **First-class, tracked `Function` values (#368).** `Function` is now an
   immutable `Node` / `Tracked` / `Annotated` object with a construction-time
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -205,11 +205,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **Renamed, for the storage rule (#381):** `FunctionSpec.output_template` is
   now **`output_spec`**, storing any `ValueSpec` or `None`, and
   `DistributionSpec.event_template` is now **`event_spec`**, storing a
-  `RecordSpec`. Both constructors still accept an `EventTemplate` and wrap it, so
-  existing positional and keyword construction from a template keeps working;
-  reads of the old attribute names do not. The names now carry their content:
-  `*_template` is always a record schema, `*_spec` a declaration of any kind —
-  which is why `RecordSpec.event_template` keeps its name.
+  `RecordSpec`. The names now carry their content: `*_template` is always a
+  record schema, `*_spec` a declaration of any kind — which is why
+  `RecordSpec.event_template` keeps its name.
+
+  To migrate: **positional** construction is unchanged, and both constructors
+  still accept an `EventTemplate` and wrap it, so `DistributionSpec(tau)` and
+  `FunctionSpec(tau, tau)` keep working. **Keyword** construction moves to the
+  new parameter name, and so does every **read** of the old attribute:
+
+  ```python
+  DistributionSpec(event_template=tau)    ->  DistributionSpec(event_spec=tau)
+  FunctionSpec(tau, output_template=tau)  ->  FunctionSpec(tau, output_spec=tau)
+  spec.event_template                     ->  spec.event_spec.event_template
+  ```
+
+  Each field is now declared at the type it *stores* — `event_spec: RecordSpec`,
+  `output_spec: ValueSpec | None` — with the wider template sugar carried by the
+  constructor signature, so a type checker and the generated API reference both
+  read the post-construction guarantee. `ArraySpec` follows the same split, its
+  `dtype` field declared as the `numpy.dtype` it stores rather than the
+  `DTypeLike` spellings it accepts.
 
 - **Function calls establish a new result identity and provenance boundary
   (#368, breaking).** Existing operations such as `condition_on` and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,8 +23,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   bare `EventTemplate` is accepted as construction sugar wherever a record
   declaration is meant and normalised to `RecordSpec(template)`, so after
   construction the declared kind is simply the stored spec's class. This makes
-  `sample`'s draw kind and an operation's result kind a structural test rather
-  than an inference from a runtime value.
+  an operation's result kind a structural test rather than an inference from a
+  runtime value. A callable's `output_spec` accepts any kind; an *event*
+  declaration is record-valued for now, since a `Distribution` exposes an
+  `EventTemplate` and nothing that reports a term-valued draw kind, so a
+  random-measure declaration is refused at construction rather than accepted
+  and always reported invalid. Fingerprinting recurses into declarations and
+  covers `RecordSpec`, so equal declarations hash equally.
 
 - **First-class, tracked `Function` values (#368).** `Function` is now an
   immutable `Node` / `Tracked` / `Annotated` object with a construction-time
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   list, or mapping — records its type and declaration fields instead of falling
   through to identity hashing. Previously such a spec hashed weakly, so two
   *equal* declarations produced different fingerprints and silently broke jit
-  cache keys and provenance.
+  cache keys and provenance. Because a record declaration is now stored as a
+  `RecordSpec`, the digest of a template carrying a `DistributionSpec`, or a
+  `FunctionSpec` with a declared output, also changes value; fingerprints are
+  in-memory jit cache keys and provenance only, never persisted.
 
 ### Added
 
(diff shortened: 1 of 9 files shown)


probpipe/core/event_template.py
```diff
@@ -288,22 +328,107 @@ class DistributionSpec(ValueSpec):
...
+    Raises
+    ------
+    TypeError
+        If ``event_spec`` is neither an :class:`EventTemplate` nor a
+        :class:`RecordSpec`.
+    """
+
+    # ``_EventDecl`` types the constructor *input* (a template or a RecordSpec);
+    # ``__post_init__`` normalises it, so the stored value is always a RecordSpec.
+    event_spec: _EventDecl
```
SoymilkWinsAgain: The type annotations for public fields still describe "construction parameters". `DistributionSpec.event_spec` is exposed as `EventTemplate | RecordSpec`, and `FunctionSpec.output_spec` is exposed as `EventTemplate | ValueSpec | None`. However, the core guarantee of the PR is that after construction, they can only be `RecordSpec` and `ValueSpec | None`, respectively. Therefore, the type checker and the generated API documentation cannot utilize this guarantee. It is recommended to annotate storage fields with their actual types and express broader construction inputs separately through custom constructors, `InitVar`, or read-only properties.

CHANGELOG.md
```diff
@@ -159,6 +202,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **Renamed, for the storage rule (#381):** `FunctionSpec.output_template` is
```
SoymilkWinsAgain, line 205: The actual migration rule is: positional parameters can still be passed to templates, but keyword calls must be changed to `event_spec=` and `output_spec=` respectively. The current explanation will directly mislead users affected by the breaking change.

probpipe/core/event_template.py
```diff
@@ -288,22 +328,107 @@ class DistributionSpec(ValueSpec):
...
+        if not isinstance(value, Record):
+            return False
+        try:
+            template = value.event_template
+        except (AttributeError, TypeError):
+            return False
+        return template == self.event_template
+
+
+# A *declaration* — of an event or of an output — is stored as a ``TermSpec``,
```
SoymilkWinsAgain: This seems to relate to the later allowance of output declarations as arbitrary `ValueSpecs`. It is recommended to uniformly describe it as follows: declaration is stored as `ValueSpec`; template sugar is normalized as `RecordSpec`; among them, only `TermSpec` can declare `ProbPipe` kind through its specific type

SoymilkWinsAgain (commented):
Also, PR body should be updated

jhuggins:
@SoymilkWinsAgain All three were right, and the first two were defects rather than wording. Addressed in 179b55fe.

**Field annotations vs. stored types (`event_template.py:410`).** Correct, and it was our own design text we were diverging from: #382 already writes `output_spec: ValueSpec | None` with the wrap as a construction note. The `cast(RecordSpec, self.event_spec)` in `is_valid` was the tell — the guarantee this branch exists to establish was being *asserted* to the type checker rather than expressed.

Both fields are now declared at the type they hold, with the wider template spelling carried by an explicit `__init__` under `dataclass(init=False)`. That keeps `eq`, `hash`, `repr`, and `dataclasses.replace` intact (verified), and it means the constructor signature is where the sugar is documented — which is also where a reader looks for it. A narrow `_to_record_declaration` retires the `cast` and its import instead of relocating it.

I took the same fix to `ArraySpec`, which had the same mismatch on `dtype` (`DTypeLike` annotated, `numpy.dtype` stored) and on `shape` (tuple annotated, any iterable accepted). It predates this branch, but leaving it would have made the rule hold for two specs out of three in one file.

Of the alternatives you listed I did not use `InitVar` or a read-only property: a private stored field renames the dataclass field, which changes `repr`, the `replace()` keywords, and the public read path, for nothing a typed `__init__` doesn't already give.

**The migration rule (`CHANGELOG.md:205`).** Right, and this was the most user-visible of the three — the entry claimed keyword construction still worked. Verified that it does not:

```
DistributionSpec(event_template=tau)    -> TypeError: unexpected keyword argument
FunctionSpec(tau, output_template=tau)  -> TypeError: unexpected keyword argument
```

The entry now states the three cases separately — positional unchanged, keyword moves to the new parameter name, reads move too — with a before/after block. A test pins it, since nothing did.

**"stored as a `TermSpec`" (`event_template.py:356`).** Agreed, and it was a genuine inconsistency rather than imprecision: the sentence predates `output_spec` widening to `ValueSpec`, and the paragraph directly beneath it says the opposite. I adopted your three-clause formulation almost verbatim — stored as a `ValueSpec`; template sugar normalised to `RecordSpec`; only a `TermSpec` names a kind, by its class. The same stale claim was in a test section comment and in the PR body; all three are corrected.

**PR body** updated to the full diff.

Two notes on scope. `pyright` on `event_template.py` reports the same two pre-existing errors as before, both outside this diff — no new ones, one fewer suppression. And #386 has this branch merged into it, so the same field-declaration split applies to `BatchSpec`, whose `axis_groups` accepts any iterable-of-iterables while declaring the stored tuple; that follows there rather than here.

PR #385, opened 2026-07-27
refactor(core): rename Tracked to TrackedTerm

jhuggins:
## Summary

Renames the identity mixin `Tracked` to **`TrackedTerm`**, and its private metaclass `_TrackedMeta` to `_TrackedTermMeta`. Mechanical and behaviour-preserving.

The name says what the mixin makes an object: a tracked *term*, the kind an operation consumes and produces, as distinct from the templates and specs that are structural helpers and are deliberately not tracked. The design reference already calls it `TrackedTerm`, so this brings the code into line with settled text rather than proposing anything new.

## Scope

92 occurrences of the class and 6 of the metaclass, across 27 files: the mixin itself, the distribution and record layers, the workflow layer's `isinstance` checks and type annotations, `provenance.py`, two backend adapters, the tests, `docs/api/*`, and the convention docs.

**Breaking.** `probpipe.Tracked` was exported, so `probpipe.TrackedTerm` replaces it with no compatibility alias — at `0.1.0` an alias would preserve exactly the name the design rejects.

## Deliberately not here

- **The module keeps its name.** `probpipe/core/tracked.py` holding `TrackedTerm` is a small inconsistency, but `package-structure.md` already targets `core/_identity.py` for this content, so moving the file belongs with that layout change, not with a rename. `tests/core/test_tracked.py` mirrors the module and stays with it.
- **`design/` is untouched.** #382 renames the design text; doing it here too would only manufacture a conflict.
- **The `spec` slot** the design gives `TrackedTerm` is separate work: it depends on `RecordSpec` from #383 and has to absorb the scattered `self._event_template = …` assignments in `_empirical`, `_joint_empirical`, `_broadcast_distributions`, and `_record_distribution`. That should be revertable on its own.

## `git blame`

The rename is listed in `.git-blame-ignore-revs`, beside the `ruff format` reflow and the `WorkflowFunction` → `Function` rename. Verified: with the ignore file active, `class TrackedTerm` attributes to `f3e92af5` — the commit that actually wrote the class — rather than to the rename, and every SHA in the file resolves.

GitHub applies the file automatically. Locally:

```
git config blame.ignoreRevsFile .git-blame-ignore-revs
```

## Interaction with the open PRs

Independent of both, and safe to review in any order. The overlap with #383 is five files, and the conflicts are mechanical: `probpipe/__init__.py` (#383 inserts `"TermSpec"` directly above `"Tracked"` in `__all__`) and `CHANGELOG.md` (both add `[Unreleased]` entries). Whichever merges second rebases and resolves those two by hand.

## Verification

`4050 passed`, 53 skipped in the full suite. The 5 failures are the subprocess process-stability tests that fail identically on pristine `origin/main`, so they are unrelated to this change. `ruff check` and `ruff format` clean. `probpipe.TrackedTerm` imports, the old attribute is gone, and `issubclass(Normal, TrackedTerm)` holds.


SoymilkWinsAgain (approved):
Looks great! No stuck

PR #386, opened 2026-07-27
feat(core): add the Batch[E] ABC with named axis levels

jhuggins:
Adds `Batch[E]`, the tracked nd-collection ABC the concrete batch types will specialize, and `BatchSpec`, the term spec that types a batch. The contract is written up in `design/02-shared-abstractions.md` §II.5, which this PR also corrects; the notes below cover only what is worth a reviewer's attention.

```python
class Batch[E](TrackedTerm, ABC):
    spec: BatchSpec              # the batch's own type, at the family kind
    element_spec: ValueSpec      # view on spec: what every element satisfies
    batch_shape: tuple[int, ...] # flat; axis_groups tiles it into levels
    axis_groups: tuple[tuple[int, ...], ...]
    level_names: tuple[str, ...]
    def with_level_names(self, mapping=None, /, **kwargs) -> Self
    def __getitem__(self, key) -> Any          # a position indexes axes, a name indexes fields
    def at_levels(self, /, **levels) -> E | Self
```

`FunctionBatch` / `OpaqueBatch` (#394), `RecordBatch`, and `DistributionBatch` follow separately, as does converting the existing `*Array` types. Nothing is reparented here, so this is purely additive.

## Decisions worth reviewing

**A batch's type is its own.** `BatchSpec` carries the element spec together with the named multiplicity, so `spec` names the *collection* exactly as any other term's spec names the term, and the level accessors are views on it. This is what makes a batch of opaque values well typed — an `OpaqueSpec` names no kind, and the batch is specified all the same.

**Levels, and what happens on a clash.** `axis_groups` tiles `batch_shape` into contiguous named groups, so `N` laws of `S` draws each read as `(N,)` of `(S,)` rather than one anonymous `(N, S)`. Names are unique, and a duplicate **raises** rather than being resolved: an earlier revision of this branch appended an integer suffix, which is removed here, because `draw2` records the order levels were added in — a fact about the computation path, not about meaning, and levels exist so operands align by meaning.

**A view is named by what it selects,** naming the level each selection addresses: `posterior.at_levels(chain=0, draw=7).name` is `"posterior[chain=0, draw=7]"`. Levels selected whole are left out and the rest are listed in the batch's own order, so the name is a function of the selection: two routes to one selection read alike, and two selections never do. This replaces a positional `name[i]`, which cannot say which axis `i` addressed.

**Selections carry positions, not slices.** A slice states a descending stop as `-1`, a bound meaningful only once, so anything resolving it twice read that as the last position — `flat[::-1]` had `batch_shape == (0,)` and handed storage an equally empty slice, spec and store agreeing on the wrong answer with nothing raised. A `range` carries the positions and their order, and becomes a slice only at the storage seam.

**Storage is the only thing left to a subclass**, through `_element_at` and `_sub_batch_at`, plus `_at_fields` where the elements have fields. `_sub_batch_at` is contracted to return a **view** sharing the store, which is why selecting all of a batch needs no special case. An element is handed back as it was stored, under its own name and lineage, where the batch stores rather than materializes — the one place storage affects identity.

## Tests

`tests/core/test_batch.py` — **232 tests** over eight doubles, since the ABC has no storage of its own: elements that are tracked terms, bare values, batches, or answer a named key; one that stores no elements at all; one that returns the caller's own element; and two that declare their slots unusually.

Beyond the level algebra and every construction-time invariant, three things get dedicated coverage because they are claims rather than values:

- **Naming**, since a derived name asserts identity: that it identifies the object and not the route, across chains mixing reversal, stepping and forward selection, and that a sliced view's *elements* are the positions its name claims. A mutation check found the earlier suite could not see this — it had no negative-step case, and 102 of its 103 tests stayed green when sliced views were handed the leading-k elements instead.
- **The storage contract from both sides**: a copying double and a no-storage double run through the same selections and asserted to agree on names, shapes, levels, and elements, so the algebra is shown implementable as a view rather than only as a copy.
- **State round-trips per slot form**, after two silent-loss modes: a bare-string `__slots__` iterates into characters, and a subclass declaring no `__slots__` keeps attributes in a dict — both dropped storage through `copy`, `pickle`, `with_name`, and `with_level_names`.

Verified by mutation: 18 targeted breaks of the naming, dispatch, indexer-validation, error-placement, storage-order, slot-walk and `repr` paths were each caught. **2205 pass** in `tests/core`; `ruff check` and `ruff format` clean.

## Review history

@SoymilkWinsAgain found four defects, all fixed: the hand-rolled slot walk in `__getstate__` and `_shallow_copy` (silent state loss, now delegated to `object.__getstate__`), a regression double too weak to catch it, `with_level_names` reporting a non-string name as empty, and four incomplete NumPy docstring sections. A review sweep over the whole diff added three more: a non-integer slice bound, a bare-string `level_names`, and a flat `batch_shape` passed as `axis_groups` all escaped the module's located error messages.

Known follow-up: `_with_level_names` does not decouple the `_annotations` container as `with_name` does. Latent — `Batch` does not mix in `Annotated` — and the right fix belongs in `_shallow_copy`, which #395 covers.

Completes the `Batch[E]` item in #350's Wave 1.


jhuggins:
Note on the missing `spec`, now that the design question behind it is settled.

This ABC stores `_axis_groups` and `_level_names` directly and carries no `spec` at all. That was deliberate but unresolved: `TrackedTerm` gains its `spec` slot separately (#385 defers it), and until today the design reference used a batch's `spec` for its *element's* type, which could not be implemented — `OpaqueBatch.spec -> OpaqueSpec` is not a `TermSpec`, and `RecordBatch` declared no spec at all.

#382 resolves it with **`BatchSpec`**: a term spec in its own right, carrying the element spec together with the named multiplicity. So a batch's `spec` names the batch's own type, exactly as on any other term, and `element_spec` is a view on it.

For this PR that changes nothing — the level algebra, the invariants, and the two storage hooks are the same either way. It does fix the shape of the follow-up: `axis_groups` and `level_names` become views on a stored `BatchSpec` rather than slots, `_init_batch` builds that spec (or accepts one) instead of validating loose tuples, and `with_level_names` returns a copy carrying a renamed spec. That is one commit against `_batch.py` once the `spec` slot exists on `TrackedTerm`, and it needs `element_spec`, so it belongs with the first concrete batch type (W3.1) rather than here.

Reviewing against the current `_batch.py` is therefore still the right thing; the absent `spec` is not an oversight to flag.

jhuggins:
Superseding the comment above: `BatchSpec` is now **in** this PR (838bfdb7), not deferred to W3.1.

The argument for holding it back was that the refactor needs `element_spec`, which only a concrete batch can supply. That is true of *building* a spec and irrelevant to *storing* one — a subclass passes the spec in, and `axis_groups` / `level_names` are precisely the multiplicity half of it. Keeping them as slots here and converting them to views later would churn the one class whose entire job is that data, so it belongs in the same PR as the level algebra.

The cost is that `BatchSpec` subclasses `TermSpec`, so #383 is merged into this branch and the stack is three deep until #383 and #385 land. The PR body records that.

probpipe/core/_batch.py
```diff
@@ -0,0 +1,989 @@
...
+    def __getstate__(self) -> tuple[None, dict[str, Any]]:
+        """Every assigned slot, for ``pickle`` and ``copy``.
+
+        A batch has no instance dictionary, so the state is the slots declared
+        across the class hierarchy — a subclass's storage included, without it
+        having to say so.
+        """
+        state: dict[str, Any] = {}
+        for klass in type(self).__mro__:
+            for slot in getattr(klass, "__slots__", ()):
```
SoymilkWinsAgain: Python allows `__slots__ = "_store"`; however, this loop returns a single character instead of the slot name. In practice, `_store` is lost after both `copy.copy()` and `pickle` round trips, without any exceptions.

It is recommended to use `object.__getstate__()`, or to normalize the string slot name and properly handle name rewriting. Regression tests for copy/pickle with the added string form of `__slots__` are needed.

probpipe/core/_batch.py
```diff
@@ -0,0 +1,989 @@
...
+                        if len(group) == 1
+                        else f"axis {offset} of level {level_name!r}, axes {group}"
+                    )
+            start += len(group)
+        return self._at_axes(tuple(axis_index), where=tuple(where))
+
+    # -- the concrete-storage seam ------------------------------------------
+
+    @abstractmethod
+    def _element_at(self, index: tuple[int, ...], *, name: str) -> E:
```
SoymilkWinsAgain, line 693: This requirement calls for a stored implementation that directly returns an existing element without modifying its provenance. Currently, test surrogates either create a new object or copy the object first, thus failing to prevent the base class from re-attaching provenance to the borrowed object in the future.

It is recommended to add a batch that saves the `_Leaf` instance provided by the caller and returns that instance as is, asserting that the selection operation preserves the object's identity and does not change its provenance.

SoymilkWinsAgain (commented):
Just some minor bugs can be fixed and then merged. Note that merging this PR will close #340 incorrectly.

AI indicates that the newly added public API documentation does not fully comply with the project's NumPy docstring specifications.

1. `BatchSpec` is missing Parameters;
2. `with_level_names` is missing Parameters and Returns;
3. `with_name` is missing a complete contract;
4. `batch[::0]` and `at_levels` with zero steps actually throw ValueError, but the corresponding Raises are not documented.

This should be quickly fixed using AI.

probpipe/core/_batch.py
```diff
@@ -0,0 +1,1064 @@
...
+        if len(set(root_names)) != len(root_names):
+            taken = sorted({name for name in root_names if root_names.count(name) > 1})
+            raise ValueError(
+                f"level name {taken[0]!r} is already used by a level this view derives its "
+                f"name from but no longer carries; renaming onto it would make the derived "
+                f"name ambiguous. Rename it on the batch this view came from, or give the "
+                f"level another name"
+            )
+
+        renamed = self._shallow_copy()
```
SoymilkWinsAgain, line 921: `TrackedTerm._shallow_copy()` traverses

```python
for slot in getattr(klass, "__slots__", ()):
```

When a valid subclass declares __slots__ = "_store", the string is being iterated over, causing _store to not be copied.

`_StringSlotsBatch` did not capture this issue because it inherits from `_ListBatch`, which already declares a `_store` with the same name as a tuple; and the relevant tests only cover copy/pickle, not with_name() or with_level_names().

probpipe/core/_batch.py
```diff
@@ -0,0 +1,1064 @@
...
+                f"by a keyword; give it one new name"
+            )
+        renames: dict[str, str] = {**positional, **kwargs}
+        unknown = set(renames) - set(self.level_names)
+        if unknown:
+            raise KeyError(
+                f"not levels of this batch: {sorted(unknown)}; have {list(self.level_names)}"
+            )
+        for new in renames.values():
+            if not new:
```
SoymilkWinsAgain, line 477: It is recommended to check `isinstance(new, str)` first, otherwise `None`, `0`, and `[]` will all throw a `ValueError`.

SoymilkWinsAgain (commented):
Two small bugs. It should be able to be merged after the repair.

Also, PR body needs cleaning

SoymilkWinsAgain (approved):

SoymilkWinsAgain:
Actually, there were two docstrings that needed to be modified. I will change them together in the next PR.

Issue #389, opened 2026-07-28
Make lifting RNG workflow-owned and structurally reproducible


SoymilkWinsAgain:
## Summary

Move all ProbPipe-owned automatic randomness into workflow execution contexts. Distinct stochastic call occurrences receive fresh random realizations, seeded workflow structure is reproducible, managed concurrency is independent of worker scheduling, and repeated references to the same random quantity remain joint within one lifted invocation.

The design preserves the immutable, first-class `Function` model established by #380 while addressing the repeated-realization behavior in #384, the jointness requirements in #388, and the process-global state used by omitted-key operations such as `sample(dist)`.

The public model is:

> A `Function` is a mathematical map. Random realizations belong to executions of that map. Reproducible sequences belong to workflow contexts.

## Goals

- Give distinct lifted stochastic occurrences fresh Monte Carlo realizations.
- Reproduce the same stochastic structure from the same workflow seed.
- Preserve immutable, reentrant `Function` objects with no advancing RNG state.
- Use one authority for every key ProbPipe synthesizes because a supported operation omitted an explicit key.
- Preserve the joint law of repeated aliases, views, and approved deterministic descendants within one invocation.
- Make managed sequential, threaded, JAX, and orchestration execution derive the same logical random-event mapping independently of scheduling and retry.
- Record a compact, versioned random-stream recipe for audit and supported artifact replay.
- Reject incompatible replay before stochastic commit whenever compatibility is knowable during preflight.
- Keep the public randomness API small and unambiguous.

### Bare stochastic calls are fresh

```python
@function
def twice(x):
    return 2 * x

a = twice(dist)
b = twice(dist)
```

Each root bare call receives an independent ephemeral workflow context. Its RNG root is generated lazily from OS entropy when the call first commits a workflow-owned random event. `a` and `b` therefore use different Monte Carlo realizations.

The generated root is retained in recipe-capable provenance, so a supported artifact can reconstruct its historical random-event mapping:

```python
with replay_run(a.provenance):
    reproduced_a = twice(dist)
```

### Seeded workflows reproduce stochastic structure

```python
with workflow_run(seed=114514):
    results = [twice(dist) for _ in range(100)]
```

The 100 calls are distinct stochastic occurrences. Repeating the block with the same seed and the same committed stochastic structure reproduces occurrence `i` as occurrence `i`.

[...]

Issue #384, opened 2026-07-26
Repeated lifted `Function` calls reuse the same sample batch after PR #380


SoymilkWinsAgain:
## Summary

After #380, separate lifted calls to the same `Function` deterministically reuse the same finite sample batch when they use the same construction seed.

This means the two results below represent the same pushforward law but also contain exactly the same Monte Carlo realization. Repeating the call in a loop likewise produces clones of one batch rather than separate realizations.

## Reproduction

```python
import numpy as np

from probpipe import Normal, function


@function(seed=114514, n_broadcast_samples=16, dispatch="sequential")
def twice(x):
    return 2 * x


dist = Normal(0.0, 1.0, name="x")
a = twice(dist)
b = twice(dist)

print(np.array_equal(np.asarray(a.samples), np.asarray(b.samples)))
# True
```

Before #380, two such calls normally produced different sample batches because each call advanced the `Function`'s random key.

## Regression

This behavior was introduced by #380.

Before that PR, `Function` stored a mutable `self._key`, and `_get_key()` split it while advancing the stored key. #380 removed `_key`, stored only `self._seed`, and changed each invocation to initialize a new call-local key from that unchanged seed.

## Root cause

Every invocation now starts from the same construction seed. When the lifted input structure and sampling order are the same, the call-local key is split along the same path and yields the same sampling subkeys. Consequently, the distributions are sampled identically across otherwise separate calls.

PR #392, opened 2026-08-01
fix(core): co-sample aliased lifted arguments; lift record-valued laws

### LINKED ISSUE #388: Repeated plain `Distribution` arguments are sampled independently within one lifted call
## Summary

Passing the same plain `Distribution` object to more than one argument of a lifted `Function` incorrectly samples it independently for each argument.

Within one invocation, repeated references to one distribution denote aliases of the same random variable. The current behavior instead treats them like independent, identically distributed variables.

## Steps to reproduce

```python
import numpy as np

from probpipe import Function, Normal


dist = Normal(loc=0.0, scale=1.0, name="x")

difference = Function(
    func=lambda a, b: a - b,
    dispatch="sequential",
    n_broadcast_samples=8,
    seed=0,
    include_inputs=True,
)

result = difference(dist, dist)
a = np.asarray(result.input_samples["a"])
b = np.asarray(result.input_samples["b"])
out = np.asarray(result.output.samples)

print(np.array_equal(a, b))
print(np.array_equal(out, np.zeros(8)))
print(out)
```

## Expected behavior

The distribution should be sampled once for this invocation and the same realization batch should be supplied to both arguments:

```text
True
True
[0. 0. 0. 0. 0. 0. 0. 0.]
```

Equivalently, `difference(dist, dist)` should approximate the diagonal pushforward `X - X`, not `X1 - X2`. Two distinct distribution objects may still denote independent variables; the aliasing signal here is that both arguments reference the same object.

This is also the behavior originally described in #13: assigning the same distribution to multiple arguments should let the backend preserve their joint use.

## Actual behavior

On commit `8bf5f252a8ebe4c8b7e2cf19a626d516963f1bc3`:

```text
False
False
[-1.8929423  -0.41540334 -2.0018601  -0.43072104  0.1886251
 -0.06000143  0.3593523   0.3221799 ]
```

The output exactly matches `a - b`, where `a` and `b` are two independently sampled batches.

## ProbPipe version

`0.1.0`, source commit `8bf5f252a8ebe4c8b7e2cf19a626d516963f1bc3`

## Python version

`Python 3.12.11`

## JAX version

`0.8.3`

## Additional context

The sampling planner groups the two references by the identity of their source object, but `_sample_broadcast_args` only reconnects groups whose first member is a `_RecordDistributionView`. For a plain distribution it iterates over every argument reference, splits a new subkey, and samples the distribution again. The implementation comment explicitly describes plain non-view distributions as sampled independently per kwarg even when the same object is passed under multiple names.

This behavior predates PR #380 and should not be classified as a #380 regression. The first parent of the #380 merge already contains the independent-per-argument branch.

jhuggins:
## Summary

Within one lifted call, two references to the same law denote **one** random variable. Passing the same `Distribution` to two arguments sampled it twice, so `f(d, d)` approximated `f(X1, X2)` rather than `f(X, X)` — a silently wrong answer, not a convenience issue.

```python
difference = Function(func=lambda a, b: a - b, n_broadcast_samples=8, seed=0)
difference(dist, dist).samples
# before: [-1.89, -0.42, -2.00, ...]      after: [0. 0. 0. ...]
```

This is the contract in `design/04-functions.md` IV.2, which names the case: "The lifted arguments are grouped by **root ancestor**, transitively: sibling views of one parent, **the same distribution passed twice**, and a parent passed alongside its own view all fall in one group. Each group contributes **one joint draw** per repetition."

**Six commits.** The first shares the draw within a group, which is #388 and two further cases of the same defect. The rest fix the places broadcast assembly assumed every argument's samples were an array — without which the last of those cases would be fixed at the co-sampling layer and still unrunnable above it. That assumption is older and wider than aliasing, and each commit is separable — see below.

Fixes #388.

## Root cause — the grouping was computed and then discarded

`group_by_parent` already put aliases in one group (both references resolve to `id(dist)`). `_sample_broadcast_args` then branched on whether the group's **first** member was a field view: if it was not, it looped the members and split a fresh subkey for each; if it was, it projected *every* member as a view. One decision, taken from one member, applied to all.

That single defect produced three distinct symptoms, all confirmed on `main`:

| design IV.2 case | before | after |
|---|---|---|
| same plain distribution twice | independent draws | one draw |
| same view twice | correct | correct |
| sibling views of one parent | correct | correct |
| parent alongside its own view | `AttributeError: 'JointGaussian' object has no attribute 'field'` | one draw, parent whole and view projected — and now runs end to end |
| same empirical twice | 3×3 grid, weights squared | 3-point diagonal, weights once |
| unrelated laws | correct (product law) | unchanged, down to the subkeys |

The empirical case is a second site with the same cause: `_split_empirical_args` gave each *reference* its own enumeration axis, so an alias multiplied the grid instead of sharing an atom.

## The change

Each group is drawn **once**, from its root, and every member takes its own value out of that draw — the root whole, a view through its field path. That one shape of code replaces the branch, so all three symptoms follow from the contract rather than from three separate patches.

- `group_by_parent` → **`group_by_root`**, returning the root object alongside its references. The design's own vocabulary, and more accurate: a plain distribution is its own **root**, not its own parent.
- `_split_empirical_args` returns co-sampling **groups**. A group is enumerable only when its root is a small-enough empirical *and* every member is that root; a group holding a view goes to the sampling path whole, which prevents a parent being enumerated while its view is sampled from a different draw.
- `_broadcast_enumerate` builds one product axis per group, and counts each atom's weight once.
- The `_sample_broadcast_args` docstring stated the bug as the contract ("Plain non-view distributions are sampled independently per kwarg, even if the same object is passed under multiple names") and now states the rule.

## Commits two and three: a record-valued law can be lifted

Sharing the draw was necessary but not sufficient for `f(d, d["x"])`: with the parent's draw in hand, assembly then refused it. `BroadcastDistribution` took its row count from the samples' `shape`, which a record batch refuses unless it holds exactly one leaf — and refuses with a `TypeError`, which `hasattr` propagates rather than swallowing, so the `len` fallback written beside it was never reached.

The count now comes from one helper that reads whichever accessor means *rows* for the container it is handed: `batch_shape` for a record batch, a leaf's leading axis for a plain record batched on its leaves — which is what `RecordEmpiricalDistribution._sample` returns, so any empirical with more atoms than the sample budget takes that branch — and `shape` or `len` otherwise. Deliberately **not** `len` throughout, which is shorter and covers arrays: on a `RecordArray` that is the *field* count, documented as such at `_record_array.py:271`, so it would have returned 2 where the answer was 7 — a silently wrong `num_atoms` instead of a raise. `batch_shape` is the one accessor that means the same thing for every batched value. (This call site is one more consumer that W3.2's "a batch is a collection, not a named tree" work will simplify.)

A second site had the same assumption: enumeration stacked each argument's per-row values with `jnp.stack`, which a `Record` row is not, so a record-valued *empirical* raised even though its atoms were right. Those rows now stack through `RecordArray.stack`, the record counterpart that already exists — after which the row count above reads the resulting `batch_shape`. Record-valued empiricals therefore enumerate, and co-sample when passed twice.

The defect is older and wider than aliasing — `f(j)` with a *single* record-valued argument fails identically on `main` — but the two fixes are ~10 lines, and without them this PR would ship the parent-and-view case as fixed at the co-sampling layer and unrunnable above it. Each is its own commit, so either can be reverted independently.

## The joint also has to resample

Lifting with `include_inputs=True` returns a joint over the inputs and the output, and drawing from it gathers the same rows from every component so a drawn tuple stays paired. That gather assumed an array too, so `sample(lifted(...), sample_shape=...)` raised `TypeError: key must be str, tuple, or int, got ArrayImpl` the moment an input arrived as a record batch — a line unchanged from `main`, made reachable by the record-valued lifting above. Reading `.samples` went on working throughout, since that goes through the output marginal, which is why the record-valued test added earlier did not catch it.

One gather now reads the container it is given: an array indexes directly, a list of per-row objects gathers positionally, and a record is rebuilt from its gathered leaves. The rebuild is deliberate rather than a `jax.tree.map` — a `RecordArray` stores its row count and a `Record` its event template, both in pytree aux data, so mapping the leaves alone yields a batch quietly claiming the rows it started with. Measured: a tree-map gather of three rows by four indices leaves `batch_shape` `(3,)` over leaves of shape `(4,)`, with nothing raised.

The same gather covers the output side, where a vectorized broadcast over a record-returning function leaves the output a batched `Record`. That failure predates the record-lifting work and reproduces on `origin/main`; it is fixed here because it is the same three lines.

**Two shapes are still unsupported**, both stated in the CHANGELOG rather than implied. A record with **nested** fields cannot be batched at all, since a record batch keys its store by top-level child rather than by leaf path — `RecordArray.stack` refuses the template outright. Lifting such a law now says which argument it was and that the shape is unrepresented rather than mistaken, instead of surfacing the container's own message; a strict `xfail` marks the working case so it reports the day record batches become leaf-keyed. That is #340, not this PR's to close.

The other: a record-valued empirical passed alongside a field view of itself. That group routes to sampling rather than enumeration, and `RecordEmpiricalDistribution._sample` returns a single `Record` rather than a batch of them, so there are no rows to index. Distribution-layer contract gap, not a broadcast one; tracked in #393.

Verified working with a record-valued argument: `auto` / `sequential` / `thread` dispatch, `include_inputs`, `marginalize`, `mean`, and record-valued outputs. Explicit `dispatch="jax"` reports its usual not-traceable error when the wrapped function indexes a record, which is the designed path, and `auto` falls back correctly.

## What cannot move

Arguments with no common root are unaffected **down to the subkeys**: a group of one consumes exactly one key split, exactly as the old per-member loop did. Verified by digesting the draws of an unrelated pair, a single-argument lift, and sibling views before and after — byte-identical in every case. So no currently-correct numeric result changes; only calls that were already wrong do.

`test_arguments_with_no_common_root_are_drawn_independently` pins this in the suite by reproducing the expected subkeys explicitly rather than only asserting that the two draws differ.

## Tests

`tests/core/test_workflow_distribution_broadcast.py`, **20 new test functions** in two classes (one parameterized over three sample budgets, so 22 cases), plus five in `tests/core/test_broadcast_distribution.py` covering the gather directly.

`TestCoSamplingGroups` drives `_sample_broadcast_args` directly, over the IV.2 cases: an alias drawn once, a parent and its own view sharing a draw, the same result under either argument order, sibling views projecting one parent draw while differing from each other, and the subkey-level independence above. **Four of the five fail against unfixed source**; the fifth is the stream-preservation test, which must pass before and after by construction.

`TestCoSamplingThroughACall` asserts the same contract as a caller sees it: `f(d, d)` returns zeros, `include_inputs` reports one realization under both names, unrelated laws still sample the product, an aliased empirical enumerates 3 points across three different `n_broadcast_samples` budgets — parameterized because the old behaviour degraded *differently* as the budget fell below the product size — and its weights are `1/3` each rather than squared. Five more cover the assembly commits: a record-valued law lifts to the right row count, `f(d, d["x"])` returns zeros end to end, a record-valued empirical enumerates and co-samples when passed twice, and grouping is object identity rather than structural equality — two separately constructed laws with equal parameters *and* equal names are two random variables, and only a shared object is one.

## Out of scope

- **#384 / #389 — freshness between calls.** Untouched. This PR changes only *within*-invocation behaviour, which IV.2 already settles. Whether a separate invocation is a distinct structural position is a design question the merged IV.3 text does not answer, and #389 proposes to answer it by removing `Function(seed=)`. Independent of this fix either way.
- **Transitive grouping.** IV.2 says grouping is transitive; `group_by_root` does a single `parent` lookup. Exact today, since a view's parent is always a distribution and `select()` returns a plain dict of depth-1 views. A comment marks the assumption for whenever a nested view type arrives.

## Verification

`1999 passed, 1 xfailed` in `tests/core` (excluding the subprocess process-stability tests, which fail identically on pristine `origin/main`) and `1061 passed`, 49 skipped across `tests/distributions`, `tests/record`, `tests/modeling`, and `tests/validation`. `ruff check` and `ruff format` clean at the pinned 0.15.16.

### DIFF
--- a/probpipe/core/_broadcast_distributions.py
+++ b/probpipe/core/_broadcast_distributions.py
@@ -886,6 +886,60 @@ def _make_stack(
 # ---------------------------------------------------------------------------
 
 
+def _record_rows(record: Record, rows: Any) -> Record:
+    """*record* with every leaf reduced to *rows*, its specification re-derived.
+
+    A record's pytree aux data carries its event template, whose shapes describe
+    the leaves it was built from, so ``jax.tree.map`` would rebuild a record
+    claiming the shapes it started with while holding gathered ones. Rebuilding by
+    path lets the template be re-derived from the leaves that survived.
+    """
+    paths = tuple(record.event_template.keys())
+    return type(record)(
+        record.name, {path: record[path][rows] for path in paths}, name_is_auto=True
+    )
+
+
+def _take_rows(component: Any, indices: Array) -> Any:
+    """Gather the rows *indices* of one batched component, keeping its container.
+
+    A component of a broadcast is batched along its leading axis, but what carries
+    that axis depends on what the component holds. An array is indexed directly. A
+    list holds one object per row and is gathered positionally. A record has fields
+    rather than a shape, so its leaves are gathered and the record rebuilt around
+    them — a ``RecordArray`` stating the row count it now holds, since that count
+    is stored rather than read off the leaves.
+    """
+    if isinstance(component, list):
+        return [component[int(i)] for i in indices]
+    if isinstance(component, RecordArray):
+        return type(component)(
+            {path: component[path][indices] for path in component.fields},
+            batch_shape=(indices.shape[0], *component.batch_shape[1:]),
+            template=component.template,
+        )
+    if isinstance(component, Record):
+        return _record_rows(component, indices)
+    return component[indices]
+
+
+def _one_row(component: Any) -> Any:
+    """One drawn row of a component, presented on its own.
+
+    ``sample_shape=()`` asks for a single draw rather than a batch of one, so the
+    row is unwrapped: a record of that row rather than a one-row batch of records,
+    and the object itself rather than a one-element list. Field names survive —
+    one draw of a record-valued component is a record, whatever its field count.
+    """
+    if isinstance(component, RecordArray):
+        return component[0]
+    if isinstance(component, Record):
+        return _record_rows(component, 0)
+    if isinstance(component, list):
+        return component[0]
+    return component[0] if hasattr(component, "__getitem__") else component
+
+
 class BroadcastDistribution(Distribution[dict], SupportsSampling):
     """Joint distribution over broadcast inputs and function output.
 
@@ -1014,25 +1068,22 @@ class BroadcastDistribution(Distribution[dict], SupportsSampling):
     # -- joint sampling -----------------------------------------------------
 
     def _sample(self, key, sample_shape=()):
-        """Resample paired input–output tuples."""
+        """Resample paired input–output tuples.
+
+        Every component is batched along the same leading axis, so one set of
+        drawn rows is gathered from each — the inputs and the output alike, which
+        is what keeps a drawn tuple paired.
+        """
         n_draws = prod(sample_shape) if sample_shape else 1
         indices = self._w.choice(key, shape=(n_draws,))
 
-        result = {}
-        for arg_name in self._broadcast_args:
-            arr = self._input_samples[arg_name]
-            result[arg_name] = arr[indices]
-
-        # Output
-        if isinstance(self._output_samples, jnp.ndarray):
-            result["_output"] = self._output_samples[indices]
-        elif isinstance(self._output_samples, list):
-            result["_output"] = [self._output_samples[int(i)] for i in indices]
-        else:
-            result["_output"] = self._output_samples[indices]
+        result = {
+            name: _take_rows(self._input_samples[name], indices) for name in self._broadcast_args
+        }
+        result["_output"] = _take_rows(self._output_samples, indices)
 
         if sample_shape == ():
-            return jax.tree.map(lambda x: x[0] if hasattr(x, "__getitem__") else x, result)
+            return {name: _one_row(component) for name, component in result.items()}
         return result
 
     # -- marginalization ----------------------------------------------------
--- a/probpipe/core/_broadcast_distributions.py
+++ b/probpipe/core/_broadcast_distributions.py
@@ -954,10 +954,19 @@ class BroadcastDistribution(Distribution[dict], SupportsSampling):
         self._output_distributions = output_distributions
         self._output_template = output_template
 
-        # Determine n from first broadcast arg
+        # The row count, taken from the first broadcast arg. A record batch
+        # answers neither of the obvious questions the way one would hope: its
+        # ``len`` is the field count, and its ``shape`` raises unless it holds a
+        # single leaf — and raises a ``TypeError``, which ``hasattr`` propagates
+        # rather than swallowing. ``batch_shape`` is the one accessor that means
+        # the same thing for every batched value.
         first_key = next(iter(broadcast_args))
         first_arr = input_samples[first_key]
-        n = first_arr.shape[0] if hasattr(first_arr, "shape") else len(first_arr)
+        batch_shape = getattr(first_arr, "batch_shape", None)
+        if batch_shape:
+            n = batch_shape[0]
+        else:
+            n = first_arr.shape[0] if hasattr(first_arr, "shape") else len(first_arr)
         self._w = Weights(n=n, weights=weights, log_weights=log_weights)
         self._broadcast_args = list(broadcast_args)
         name, name_is_auto = auto_name(name, "broadcast")
--- a/probpipe/core/_broadcast_distributions.py
+++ b/probpipe/core/_broadcast_distributions.py
@@ -913,8 +913,12 @@ def _take_rows(component: Any, indices: Array) -> Any:
     if isinstance(component, list):
         return [component[int(i)] for i in indices]
     if isinstance(component, RecordArray):
+        # Keyed by the template, which is leaf-keyed, rather than by ``fields``,
+        # which names the top-level children and is retained only for the
+        # migration. The two agree while a record batch is flat and will not once
+        # one can nest; ``_RecordMarginal`` peels a batch the same way.
         return type(component)(
-            {path: component[path][indices] for path in component.fields},
+            {path: component[path][indices] for path in component.template},
             batch_shape=(indices.shape[0], *component.batch_shape[1:]),
             template=component.template,
         )
@@ -968,10 +972,15 @@ class BroadcastDistribution(Distribution[dict], SupportsSampling):
 
     Parameters
     ----------
-    input_samples : dict[str, Array]
-        ``{arg_name: (n, *event_shape)}`` for each broadcast argument.
-    output_samples : Array or list
-        ``(n, *event_shape)`` for array outputs, or a list of length *n*.
+    input_samples : dict[str, Array or RecordArray or list]
+        ``{arg_name: rows}`` for each broadcast argument, every value batched
+        over the same leading axis of length ``n``: an array of shape
+        ``(n, *event_shape)``, a record batch of ``batch_shape == (n,)`` for a
+        record-valued argument, or a list of ``n`` objects.
+    output_samples : Array, Record, or list
+        The outputs, batched over the same axis: ``(n, *event_shape)`` for array
+        outputs, a record whose leaves carry that axis for record-returning
+        functions, or a list of length *n*.
     weights : array-like, :class:`~probpipe.Weights`, or None
         Non-negative weights (normalized internally).  A pre-built
         :class:`~probpipe.Weights` object is also accepted.  Mutually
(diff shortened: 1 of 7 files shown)


probpipe/core/_broadcast_distributions.py
```diff
@@ -886,6 +886,60 @@ def _make_stack(
...
+    them — a ``RecordArray`` stating the row count it now holds, since that count
+    is stored rather than read off the leaves.
+    """
+    if isinstance(component, list):
+        return [component[int(i)] for i in indices]
+    if isinstance(component, RecordArray):
+        return type(component)(
+            {path: component[path][indices] for path in component.fields},
+            batch_shape=(indices.shape[0], *component.batch_shape[1:]),
+            template=component.template,
```
SoymilkWinsAgain, line 943: This method iterates through `component.fields` and then directly executes `component[path]`. For nested structures, `fields` will contain subtree names like "group", while `RecordArray.__getitem__` only allows leaf fields.

Using a nested `ProductDistribution` lift and setting `include_inputs=True`, then `sample()` again, we get a stable result: KeyError: "'group' is a subtree, not a field; use at_path() to navigate to it"

probpipe/core/_workflow_distribution_broadcast.py
```diff
@@ -479,6 +507,22 @@ def _broadcast_sample(
...
+
+    ``jnp.stack`` covers array-valued rows. A record-valued row is not an array
+    — a ``Record`` has fields, not a shape — so those stack through
+    ``RecordArray.stack``, giving a batch whose ``batch_shape`` is ``(n,)`` and
+    whose fields are the per-row leaves.
+    """
+    from ._record_array import RecordArray
+    from .record import Record
+
+    if rows and isinstance(rows[0], Record):
```
SoymilkWinsAgain, line 526: Calling `RecordArray.stack(rows)` directly on a Record row does not explicitly support nested templates.

Passing a `RecordEmpiricalDistribution` containing nested fields such as `group/x` and `group/y` to a sequential function will result in a `TypeError`.

probpipe/core/_broadcast_distributions.py
SoymilkWinsAgain: Needs update here

arob5:
The fix lives in the shared `_sample_broadcast_args`, which `_broadcast_jax` calls as well, so the JAX path co-samples too — but every new test in `TestCoSamplingThroughACall` runs `dispatch="sequential"`. `test_same_view_twice_jax_backend` covers the view case under JAX, but the plain-alias case (`f(d, d)`) has no JAX-path test, and CONTRIBUTING calls out jax/sequential equivalence as high-value (path divergence has produced silently-wrong-results bugs before).

A one-line `dispatch="jax"` variant of `test_a_law_passed_twice_approximates_f_of_one_variable` would close the gap.

arob5:
Minor PR-body hygiene: the "commits two and three" section says "A second site had the same assumption, **found by review**." Per CONTRIBUTING, the description should read for an outside audience without review-process narration — suggest dropping "found by review". Everything else in the body reads as final-state and matches the diff.

arob5:
Nit: `_project_from_root` is annotated `-> Array`, but when it projects a nested field it can return a sub-`Record` rather than an array. The sibling helpers (`_take_rows`, `_one_row`) use `-> Any`; matching that here would be more honest. Cosmetic — no behavior implication.

jhuggins:
@arob5 All three addressed in 7452432f. Thanks — the first one was a real hole and the third is a genuine mis-annotation.

**JAX-path coverage.** Right on every count, and I checked the reasoning rather than taking it: `_sample_broadcast_args` is called from `_broadcast_jax` (line 347), `_broadcast_enumerate` (414), and `_broadcast_sample` (483), so the grouping is shared by all three, and **all 20** co-sampling tests drove `sequential` or the helper directly. Not one touched `dispatch="jax"`.

Rather than add a variant, I parametrized the two tests that state the contract a caller actually depends on — `f(d, d)` drawing one variable, and `include_inputs` reporting one realization under both names — over both dispatches. That way the equivalence is one assertion rather than a pair of tests that could drift apart.

Worth saying plainly: this is coverage over behavior that already worked, not a latent bug. I verified before writing the test:

```
dispatch=sequential  all-zero=True
dispatch=jax         all-zero=True
```

You also had `test_same_view_twice_jax_backend` placed correctly — it is in `tests/distributions/test_product.py:604`. I looked in the broadcast test file first and briefly thought it did not exist.

**`_project_from_root`.** Correct: annotated `-> Array`, but the fallback walk down `_key_path` yields a sub-`Record` when a view names a field group. Now `-> Any`, with a line saying why. The two siblings you named have since been renamed, but the surviving ones (`_stack_rows`, `_index_sample`) both use `Any`, so it was the odd one out.

**PR body.** Fixed, and the same phrasing was in two other open PRs of mine, so I swept those too rather than only the one you read — the descriptions now state the final situation without the discovery story.

`tests/core`: **2001 passed**, 1 xfailed. `ruff check` and `ruff format` clean.

PR #394, opened 2026-08-01
feat(core): add FunctionBatch and OpaqueBatch, the batch forms that store objects

### LINKED ISSUE #340: Nested empirical / nested-prior inference broken: NumericRecordArray not leaf-keyed

jhuggins:
## Summary

The batch forms of the two value kinds that have no native stacked form. A numeric array batches natively, with the batch axes leading, so it needs no class; a callable and an opaque object have nothing to stack *into*, so each gets a thin `Batch` that stores its elements and carries the one `element_spec` they all satisfy, adding no interface beyond it — design III.1.

```python
FunctionBatch([lambda x: x, lambda x: 2 * x], "variant", name="f")[1](3)   # 6
OpaqueBatch(["north", "south"], "site", name="labels")[0]                  # 'north'
```

> **Stacked on #386.** These are the first implementers of the `Batch[E]` ABC, so this branches off `dev/batch-abc`. Review #386 first; once it lands this rebases onto `main` and the diff reduces to its own commit.

Implements **W3.1** of #350.

## The ABC's storage contract, checked against real storage

Worth stating plainly because it is this PR's second purpose. `Batch` has so far been exercised only through the five doubles in `test_batch.py`; these are the first real implementers, so the value here is partly in finding out whether the hook shapes hold up. They did — no change to `_batch.py` was needed — and the storage contract is now asserted against something that actually stores.

`_sub_batch_at` is contracted to present a **view** sharing the parent's store, with the index presented *in the order the view presents them*, a descending slice included. A numpy object array answers both: basic indexing returns a view, and it honors a descending or stepped slice in place. That is why the store is an object array — the contract, not any interest in arrays.

Both classes share that storage through a private `_ObjectBatch`. The alternative was ~40 lines duplicated per class, and the lines in question are exactly the contract-bearing ones: a copy could creep into one while the other stayed a view, which is the failure the ABC's docstring warns about.

## Deliberate choices, each of which could reasonably go the other way

- **A level name is required, not defaulted.** `FunctionBatch(elements, "variant")`, never `FunctionBatch(elements)`. Same reason #386 removed suffixing: a placeholder like `element` reads as meaning something while naming nothing, and levels exist so that operations can align operands by meaning. The caller minting the level knows what it means.
- **An element comes back as the object that was stored, untouched.** Neither renamed nor given provenance. Identity and lineage are derived where a batch *materializes* an element per index — a row of columnar storage has no identity until it is built — and these store theirs, so `batch[0] is element` and what the caller put in is what comes out. That has a consequence for provenance too: the ABC used to stamp whatever `_element_at` returned, which for a storing batch reached into the caller's own object. #386 now makes a selection *inherit* its batch's lineage rather than record a node, and element provenance the hook's own; W3.2's materialized rows will inherit, these do not.
- **Elements are never unpacked.** `np.asarray([jnp.zeros(3), jnp.ones(3)], dtype=object)` stacks those into one 2-d array; allocating empty and assigning keeps each object whole. A batch of two arrays stays a batch of two things.
- **Every element is checked at construction**, naming the position that failed, because a batch asserts its `element_spec` of *all* of them — an element that fails makes the batch's own spec a false statement. `FunctionSpec.is_valid` is callability; `OpaqueSpec.is_valid` is anything but a mapping, which the value layer reads as a subtree. The check lives once on the base, each class supplying only the phrase its refusal reads with.
- **The store is frozen and a supplied array copied**, so the batch holds the elements it validated and a view cannot write through to its parent. Only the pointer array is copied; the elements stay shared.
- **`axis_groups` must tile the shape the elements are stored in**, so the spec cannot describe a shape the storage does not have.
- **A container that iterates into its parts is refused** — a string, a mapping, a numeric array — since each would give a batch of pieces of one object rather than a batch of objects. The mapping case would otherwise slip past the check that refuses a mapping *as* an element.
- **Shape comes from the elements.** An object array carries its own; a flat sequence is one axis. A nested sequence is *not* unpacked into axes, since what nesting means for an arbitrary object is the caller's to decide — build the array to state more than one axis. `axis_groups` defaults to one axis per level and is given explicitly for a level spanning several.

## Deviation from the target layout

`design/package-structure.md` places these in `values/_function_batch.py` and `values/_opaque_batch.py`. The `values/` split has not happened, so they sit in `core/` under the same filenames, as the other in-place migrations do. The shared `_object_batch.py` is a third module the layout did not name; it holds the private base only, and the two public classes keep the files the design names. The layout now lists it, since `LinOpBatch` is a likely second consumer.

## `OpaqueBatch` closes a loop

It is the case a batch's own spec was introduced for. An `OpaqueSpec` names no ProbPipe kind, so under the old reading — a batch's spec being its *element's* — `OpaqueBatch.spec` would not have been a `TermSpec` at all. With `BatchSpec` it is specified like any other term, at the family kind over its element's. That argument was made in #382 and is now a class.

## Tests

`tests/core/test_function_opaque_batch.py` — **77 tests**. Construction from a sequence and from an object array, multi-level and multi-axis-level, auto vs. given names; every construction refusal (non-callable element, mapping element, no elements, a single object, a numeric ndarray, too few level names, a wrong-kind `element_spec`), each asserting the position or the count it reports; the spec at the family kind and `element_spec` as a view of the declared kind; elements returned as stored, including a tracked one keeping its own name and elements holding arrays or lists staying whole; the storage contract above; derived names, level renaming, and `with_name` re-rooting; the field-key error a batch with no fields gives; `pickle` / `copy` of a batch and of a view, including that `_store` travels without being declared; immutability; and that neither class adds public interface beyond `element_spec`.

Verified by mutation, **13 breaks each caught**. Five were chosen by me: a view turned into a copy, a descending index re-sorted, the element check dropped, `asarray` used for the store, and an element renamed to its position. The other eight close gaps the first suite left — a *partial* view regression that copied only when the index contained an integer, `is_valid` blind to `level_names`, `OpaqueBatch.element_spec` not pinned as a view, `name_is_auto` and `provenance` never read, generators rejected, the non-iterable message gutted, and `FunctionBatch`'s multi-axis position branch, which had no 2-d `FunctionBatch` to exercise it. The sharing assertions had all used pure slices and are now parametrized over every indexing form.

**2262 pass** in `tests/core`; **1061 pass**, 49 skipped across `tests/distributions`, `tests/record`, `tests/modeling`, `tests/validation`. `ruff check` and `ruff format` clean at the pinned 0.15.16.

## Follows separately

`RecordBatch` / `NumericRecordBatch` (W3.2, which fixes #340) and `DistributionBatch` (W3.3, which also needs W2.1), plus converting the existing `RecordArray` / `NumericRecordArray` / `DistributionArray`. Nothing is reparented here, so this PR is purely additive.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,30 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **`FunctionBatch` and `OpaqueBatch` — the batch forms that store objects.** A
+  numeric array batches natively, with the batch axes leading, so it needs no
+  class; a callable and an opaque object have no such form, so each gets a thin
+  `Batch` that stores its elements and carries the one `element_spec` they all
+  satisfy, adding no interface beyond it. Elements go in as a flat sequence or
+  as an object array of any shape, with a name required for every level — a
+  placeholder would read as meaning something while naming nothing, the same
+  reason a level clash is not resolved by suffixing.
+
+  Storage is a numpy object array, chosen for the contract rather than for
+  arrays: numpy basic indexing returns a **view**, so a sub-batch shares its
+  parent's store, and it honors a descending or stepped slice in the order
+  given, which is the order a view's derived names are stated in. Elements are
+  never unpacked, so a batch of arrays or of lists stays a batch of two things
+  rather than becoming one 2-d array. An element comes back as the object that
+  was stored: identity is derived where a batch materializes an element per
+  index, and these store theirs.
+
+  `OpaqueBatch` is the case a batch's own spec exists for — an `OpaqueSpec`
+  names no ProbPipe kind, yet the batch is specified all the same, at the family
+  kind over it. Every element is checked against the shared spec at construction,
+  reporting the position that failed, since a batch asserts that spec of *all*
+  of them.
+
 - **`TermSpec` — the term-spec sub-hierarchy, and declarations stored as specs
   (#381).** `ValueSpec` now splits into *raw-value specs* (`ArraySpec`,
   `OpaqueSpec`), which name no ProbPipe kind, and *term specs*, one per kind,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,18 +35,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
   Storage is a numpy object array, chosen for the contract rather than for
   arrays: numpy basic indexing returns a **view**, so a sub-batch shares its
-  parent's store, and it honors a descending or stepped slice in the order
-  given, which is the order a view's derived names are stated in. Elements are
-  never unpacked, so a batch of arrays or of lists stays a batch of two things
-  rather than becoming one 2-d array. An element comes back as the object that
-  was stored: identity is derived where a batch materializes an element per
-  index, and these store theirs.
+  parent's store in every indexing form, and it honors a descending or stepped
+  slice in the order given, which is the order a view's derived names are stated
+  in. The store is frozen and a supplied array is copied — only the pointer
+  array, so the elements stay shared — so a batch holds the elements it
+  validated and a view cannot write through to its parent. Elements are never
+  unpacked: a batch of arrays or of lists stays a batch of two things rather than
+  becoming one 2-d array, and a container that iterates into its *parts* rather
+  than into elements (a string, a mapping, a numeric array) is refused, since
+  each would otherwise yield a batch of pieces of one object.
+
+  An element comes back as the object that was stored, untouched — neither
+  renamed nor given provenance. Identity and lineage are derived where a batch
+  *materializes* an element per index; these store theirs, so what the caller put
+  in is what comes out.
 
   `OpaqueBatch` is the case a batch's own spec exists for — an `OpaqueSpec`
   names no ProbPipe kind, yet the batch is specified all the same, at the family
   kind over it. Every element is checked against the shared spec at construction,
   reporting the position that failed, since a batch asserts that spec of *all*
-  of them.
+  of them, and `axis_groups` must tile the shape the elements are stored in, so
+  the spec cannot describe a shape the storage does not have.
 
 - **`TermSpec` — the term-spec sub-hierarchy, and declarations stored as specs
   (#381).** `ValueSpec` now splits into *raw-value specs* (`ArraySpec`,
--- a/CONTRACTS.md
+++ b/CONTRACTS.md
@@ -97,6 +97,7 @@ if it were user-guide reference text.
 | `Record` / `NumericRecord` | docstrings in `probpipe/core/record.py`, `_numeric_record.py`; #235 Chapter 2 |
 | `Batch` / `BatchSpec` (the multiplicity axis: levels, level names, view identity) | docstrings in `probpipe/core/_batch.py`; #235 Chapter 2 |
 | Batch types (`RecordArray`/`NumericRecordArray`/`DistributionArray` → `*Batch`) | docstrings in `_record_array.py`, `_distribution_array.py`; #235 Chapter 2 |
+| `FunctionBatch` / `OpaqueBatch` (the batch forms that *store* their elements, over shared object-array storage) | docstrings in `probpipe/core/_function_batch.py`, `_opaque_batch.py` (storage in `_object_batch.py`); #235 Chapter 2 |
 | `Function` & ops (`sample`, `log_prob`, …) | docstrings in `core/node.py`, `core/ops.py`, `_workflow_result.py`; #235 Chapter 3 |
 | Naming / provenance / annotations (`TrackedTerm` / `Annotated` mixins) | docstrings in `probpipe/core/tracked.py` (and `provenance.py` for `Provenance` / `ParentInfo`); the naming contract in #235 Chapter 5 |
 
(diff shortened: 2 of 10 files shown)


arob5:
**Coverage gap: the object-ndarray input path with array-valued elements**

`test_elements_holding_arrays_are_not_unpacked` exercises only the *flat-sequence* input path (`_from_iterable`, allocate-empty-and-assign). The *object-ndarray* input path is a distinct branch — `np.array(elements, dtype=object, subok=False)` in `_object_batch._as_object_array` — and it's the branch most at risk of numpy silently re-stacking array-valued elements into one N-d numeric array.

The code is correct here (I confirmed `np.array(src, dtype=object, subok=False)` preserves shape and does not re-descend into element arrays), but that path isn't directly guarded by a test, so a future refactor of `_as_object_array` could regress it without a failure. Suggest a one-liner alongside the existing case:

```python
def test_an_object_array_of_arrays_is_not_unpacked(self):
    store = np.empty(2, dtype=object)
    store[0], store[1] = jnp.zeros(3), jnp.ones(3)
    batch = OpaqueBatch(store, "site")

    assert batch.batch_shape == (2,)
    np.testing.assert_array_equal(np.asarray(batch[1]), np.ones(3))
```

Minor / nice-to-have.

jhuggins:
@arob5 Added in 2c8c6a58, essentially as you wrote it — and confirming your correctness read independently: `np.array(src, dtype=object, subok=False)` keeps shape `(2,)` with the elements intact.

You are right that the two input branches are distinct and only one was covered. Every existing object-array fixture in that file fills with strings, lambdas, or dicts, so nothing put an array-valued element through the supplied-ndarray path.

One detail worth recording, since it changes *which* regression the test catches. The first mutation I reached for — `np.asarray(elements)` — does **not** re-stack, because `asarray` on an array that is already `dtype=object` is a no-op; it killed only `test_the_array_a_caller_passes_is_copied`. What does re-stack is normalizing through a list:

```
np.array(src, dtype=object, subok=False)   ->  shape (2,)     dtype=object    (current)
np.asarray(src)                            ->  shape (2,)     dtype=object
np.array(src.tolist(), dtype=object)       ->  shape (2, 3)   dtype=object    <- re-stacked
np.stack(list(src))                        ->  shape (2, 3)   dtype=float32
```

Under the `.tolist()` mutation the new test is the **only** one that fails, which is the outcome you were after. And that is not a contrived refactor: routing through a list is exactly how `_from_iterable` builds the store for the other branch, so unifying the two is the obvious simplification someone would attempt.

`tests/core`: **2272 passed**. `ruff check` and `ruff format` clean.

PR #396, opened 2026-08-02
fix(core): report and bind symbolic dimensions wherever they are declared

### LINKED ISSUE #390: EventTemplate.is_concrete reports True when the symbolic dimensions are inside a term spec
## Summary

`EventTemplate.is_concrete` reports `True` for templates that are polymorphic, whenever the symbolic dimensions sit inside a term spec. `free_dims` recurses into nested `EventTemplate` nodes and reads `ArraySpec.shape`, but passes a `RecordSpec` / `DistributionSpec` / `FunctionSpec` leaf through untouched, so the inner schema's unbound dimensions are never reported.

```python
sym = EventTemplate(x=ArraySpec(shape=("obs",)))

EventTemplate(x=ArraySpec(shape=("obs",))).free_dims   # {'obs'}  is_concrete=False  ✔
EventTemplate(law=DistributionSpec(sym)).free_dims     # set()    is_concrete=True   ✘
EventTemplate(r=RecordSpec(sym)).free_dims             # set()    is_concrete=True   ✘
EventTemplate(f=FunctionSpec(sym, sym)).free_dims      # set()    is_concrete=True   ✘
```

The specs themselves answer neither question: `DistributionSpec(sym)` has no `free_dims` and no `is_concrete`, so there is nowhere to ask.

## Why this surfaces now

The blind spot predates #383, but that PR makes it reachable in a new way. Declarations are now *stored as specs* — `DistributionSpec.event_spec` is a `RecordSpec`, `FunctionSpec.output_spec` is any `ValueSpec` — so the symbolic dimensions of a declared inner schema are exactly what a caller has reason to ask about. Before, a declaration was a bare `EventTemplate` in a position `free_dims` did not visit either; now it is a spec that structurally *could* be visited.

The batch work in #386 runs into the same wall from the other side: `BatchSpec` carries an `element_spec`, and "is this batch's element schema concrete?" has no accessor to answer it.

## Cause

`free_dims` (`probpipe/core/event_template.py:856`) and `_replace_template_dimensions` (`:1339`) are **both** shallow, and consistently so — the latter's `else: children[name] = spec` branch passes a term spec through exactly as the former skips it. So today the two agree: a dimension that is not reported is also not substitutable. That consistency is why nothing is visibly broken, and it is also why the fix cannot be half-applied.

## Proposal

Lift `free_dims` onto `ValueSpec` and derive `is_concrete` from it once:

| Spec | `free_dims` |
|---|---|
| `ArraySpec` | the symbolic entries of `shape` |
| `OpaqueSpec` | empty |
| `RecordSpec` | `event_template.free_dims` |
| `DistributionSpec` | `event_spec.free_dims` |
| `FunctionSpec` | `input_template.free_dims` ∪ `output_spec.free_dims` |
| `BatchSpec` (#386) | `element_spec.free_dims` |

`EventTemplate.free_dims` then collapses to a union over its children's `free_dims`, and `is_concrete = not free_dims` is stated once rather than per class. Each spec reports the dimensions of what it declares, which is the rule the hierarchy already follows for `is_valid`.

**Substitution must go deep in the same change.** `_replace_template_dimensions` needs to rebuild a term spec with its inner template substituted, or `_bind_template_dimensions` will report a successful binding while leaving symbolic dimensions in place. Reporting a dimension that cannot be bound is worse than not reporting it.

## Blast radius

- `probpipe/core/record.py:433` — `Record` construction unifies when `event_template.free_dims` is non-empty. A template with a symbolic-inner term spec would newly enter unification, so `_unify_event_template_with_value` needs to either bind through the term spec or leave those dimensions to the caller with a clear error.
- `probpipe/core/_function_contract.py:169-170` — per-invocation input/output unification would see the additional dimensions.
- `NumericEventTemplate.vector_size` is **not** affected: a term-spec leaf makes a template non-numeric, so the numeric path never sees these dimensions.

## Out of scope

Whether a batch's *multiplicity* may be symbolic. `ArraySpec.shape` admits `("obs",)` while `BatchSpec.axis_groups` is integer-only, so a batch spec is strictly less expressive than an array spec on the same axis-size question — relevant to declaring "returns a batch of `S` draws" before `S` is known. That changes what a batch is, not just what it reports, and deserves its own issue.

## Acceptance

- `free_dims` on `ValueSpec`, implemented per spec as above; `EventTemplate.free_dims` a union over children; `is_concrete` derived once.
- Substitution recurses into term specs, so `free_dims` and `_bind_template_dimensions` agree — every reported dimension is bindable.
- A test for each of the three cases in the repro above, asserting the *old* answer was wrong.
- `Record` construction and the function contract exercised with a symbolic-inner term spec.
- CHANGELOG entry under Fixed: `is_concrete` changes its answer for existing templates.

jhuggins:
## Summary

`EventTemplate.is_concrete` reported a polymorphic template as concrete whenever its symbolic dimensions sat inside a term spec:

```python
EventTemplate(x=ArraySpec(shape=("obs",))).is_concrete    # False  ✔
EventTemplate(law=DistributionSpec(sym)).is_concrete      # True   ✘
EventTemplate(r=RecordSpec(sym)).is_concrete              # True   ✘
EventTemplate(f=FunctionSpec(sym, sym)).is_concrete       # True   ✘
```

Design II.3 draws no line at a term-spec boundary — "A template with **any** symbolic entry is polymorphic, with `is_concrete` false and `free_dims` listing the unbound names." So this is a divergence from a documented contract rather than a design change.

Fixes #390.

> **Stacked on #386**, where `BatchSpec` lives, so the batch half of the same feature lands here rather than being retrofitted. Sibling to #394 over the same base — different files, so neither blocks the other.

## Four parts, and the third is the substance

The issue as filed described two. There are four, and they were *consistently* shallow, which is why nothing observed the bug.

1. **Reporting** — `free_dims` skipped term specs.
2. **Substitution** — `_replace_template_dimensions` skipped them identically.
3. **Matching** — `_unify_spec_with_value` short-circuited any non-`ArraySpec` leaf to `spec.is_valid(value)`, and a term spec's `is_valid` compares inner templates for **exact equality**, so a symbolic declaration never matched a concrete value.

Part 3 meant **a polymorphic term-spec leaf could not be satisfied at all**, so no working program could hold one, so parts 1 and 2 never bit. The feature was absent rather than half-built.

4. **Spec-vs-spec matching** — `_unify_specs`, the path the function contract uses for output validation, compared term specs by exact equality too, which left the boundary permeable in one direction only. It now unifies term-spec schemas as well.

**Part 3 is not new semantics.** `ArraySpec` has always done the right thing:

```python
ArraySpec(("obs",)).is_valid(zeros(5))        # True  — a symbolic dim accepts any size
ArraySpec(("n","n")).is_valid(zeros((3,3)))   # True  — repeated symbols must agree
ArraySpec(("n","n")).is_valid(zeros((3,4)))   # False   within the one value
DistributionSpec(x=("obs",)).is_valid(d)      # False ← the odd one out
```

II.3 divides the work: "A leaf's `is_valid` checks its own rank and dtype, and nothing else. **Sizes belong to the one pass**, since only it sees every occurrence of a name." `ArraySpec` follows that; the term specs folded sizes into `is_valid`. This PR brings three specs into line with a rule the fourth has always followed.

## The change

**`free_dims` is a property of every `ValueSpec`** — an `ArraySpec` reports its symbolic shape entries, a term spec those of the schema it carries, a spec declaring neither reports none. `EventTemplate.free_dims` collapses to a union over its children, and `is_concrete` derives from it once.

**Substitution reaches through**, and is a `ValueSpec` hook rather than a chain: `with_bound_dims` sits beside the `free_dims` each spec answers, so what a spec reports it substitutes. That matters structurally: a closed `isinstance` chain would have to live in `event_template.py`, which cannot name `BatchSpec` because `_batch.py` imports from it — leaving `with_dims` to return a still-polymorphic template for a batch axis, which is issue #390's own defect one level over. Every future `TermSpec` outside this module would have hit the same wall.

**Substitution happens once the scope is closed.** A term spec's dimensions may be bound by a field the walk reaches later, so no leaf can know at its own turn whether a name is bound. The first revision substituted per-leaf, which raised a raw `KeyError` out of `Record(...)` for a bare callable and made the same declaration behave differently depending on the order its fields were written in.

**The kind is still checked.** The polymorphic path skipped `is_valid`, so making a declaration polymorphic disabled its kind check — a `RecordSpec` declaration accepted a `MultivariateNormal`, since the schema is read by duck typing and nine classes in the package expose an `event_template`. The check now asks the spec's own `is_valid` with the schema the value carries, leaving the kind as the only thing under test.

**Unification reaches through**, in the *shared* binding scope. That last point is the load-bearing one: the declared schema unifies through `_unify_template_node` with the caller's own mutable `bindings`, not through `_unify_event_templates`, which copies them. So a name inside a term spec is genuinely the same dimension as that name beside it:

```python
declared = EventTemplate(
    data=ArraySpec(shape=("obs",)),
    law=DistributionSpec(EventTemplate(x=ArraySpec(shape=("obs",)))),
)
Record("r", data=zeros(3), law=law_over_3, event_template=declared)   # binds obs=3
Record("r", data=zeros(5), law=law_over_3, event_template=declared)   # raises: 'obs' → 3 vs 5
```

**`BatchSpec` axis sizes may be symbolic**, joining the same scope, so a batch of `("n",)` over arrays of shape `("n",)` is square by declaration. Only the **multiplicity** must be concrete for a live `Batch`, which holds elements at positions: `_init_batch` and `batch_size` gate on `free_axis_dims`, not on `free_dims`. An element's own schema may stay polymorphic — how many elements there are is a different question from what one of them looks like, and the first revision conflated them, refusing a batch whose axes were `(4,)`.

**`EventTemplate.with_dims(**sizes)`** binds explicitly, returning a new template so refinement stays monotone, naming any dimension left unbound, and auto-promoting to `NumericEventTemplate` when the result is all-numeric. Design 03 names `with_dims` on `Distribution`; that law-level form will delegate here.

## Out of scope

- **`FunctionSpec` variance.** A function is contravariant in its input, so "does this callable conform" is not one unification. The two sides bind independently against the value's own declaration; the docstring says so rather than implying a conformance check.
- **A polymorphic *actual*.** Refused before and still refused: the value side must be concrete.
- **`Distribution.with_dims`.** Law-layer — every distribution rebuilding itself under substituted dimensions, which wants W2.1's single schema source first.

## Design

`design/02-shared-abstractions.md` in four places: the `with_dims` signature in the II.3 listing; a paragraph stating that every spec reports its own dimensions and the scope reaches through a term-spec boundary, since the existing text said "a leaf" and the implementations disagreed about whether that included them; the binding prose naming `with_dims`; and II.2's `BatchSpec` fields plus an II.5 paragraph on polymorphic multiplicity.

## Tests

**+24**, and two of #386's updated where they pinned the old int-only axis contract.

`TestFreeDimsReachThroughTermSpecs` parametrizes the six declaration shapes — array, record, distribution, function-in, function-out, and nested — asserting each reports `obs`, plus one scope across the boundary and the concrete cases reporting nothing. `TestWithDims` covers binding through a term spec, a new template returned, the numeric promotion, the unbound-dimension message, and irrelevant names ignored. `TestInferenceThroughTermSpecs` covers binding from a distribution, a name shared across the boundary binding once, **the disagreement raising**, and a concrete declaration still demanding an exact match. `TestSymbolicMultiplicity` covers a symbolic axis reported free, the shared scope, `batch_size` raising, and a live batch refusing a polymorphic spec.

Verified by mutation: **35 deliberate breaks, all caught**. The 13 that a first pass let live clustered in `FunctionSpec` (8), the binding scope (1), and the schema helpers (4); each is now pinned, along with the four defects above.

The one worth naming: **M6b**, an isolated inner binding scope. My shared-scope test declared the array *before* the term spec, so it exercised only outer-to-inner propagation — which survives isolation, because the copy flows inward. The reversed order is the direction that proves one scope, and it is now tested both as a disagreement and as order-independence. Two other mutants had to be re-pinned at the unifier rather than through `Record`, whose later validation was masking the unifier's own kind check.

**2263 pass** across `tests/core` and `tests/record`. `ruff check` and `ruff format` clean at the pinned 0.15.16. `pyright` drops from **5 errors to 2** on the two changed files — the widening now reaches the parameter annotations and the two `range()` sites, with casts where a live batch's concreteness is guaranteed by construction.


probpipe/core/event_template.py
```diff
@@ -1287,6 +1426,8 @@ def _unify_spec_with_value(
     path: str,
 ) -> ValueSpec:
     """Validate and concretize one declared leaf spec against an actual value."""
+    if spec.free_dims and not isinstance(spec, ArraySpec):
```
SoymilkWinsAgain: The symbol axis of `BatchSpec` cannot be uniformly bound from the actual `Batch`. The new logic sends all non-`ArraySpec`s with `free_dims` to `_unify_term_spec_with_value()`, but `_declared_schema_of()` only supports `RecordSpec` and `DistributionSpec`.

Consider the following example:

``python
EventTemplate(batch=BatchSpec(OpaqueSpec(), [("n",)], ["item"]))

``` A `ValueError` will be thrown directly when matching an actual `OpaqueBatch` of size 3.

Even though `ArraySpec(("n",))` in the same template has bound n to 3, it still fails. This violates the design of this PR and the promise in the CHANGELOG of "adding a shared binding scope to the BatchSpec axis" and "a value binds by unification".

SoymilkWinsAgain (changes requested):
Additionally, AI recommends adding true end-to-end regression testing. Currently, the tests only verify `free_dims`, `with_bound_dims()`, and explicit `with_dims()`, without verifying consistency between the declared values ​​and the actual batch size. At a minimum, the following should be covered:

- The size of the separate symbolic `BatchSpec` should be inferred from the actual `Batch`.
- Array fields and Batch axes should share the same name, and the order of the two fields should be consistent.
- Explicitly report errors when the sizes on both sides are inconsistent.
- The templates for polymorphic `BatchSpec` and concrete `BatchSpec` should be consistent.

jhuggins:
Good catch — confirmed, and worse than it looked: with `n` already bound by the field beside it, the axis still refused, so "joins the same scope" only ever held for explicit `with_dims`. `_unify_specs` had the same hole, so a symbolic `BatchSpec` didn't unify against a concrete one either.

Fixed in `f95cbba8`, `ff9da910`, `feb8cbfe`. Binding is now a method every `ValueSpec` answers, beside `free_dims` and `with_bound_dims`, so the spec that declares a dimension resolves it — which is what reaches `BatchSpec` from `_batch.py`. `_declared_schema_of` and `_spec_around` are gone.

All four of your cases are covered, plus order-independence in both directions; eight of the nine fail against `31ed8a9c`. And you were right about why it survived — `TestSymbolicMultiplicity` never put a real `Batch` against a declaration.

Two more the same pass turned up: `ArraySpec` and `FunctionSpec` reported dimensions but inherited the base class's refusal (invariant test added over the live subclasses), and a non-record `FunctionSpec` output bound the input alone, reporting an output schema that was wrong rather than unbound.

SoymilkWinsAgain (approved):
Looks good!

PR #397, opened 2026-08-03
feat(core): a Record stores its RecordSpec, with event_template a view on it

jhuggins:
## Summary

A `Record` now stores its type as a `RecordSpec`, and `event_template` becomes a
view on that one object, so the two cannot disagree. This is the storage rule
`design/02-shared-abstractions.md` states in §II.2 and §II.4 — a tracked term
carries the spec of its kind as the single stored source of its type, and its
schema accessors are views on it — reaching the record side.

#383 landed the spec classes (`TermSpec`, `RecordSpec`, …) but no term adopted
them: `Record` still stored a bare `EventTemplate`, `Distribution` stores
`_event_template`, and `Function` stores `_input_template` / `_output_template`.
This closes the record half. Concretely:

- `Record._spec` (a `RecordSpec`) replaces the `_event_template` slot. New
  `Record.spec` reads it; `event_template` returns `self._spec.event_template`.
- Construction accepts a record declaration in **either** form, as §II.2
  specifies: a `RecordSpec` is stored as given, a bare `EventTemplate` is
  wrapped. A supplied spec is kept **by identity** rather than re-wrapped,
  matching the existing guarantee that a supplied template comes back verbatim.
- Three module-private helpers in `core/event_template.py` mediate the two
  forms: the pre-existing `_to_record_declaration` (template → spec), plus
  `_record_declaration_template` (read the structure out of either form) and
  `_record_declaration_for` (the spec to store, reusing a supplied one).
- The JAX pytree aux data is now `(spec, name, name_is_auto)`, and `__reduce__`
  serializes the spec.

Everything that reads `event_template` is untouched — that is the point of the
view.

### Scope

Deliberately the record side only. `Distribution.event_spec` and `Function`'s
`output_spec` belong with their own layers, and the `spec` slot is **not**
hoisted onto the `TrackedTerm` mixin yet: two of the four kinds have not
adopted, and an optional `spec` that is `None` half the time would be worse
than the current gap. `Batch` (#386) already stores its own spec on the same
per-class pattern; the base declaration lands once every kind carries one.

### The batched-record exception

`RecordArray` subclasses `Record` without being one record, so it overrides
`spec` to raise rather than reporting an element's spec as the collection's own
type. `Record.spec -> RecordSpec` is wrong for a batch under **any** option
here, not just the raising one: `Batch.spec` already returns a `BatchSpec`, so a
migrated batch will not answer with a `RecordSpec` either. The defect is the
subclassing, and it is being removed (#340, #386).

So the exception is written entirely into the files that go with it — the
override and its notes in `_record_array.py`, the test that pins it in
`tests/core/test_record_array.py`. `Record.spec` states only what stays true and
does not mention the batch at all. Removing `RecordArray` is then removing its
files, with nothing to hunt for in `record.py` or `test_record.py`.

Returning a real `BatchSpec` instead was considered and rejected:
`BatchSpec.__init__` requires a name for every level, and `RecordArray` carries
a flat `_batch_shape` with no notion of levels, so it would mean minting level
names and pre-empting the naming decision that belongs with the migration.

Verified safe: no `hasattr` / `getattr(_, "spec")` probe exists anywhere in
`probpipe/` or `tests/`, nothing reads `.spec` outside `_batch.py`, and
`pickle`, `copy.deepcopy`, `tree_leaves`, `repr`, and
`TrackedTerm._shallow_copy` all still work on a batch.

## Linked issue(s)

Part of #235 — the value-model refactor's contract discipline (`CONTRACTS.md`).
Part of #350 — substrate work the concrete batch types and the `Distribution`
base build on. This item is not on that checklist yet; see the review comment.
Refs #383 — landed the spec classes this adopts.

## Test plan

New: `tests/core/test_record.py::TestSpecStorage`, 17 tests covering the stored
spec and its view, both declaration forms, identity-vs-wrapping (including a
polymorphic declaration stored *bound*), numeric promotion through a spec,
nested children, the pytree aux, treedef and jit-cache behavior across
declaration forms, and the pickle round-trip. The batch's refusal is pinned in
`tests/core/test_record_array.py`, which goes away with the class.

Three of them assert the invariants that would otherwise rot silently:

- `test_the_spec_rides_in_the_pytree_aux` compares by **`is`**, not `==`. The
  aux spec is reused rather than re-wrapped, which is what keeps unflatten
  inside a `jit` / `vmap` / `grad` trace allocation-free; an equality check
  would still pass with that reuse lost.
- `test_a_pickle_carrying_a_bare_template_still_loads` drives the
  reconstructors directly. No current code emits that payload, so nothing else
  would reach the path the compatibility claim is about.
- `test_the_declaration_form_does_not_reach_the_treedef` builds one record four
  ways — a bare template, the spec wrapping it, and an equal-but-distinct of
  each — and demands **one** treedef with **one** hash. A treedef is a jit cache
  key, so a form-dependent split would silently retrace.
  `test_jit_traces_once_across_the_declaration_forms` states that end directly,
  and `test_a_different_declaration_gives_a_different_treedef` keeps the pair
  from passing vacuously (the trace counter reads 2 for genuinely different
  declarations).

Existing: `tests/core/test_record.py::TestEventTemplateStorage` and the rest of
the record suite unchanged and green.

Verified beyond the suite:

- **jit cache integrity.** Treedefs compare equal across equal-but-distinct
  templates *and* across the two declaration forms, unequal across different
  declarations, and hash consistently. A `@jax.jit` function traced **once**
  across records declared three different ways — switching declaration form
  does not fragment the cache.
- **Cost.** One 181 ns allocation against an 18 µs construction (~1%);
  `hash(RecordSpec)` 512 ns vs `hash(EventTemplate)` 453 ns. Unflatten allocates
  nothing.
- **Transforms.** `map`, `with_path_names`, `replace`, `without`, `merge`,
  `with_name`, `at_path`, `ensure`, `to_numeric` each leave a coherent spec —
  III.2 invariant 1 holds after every one.
- Full suite: 4065 passed, 54 skipped. `ruff check` / `ruff format` clean.
  Advisory pyright: no new errors (424 → 423).

## Breaking changes

Not user-facing API, but two stored representations change:

- **JAX pytree aux** is `(spec, name, name_is_auto)` instead of
  `(event_template, name, name_is_auto)`. Aux stays hashable and equal for
  equal declarations, so treedefs still compare by value.
- **Pickle payloads** carry the spec. A pickle written before this change still
  loads: its bare template is accepted as the declaration it is (tested).

No `kind:breaking-change` label — the public surface only gains `Record.spec`,
and `event_template` is unchanged in type and value.

## Documentation

`CONTRACTS.md`'s abstractions table now names the storage rule for
`Record` / `NumericRecord`. `docs/api/records.md` renders `Record`'s members via
mkdocstrings, so `Record.spec` and its docstring appear without an edit. The
user-guide notebook `02_records.ipynb` said the schema is inferred "then
cached", which is the wrong shape now; §4 names `record.spec` and
`event_template` as the view on it. Markdown only — outputs are committed and
mkdocs-jupyter runs with `execute: false`, so a code cell would mean re-running
the notebook for a large output diff with nothing extra pinned.

### Known adjacent issue, not fixed here

`RecordSpec.is_valid` admits a batch: `one_record.spec.is_valid(batch_of_two)`
returns `True`, because the guard is `isinstance(value, Record)` and
`RecordArray` subclasses `Record`. That predates this PR (#383) and resolves
itself when the batch types stop being records, so it is left alone rather than
patched here.


probpipe/core/_record_array.py
```diff
@@ -182,6 +183,30 @@ def template(self) -> EventTemplate:
...
+        AttributeError
+            Always, so ``hasattr`` reports ``False``.
+
+        Notes
+        -----
+        Interim: a batch subclasses :class:`Record` without being one record.
+        Both the subclassing and this override end when the batch types become
+        collections in their own right.
+        """
+        raise AttributeError(
```
SoymilkWinsAgain, line 205: `RecordArray` violates the public contract of `Record.spec`. `Record.spec` declares this property "Fixed at construction and always present," but throws an `AttributeError` here. However, the `RecordArray` documentation explicitly defines it as a `Record` and encourages receiving both via `isinstance(x, Record)`.

Therefore, any legitimate code that accepts a `Record` and reads a `.spec` will fail when it receives a `RecordArray`. Worse still, the type annotation incorrectly promises to return a `RecordSpec`. Future migration to Batch cannot fix the currently released API. The solution should be to first complete the hierarchical split, provide a true batch spec, or postpone the public release of `Record.spec`.

probpipe/core/record.py
```diff
@@ -580,25 +591,42 @@ def _check(record: Record, template: EventTemplate, prefix: str) -> None:
...
-        returned; otherwise one was inferred from the data at construction (via
-        :meth:`EventTemplate.infer_from`) and stored.
+        A view on :attr:`spec`, fixed at construction and always present. When a
+        declaration was supplied (carried forward from the producing generator)
+        it describes that structure; otherwise one was inferred from the field
+        data at construction (via :meth:`EventTemplate.infer_from`).
 
         Notes
         -----
         Inference is a lossy fallback (it cannot recover an ``ArraySpec``'s
```
SoymilkWinsAgain, line 615: It looks like this hasn't been updated? The original spec can be restored. The corresponding test name `test_pytree_roundtrip_reinfers_template` is also outdated.

SoymilkWinsAgain (changes requested):
The `docs/user_guide/02_records.ipynb` file needs updating because the API has changed.

AI recommends adding JAX regression tests across declarative forms: directly verify that records constructed using `EventTemplate` and the equivalent `RecordSpec` have the same treedef. Existing tests only compare record values; tests claiming to share a spec actually use the bare template.

jhuggins:
Thanks — all three taken, and `main` merged now that #396 has landed (only the changelog collided).

**`RecordArray.spec`.** You're right, and it's worse than my note admitted: `-> RecordSpec` is wrong for a batch under *any* option, since `Batch.spec` already returns a `BatchSpec`. The defect is the `RecordArray(Record)` subclassing, which #404 removes. A real `BatchSpec` here isn't available either — its constructor needs a name per level and `RecordArray` has no levels.

So instead of defending the text I made the exception free to delete: `Record.spec` no longer claims "always present" or mentions batches, and the override plus its test now live only in files that go with the class. At #404 the accessor is total and the claim comes back, true this time.

**The unpickling clause.** Right on both counts — `__reduce__` has serialized the declaration all along. Rewritten to say both round trips carry it, and the test is now `test_pytree_roundtrip_threads_the_declaration`.

**Notebook.** §4 updated to name `record.spec` with `event_template` as the view. Markdown only on purpose: outputs are committed and mkdocs-jupyter runs `execute: false`, so a code cell means re-running it for a large diff.

**Treedefs.** Good catch — the old test passed the same bare template to both records, so it compared a declaration to itself. Three added: one treedef and one hash across four ways of writing the declaration, the unequal case, and a `jit` trace count (which reads 2 for genuinely different declarations, so it isn't vacuous).

Separately, I found that `Record`'s pytree unflatten threads the stored declaration over whatever children a transform produced — reproducible on `main`, so not from this PR, but this is the code that owns the claim. Filed as #406.

4615 passed, ruff clean.

SoymilkWinsAgain (approved):
Looks fine!

PR #399, opened 2026-08-03
feat(core): add RecordBatch

jhuggins:
## Summary

Adds `RecordBatch` / `NumericRecordBatch` — a batch of `Record`s conforming to one
`EventTemplate`, on the `Batch[E]` ABC from #386. This is `design/` §III.3, and the
first half of W3.2 in #350.

**A field is stored once, for the whole batch.** `batch["x"]` hands back that
field's values across every element, shaped `(*batch_shape, *event_shape)`, and an
element is assembled on demand rather than stored a second time. Fields are keyed
by **leaf path**, which is what makes a nested one reachable at all:
`batch["outer/a"]` reads a field like any other and `batch["outer"]` gives the
sub-batch of the fields beneath it, so a nested record batches and reads back.
That is the storage the leaf-keying bug (#340) needs; **this PR does not close
it** — `_empirical.py` and the rest still use `RecordArray`, and the cutover that
migrates them is the follow-up.

A field comes back in the form its **spec** calls for: the array itself for an
array field, and a `FunctionBatch` or `OpaqueBatch` (#394) for a field with no
native stacked form.

**A batch is a collection, not a named tree.** No field-keyed `Mapping` protocol —
no `keys()` / `values()` / `items()` / `children` / `at_path` — so `len` and `iter`
speak unambiguously about the batch, and the field structure is read from
`event_template`. `[]` dispatches on the key: a position addresses the batch axes,
a name addresses a field. `select` / `select_all` survive as the field-splatting
selector, resolving a path as `Record.select` does and returning batch *views* that
carry the parent's level names; `select_all` keys by top-level name, since a
`/`-path could not bind to a parameter.

What does carry over is the **structure-preserving transforms** —
`with_path_names`, `without`, `replace`, `merge`, `map`, `map_with_keys` — because
they act on the elements rather than navigating the batch as a tree. Each is its
record counterpart applied to every element at once, and for field-wise storage the
two readings coincide: what a batch presents at a field *is* that field's values
across the batch. Where it shows is that `map` gets one call per field rather than
one per element, so it vectorizes, and `replace` takes a field's values for the
whole batch rather than one element's value. `with_path_names` renames fields
within the elements and `with_level_names` renames the levels; the two namespaces
are independent.

An element is **materialized** rather than stored — the other side of the rule
`Batch._element_at` states. It takes the derived name (`"post[draw=1]"`) marked
auto, inherits the batch's provenance, and is built against the batch's own
`element_spec`, so batch and element share one spec object: schema agreement is
structural, and a row costs no declaration to build. That reuse is asserted with
`is`, since an equality check would pass with it lost and every row inside a trace
would allocate.

**A batch holds what it validated.** Every field is checked against what it
declares: an array field for a numeric dtype its declaration admits, by the same
same-kind rule `ArraySpec.is_valid` applies to one value, and every other field
value by value against its spec, naming the field and the position that failed. A
field with no stacked form is given an object array rather than a dense one, and
stored frozen, so a caller keeping a handle cannot afterwards write in a value the
spec refuses.

`NumericRecordBatch` adds the batched flat layout — `to_vector` giving
`(*batch_shape, vector_size)` with the flat dimension last, `from_vector`
inverting it and restoring each field's declared dtype — and is a bare array
pytree. Under `vmap` the body receives an **element**, not the batch: the
transform strips the mapped axis, so the multiplicity is re-derived from what
arrives and a `Record` is handed over once no batch axis remains (a two-level
batch leaves the inner level).

### Following the object batches rather than inventing a second idiom

`FunctionBatch` / `OpaqueBatch` (#394) are the reference implementations, and this
copies their shape: the same constructor signature, a narrowing `element_spec`
property, `object.__new__` view construction, `__slots__` declaring only storage,
a required level name with no default. **The class is its own view type**, as they
are, so there is no view class — a field view and a sub-batch are both
`RecordBatch`. Three helpers are shared rather than reimplemented:
`_axis_groups_for` moves from `_object_batch.py` to `_batch.py` (it is about levels,
not storage), and the object-array construction and freezing move the other way,
into `_object_batch.py` beside the array building they belong with.

Where the storage forces a divergence it is stated: validation is per field and
names the leaf path that failed, and this is the first batch to implement the ABC's
`_at_fields` hook, which the object batches decline because their elements have no
fields.

### Layout

`RecordBatch` in `probpipe/core/_record_batch.py`, `NumericRecordBatch` in
`_numeric_record_batch.py` — the split `Record` / `NumericRecord` already
makes, for the same reason: the numeric specialization adds a flat-vector contract
of its own and nothing in the general case depends on it.

### Scope

**Additive.** `RecordArray` / `NumericRecordArray` are untouched and still what the
library uses, and the new classes are deliberately **not** exported from
`probpipe/__init__.py`. The cutover — migrating every consumer, deleting
`_record_array.py`, un-`xfail`ing the #340 test, and the docs/notebook pass — is
the follow-up PR, so this one is reviewable on the contract alone.

Stacked on #397 (`Record` stores its `RecordSpec`), which this uses: because a
record keeps a supplied declaration verbatim, `_element_at` can hand the batch's
own `element_spec` to each row.

## Linked issue(s)

Part of #235 — the value-model refactor's contract discipline (`CONTRACTS.md`).
Part of #350 — W3.2, first half; the cutover follows.
Refs #340 — this is the leaf-keyed storage that fixes it; the follow-up closes it.
Refs #386, #394 — the ABC and the object batches this builds on.

## Test plan

New: `tests/core/test_record_batch.py`, **151 tests** — construction and its
refusals, leaf-path field access (including the nested-template case #340 is
about), the form each field kind comes back in, per-field validation, the absent
`Mapping` protocol, the structure-preserving transforms, element identity and spec
sharing, the level algebra, `select`, the flat layout, `stack`, equality and
copying, and JAX.

Existing: full suite green, **4554 passed / 54 skipped**. `uvx ruff check` and
`ruff format` clean.

## Breaking changes

None. The public surface is unchanged — the new classes are not exported, and
`RecordArray` is untouched.

## Documentation

`CONTRACTS.md` gains an index row. `design/03-values-and-distributions.md` §III.3
and `design/package-structure.md` are corrected rather than left contradicted (per
`CONTRACTS.md` directive 4): §III.3 showed `from_vector` without the level names
the implementation requires and did not mention `stack`, and the target layout did
not have the numeric module. The user-guide notebooks belong with the cutover,
which is what changes what a user writes.


probpipe/core/_record_batch.py
```diff
@@ -1050,12 +1096,21 @@ def _unflatten(aux: tuple[BatchSpec, str, bool], children: list) -> RecordBatch
                 name_is_auto=name_is_auto,
                 _validate_leaves=False,
             )
-        # Reuse the stored spec where the multiplicity is unchanged, which is the
-        # ordinary round trip: rebuilding an equal one would allocate at every
-        # transform boundary and cost the identity a caller can rely on.
-        rebuilt = (
-            spec if rank == len(spec.batch_shape) else replace(spec, **_levels_for_rank(spec, rank))
-        )
+        if rank == len(spec.batch_shape):
```
SoymilkWinsAgain: I don't think comparing `len(spec.batch_shape)` is a good idea. Suppose `jax.tree.map` cuts (3, 2) to (1, 2) and the returned object still claims `batch_shape == (3,)`. An index might silently repeat a single row under JAX's out-of-bounds constraint behavior, instead of throwing an error.

probpipe/core/_record_batch.py
```diff
@@ -1087,26 +1142,47 @@ def _surviving_batch_rank(
...
-        kept = group[-remaining:] if remaining < len(group) else group
-        groups.insert(0, tuple(kept))
-        names.insert(0, level_name)
-        remaining -= len(kept)
-    return {"axis_groups": tuple(groups), "level_names": tuple(names)}
+    original = spec.batch_shape
+    dropped_count = len(original) - len(surviving)
+    if dropped_count < 0:
+        return None
+    for dropped in combinations(range(len(original)), dropped_count):
```
SoymilkWinsAgain: For a batch shape of (2, 2) and levels `("chain", "draw")`, after `vmap(..., in_axes=1)` only `"chain"` remains, but the code reports `"draw"`. This will cause a silent error when aligning by level name later.

probpipe/core/_record_batch.py
SoymilkWinsAgain: This calls `jnp.asarray` directly instead of using the existing `_to_jax_array`. Therefore, a valid NumericRecord leaf that can only be transformed via a registered `ArrayBackend.to_jax` would fail in `NumericRecordBatch.stack()`.

SoymilkWinsAgain, line 977: `shape=(0,)` can pass this

SoymilkWinsAgain: `batch["field"]` creates a `FunctionBatch`/`OpaqueBatch` through a normal constructor, which defensively copies the array of objects. Field access is therefore O(batch_size) and is no longer the shared store view promised by the document.

SoymilkWinsAgain (changes requested):
The AI ​​report states that the `float64 → float32` test case in `test_a_same_kind_cast_is_admitted` is truncated to `float32 → float32` when JAX x64 is disabled by default, and the narrowing path is not actually covered. A warning has been generated during test runtime.

probpipe/core/_record_batch.py
SoymilkWinsAgain, line 1177: When executing `jax.vmap(lambda row: row)(batch)`, the outer `vmap` will re-stack these Records, but still use the `RecordSpec` of a single element. As for reference, the AI ​​report shows that the field shape of `NumericRecord` is `(3,)`, but `event_template` is still declared as `()`; `event_template.vector_size == 1`, while `record.vector_size == 3`. This is a silent type invariant violation.

SoymilkWinsAgain, line 377: I guess there should be a `DistributionSpec` field here. Because the current `RecordBatch` can legally construct the `DistributionSpec` field, but cannot read it.

SoymilkWinsAgain, line 892: Since this function saves `FunctionSpec` only when `entries` is not empty, `empty.replace({"f": empty["f"]})` would change `FunctionSpec` to `OpqqueSpec` silently. When unpacking a `FunctionBatch`, its element spec should be preserved; it should not be inferred again based solely on empty data.

SoymilkWinsAgain (changes requested):

jhuggins:
Thanks — all eight taken across both rounds. Two changed the design, so flagging those first.

**Rank comparison and level recovery.** Both reproduce, both worse than reported. The rank check also mistook an *added* axis for an unreadable shape, so a batch-returning body produced a batch claiming two elements while holding six. And my first fix — enumerate the unsound cases and refuse them — was the wrong shape of answer: a drop combined with a resize lands on a *unique* candidate, so the ambiguity guard never fired and `x[0, :2, :]` on a `(2,3,4)` batch reported the survivors as `(a, c)` where they were `(b, c)`.

A shape isn't a provenance. So the hook no longer infers which axis went: it supports keeping every batch axis (reuse the spec) or removing all of them (yields a `Record`), and refuses the rest. `_levels_for_shape` is gone. An equal-size permutation is undetectable, so it's documented as unsupported rather than claimed as refused. Cost: raw `jax.vmap` over one level of a multi-level batch is now refused, default `in_axes` included — nothing in the library relied on it, since the sweep maps raw columns.

**`jnp.asarray`.** Right, and sharper than the note — `_to_jax_array`'s docstring names batch stacking as one of its callers, so the code contradicted the doc. Routed through it.

**`shape=(0,)`.** I read this one differently: the guard tests axis *count*, and `(0,)` is one axis, so an empty batch is admitted deliberately. But probing it found a real problem — `batch["f"]` then raised `requires at least one element`. Fixed by the next item.

**O(batch_size) field access.** Confirmed; it enters through a private no-copy path now, store shared and still frozen.

**The x64 test.** Right — the pairs stay inside 32 bits, and the truncation warning is an error so it can't come back.

**`DistributionSpec` field.** Confirmed. A batch now admits only the kinds it can present, so the refusal moves to construction.

**Empty `FunctionSpec` → `OpaqueSpec`.** Confirmed, and broader than emptiness: the unwrap discarded the `element_spec` it was handed, so a callable-valued `OpaqueSpec` field promoted the *other* way. The spec is carried now; both directions tested.

**`vmap(lambda row: row)`.** Confirmed exactly, numbers included. Not fixable here though — it's `Record`'s own unflatten and reproduces on `main` with no batch involved. Refusing it needs a migration first (the library routes records through transforms in five places); minibatch is already converted, which cleared SGMCMC and `condition_on` too. Written up in #406.

4784 passed, ruff clean, `main` merged.

SoymilkWinsAgain (approved):
Looks good

PR #400, opened 2026-08-03
feat(core): new structured RNG system in ProbPipe core

### LINKED ISSUE #384: Repeated lifted `Function` calls reuse the same sample batch after PR #380

### LINKED ISSUE #389: Make lifting RNG workflow-owned and structurally reproducible

SoymilkWinsAgain:
## Summary

This PR implements the workflow-owned RNG design from #389 and integrates it throughout ProbPipe’s existing workflow machinery.

The main change is not simply a new seed API. ProbPipe now explicitly models a workflow-owned stochastic event: how it is identified, planned, admitted, coupled to related consumers, transported across execution routes, claimed by retries, recorded in provenance, and closed at the end of its owning scope.

The implementation preserves `Function` as an immutable mathematical computation while restoring fresh stochastic realizations across calls and making automatic randomness structurally reproducible.

## Background

Before first-class immutable `Function`, a construction seed initialized mutable state on the Function:

```text
Function seed
    → Function._key
    → split and advance on every lifted call
```

This produced a useful sequence of fresh realizations, but made later results depend on hidden call history. Concurrent callers raced on the same state, execution order determined key ownership, and `with_options(seed=...)` modified the shared Function.

The immutable Function refactor correctly replaced that state with an invocation-local key reconstructed from the Function seed. That made direct calls reentrant, but every invocation restarted from the same key, so repeated lifted calls collapsed onto the same Monte Carlo realization.

At the same time, omitted-key operations such as `sample(dist)` used a separate process-global `_AUTO_KEY_COUNTER`. Function lifting and other automatic randomness therefore had different authorities, and the latter remained schedule-dependent, process-local, invisible to provenance, and outside any meaningful retry or distributed-execution contract.

This PR replaces both mechanisms with one relation:

```text
event key = derive(
    workflow root,
    occurrence path,
    stochastic source identity,
    logical unit identity,
)
```

Randomness belongs to an execution of a Function, not to the immutable Function object itself.

## User-visible behavior

A reproducible execution is now expressed with `workflow_run`:

```python
with workflow_run(seed=42):
    result = workflow(distribution)
```

The behavior is:

- Repeating the same seeded workflow structure reproduces the same workflow-owned random events.
- Consecutive stochastic calls within one run receive distinct occurrence paths.
- A bare stochastic call receives a fresh ephemeral root.
- `workflow_run(seed=None)` receives anonymous OS entropy lazily, only when its first automatic stochastic event is committed.
- Nested workflow scopes form lazy structural subpaths. Empty, deterministic, exact, caller-keyed, and preflight-failing scopes consume no RNG position.
- An explicit caller-owned key passes through unchanged, creates no automatic event, and does not move the following automatic stream.
- Repeated references to the same stochastic root are co-sampled; equal but distinct distribution objects remain independent.
- Managed retries of the same logical work item reuse its event identities and keys.
- Supported execution-route changes preserve logical RNG identity, although cross-route floating-point results are not promised to be bitwise identical.

The following Function controls are removed without a compatibility layer:

```python
Function(..., seed=...)
@function(seed=...)
function.with_options(seed=...)
```

A wrapped callable’s own parameter named `seed` remains an ordinary user parameter.

This PR also adds:

```python
workflow_run(seed=None)
replay_run(provenance)
```

and the public exceptions:

```python
UnmanagedConcurrentWorkflowEntryError
ReplayCompatibilityError
ReplayUnsupportedCallableError
```

## Architecture and module map

The implementation is divided into four layers:

```text
RNG identity and ownership
    ↓
Immutable stochastic planning and jointness
    ↓
Managed execution and transport
    ↓
Provenance and replay
```

You can think of it as a tree of ownership and a DAG of random source.

```text
                   random source DAG
                   source 0  source 1
ownership tree
occurrence 0        key 00    key 01
occurrence 1        key 10    key 11
```

#### `probpipe/core/_workflow_rng.py`

Defines the low-level structural RNG ABI.

It validates unsigned 64-bit workflow seeds, canonically encodes event identities, and derives two raw key words using keyed BLAKE2s. Those words are wrapped as a typed JAX Threefry key through `jax.random.wrap_key_data`.

A stochastic event identity has exactly three semantic components:

```text
occurrence_path
stochastic_source_id
logical_unit_id
```

Operational details such as thread IDs, process IDs, object addresses, worker names, UUID tokens, and completion order do not enter key derivation.

The JAX adapter is certified once per thread against the official typed-key and key-data round-trip contracts.

#### `probpipe/core/_workflow_context.py`

Defines workflow frames, lazy roots, occurrence ledgers, nested paths, and execution ownership.

A frame points to its parent so the active execution can:

- inherit or replace the effective root;
- lazily materialize nested scope paths;
- allocate stochastic child ordinals;
- verify process/thread/task admission;
- reject saved contexts after their installing scope has closed.

The frame graph is intentionally an upward context chain rather than a parent-owned child tree. Synchronous scopes are governed by context-manager nesting; concurrent children are tracked separately by the broker’s managed registry; remote workers cannot be represented by in-memory child references.

Anonymous and ephemeral roots read eight bytes of OS entropy only when a real automatic stochastic event requires a key.

#### `probpipe/core/_workflow_broker.py`

Provides the single authority for every ProbPipe-synthesized key.

Each automatic operation first creates a validated `StochasticEffectPlan` describing its operation mode, planned source/unit event, sample shape, sampling/provider ABI, record path, and optional descendant descriptor.

The broker then performs, in order:

```text
workflow ownership admission
→ broker lifecycle check
→ effect and event validation
→ JAX/probe guard
→ replay compatibility check
→ lazy occurrence commit
→ managed/retry claim
→ effect recording
→ key derivation
```

One broker commits at most one own occurrence. That occurrence may contain multiple planned source/unit events.

Broker lifecycle is explicit:

```text
OPEN → FINALIZING → SEALED
```

A sealed broker cannot register work, begin attempts, claim effects, accept remote reports, or publish additional successful effects. Finalization also requires all managed work items to have joined or been cancelled.

The legacy `_AUTO_KEY_COUNTER`, `_auto_key`, and direct automatic `PRNGKey(...)` fallbacks are removed.

#### `probpipe/core/_workflow_plan.py`

Builds the immutable batch and stochastic plans after call normalization and before any key, entropy, sampling, JAX probe, or managed submission.

`BroadcastPlan` fixes deterministic sweep geometry:

- the scalar, distribution, sweep, or nested regime;
- final sweep batch shape;
- `RecordBatch` level names and axis partitions;
- zip groups for operands with the same named levels;
- Cartesian-product groups for operands with disjoint levels.

Partially overlapping level sets and incompatible partitions of the same level are rejected during planning.

`StochasticPlan` independently fixes:

- source groups and ordered consumers;
- singleton or canonical row-major sweep logical units;
- sampled, exact, and mixed execution modes;
- sample shape;
- exact empirical group and combination order;
- repetitions and total evaluation count;
- planned source × logical-unit random events.

Named levels determine deterministic batch alignment. Plan-local stochastic-root identity and supported descendant capture determine realization sharing. Level names do not enter source IDs, logical-unit IDs, key derivation, provenance controls, or replay compatibility.

Exact empirical selection remains budget-aware, but the decision is made once in the planner rather than rediscovered during execution. Python object identity is used only within one plan build to detect aliases; it never enters canonical plans, key identities, provenance, or replay.

#### `probpipe/core/_workflow_descendants.py`

Captures the co-sampling graph for direct roots, record views, and the approved transformed-descendant set.

Supported TFP bijectors are deliberately closed:

```text
Identity
Exp
Square
Shift
Scale
Softplus
Sigmoid
Chain
```

The adapter validates exact bijector types, event rank, parameter surfaces, overrides, and cycles before sampling. Approved bijectors are copied, and both the strong descriptor and bound evaluator come from the same isolated snapshot, preventing live-object mutation from drifting away from the recorded descriptor.

Record views preserve canonical field paths, including transitive nested projections.

Capture uses a plan-local identity memo so aliases and shared transformed ancestors are inspected and frozen once without creating a global cache or merging equal-but-distinct objects.

#### `probpipe/core/_workflow_distribution_broadcast.py`

Executes the immutable stochastic plan for one logical unit.

For each source group it performs exactly one root sample:

```python
root_sample = sample_root(key, sample_shape)
consumer_values = tuple(
    evaluator(root_sample)
    for evaluator in source_group.consumers
)
```

This is the core jointness rule. Repeated aliases, record projections, and approved transformed descendants receive values derived from the same root realization rather than independently sampling their descendants.

Record-valued draw rows are stacked into a `RecordBatch` with one `draw` level; nested fields remain leaf-path keyed. Wholly exact groups enumerate support without requesting a key, while mixed plans request events only for sampled roots. Execution consumes planner decisions directly and never re-groups live distributions.

#### `probpipe/core/_workflow_sweep.py`

Executes pure batch sweeps and the outer batch layer of nested distribution sweeps.

Operands sharing the same named levels zip over the same cells; operands with disjoint levels form a Cartesian product. Multi-axis batches are sliced by positional tuples, and aggregate results preserve the final `level_names` and `axis_groups` from `BroadcastPlan`.

For nested sweeps, `StochasticPlan.logical_units` are the canonical row-major coordinates of that final geometry. Pure and nested paths therefore use the same cell identities, managed work-item representation, execution contract, and broker as ordinary distribution lifting without making level names part of random identity.

The local JAX path remains limited to one structurally supported batch operand. Capability checks, explicit-dispatch errors, Prefect transport, and the runtime RNG guard use the same execution-contract machinery as other workflow routes.

#### `probpipe/core/_workflow_execution_contract.py`

Defines the execution capability contract:

```text
probpipe.workflow_rng_execution/v1
```

Execution is separated into two axes:

```text
evaluator: rowwise | jax_vmap
transport: local_inline | local_thread | prefect_task | prefect_flow
```

A contract also binds the plan’s evaluation mode and required RNG, sampling, provider, descendant-adapter, and JAX-key ABIs.

Capability checks are reconstructed from the actual stochastic plan. Noncanonical ABI collections, plan-mode drift, unsupported provider/adapter combinations, and unsupported routes fail before broker activity or user execution.

JAX-in-thread is not part of the supported route matrix.

#### `probpipe/core/_workflow_managed.py`

Defines immutable, runtime-validated transport values for managed work:

- work-item tokens;
- unit frames;
- work items;
- attempt states;
- parent authority envelopes;
- effect claims;
- claim reports;
- Prefect payloads and outcomes.

A work-item UUID provides operational uniqueness only. It does not enter RNG identity or provenance.

Transport values are validated both at construction and again at receive boundaries so type annotations and pickle payloads are not trusted as authority.

#### `probpipe/core/_workflow_execution.py`

Transports canonical managed work items across sequential, local-thread, and Prefect routes. The final execution configuration, including Prefect names, decorators, and task-runner construction, is validated before key derivation or provider sampling.

The parent broker registers every logical unit before submission. Each attempt reserves exactly one unit, installs a managed workflow frame, and either joins or aborts that reservation.

Remote anonymous execution uses a rootless coordination probe. A deterministic worker can finish without forcing entropy. If it observes an automatic stochastic effect, it reports coordination instead of deriving a key; the parent then materializes authority and resubmits the same work item under a new reserved attempt.

Remote reports are validated before ledger mutation:

- frame, work-item token, and attempt must match the reservation;
- claimed effects must belong to the unit’s namespace;
- identities must be unique;
- successful effects must be a subset of claimed effects;
- retry and replay claims must remain compatible.

Accepted reports update managed, replay, and recipe-visible state transactionally. Results are reordered by canonical work-item index rather than completion order.

This provides stable random identity across supported routes and retries, but does not claim exactly-once user side effects.

#### `probpipe/core/_workflow_callable.py`

Builds a strong definition anchor for replayable Functions.

Version 1 supports importable, module-level, closure-free Python `def` callables. The anchor covers the code object, constants, referenced names, nested code, defaults, keyword defaults, signature, annotations, and Function templates.

Lambdas, closures, local functions, bound methods, partials, callable objects, builtins, and dynamic/private implementations remain executable normally but are recorded as unsupported for standalone replay.

The anchor deliberately does not recursively freeze globals, helpers, imports, files, native libraries, or external services. Replay reconstructs a definition-compatible random stream; it does not prove semantic equivalence of the entire computation.

#### `probpipe/core/_workflow_recipe.py`

Builds the successful workflow RNG recipe stored in provenance controls.

A recipe contains:

- normalized root words and RNG origin;
- occurrence path;
- canonical stochastic plan;
- canonically sorted successful events;
- callable anchor;
- replay compatibility ABIs;
- requested and resolved execution diagnostics.

Event order is canonical and independent of worker completion order.

Operational state is never persisted:

```text
owner
thread/task/process identity
work-item token
attempt token
worker
private ledger
```

Deterministic, wholly exact, and caller-keyed-only results do not produce an RNG recipe.

#### `probpipe/core/provenance.py`

Adds exact JSON-native `controls` and `diagnostics` fields after the existing positional fields.

The intended authority boundary is:

- `metadata`: descriptive operation metadata, retaining its historical compatibility behavior;
- `controls`: exact replay-authoritative data;
- `diagnostics`: non-authoritative execution and drift information.

`FULL` and `LIGHTWEIGHT` preserve equivalent compact RNG controls. `OFF` records no replay recipe.

#### `probpipe/core/_workflow_replay.py`

Implements `replay_run(provenance)`.

Replay validates the provenance envelope, RNG and recipe ABIs, root material, callable anchor, canonical plan, execution capabilities, and expected-event registry before allowing stochastic commit.

Every actual effect is checked before key derivation. Managed success is bound to the specific attempt that claimed the event, and remote replay namespaces must be complete when their scope exits.

A replay scope permits exactly one top-level `Function.__call__`. Unknown, missing, duplicate, malformed, or unexpected events fail closed. Mutable replay authority containing non-finite numeric values raises `ReplayCompatibilityError` before key derivation.

The recorded route is diagnostic rather than authoritative: replay may use another supported route if it satisfies the same stochastic plan and capability contract.

#### `probpipe/core/_workflow_errors.py`

Defines the three new public failure categories:

- `UnmanagedConcurrentWorkflowEntryError`;
- `ReplayCompatibilityError`;
- `ReplayUnsupportedCallableError`.

#### `probpipe/core/node.py`

`Function` remains the central public computation object. The RNG system is integrated into its existing call lifecycle rather than introducing a second workflow runner.

A `Function.__call__` now follows this order:

```text
workflow admission
→ install an ephemeral run if no run is active
→ install a Function stochastic broker
→ bind replay to the top-level call
→ resolve and normalize inputs
→ classify broadcast/sweep structure
→ build the immutable batch geometry
→ build an immutable stochastic plan when distributions are present
→ validate replay and execution capability
→ execute planned source/unit events
→ assemble result provenance
→ join managed work and seal the broker
```

The planner runs before JAX probing, key claims, entropy reads, managed submission, or sampling.

`Function.apply` uses the same workflow ownership and broker boundaries for nested omitted-key operations while retaining its raw single-evaluation semantics.

#### `probpipe/core/_workflow_call.py`

Continues to own normal Python signature binding, construction bindings, defaults, dependency resolution, and Function call options.

The removal of Function RNG seed does not reserve or reinterpret a wrapped callable’s parameter named `seed`.

## One end-to-end example

Consider one record-valued stochastic root, a field view, and a supported transform of that view:

```python
record_root = ...
value_view = record_root["value"]
exp_value = TransformedDistribution(value_view, Exp())

with workflow_run(seed=42):
    result = f(
        record=record_root,
        value=value_view,
        exp_value=exp_value,
    )
```

Internally this proceeds as follows:

```text
1. workflow_run installs a seeded frame but commits no random event.

2. Function installs its broker, binds arguments, and normalizes values.

3. The planner identifies record_root, value_view, and exp_value as consumers
   of one stochastic root.

4. The descendant adapter captures:
   - an identity evaluator for record_root;
   - a canonical record projection for value_view;
   - a frozen Exp.forward evaluator for exp_value.

5. The planner creates one source group and one singleton logical unit.

6. The execution contract verifies that the selected evaluator and transport
   support this exact plan and its provider/adapter ABIs.

7. The broker admits the owner and commits one Function occurrence.

8. The broker claims exactly one event for:
       source-group 0 × singleton unit

9. The workflow context derives its key from:
       root seed
       + occurrence path
       + source-group identity
       + logical-unit identity

10. The root distribution is batch-sampled once.

11. The three captured evaluators produce record, value, and exp_value from
    the same root realization.

12. The wrapped Function runs over the resulting values.

13. Successful planned events and the canonical plan are written to provenance.

14. Managed work is joined and the broker is permanently sealed.
```

The number of consumers and Monte Carlo draws does not create additional workflow events. Jointness comes from sampling the root once and applying deterministic consumer evaluators to the same realization.

## Automatic-key integrations outside the Function lifting path

The broker is also used by existing public subsystems when, and only when, they actually need omitted-key randomness.

### Core operations and distributions

`sample(key=None)`, empirical subsampling, bootstrap expectation, numeric Monte Carlo expectation, and numeric-record expectation use managed singleton effects.

Exact branches and explicit-key paths do not create workflow events.

### Converters

Converter resolution completes before key ownership is decided. The registry prepares and certifies one execution plan for the winning built-in converter, and that converter consumes the prepared plan without replanning before sampling.

- Exact, same-class, and analytic-moment conversions receive no automatic key.
- Certified ProbPipe and TFP sampled conversions use the broker.
- SciPy fallback uses a fixed `SeedSequence + PCG64` adapter from the derived raw key words.
- Opaque third-party sampled converters require an explicit key.
- Protocol delegation does not plan or claim the same effect twice.

### Validation

- Predictive checks use one planned event.
- Simulation-based calibration uses one event and performs its documented internal splits.
- Posterior scoring requests randomness only for the sliced-Wasserstein branch with reference draws.
- Omitted-key validation requires a certified provider; opaque providers require explicit ownership.

### Diagnostics

PPC diagnostics validate the full test-function sequence before sampling, then assign one stable source event per test function in input order. Replication count affects batch shape, not event count.

Annotations are written only after the complete diagnostic computation succeeds.

## JAX behavior

Automatic keys are derived and wrapped outside JAX transformations.

Traceability probes mirror the transform selected for execution. Distribution lifting probes the scalar body with synthetic `ShapeDtypeStruct` roots derived from closed event templates, while a single-`RecordBatch` sweep probes a one-row `jax.vmap` over flattened batch leaves. The probe never invokes provider sampling, and the complete probe runs inside the workflow side-effect guard. If a source cannot declare concrete shape and dtype information, automatic dispatch falls back and explicit JAX dispatch fails before stochastic commit.

- `dispatch="auto"` falls back safely if omitted-key randomness or an unsupported transform is observed;
- explicit JAX dispatch rejects unsupported automatic effects before commit;
- the actual JAX body retains a runtime guard against dynamic omitted-key requests or managed submission;
- caller-owned trace-compatible keys remain supported.

Planning, probing, and failed route selection do not consume RNG positions or write recipes.

## Intentional boundaries

This PR does not:

- intercept arbitrary NumPy, JAX, user-code, external-service, or backend-owned randomness;
- provide exactly-once semantics for user-visible side effects;
- guarantee semantic or bitwise result equality across execution environments;
- support arbitrary bijectors or transformed event ranks;
- provide standalone replay for nested automatic-call parents;
- freeze transitive globals, helpers, files, native libraries, or external services;
- restore Function-local seed or introduce a public call-local common-random-number control.

Caller-supplied sampling keys and explicit inference `random_seed` arguments retain their existing ownership.

## Suggested review order

The implementation is easiest to review in this order:

1. `_workflow_rng.py`, `_workflow_context.py`, `_workflow_broker.py` — identity, ownership, and lifecycle invariants.
2. `_workflow_plan.py`, `_workflow_descendants.py`, `_workflow_distribution_broadcast.py` — batch geometry, stochastic planning, and jointness.
3. `node.py`, `_workflow_sweep.py` — Function integration, named-level sweeps, and JAX route selection.
4. `_workflow_execution_contract.py`, `_workflow_managed.py`, `_workflow_execution.py` — route capability, concurrency, transport, and retries.
5. `_workflow_callable.py`, `_workflow_recipe.py`, `_workflow_replay.py`, `provenance.py` — audit and replay.
6. Core operations, converters, validation, and diagnostics — migration of the remaining automatic-key entry points.

## Linked issue(s)

Closes #389.

Related background: #384, #388, #368.

## Test plan

- Focused workflow RNG, broker, replay, converter, validation, diagnostics, and Prefect/JAX conformance tests.
- `uv run pytest -n 4 --dist worksteal`
- `uv run ruff check .`
- `uv run ruff format --check .`
- `git diff --check origin/main...HEAD`
- `uv run mkdocs build --strict`
- Pyright 1.1.410 as an advisory check.

## Breaking changes

- Removes `seed` from `Function`, `@function(...)`, and `Function.with_options(...)`.
- Reproducible automatic randomness now uses `workflow_run(seed=...)`.
- Adds `replay_run` and the three documented workflow and replay exceptions.

## Documentation

- [x] User Guide and tutorials updated where relevant
- [x] CHANGELOG updated

## Checklist

- [x] Linked to issue #389


SoymilkWinsAgain:
Pushing a version to see it if can pass the checks

SoymilkWinsAgain:
Changed `derive_event_key_words` algorithm. Now 60x faster

SoymilkWinsAgain:
ABI collection is new 100x faster

SoymilkWinsAgain:
The update conflicted with the main update. There are no substantial changes because the two updates are orthogonal. The problem mainly stemmed from the dependency on the old `RecordArray`, which can be resolved by replacing it with the new `RecordBatch`. While not a mechanical replacement, the new class inherits most of the old class's functionality in the project, so a redesign isn't necessary.

The most significant changes are in `_workflow_plan.py`, but it only needs to explicitly distinguish:
- stochastic grouping: by random root identity;
- batch alignment: by level names.

There are also some helper changes, but these are brainlessness.

SoymilkWinsAgain:
This is not the final version, just push to see if it can pass CI

jhuggins (commented):
`register_managed_work_items` is still called before the `try` in `probpipe/core/_workflow_execution.py:99`. A partial registration failure therefore leaves units `ISSUED`, skips the `finally`, and lets `_begin_finalization` raise "cannot exit before all managed work items join" over the real cause, with the broker left `OPEN`. Either move registration inside the guarded region, or have `_finalize(successful=False)` cancel outstanding reservations rather than asserting they joined.

## Two from the audit

1. `test_rootless_task_coordinates_nested_managed_thread` sits in `TestPrefectProvenance` ("Verify provenance includes orchestration info") rather than `TestPrefectRngConformance`. It runs in a full-file CI run, but running the conformance class alone collects three tests and silently skips the blocker regression. Worth moving.
2. Moving `_source_artifact` after the form check means unsupported callables no longer record `source_location` / `source_artifact_digest`. That is the right trade for the hot path and the tests assert it, but it removes the "where was this lambda defined" hint when diagnosing a replay refusal — flagging in case it was not intended.

Minor leftovers, all cosmetic: the converter execution plan is still derived twice, `_canonical_json` in `_workflow_recipe.py` still lacks the `allow_nan=False` that replay's copy sets, and the source-scan test still asserts exact source substrings from `_glm.py`.

Move the regression test and close the registration/finalization gap and this looks good.

jhuggins (changes requested):
The findings below come from a Claude-assisted review — four independent passes over the delta (conventions and docs, ownership and concurrency, replay validation, tests), with most items confirmed by executing code.

## Provenance mode: snapshot it, don't read it twice

`node.py:695` samples `provenance_config.mode` on entry while the recipe is built on exit (gate at `_workflow_recipe.py:24`). If the mode goes OFF → non-OFF across that window, the recipe records `callable: {supported: False, form: "missing_function_anchor"}` beside a complete `randomness` and `plan` envelope — a record that passes the entire structural validator yet can never be replayed, with `ReplayUnsupportedCallableError` surfacing only later. `provenance_config` is a plain global rather than a `ContextVar`, so a concurrent flip from another thread reproduces it too. This was impossible before the gating change, when the anchor was always captured.

The fix I'd like isn't a new rule for users — it's having the run read the mode once. Let the workflow run capture `provenance_config.mode` at entry, and have both `node.py:695` and `_workflow_recipe.py:24` read that snapshot, with nested runs inheriting the outer value. A mid-run flip then becomes harmless rather than forbidden: the run behaves consistently at the mode it started with, and the change takes effect on the next run. There are only three functional read sites of the mode — those two plus `Provenance.create` at `provenance.py:237` — so the plumbing is small, and the active frame is already the natural carrier (`_workflow_context` doesn't import `config`, so there's no cycle to work around).

This is also the discipline the rest of the PR already follows: the stochastic plan is frozen once, root words resolve once and cache, the execution contract is fixed at plan time. Provenance mode is the one control still read live at two separate moments.

Worth keeping a cheap backstop: if a non-OFF snapshot has no callable anchor at recipe build, raise rather than emit a record that claims replayability it doesn't have.

## Still open from last round

- `register_managed_work_items` before the `try` at `_workflow_execution.py:99`.
- `test_rootless_task_coordinates_nested_managed_thread` in `TestPrefectProvenance` rather than `TestPrefectRngConformance`.

## Tests

- `test_numpy_integer_counts_are_normalized_before_event_commit` (`tests/validation/test_workflow_validation_rng.py:269`) passes `key=jax.random.key(11)` with no `workflow_run`, so it commits no event and never reaches `encode_random_event` — the code that actually rejects `np.int64`. The sibling at `:257` asserts precisely that no commit happens. The PPC twin at `:119` is correct; this one needs the explicit key dropped. The code itself is fine: `operator.index()` was verified on all three sites.
- Cross-route conformance covers OFF/sequential, OFF/thread, and TASK/sequential. There's no `FLOW` case asserting equality with local_inline under one seed, and no TASK + `dispatch="thread"`.
- Of the five new coordination `RuntimeError`s, "already installed" and "must exit in nesting order" cannot fire — a fresh frame and observation are allocated per payload, including retries and Prefect-in-Prefect — so mark those. `_workflow_broker.py:852` is reachable and worth a test.
- The golden vector's literals carry no regeneration note and no gate on JAX version or `jax_default_prng_impl`; `test_workflow_recipe.py:509` is the better in-repo idiom.

## Duplication

`_validate_version_one_structure` repeats the compatibility-fields and per-effect field-set checks with byte-identical messages, so the copies in `_validate_provenance` (`:1057`) and `_validate_effect` (`:1322`) can never fire first. `_validate_positive_int` and `_validate_conversion_sample_count` are now the same function differing only in a message noun. And there is a third `_canonical_json` without `allow_nan=False` at `_workflow_callable.py:393` — that one feeds the callable-anchor `sha256`, so it's worth fixing regardless of the other two.

## Smaller

- `retry_ordinal = max(...)` at `_workflow_execution.py:388` charges the whole batch the most-retried item's delay, skipping the ramp with a list schedule; `min` or a per-item delay is closer to Prefect's per-task-run semantics.
- `_find_remote_coordination_observation` acquires a lock per hop on every `key_for`, replacing a lock-free read. The field is a single attribute and the walk isn't atomic anyway.
- Version-1 exactness runs before the schema ABI checks (`_workflow_replay.py:975`), so a future `/v2` record reports "does not match the version-1 schema" rather than the schema error written for that case.
- PPC name uniqueness keys on `__name__`, which rejects two bound methods of different instances; `__qualname__` plus rejecting the address-bearing `repr` fallback would be tighter.
- Now dead: `_RemoteCoordinationObservation.attempt`, and `_unsupported_anchor`'s `source_location` / `source_artifact_digest` parameters.
- Docs: `docs/api/validation.md:33` narrates release history where the docstring correctly states the current contract; neither it nor the CHANGELOG mentions the `ValueError` raised when `provider._x is None`; and the duplicate-`test_fns`-name rejection isn't in the CHANGELOG.

SoymilkWinsAgain:
A bug was discovered: Prefect route preflight failures occur after sampling, incorrectly advancing the RNG flow.

Trigger scenario: TASK/FLOW lifting fails during the construction phase of the Prefect decorator, flow/task name, or task runner. For example, the default lambda name "<lambda>_map" is rejected by the Prefect, or an invalid task runner is configured.

The actual order is: first, `_sample_planned_source_groups()` is called, submitting the invocation, derived key, and sampling; only then are the execution config, Prefect wrapper, and managed request constructed.

This violates the contract that "preflight failures do not consume stochastic positions".

SoymilkWinsAgain:
There's another rather silly bug: the JAX route probe actually calls the provider, causing an event to be sampled twice. When `dispatch="auto" or "jax"`, the probe calls the actual `_sample(..., (1,))` with a fixed key; the actual execution then calls `_sample(..., (n,))`.

SoymilkWinsAgain:
This PR looks large, but it actually only involves changing imports in many places and adding `slot=True` to many `@dataclass` tags. However, it would be better to split it into several smaller ones in the future.

SoymilkWinsAgain:
Reduced the number of file verifications significantly. Although it is not the theoretical lower bound, it is a relatively good choice of security and efficiency under the existing framework.

jhuggins (approved):

PR #401, opened 2026-08-04
refactor(core): recognize a batch of records wherever a batched record was

jhuggins:
## What this is

A `RecordBatch` is deliberately **not** a `Record`. Every place in the package
that recognized a batched value by `isinstance(x, Record)`, by a `RecordArray`
subclass check, or by duck-typing on `.fields` therefore stops recognizing one
when a batch arrives.

None of those gates raise. They take the other branch. A batch reaching
`_wrap_as_record` would be re-wrapped as a single opaque field named after the
function; a batch reaching `_data_size` would be minibatched by its *field*
count; a batch reaching the ArviZ bridge would be read as a bare array under the
variable name `"x"`. Each of those is a wrong answer, not an error.

This PR opens those gates and pins them with tests. **No producer returns a batch
yet**, so the suite is unchanged in behavior and every existing path still runs
on `RecordArray`.

Stacks on #399 (`dev/record-batch`), which adds the classes.

## The gates, and what each now does

**The `Function` boundary** (`_workflow_result`, `_function_contract`). A returned
batch stays the batch it is instead of being wrapped, and becomes an independent
result copy: for a batch, retyping under a declared output template replaces the
`element_spec` inside its `BatchSpec` rather than overwriting a template, since
the batch's own type is the `BatchSpec`. Declared-output validation accepts a
batch as a schema-carrying result and checks each column against its field spec.

**Broadcast planning** (`_workflow_plan`, `node`). A batch argument is a batched
argument: it sweeps its rows, the trace probe takes row 0 for the dummy call, and
the hint check accepts a `RecordBatch` annotation. The plan's flag is now
`is_batched_record`, which is what it has always meant.

**The broadcast helpers** (`_broadcast_distributions`). `_row_count`,
`_take_rows`, and `_one_row` count, gather, and unwrap a batch's rows. A gather
rebuilds the batch with the same levels and only the leading axis's size
rewritten, so which level holds that axis is preserved. `_RecordMarginal` peels a
batch's rows axis into a record of batched leaves, reading `event_template` —
the accessor that means the same thing on a `RecordArray` and a batch.

**The value boundaries.** `NumericRecordDistribution.flatten_value` and the
dynamic `_log_prob` take a `NumericRecordBatch` through `to_vector`; a
`_RecordDistributionView` reads its column out of a batch; the joint Gaussian and
product / sequential-joint log-densities accept one, the two joints re-nesting a
batch's leaf-keyed columns so the tree map still pairs each column with the
component that declared it; and the GLM design coercion extracts or stacks a
batch's columns.

**Minibatching** (`_minibatch`). Row count from `batch_shape`; row gathering a
field at a time, since neither a `RecordArray` nor a batch indexes by an array of
positions.

**The ArviZ bridge** (`_arviz_bridge`). Variables are found through
`event_template`, which a batch has, *or* `.fields`, which a duck-typed
record-like may be all that has — `_leaf_keys` already reads whichever is
present.

## Deliberately left for the cutover

Two paths need a decision rather than a wider gate, and are called out in the
changelog:

- Stacking a list of batched records from a broadcast has to name the levels the
  broadcast grid mints. That is a naming decision, not a widening.
- `RecordEmpiricalDistribution` requires a `Record` outright; migrating it is
  #340.

## Testing

`tests/core/test_record_batch_interop.py` — 20 tests, one per gate, each driving
a hand-built batch through the consumer rather than through a producer. That is
the only way to reach these branches while producers still return `RecordArray`,
and it is why they get their own file.

**Two of the gates are the sweep planner's.** Levels align by name, and only
whole: a level shared by operands whose other levels differ is refused rather
than silently producted, operands naming the same levels must hold them on the
same axes, and the annotation that takes a batched value whole answers `Batch`,
generic aliases, and optional unions (`Batch | None`) alike — while an element
annotation such as `p: Record` keeps the sweep.

**Two of the gates are the sweep planner's.** A batch view has no parent to look
up — a view over a batch is another batch over the same axes — so the planner
groups batches by their **level names**, the level algebra's own alignment rule:
sibling `select_all` views zip, batches with no level in common form a product,
and one level name at two sizes is refused rather than producted silently. And
the annotation that lets a body take a batched value whole is answered by the
value, and only when it names a batched container: family membership alone
delivered a batch whole to a body annotated with the other batched-record class,
while an element annotation like `p: Record` — which a record array satisfies by
subclassing — keeps the sweep, since asking for an element is asking for rows.

Existing: full suite green, **4730 passed / 48 skipped / 1 xfailed / 1 xpassed**
with every optional backend installed. `uvx ruff check` and `uvx ruff format`
clean.

Part of #350 (W3.2) and a prerequisite for #340.

### DIFF
--- a/probpipe/core/_workflow_plan.py
+++ b/probpipe/core/_workflow_plan.py
@@ -10,7 +10,7 @@ from __future__ import annotations
 from collections.abc import Mapping, Sequence
 from dataclasses import dataclass
 from math import prod
-from typing import Any, Literal
+from typing import Any, Literal, get_origin
 
 from . import _workflow_call, _workflow_distribution_normalization
 from ._batch import Batch
@@ -143,7 +143,19 @@ def group_array_args_by_parent(
         batch_shape = tuple(first.batch_shape)
         for ref in arg_refs[1:]:
             other = _workflow_call.input_ref_value(values, ref)
-            if tuple(other.batch_shape) != batch_shape:
+            if isinstance(first, Batch):
+                # Two operands naming the same levels claim the same axes, group
+                # by group: agreeing on the flat shape alone would zip a
+                # ((2,), (3, 4)) partition with a ((2, 3), (4,)) one and hand the
+                # output whichever arrived first.
+                if tuple(getattr(other, "axis_groups", ())) != tuple(first.axis_groups):
+                    raise ValueError(
+                        f"{arg_refs[0].label!r} and {ref.label!r} carry the same levels but "
+                        f"are batched differently: {tuple(first.axis_groups)} against "
+                        f"{tuple(getattr(other, 'axis_groups', ()))}. Levels align by name, "
+                        f"so operands naming the same levels must hold them on the same axes"
+                    )
+            elif tuple(other.batch_shape) != batch_shape:
                 raise ValueError(
                     f"{arg_refs[0].label!r} and {ref.label!r} are zipped together but are "
                     f"batched differently: {batch_shape} against {tuple(other.batch_shape)}. "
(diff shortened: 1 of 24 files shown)


probpipe/core/_workflow_plan.py
SoymilkWinsAgain: The parameter name of the old `RecordArray` is mistakenly treated as a level name. Non-Batch groups use parameter names to forge names. Therefore, the parameter `draw=<RecordArray>` and another `RecordBatch` with level `draw` will be incorrectly judged as overlapping levels and throw an error, instead of forming a Cartesian product of (2, 3).

SoymilkWinsAgain (changes requested):

jhuggins:
Thanks — confirmed and fixed. The error message was self-refuting, which is a good tell that the level was fabricated:

```
'draw' and 'other' share the level 'draw' without sharing all their levels
(('draw',) against ('draw',))
```

Two identical tuples, reported as not sharing all their levels.

Cause is the one you name: a batched operand carrying no levels of its own was entered into the overlap check under its *parameter* name. Its multiplicity is anonymous — it aligns with nothing by name and products with everything — so it belongs in no owner. The check reads the levels an operand actually has now, and `draw=<levelless>` beside a batch with a real `draw` level plans as the (2, 3) product. The genuine overlap is still refused, test unchanged.

One correction to something I said earlier: I'd assumed a plain array was the levelless operand that outlives the cutover. It's `DistributionArray` — swept by its `batch_shape` without being a `Batch`. The test is written against that, so it keeps covering the case after #404 rather than disappearing with `RecordArray`.

4819 passed, ruff clean, `main` merged.

SoymilkWinsAgain (dismissed):
Looks good

PR #402, opened 2026-08-04
feat(core): a joint law draws a record batch

### LINKED ISSUE #403: The JAX traceability probe traces a bare call, not the vmap dispatch actually runs
## What happens

`Function`'s automatic dispatch decides between `jax` and `sequential` by probing
whether the body is JAX-traceable. The probe traces a **bare call**:

```python
jax.make_jaxpr(lambda kw: func(**kw))(dummy_kw)
```

but what it is deciding about is a **`jax.vmap`** over that body. A body that
traces fine unbatched but is incompatible with `vmap` therefore passes the probe,
`jax` is chosen, and the call fails at execution instead of falling back.

## A body that hits it

A function whose body returns a `RecordBatch`:

```python
@function
def f(v):
    return NumericRecordBatch(
        {"s": jnp.asarray(v["x"]) + jnp.zeros(2)},
        "inner",
        element_spec=EventTemplate(s=()),
        axis_groups=((2,),),
    )

f(v=some_batch)   # ValueError from unflattening
```

`vmap` adds an axis to each column on the way out. That axis belongs to no level,
and unflattening has no name to give one, so it raises. The sequential and thread
paths handle the same call correctly, returning a batch over
`("<the swept level>", "inner")`.

## Why the fallback does not catch it

`Function._resolve_dispatch` already falls back on a probe error:

```python
if self._jax_traceability_error(values, broadcast_args, func=func) is None:
    return "jax"
logger.info("Function '%s' is not JAX-traceable; using sequential dispatch.", self._name)
return "sequential"
```

The machinery is right; the probe is just not representative of what runs.

## Proposal

Trace the probe under `vmap` over a stacked dummy row, so it exercises the
transform the dispatch is choosing. The existing probe-to-sequential fallback then
covers this case and any other `vmap`-only incompatibility, with the INFO log it
already emits.

Worth reviewing on its own rather than inside a value-type change: the probe
governs every automatic dispatch decision, not only ones involving batches, and
tracing a `vmap` costs more than tracing a bare call.

## Scope

- `probpipe/core/node.py` — `Function._jax_traceability_error`
- Tests: a body that traces bare but not under `vmap` should resolve to
  `sequential` rather than raise; bodies that vmap fine should still resolve to
  `jax`, so the change does not silently move everything to the slow path.

jhuggins:
## What this is

The joint laws now draw a `RecordBatch`. `_sample` on product, sequential,
Gaussian, and empirical returns a `NumericRecordBatch` for a batched draw — a
`RecordBatch` when a leaf is non-numeric — over a single `draw` level spanning
however many axes the `sample_shape` had. An unbatched draw is still a `Record`.
The class follows the leaves, so an object-valued law draws a batch on the same
terms as a numeric one.

Stacks on #401, which opens the gates a batch arrives at.

## The two real changes

**A nested draw is one flat batch, not a batch per subtree.** A batch stores one
column per *field*, so a nested product draws into one mapping over `/`-paths and
the result is a single store whose `batch["outer"]` is a *view* over the columns
beneath `outer`. The recursion that used to build a record-array per subtree now
collects columns, splitting keys per level exactly as the single-draw recursion
does, so a batched draw and a single draw derive their subkeys the same way. This
is also what makes a nested field reachable by path, and it removes the
`NotImplementedError` that refused nested batched records.

**A broadcast mints the levels it swept.** Stacking rows that are each a batch
puts the swept levels in front of the levels a row already carried. An operation
that mints a level takes the name to give it (design/02, level names), so
`_make_stack` requires `level_names` — one per swept group, with each group's
own axis partition carried through the plan — rather than naming a level itself:
the sweep passes the operand's own names when it carries them, so the aggregate
aligns by name with the batch it came from, and the argument's label otherwise.
Leaf-keyed columns are why the nested case needs no special handling here either.
The sweep addresses a multi-level batch by unraveled position, one indexer per
batch axis; a flat index read the leading axis alone and ran off its end. An
empty declared sweep builds its aggregate from the output template, every
declared field present at zero rows.

**Rearranging a batch reads its columns, not its fields.** A field that is not an
array presents as the batch of its element kind — a `FunctionBatch` or an
`OpaqueBatch` — which is what reading one field wants and the wrong thing for an
operation moving the storage around. Four such operations were reading the
presented form: gathering rows for a broadcast, peeling the batch axis for a
marginal, indexing a minibatch, and splicing a component's draw into a product.
`_raw_column` / `_raw_columns` name the seam, and an object column gathers through
numpy so the positions are taken without trying to make an array of the values.

**A declared output template that reorders the same fields no longer swaps
columns.** A batch flattens in its spec's leaf order, so retyping a copied batch's
spec alone left the next unflatten pairing every value with the wrong key.

**Stacking a batch per row requires the rows to agree** on element spec, level
names, and axis groups, not merely on batch shape — otherwise the first row's
schema and level names were taken for all of them, dropping a field the others
carried and misnaming their axes.

**Levels align by name, and only whole** (with #401): a level shared by
operands whose other levels differ is refused rather than silently producted,
operands naming the same levels must hold them on the same axes, and a
parameter annotated `Batch` or `Batch[Record]` takes the value whole.

**A transform cannot lie about a batch's multiplicity — or its element.** `vmap` strips the
mapped axis on the way in, which unflattening handles by re-deriving the
surviving levels; on the way out it *adds* one, and an added axis belongs to no
level, so unflattening raises rather than keeping the stored spec — which
returned a batch whose `batch_shape` its own columns contradicted. A same-rank
transform is held to the same standard: slicing keeps every axis, so the levels
carry over onto the sizes the columns actually have, and columns left
disagreeing about their batch axes are refused. The element's own axes are the
element type's: a per-column slice that shrinks the event, a transpose that
reads an event axis as a batch axis, and a reduction below the event rank all
raise rather than rebuild a batch declaring an element it does not hold.

**Shape cannot recover axis provenance, so ambiguity refuses.** A removal whose
size matches more than one level and a permutation of the batch axes both raise
rather than guess — the leftmost-fit answer handed a `vmap(in_axes=1)` body a
batch named for the level that was *removed*, and a transpose kept each level's
name on an axis holding another's rows. A distinctly-sized removal names the
level that survived. A pinned dtype is held like the event axes, under the
constructor's same-kind rule.

**Zero rows is not missing output — and one schema under every dispatch.** An empty sweep builds its declared fields
only when zero rows were *expected*; an empty list where rows were expected
reports the count mismatch instead of fabricating a well-typed result. Zero-row
sweeps take one aggregation path under every dispatch, so the output schema no
longer depends on how the rows would have been executed. The
dispatch probe states the flat batch size exactly as the executor does, so a
zero-width event column passes under explicit `dispatch="jax"`.

**Automatic dispatch probes the vmap it is choosing.** The traceability probe
traced a bare call, so a body that traces cleanly but cannot run under `vmap` —
one returning a batch — passed the probe, took the jax path, and failed
mid-call. The probe now feeds a batched argument the way `execute_sweep_rows_jax`
will — leaf columns over one row, rebuilt into a `Record` inside the traced call,
under `vmap` — and a probe failure means sequential dispatch, which agrees on
results by the dispatch-equivalence contract. Fixes #403.

**Rows holding opaque or callable fields stack, and marginalize.** Stacking read
each field through the presenting accessor and handed `OpaqueBatch` wrappers to
`jnp.stack`; the raw columns stack, through numpy. An object-valued batch
marginal routes to the general list marginal rather than inheriting numeric
empirical validation.

**A batch is fingerprinted by its levels and its columns.** A multi-field batch
failed fingerprinting outright, so provenance recorded no input; a single-field
one hashed as its sole column, omitting schema and levels.

**A declared `support` on a batched output is checked column by column.** Walking
a batch as a named tree found no children and asked a multi-field batch to convert
to one array, so two valid columns raised.

## Consequences for calling code

Three, all stated in the changelog:

- A batched draw is not a `Record`, so `isinstance(draw, Record)` is now `False`
  where it was `True`. Ask for `RecordBatch`, or for either.
- A batch is a collection, not a named tree: a draw's fields come from
  `draw.event_template`, not `draw.fields` / `.items()` / `.at_path()`.
  `draw["x"]` and `draw["outer/a"]` are unchanged.
- A batch has no `mean` / `var` over its batch axis. Reduce the column:
  `jnp.mean(draw["x"], axis=0)`.

Both batch classes are now exported from `probpipe`. A batched draw is what a
joint law hands back, so a caller has to be able to name its type — the other
`*Batch` classes were already public.

## `RecordArray` is untouched

It still exists, is still exported, and behaves exactly as before for direct use.
No producer returns one.

One coupling did need attention: `NumericRecordArray.from_vector` delegated to the
shared flat-vector reconstruction, which now builds a batch — so the classmethod
was returning the other class, and its nested accessors (`at_path`, `mean`,
`var`) disappeared with it. It now renests the batch's flat columns itself. The
flat split belongs on the batch; only the renesting is on the class, and it goes
when the class does.

**Deleting `RecordArray` is a further step, not this one**: 670 references across
the package and its tests. Worth its own PR now that nothing produces one.
`RecordEmpiricalDistribution` (#340) and reparenting `Design` belong with it.

## Testing

Full suite green with every optional backend installed: **4777 passed / 48
skipped / 1 xfailed / 1 xpassed**, no failures and no collection errors.
`uvx ruff check` and `uvx ruff format` clean.

Test changes follow the contract rather than the class — a batched draw is a
`RecordBatch`, its fields are read from `event_template`, a row is gathered by
column. `from_vector`'s level rule gets the coverage it was missing: one name is
one level however many axes the vector carried, several names take one axis each,
and more names than axes still raises.

All 12 user-guide notebooks and both tutorials execute.

Part of #350.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -554,6 +554,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   the executor does, so a zero-width event column passes under explicit
   `dispatch="jax"`.
 
+  **Shape cannot recover axis provenance, so ambiguity refuses.** A removal
+  whose size matches more than one level (``vmap(..., in_axes=1)`` over equal
+  sizes) and a permutation of the batch axes (a transpose, which shape alone
+  cannot tell from a per-axis resize) both raise rather than guess; a
+  distinctly-sized removal now names the level that *survived*, not the
+  leftmost that fits. A pinned dtype is held like the event axes, under the
+  constructor's same-kind rule. And zero-row sweeps take one aggregation path
+  under every dispatch, so the output schema does not depend on how rows would
+  have been executed.
+
   **A same-rank transform cannot lie about sizes.** Slicing a batch's columns
   keeps every axis, so the levels carry over onto the sizes the columns actually
   have; columns left disagreeing about their batch axes are refused rather than
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -536,6 +536,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   produces the same result by the dispatch-equivalence contract, instead of
   failing mid-call.
 
+  **Levels align by name, and only whole.** Operands carrying the same level
+  names zip; operands with no level in common form a product; a level shared by
+  operands whose other levels differ is refused — aligning it would broadcast
+  the rest, which is not built, and a product would read the shared name as two
+  unrelated axes. Operands naming the same levels must also hold them on the
+  same axes: the flat shape can agree while the partition does not. A parameter
+  annotated `Batch` (or a generic alias such as `Batch[Record]`) takes the value
+  whole, since it names a batched container.
+
+  **A transform never resizes the element's own axes.** A per-column slice can
+  pass the rank check while shrinking the event, and a transpose reads an event
+  axis as a batch axis; both now raise, as does a reduction below the event
+  rank. An empty sweep builds its declared fields only when zero rows were
+  *expected* — an empty list where rows were expected is a missing-output error,
+  not a fabrication. The dispatch probe states the flat batch size exactly as
+  the executor does, so a zero-width event column passes under explicit
+  `dispatch="jax"`.
+
   **A same-rank transform cannot lie about sizes.** Slicing a batch's columns
   keeps every axis, so the levels carry over onto the sizes the columns actually
   have; columns left disagreeing about their batch axes are refused rather than
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -514,6 +514,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   follows the leaves — `NumericRecordBatch` when the template is numeric, the
   permissive `RecordBatch` otherwise — but a batched draw is a batch either way.
 
+  **A transform cannot add a level.** `vmap` strips the mapped axis on the way in,
+  which unflattening handles by re-deriving which levels survived; on the way out
+  it *adds* one, and an added axis belongs to no level. Unflattening has no name
+  to give, so it now raises instead of keeping the stored spec — which returned a
+  batch whose `batch_shape` its own columns contradicted, making every method that
+  reads the shape quietly wrong. Map over a batch's columns, or build the batch
+  where the axis is added.
+
+  **A batch is fingerprinted by its levels and its columns.** A multi-field batch
+  failed fingerprinting outright, so provenance omitted it; a single-field one was
+  hashed as its sole column, omitting the schema and the levels. Both are fixed:
+  the spec, the level names and their axis groups, and the raw columns in leaf
+  order all contribute.
+
+  A declared `support` on a batched output is checked column by column. Walking a
+  batch as a named tree found no children and asked a multi-field batch to convert
+  to one array, so two valid columns raised.
+
   `RecordArray` / `NumericRecordArray` still exist and are unchanged for direct
   use — no producer returns one. `NumericRecordArray.from_vector` keeps returning
   a `NumericRecordArray`, renesting the batch's flat columns itself, so the class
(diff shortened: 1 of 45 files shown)


SoymilkWinsAgain (dismissed):
Looks good

PR #404, opened 2026-08-06
refactor!: delete RecordArray, the batch is the batched record

### LINKED ISSUE #340: Nested empirical / nested-prior inference broken: NumericRecordArray not leaf-keyed

jhuggins:
## What this is

`RecordArray` and `NumericRecordArray` are deleted, along with their view class
and root exports. `RecordBatch` / `NumericRecordBatch` are the batched record
forms everywhere: every aggregation a broadcast builds, the empirical and PyMC
data intakes, the parameter-sweep `Design`, the notebooks, and the API docs.

Followed #402, which left no producer returning the deleted classes; this PR
removes them and migrates everything that still consumed or documented them.
#399, #401, and #402 have all merged, so this now targets `main` directly and
is the last PR of that stack.

Closes #340: a column is keyed by leaf path, so a nested record batches, lifts,
and stacks like a flat one — the refusals that issue tracked are gone, and their
tests now pin the working contract.

## What replaces what

- **Broadcast aggregation** builds batches that mint one level per swept group
  (`_stack_declared_columns`, the record/array/opaque stacking paths, and the
  empty declared aggregate), with the swept groups' own axis partitions.
- **`Design` is a `RecordBatch`** carrying a single `design` level; its views
  are plain batches (a view holds none of the marginals, so `Batch._view_type`
  lets a stateful subclass hand views to the class that makes no such claim).
  The sweep zips sibling views **by level name** — the level algebra's rule —
  rather than by parent identity, which a batch view does not have.
- **`EmpiricalDistribution` takes a batch**, routing to
  `RecordEmpiricalDistribution` and peeling it to the leaf-rows form the class
  stores (raw columns, so a non-array field rides as its object column);
  resampling keeps rows paired. The PyMC model builder reads observed variables
  off a batch by raw column.
- **`NumericRecordBatch.stack` converts registered backend leaves through the
  array backend** (one `to_jax` per leaf), which `jnp.asarray` cannot see.
- **The declared marginal names its `draw` level** and carries its partition;
  the missing arguments were previously swallowed by a permissive `except` and
  the marginal silently degraded to the list form.

## Consequences for calling code

- A record's field can hold a batch, but a path does not descend *through* it:
  `drawn["a"]["x"]`, not `drawn["a/x"]` — a batch is a leaf with its own `[]`.
- The mapping surface (`keys` / `len` / `in` / `items`) and `at_path` / `view` /
  `template` / `mean` / `var` are gone with the classes: fields come from
  `event_template`, an interior node indexes to a sub-batch view, columns
  reduce directly (`jnp.mean(batch["x"], axis=0)`), and iteration yields
  leading-axis views.
- `from_vector` requires `level_names`; `stack` requires `level_name`. The
  operation minting a level names it.
- A batch carries no `annotations`; they remain a record/distribution feature.

## Docs

The records notebook's batch section teaches `RecordBatch` /
`NumericRecordBatch` (columns under named levels, level-named stacking, column
reductions, leaf-keyed nesting with sub-batch views); the diagnostics guide and
getting-started tutorial build their data batches on an `obs` level. All 12
user-guide notebooks and both tutorials execute end to end. CONTRIBUTING,
STYLE_GUIDE, and the API pages describe only the surviving contract; the strict
docs build passes. No `RecordArray` reference remains in the repository.

## Testing

Full suite green with every optional backend installed: **4633 passed / 48
skipped / 1 xpassed**, `uvx ruff check` and `uvx ruff format` clean. Tests that
pinned the deleted classes' transitional behaviors now pin the batch contract,
including the flipped #340 refusals and a new empirical-intake pairing test.

Part of #350 (W3.2).


CONTRIBUTING.md
```diff
@@ -533,15 +533,15 @@ uv build packaging/probpipe   # probpipe (metapackage)
    term-result planning reserved for #369.
    Authoritative nested output templates use a private recursive aggregate
    packer across sequential and JAX dispatch; the public
-   `RecordArray.stack` contract remains unchanged. The field name for inferred
+   `RecordBatch.stack` contract remains unchanged. The field name for inferred
    single-field output is always the function's own name. Single-field
-   `NumericRecord` / `NumericRecordArray` / `Record`
+   `NumericRecord` / `NumericRecordBatch` / `Record`
```
SoymilkWinsAgain: This class groups the entire set of capabilities, including `__float__` and `__call__`, under `NumericRecordBatch`. However, this class only provides array conversions and `shape`/`dtype`/`ndim`, and explicitly does not provide scalar conversions. It is recommended to describe the actual shims supported by each type separately.

SoymilkWinsAgain (changes requested):
AI report #404 leaked a debug script, causing CI failures. The `_f.py` file in the root directory is a printable reproduction script containing private module imports and does not belong to the product, test, or documentation. Current lint & formatting failures are: E401, I001, E702; formatting checks also fail. This file should be deleted directly, not just formatted. This issue needs to be fixed before any further re-reviewing.

jhuggins:
Thanks — good catch on the shims. Fixed in 6a5b5b24: `Record`, `NumericRecord`, and `NumericRecordBatch` now get separate lists, with the batch keeping the array conversions (`__array__`, `__jax_array__`, `.shape`, `.dtype`, `.ndim`) and not the scalar ones. That split is deliberate and `test_a_batch_is_never_scalar_like` already pinned it, so this was only the doc overstating.

On `_f.py` — it isn't in this branch's history (`git log -- _f.py` is empty); it was amended out before the branch was pushed, and `lint & format` has been green on `9d8d470f` since. I think the review landed on a force-pushed intermediate.

Issue #405, opened 2026-08-11
Vectorize sweep bodies that return a batch, instead of falling back to sequential dispatch


jhuggins:
### Current behavior

A sweep whose body returns a `Batch` fails the JAX trace probe in `Node._jax_traceability_error` and falls back to sequential dispatch. That is correct — the two dispatch paths agree on results by contract — but it forgoes vectorization for a whole class of bodies.

### Why the fallback exists

When `vmap` stacks the body's results it inserts an output axis and re-invokes `RecordBatch`'s pytree unflatten hook. The newly created axis has no level name, and the pytree contract gives the hook no way to learn one: the hook receives the static aux replayed from the treedef plus the transformed children, and neither `in_axes` nor `out_axes` is part of that. Without a channel, the reconstructed batch cannot state an honest `BatchSpec` for the axis `vmap` just added.

### Proposal

Add a private output adapter to the workflow executor. Instead of returning a `RecordBatch` from the mapped body, return a private pytree payload whose children are the returned batch's raw columns and whose aux carries:

- the returned batch's inner `BatchSpec`;
- the sweep's own level names and axis groups;
- the `out_axes` placement.

Its unflatten hook then builds the final batch explicitly, with levels ordered as the sweep's followed by the body's.

This puts the reconstruction where the metadata already is. `execute_sweep_rows_jax` knows the sweep's levels and axis groups, and already constructs each body argument explicitly — it flattens each batched input to raw leaf columns and rebuilds a `Record` inside the traced call — so the input side is already free of inference. The adapter does the same for the output side.

### Non-goals

- **Not a public API.** A public level-aware mapping method could later wrap this executor, but it is not required, and one that corrected only the input side would leave the output axis unnamed.
- **Not `jax.custom_batching.custom_vmap`**, which customizes one function's batching rule rather than pytree reconstruction, and does not on its own support reverse-mode differentiation.
- **Not JAX's experimental `hijax` mapping specs.** Those could carry level identity through `vmap`, but only by redesigning `RecordBatch` as a custom JAX type with its own tracing, batching, lowering, and autodiff behavior.

### Acceptance

A sweep whose body returns a batch runs under `vmap` and yields a batch whose levels are the sweep's followed by the body's, with `batch_shape` and `batch_size` consistent with its columns, and whose result matches sequential dispatch.

### Depends on

The `RecordBatch` stack — #399 introduces the class, #402 the sweep executor this would extend.

[...]

PR #408, opened 2026-08-12
fix(core): probe each mapping executor under its own map

jhuggins:
## Summary

A `Function` whose body returns a `RecordBatch`, called with a `Distribution`
argument, **crashed** instead of falling back to sequential dispatch:

```python
@function
def body(x: float) -> RecordBatch:
    return RecordBatch.stack(
        [Record("r", {"y": x * k}, name_is_auto=True) for k in (1.0, 2.0, 3.0)],
        level_name="k",
    )

body(Normal(0.0, 1.0, name="x"))
```

```
ValueError: a transform left this RecordBatch's fields with 2 batch axes where
its levels account for 1. An added axis belongs to no level, and unflattening
has no name to give one …
```

The trace probe exists so that a body which traces cleanly bare but cannot
survive the transform its executor applies is caught **while a fallback is still
available**. `Node._jax_traceability_error` modelled the transform for
`execute_sweep_rows_jax`, which maps over batch rows, but not for
`_broadcast_jax`, which also maps — over the draw axis rather than over rows.
With no batched-record argument the probe took its bare branch,
`jax.make_jaxpr(lambda kw: func(**kw))(dummy_kw)`, so the body passed and then
failed inside `jax.vmap`, past the point where anything could fall back.

Both mapping executors are now probed under a map. A plain distribution argument
is traced over a single draw taken from `_sample_broadcast_args` — the
executor's own input construction — rather than synthesized from the law's event
shape. That distinction turned out to matter: a record-valued law draws a *batch
of records*, so mapping it hands the body a record where zeros of the event
shape would hand it an array. A one-field `ProductDistribution` took the JAX
path on a body the executor could not run and crashed inside it, which is the
same failure mode again, and is now covered by a test.

Chasing that down surfaced the underlying defect: **the dispatches disagreed on
what a one-field draw even is.** Such a law draws a batch of records; the
row-wise paths presented each draw as its bare leaf via `_index_sample`, while
the `vmap` path presented the record. The record shim carries conversions and
deliberately no arithmetic, so `x * 2` succeeded under `dispatch="sequential"`
and crashed under the mapped executor.

Design II.4 leaves the choice itself open — it rides on the single-value
coercion question `Record` poses — so the rule was picked by which paths already
made it: `_broadcast_sample` and both arms of `_broadcast_enumerate` unwrap, and
only `_broadcast_jax` did not. Presenting the bare leaf everywhere changes no
behavior that worked, fixes the one that crashed, and keeps the mapped path for
these laws rather than retreating to row-wise. `_present_draw` is that single
rule; `mapped_draw_body` is the executor body the probe traces, so neither the
dispatches nor the probe and its executor are held in agreement by hand.

## Linked issue(s)

Refs #405 — that issue proposes vectorizing these bodies rather than falling
back, and describes the fallback as correct-but-slow. That holds for the sweep
executor; this is the path where it was not correct. Fixing it first means the
fallback #405 replaces is one that actually exists.

## Test plan

`tests/core/test_workflow_distribution_broadcast.py`, new class
`TestTheProbeModelsItsExecutorsTransform` — 13 tests. These fail on `main`
without the source change and pass with it:

- the regression itself: a batch-returning body over a law does not raise;
- the fallback is what ran (asserted via the dispatch log);
- the fallback's result equals explicit `dispatch="sequential"`;
- `dispatch="jax"` reports the dispatch it cannot use rather than surfacing the
  raw pytree error;
- **several distribution arguments** are all mapped, not just the first;
- a **one-field record law** presents the same value to every dispatch — the
  case that used to crash under the mapped executor and succeed row-wise;
- a **law that cannot answer `dtype`** is still probed. This is the one the
  removed preflight actually refused: `getattr(law, "dtype", None)` swallows
  only `AttributeError`, and a `SequentialJointDistribution` view raises
  `NotImplementedError`.

The rest are guards, pinning that the probe gained a transform and a real draw
without becoming a blanket refusal or disturbing neighbouring paths:

- a body that survives the transform still takes JAX dispatch;
- the mapped slice carries the declared event shape — verified by mutation:
  shortening the draw to `(1,)` kills this test and no other;
- an aliased argument still reads as one variable — `f(d, d)` is `X - X`. The
  probe now draws through `_sample_broadcast_args`, so its draws are co-sampled
  the way the executor's are;
- a **multi-field law** vectorizes — a guard, since it answers both
  `event_shape` and `dtype` and so always could;
- an **empirical law** is enumerated rather than mapped, whatever the probe
  would say, so its exact weights are preserved;
- the nested regime (a sweep crossed with a law) still composes to a result.

Full suite: 4612 passed, 54 skipped. `ruff check .` and `ruff format --check .`
clean from the repo root.

## Breaking changes

`dispatch="jax"` now accepts cases it used to reject — views of a
`SequentialJointDistribution` above all, whose mapped result is measured
identical to the row-wise one. That is strictly more permissive, and the two
tests that asserted the refusal and the matching auto-fallback were pinning an
artifact of the removed preflight; both now pin the agreement.

Otherwise none. The change moves calls that previously raised onto the
sequential path,
which the two dispatches agree on by contract.

## Documentation

CHANGELOG entry under `Unreleased → Fixed`. The probe's docstring now states the
invariant it enforces — every executor that maps is probed under a map — and
names the defect it guards, since the correspondence between probe and executor
was previously carried by a prose comment and had drifted.

### DIFF
--- a/probpipe/core/_workflow_distribution_broadcast.py
+++ b/probpipe/core/_workflow_distribution_broadcast.py
@@ -528,15 +528,15 @@ def _stack_rows(rows: list[Any], *, arg_name: str) -> Any:
     return jnp.stack(rows)
 
 
-def _present_draw(drawn: Any) -> Any:
-    """How one draw is presented to the wrapped function.
+def _sole_leaf_path(drawn: Any) -> str | None:
+    """The one leaf path a draw presents as, or ``None`` if it presents whole.
 
-    A single-field draw presents as its bare leaf. Design II.4 leaves the choice
-    open — it rides on the single-value-coercion question ``Record`` poses — so
-    what matters here is that every dispatch make the *same* one: the record
-    shim is deliberately narrow, carrying conversions but no arithmetic, so a
-    body written for the value would work under one dispatch and fail under
-    another if they disagreed. Every path presenting a draw goes through here.
+    The single place the single-field question is decided, so that every path
+    presenting a draw decides it the same way. Design II.4 leaves the choice
+    itself open — it rides on the single-value-coercion question ``Record``
+    poses — but not the agreement: the record shim is deliberately narrow,
+    carrying conversions and no arithmetic, so a body written for the value
+    would work under one dispatch and fail under another if they diverged.
     """
     from ._record_batch import RecordBatch
     from .record import Record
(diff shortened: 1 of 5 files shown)


probpipe/core/_workflow_distribution_broadcast.py
```diff
@@ -530,6 +528,26 @@ def _stack_rows(rows: list[Any], *, arg_name: str) -> Any:
     return jnp.stack(rows)
 
 
+def _present_draw(drawn: Any) -> Any:
```
SoymilkWinsAgain: The documentation, PR body, and CHANGELOG for `_present_draw()` all claim that "all four paths follow the same rule," but it's actually only used by `mapped_draw_body()`; row-wise sampling and enumeration are still determined by repeatedly checking `len(leaf_paths) == 1` within `_index_sample()`.

Current results are consistent, and tests can prevent existing regressions, but the structural guarantee of "sharing a single rule" doesn't exist, and future drift is still possible. However, this doesn't constitute a blockage.

SoymilkWinsAgain (approved):
There's a minor issue, but it's not a blockage problem. You may fix or modify the document and then merge this PR, or just ignore it...

Issue #406, opened 2026-08-11
Record's pytree unflatten threads a stale declaration; refuse raw jax.vmap once the library stops relying on it


jhuggins:
### The defect

`Record`'s pytree unflatten threads the stored declaration over whatever children a transform produced, without checking that they still fit it. `vmap` removes an axis on the way in and adds one on the way out, so the record it hands the body — and the record it rebuilds from the body's return — can report an `event_template` its own leaves contradict.

Measured on `main`, no batch involved:

```python
jax.vmap(body)(NumericRecord("r", x=jnp.arange(3.0)))
#   body sees:  template x=(3,)   leaf shape ()   vector_size 1 vs 3
```

The two errors cancel when the aux going in equals the aux coming out, which is why an identity `vmap` over a plain record looks fine. They do not cancel when the aux is an *element* declaration, so a record-returning body over a batch of records yields:

```python
out.event_template   # x=()
out["x"].shape       # (3,)
out.to_vector()      # (6,) for a two-field batch — ravels the real leaves
from_vector(...)     # ValueError: vec trailing axis is 6, expected vector_size=2
```

`to_vector` reads the leaves and `from_vector` believes the template, so they stop round-tripping. This is not a batch defect: it predates `RecordBatch` and lives in `Record`.

### Why inference is not the answer

Re-deriving the template from the children answers the shape and loses everything else. A stored declaration carries what `EventTemplate.infer_from` cannot recover:

```
stored  : x=ArraySpec(shape=(3,), dtype=float32, support=positive), tag=OpaqueSpec(meta='units')
inferred: x=(3,),                                                   tag=None
```

That loss is the reason a record serializes its spec rather than re-inferring it on load.

### The resolution

Refuse. The operation that knows the output type is the one that should state it, and by the time the pytree hook runs, nothing does:

- A `Function` with a declared output type **already carries it correctly at both ends** — verified: a declared `support=positive` survives a sweep onto the aggregate, and the sweep hands each body row a correctly typed element because it builds rows explicitly from raw columns.
- An undeclared `Function` infers the aggregate from the rows it built, which is the legitimate undeclared case.
- Raw `jax.vmap` with no `Function` and no declaration is the only path where a type has to be invented — so it is refused rather than guessed.

The refusal itself is ~15 lines in `_record_unflatten` / `_numeric_record_unflatten`: compare each field's declared shape against the child that arrived, and raise when they differ.

### Migration required before the refusal can land

[...]

PR #410, opened 2026-08-13
feat(core): vectorize a swept body that returns a batch

### LINKED ISSUE #405: Vectorize sweep bodies that return a batch, instead of falling back to sequential dispatch

jhuggins:
## Summary

A sweep whose body returns a `RecordBatch` now runs under `jax.vmap` instead of
dropping to row-wise dispatch.

`vmap` inserts an output axis and re-enters `RecordBatch`'s unflatten hook, which
has no name to give the new level: a treedef replays the static aux and the
transformed children, and carries neither `in_axes` nor `out_axes`. **A shape is
not a provenance**, so the hook refuses rather than guess — and that refusal stays
exactly as it is.

What changes is that the executor stops routing through it. The body's returned
batch is taken apart into `_MappedBatchColumns`, an inert private carrier whose
own unflatten rebuilds it verbatim and checks nothing, which is what lets it cross
a rank change the batch cannot. The executor rebuilds the batch on the far side,
where it holds the level names the hook lacked.

This is the move the input side already makes: `execute_sweep_rows_jax` flattens
each batched argument to raw leaf columns and rebuilds a `Record` inside the
traced call, so nothing infers a batch axis on the way in. Now the same holds on
the way out.

## Structure

- **The probe and the executor share one body wrapper**
  (`_workflow_sweep.mapped_row_body`) rather than being two functions kept alike
  by a comment. That correspondence is what drifted to produce #408's bug, so it
  is now structural.
- **One reconstruction serves both dispatches.** `_batch_over_swept_columns`
  builds the aggregate: the row-wise path calls it after stacking rows itself,
  the mapped path with the transform's output axis already in place. The two
  agree by construction instead of by parallel maintenance.

## Linked issue(s)

Closes #405.

Stacked on #408, which fixed the case where the same class of body did not fall
back but crashed. This PR targets that branch and will retarget to `main` when it
merges.

## Test plan

`tests/core/test_workflow_sweep.py`, new class `TestASweptBodyThatReturnsABatch`
— 9 tests:

- **The sweep takes the mapped path** — the discriminating one, verified
  before/after by spying on `execute_sweep_rows_jax`: on #408's head it is
  reached 0 times and the log says "not JAX-traceable; using sequential
  dispatch"; here it is reached once, with no fallback.
- **The levels are the sweep's followed by the body's** — `('row', 'k')`, groups
  `((4,), (3,))`.
- **The shape agrees with the columns it holds** — `batch_shape=(4, 3)`,
  `batch_size=12`, column shape `(4, 3)`, since a batch whose spec its own
  columns contradict is the failure to avoid.
- **It matches sequential dispatch** — on element spec, levels, groups, and
  values.
- **A multi-axis level reshapes on the mapped path** — one argument carrying a
  level over two axes (`batch_shape=(2, 3)`), which is what exercises the
  carrier's flatten-to-`prod`-and-restore. It takes a single argument on
  purpose: two array arguments make two zip groups, which sets `jax_supported`
  false, so the earlier two-argument version measured row-wise dispatch and left
  this uncovered. A spy asserts the mapped executor ran.
- **Two zip groups sweep as a product** — `('a', 'b', 'k')` over
  `((2,), (3,), (2,))`. Row-wise by construction, so it pins the aggregation
  rather than the mapped path.
- **A nested element survives**, since columns are leaf-keyed.
- **The carrier does not reach the caller.**
- **A raw `jax.vmap` returning a batch is still refused** — the hook is routed
  around, not softened, which #406 depends on.

Full suite: 4621 passed, 54 skipped. `ruff check .` and `ruff format --check .`
clean from the repo root.

## Breaking changes

None. Calls that previously fell back now vectorize and produce the same result;
the two dispatches agree by contract, and the tests assert that agreement
directly rather than approximating it.

## Documentation

CHANGELOG entry under `Unreleased → Added`. `_MappedBatchColumns` carries a
docstring stating why it is inert and why that is the other half of the hook's
refusal rather than an exception to it; `mapped_row_body` states why both callers
must use it.

## Checklist

- [x] Linked to the primary implementation issue above.

### DIFF
--- a/tests/core/test_workflow_sweep.py
+++ b/tests/core/test_workflow_sweep.py
@@ -12,6 +12,7 @@ import pytest
 from probpipe import (
     BroadcastDistribution,
     DistributionArray,
+    EventTemplate,
     Function,
     Normal,
     NumericRecord,
@@ -341,13 +342,49 @@ class TestASweptBodyThatReturnsABatch:
         assert mapped.axis_groups == sequential.axis_groups
         np.testing.assert_allclose(np.asarray(mapped["y"]), np.asarray(sequential["y"]))
 
-    def test_a_multi_axis_sweep_keeps_the_groups_apart(self):
-        """Two zip groups sweep as a product, so the aggregate carries both sweep
-        levels before the body's own.
+    def test_a_multi_axis_level_reshapes_on_the_mapped_path(self, monkeypatch):
+        """One level spanning two axes, swept under the map.
+
+        The mapped executor flattens the sweep to a single axis of
+        ``prod(batch_shape)`` and the carrier restores its shape afterwards, so
+        a sweep of rank greater than one is what exercises that reshape. It
+        takes a single argument on purpose: two array arguments make two zip
+        groups, which sets ``jax_supported`` false and would quietly measure
+        row-wise dispatch instead.
+        """
+        reached = []
+        real = _workflow_sweep.execute_sweep_rows_jax
+
+        def spy(**kwargs):
+            reached.append(1)
+            return real(**kwargs)
+
+        monkeypatch.setattr(_workflow_sweep, "execute_sweep_rows_jax", spy)
+
+        grid = RecordBatch(
+            {"x": jnp.arange(6.0).reshape(2, 3)},
+            "cell",
+            element_spec=EventTemplate(x=()),
+            axis_groups=((2, 3),),
+        )
+
+        mapped = Function(func=self._body, name="swept")(grid)
+        sequential = Function(func=self._body, name="swept", dispatch="sequential")(grid)
+
+        assert reached == [1]
+        assert mapped.level_names == ("cell", "k")
+        assert mapped.axis_groups == ((2, 3), (3,))
+        assert mapped.batch_shape == (2, 3, 3)
+        assert np.shape(mapped._raw_column("y")) == (2, 3, 3)
+        np.testing.assert_allclose(np.asarray(mapped["y"]), np.asarray(sequential["y"]))
+
+    def test_two_zip_groups_sweep_as_a_product(self):
+        """Two groups product, so the aggregate carries both sweep levels first.
 
         Group structure, not only total size: collapsing the sweep onto one flat
         leading axis agrees on ``batch_size`` and loses which axis belongs to
-        which level.
+        which level. Two array arguments take row-wise dispatch, so this pins
+        the aggregation rather than the mapped path.
         """
 
         def body(p, q):
--- a/tests/core/test_workflow_sweep.py
+++ b/tests/core/test_workflow_sweep.py
@@ -281,8 +281,8 @@ class TestASweptBodyThatReturnsABatch:
     ``vmap`` adds an output axis that ``RecordBatch``'s unflatten hook cannot
     name — *a shape is not a provenance* — so the executor hands the transform
     raw columns and rebuilds the batch itself afterwards, from the levels it
-    swept. The sequential path is the oracle throughout: it always produced the
-    right answer, only slowly.
+    swept. The sequential path is the oracle throughout: it builds the same
+    aggregate one row at a time.
     """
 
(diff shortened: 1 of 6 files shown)


tests/core/test_workflow_sweep.py
```diff
@@ -268,3 +273,140 @@ def distribution_broadcast(
...
+    def test_it_matches_sequential_dispatch(self):
+        mapped = Function(func=self._body, name="swept")(self._rows(4))
+        sequential = Function(func=self._body, name="swept", dispatch="sequential")(self._rows(4))
+
+        assert mapped.element_spec == sequential.element_spec
+        assert mapped.level_names == sequential.level_names
+        assert mapped.axis_groups == sequential.axis_groups
+        np.testing.assert_allclose(np.asarray(mapped["y"]), np.asarray(sequential["y"]))
+
+    def test_a_multi_axis_sweep_keeps_the_groups_apart(self):
```
SoymilkWinsAgain: This test doesn't actually go through the JAX path. It takes two arguments and two array groups as input, and the executor explicitly sets `jax_supported=False` in this case. Therefore, the result named `mapped` actually comes from sequential dispatch and doesn't cover the reshape and level reconstruction of the new carrier on the multi-axis outer sweep.

The AI ​​suggests adding a single-parameter, single-level `RecordBatch` with `batch_shape=(2, 3)` and `axis_groups=((2, 3),)`, and then using `spy execute_sweep_rows_jax` to verify the final levels, shape, columns, and sequential equivalence.

SoymilkWinsAgain (dismissed):
There's a small issue in `test_workflow_sweep.py`, no fundamental issues. It is recommended to fix and then merge

Issue #395, opened 2026-08-01
Enforce tracked-term immutability with a shared Immutable mixin


jhuggins:
## Summary

The design says every tracked term is immutable. Six classes enforce it; fifty-seven do not. Meanwhile the machinery that immutability drags along — pickling an object whose state cannot be assigned back — is implemented five times, in five different ways, one of those copies shipped a silent data-loss bug, and **three of them are dropping annotations right now** (verified below).

This proposes an `Immutable` mixin owning the assignment guard and the state round-trip, **inherited by `TrackedTerm`** so that a tracked term is immutable by construction, and mixed into `EventTemplate`, which is immutable without being a term.

**Revised 2026-08-12**, re-measured on `main` at `7e238916` after the record-batch stack (#396, #397, #399, #401, #402, #404) landed. Against the filed version: the sequencing note was stale and **no prerequisite is left**, the three open decisions have answers, the cost estimate was wrong in two directions — the work is not "56 constructors", but it is more than the six classes the 08-04 revision claimed — and the consolidation now fixes a live bug rather than being behavior-neutral.

## What the design says

Two statements, both load-bearing:

- `design/01-design-principles.md` — **`C2 – Functional interface over immutable objects`**: "operations have no side effects, and objects are never modified in place … any transformation yields a new object rather than mutating an existing one."
- `design/05-operations.md` (V.1, *Identity is attached by derivation*): "**TrackedTerm terms are immutable** and provenance is write-once. The default result is therefore a new term … The implementer's object is never modified."

The operation contract *depends* on this: it promises an implementer's object is never modified, and today nothing stops an operation from modifying one. One already does — see *Writes to another object* below.

## Where the code stands

| | filed 08-01 | `main` 08-12 |
|---|---|---|
| tracked-term leaf classes | 63 | **63** |
| … enforcing immutability | 6 | **6** |
| … not enforcing it, and assigning `self.x = …` somewhere in their own MRO | 56 | **56** |
| … not enforcing it, no plain assignment | 1 | **1** (`NumericRandomMeasure`) |

The ratio has not moved in eleven days — deletions and additions cancelled. What moved is which guard carries the weight: #404 deleted `RecordArray`, `NumericRecordArray`, and `_RecordArrayView`, and `Design` / `FullFactorialDesign` were retyped onto `RecordBatch`, so the *correct* guard now covers seven classes and the hardcoded `"Record"` one covers two.

| Guard defined in | Message | Inherited by |
|---|---|---|
[...]

PR #411, opened 2026-08-13
fix(core): preserve annotations across serialization

### LINKED ISSUE #409: copy and pickle silently drop a term's annotations
## Summary

`copy.copy`, `copy.deepcopy`, and `pickle` silently drop a term's **annotations**. Three classes are affected — `Record`, `NumericRecord`, and `ProductDistribution` — and the cause is the same in each: a hand-written `__reduce__` that lists its state fields by hand and does not list `_annotations`.

Nothing raises. A dropped annotation is indistinguishable from one that was never attached.

## Reproduction

On `main` at `7e238916`:

```python
import copy, pickle
import numpy as np
from probpipe import Record

r = Record("r", {"x": np.array([1.0, 2.0])})
object.__setattr__(r, "_annotations", {"diagnostics": "payload"})   # the documented channel

r.annotations                                  # {'diagnostics': 'payload'}
pickle.loads(pickle.dumps(r)).annotations      # None
copy.copy(r).annotations                       # None
copy.deepcopy(r).annotations                   # None
```

Same on `NumericRecord` and on `ProductDistribution`:

```python
from probpipe import Normal
from probpipe.distributions._product import ProductDistribution

d = ProductDistribution(a=Normal(0.0, 1.0, name="a"), b=Normal(1.0, 2.0, name="b"), name="j")
d._annotations = {"note": "keep me"}    # no guard on this class, so plain assignment works
copy.deepcopy(d).annotations            # None
```

`_annotations` is the *only* field lost — everything else round-trips:

```
state before: ['_annotations', '_jax_cache', '_name', '_name_is_auto', '_provenance', '_spec', '_tree', '_vector_size']
state after : [               '_jax_cache', '_name', '_name_is_auto', '_provenance', '_spec', '_tree', '_vector_size']
LOST        : ['_annotations']
```

## Why it happens

Three `__reduce__` implementations enumerate their state by hand:

| | Threads |
|---|---|
| `probpipe/core/record.py:635` | `_tree`, `_name`, `_name_is_auto`, `_provenance`, `_spec` |
| `probpipe/core/_numeric_record.py:402` | the same five (and deliberately omits `_jax_cache`, a memo) |
| `probpipe/distributions/_product.py:253` | `_components`, `_name`, `_name_is_auto`, `_provenance` |

None lists `_annotations`, and `__reduce__` governs `copy.copy` and `copy.deepcopy` as well as `pickle`, so all three paths lose it.

The omission is systematic rather than careless. `_annotations` is the one field a class never assigns in its constructor: `probpipe/core/tracked.py:338` calls it "the one documented exception to object immutability — the `_annotations` store is designed to be written *after* construction by inference backends, validators, and diagnostic operations." A state list written by reading `__init__` will therefore miss exactly this field, every time.

`Annotated` hosts **without** a hand-written `__reduce__` are unaffected — default pickling reports every assigned attribute, so `ApproximateDistribution` (whose `make_posterior` is the one in-tree writer that assigns the store, at `probpipe/inference/_approximate_distribution.py:503`) round-trips correctly. The bug tracks the hand-written lists, not the annotation writers.

## Where it bites

The library never copies or pickles a tracked term itself — `_workflow_result._copy_result_term` goes through `TrackedTerm._shallow_copy`, which delegates to `object.__getstate__` and preserves annotations correctly. So nothing in-tree catches this, and the loss surfaces at the user's boundary:

- a user calling `copy.deepcopy` or `pickle.dump` on an annotated record,
- result persistence and process-boundary transfer under orchestration (cloudpickle),
- caching layers that serialize results — #364 makes this routine.

Diagnostics store ArviZ-compatible payloads under `posterior._annotations["arviz"]`, and `probpipe/core/_distribution_base.py:191` documents reading them back, so the channel carries real data and not just user notes. In-place writes to an existing store (`term.annotations["arviz"] = …`) are unaffected — they are subscript assignments and do not go through the round-trip at all.

## Fix

Minimal: thread `_annotations` through the three `__reduce__` / `_unpickle_*` pairs, keeping `NumericRecord`'s deliberate exclusion of `_jax_cache` (a memo, rebuilt on demand).

Structural: #395 proposes a shared state round-trip delegating to `object.__getstate__`, which reports every assigned attribute — annotations included — and so fixes this **by construction**, along with the same class of omission for any field added later. This is the same failure mode #386's review found in `Batch.__getstate__`, where the fix was exactly that delegation.

The minimal fix is worth landing on its own so the bug closes with a regression test that does not depend on #395; #395 then deletes the hand-written lists rather than maintaining them.

## Test

Two tests, one narrow and one that catches the next field:

1. For every `Annotated` host, an attached annotation survives `pickle`, `copy.copy`, and `copy.deepcopy`.
2. A state-diff test: the keys reported by `object.__getstate__` before a round-trip equal those after, minus a class's declared transient state (`_jax_cache`). This is what would have caught the omission when `Annotated` was introduced, and it catches the next one too.

breezeFur:
## Summary

Preserve annotations when `Record`, `NumericRecord`, and `ProductDistribution` instances are copied or pickled. Their custom `__reduce__` implementations now include the post-construction annotation store, and the reconstruction helpers restore it while retaining defaulted parameters for older pickle payloads.

## Linked issue(s)

Closes #409

## Test plan

- `python -m pytest -p no:xdist -o "addopts=" tests/core/test_record_serialization.py -q` — 36 passed
- `python -m pytest -p no:xdist -o "addopts=" tests/core/test_record.py tests/core/test_numeric_record.py tests/core/test_tracked.py tests/distributions/test_product.py -q` — 390 passed
- `ruff check .` — passed
- `ruff format --check probpipe/core/record.py probpipe/core/_numeric_record.py probpipe/distributions/_product.py tests/core/test_record_serialization.py` — passed

`ruff format --check .` still reports nine pre-existing Markdown/design files outside this change; none of the modified Python files are among them.

## Breaking changes

None.

## Documentation

- [ ] Docstrings updated for changed public APIs

## Checklist

- [ ] At least one `area:*` label applied (the repository labeler applies these from changed paths)
- [x] Linked to an issue above — or N/A for a small standalone fix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Copying and pickling records and product distributions now preserve annotations (#409).**
+  `Record`, `NumericRecord`, and `ProductDistribution` previously omitted their
+  post-construction annotation store from their custom serialization state, so
+  `copy`, `deepcopy`, and pickle silently discarded diagnostic and backend metadata.
+
 - **`is_concrete` no longer reports a polymorphic template as concrete (#390).**
   A symbolic dimension declared inside a term spec — a `RecordSpec`'s schema, a
   `DistributionSpec`'s event declaration, a `FunctionSpec`'s either side — was
--- a/probpipe/core/_numeric_record.py
+++ b/probpipe/core/_numeric_record.py
@@ -414,6 +414,7 @@ class NumericRecord(Record):
                 self._name_is_auto,
                 self._provenance,
                 self._spec,
+                self.annotations,
             ),
         )
 
@@ -600,14 +601,17 @@ def _reconstruct_from_vector(
 
 
 def _unpickle_numeric_record(
-    store: dict, name: str, name_is_auto: bool, provenance, spec=None
+    store: dict, name: str, name_is_auto: bool, provenance, spec=None, annotations=None
 ) -> NumericRecord:
     # ``store`` holds the native leaves verbatim (they pickle themselves), so
     # reconstruction is ordinary validation-without-conversion. The threaded
     # declaration preserves an explicit schema across the round-trip; a pickle
     # written before it was serialized still loads.
     nr = NumericRecord(name, store, event_template=spec)
-    return nr._restore_identity(name_is_auto=name_is_auto, provenance=provenance)
+    nr._restore_identity(name_is_auto=name_is_auto, provenance=provenance)
+    if annotations is not None:
+        object.__setattr__(nr, "_annotations", annotations)
+    return nr
 
 
 # ---------------------------------------------------------------------------
--- a/probpipe/core/record.py
+++ b/probpipe/core/record.py
@@ -645,6 +645,7 @@ class Record(NamedTree[Any], TrackedTerm, Annotated):
                 self._name_is_auto,
                 self._provenance,
                 self._spec,
+                self.annotations,
             ),
         )
 
@@ -1245,11 +1246,16 @@ def _pack_fields(
 # ---------------------------------------------------------------------------
 
 
-def _unpickle_record(store: dict, name: str, name_is_auto: bool, provenance, spec=None) -> Record:
+def _unpickle_record(
+    store: dict, name: str, name_is_auto: bool, provenance, spec=None, annotations=None
+) -> Record:
     # ``spec`` is optional, and construction accepts either form, so a pickle
     # written before the declaration was serialized still loads.
     r = Record(name, store, event_template=spec)
-    return r._restore_identity(name_is_auto=name_is_auto, provenance=provenance)
+    r._restore_identity(name_is_auto=name_is_auto, provenance=provenance)
+    if annotations is not None:
+        object.__setattr__(r, "_annotations", annotations)
+    return r
 
 
 # ---------------------------------------------------------------------------
--- a/probpipe/distributions/_product.py
+++ b/probpipe/distributions/_product.py
@@ -253,7 +253,13 @@ class ProductDistribution(
     def __reduce__(self):
         return (
             _unpickle_product_distribution,
-            (dict(self._components), self._name, self._name_is_auto, self._provenance),
+            (
+                dict(self._components),
+                self._name,
+                self._name_is_auto,
+                self._provenance,
+                self.annotations,
+            ),
         )
 
     # -- Sampling (returns Record) ------------------------------------------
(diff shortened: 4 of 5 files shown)


jhuggins:
Heads up: I had a fix for #409 in progress when this went up, and it is now #412. Three differences:

**1. `copy.copy` shares the annotations container.** `__reduce__` args are passed to the reconstructor without copying, so assigning the container straight through leaves the copy and the original pointing at one dict. Simulating this PR's restore on a plain `Record`:

```
copy.copy shares the container: True
write on copy shows on original: True
deepcopy shares the container: False
```

The channel is written in place by design (`Annotated` documents it as append-only), so that write is observable on the original — a copy is not supposed to be a second handle on the same annotations. `with_name` already decouples the container for exactly this reason (`tracked.py`, in the `_shallow_copy` tail): entries are shared, the container is not.

**2. `_product.py` assigns the attribute directly.** `p._annotations = annotations`, where the other two helpers use `object.__setattr__`. That works only because `ProductDistribution` carries no immutability guard today; it raises the moment one lands on `TrackedTerm`, which is #395. Worth making it `object.__setattr__` regardless, so the three read the same way.

**3. The restore is repeated three times.** All three helpers already call `_restore_identity`, so the `if annotations is not None: ...` block can live there once, next to the `name_is_auto` and provenance restores it already does. This is a preference rather than a defect — but the specific duplication #395 is about is "the same state question answered separately in each class", so three more copies of it is a small step the wrong way.

Nothing above touches the core of what this PR gets right: the three classes that needed threading are exactly the three `Annotated` hosts with a hand-written `__reduce__`, and defaulting the new parameter keeps older pickles loading.

jhuggins:
Closing in favor of #412, which fixes the same bug — but taking the better half of this PR with it.

**Adopted from here:** the test design. One fixture parameterized over `Record`, `NumericRecord`, and `ProductDistribution`, driving each operation, is tighter than what #412 had — the record family covered in one file and a near-copy of the same assertions for the product distribution in another. #412 now uses this shape, so its extra coverage (cloudpickle, container decoupling, the unannotated case, the older-pickle argument list) applies to all three classes rather than two. Credited with a `Co-authored-by` on the commit that made the change.

**Why #412 rather than this one**, restating the three points from the review above: the restore lives once in `_restore_identity`, which all three helpers already call, instead of being repeated at each of them; the restored container is decoupled, so a write on a `copy.copy` does not show through on the original; and `_product.py` writes through `object.__setattr__` like its siblings rather than assigning the attribute directly.

Nothing here was wrong about the diagnosis — the three classes needing the thread are exactly the three `Annotated` hosts with a hand-written `__reduce__`, and defaulting the new parameter to keep older pickles loading is the right call, which #412 keeps. Thanks for catching this quickly after the issue went up.

PR #412, opened 2026-08-13
fix(core): carry annotations through copy and pickle

### LINKED ISSUE #409: copy and pickle silently drop a term's annotations

jhuggins:
## Summary

`Record`, `NumericRecord`, and `ProductDistribution` each reconstruct through a
`__reduce__` that lists its state by hand, and none of them listed
`_annotations`, so a copied or unpickled term came back with its annotations gone
and nothing raised. `__reduce__` governs `copy.copy` and `copy.deepcopy` as well
as `pickle`, so all three paths lost them.

The omission is systematic rather than careless: annotations are the one field
written *after* construction — `tracked.py` calls it "the one documented
exception to object immutability" — so a state list assembled from constructor
arguments misses exactly this one, every time.

The three now thread the store, and **the constructors accept it**: `Record`,
`NumericRecord`, and `Distribution` take private `_provenance` and `_annotations`
arguments (`ProductDistribution` a `_name_is_auto` too, since it derives that flag
rather than accepting it), following the `_validate_leaves` precedent for a
reconstruction-only keyword. A rebuilt term is complete when its constructor
returns.

That let `TrackedTerm._restore_identity` be **deleted**. It wrote identity onto an
already-constructed object through `object.__setattr__`, bypassing both the
immutability guard and the write-once provenance rule for any caller who found it
— written for reconstruction, but scoped to nothing. Reconstruction now has no
privileged write path, and the two JAX unflatten paths pass `name_is_auto` to the
constructor, which already accepted it.

`Annotated._init_annotations` is the counterpart to `TrackedTerm._init_tracked`,
so each mixin still initializes its own state. The container it is handed is
**decoupled** from the one it was rebuilt from, as `with_name` already does: entries are shared, the container is not, so a write on
a copy does not show through on the original. That matters because the channel is
written in place by design — the shared-container case is observable, and the
copy moves into a `_decoupled_annotations` helper the two paths share.

Pickles written before this still load: the new argument is optional, following
the `spec` precedent already there for the same reason. Annotations still do not
cross a JAX transform boundary — `tree_unflatten` rebuilds a bare term,
unchanged, and that is untouched here.

## Relationship to #411

**#411 fixes the same bug and was opened first, and its test design is better than
what this branch started with** — one fixture parameterized over the three classes,
rather than the record family in one file and a near-copy of the same assertions
for `ProductDistribution` in another. That shape is adopted here, with credit on
the commit, and the extra coverage this branch carries now applies to all three
classes instead of two.

What remains are three differences in the fix itself, each verified rather than
predicted:

1. **One restore site rather than three.** #411 repeats
   `if annotations is not None: object.__setattr__(...)` at each of the three
   unpickle helpers. Here it goes into `_restore_identity`, which all three
   already call. The repetition is the shape #395 exists to remove, so landing it
   means the mixin change there has three more copies to undo.
2. **`copy.copy` shares the annotations container in #411.** Simulating its
   restore on a plain `Record`:

   ```
   copy.copy shares the container: True
   write on copy shows on original: True     # deepcopy is unaffected
   ```

   Since writers add entries in place, that write is observable on the original.
   Decoupling the container is what `with_name` already does, for this reason.
3. **#411 assigns `_annotations` directly in `_product.py`**
   (`p._annotations = annotations`) while using `object.__setattr__` in the other
   two. It works only because `ProductDistribution` carries no immutability
   guard today, and breaks as soon as one lands on `TrackedTerm` (#395).

## Linked issue(s)

Closes #409.

Refs #395 — the structural fix there (a shared state round-trip delegating to
`object.__getstate__`) makes this class of omission impossible rather than
patching one instance of it, and deletes these hand-written lists. This PR is the
narrow fix so the bug closes with a regression test that does not depend on that
work.

## Test plan

`TestRoundTripPreservesAnnotations` in `tests/core/test_record_serialization.py`,
parameterized over `Record`, `NumericRecord`, and `ProductDistribution`:

- annotations survive `pickle`, `cloudpickle`, `copy.copy`, and `copy.deepcopy`;
- the container is decoupled — a write on the copy does not reach the original;
- an unannotated term still round-trips to `annotations is None`;
- **no state is lost at all** — the keys `object.__getstate__` reports before the
  round-trip equal those after, minus the documented `_jax_cache` memo. This is
  the general form of the bug, and it is the test that would have caught it when
  `Annotated` was introduced;
- reconstruction stays callable with the shorter argument list an older pickle
  carries;
- **the reconstruction has the same type** — added after the state-key test caught
  a class-selection regression this change introduced and the targeted tests
  missed. `ProductDistribution.__new__` has its own signature, so the new private
  keywords sat in its `**components` and reached the probe that picks the
  TFP-backed subclass; a reconstruction came back on the plain class, carrying the
  right state without `_tfp_dist`. `__new__` now binds them, and
  `TFPProductDistribution.__init__` forwards its arguments wholesale rather than
  restating a signature that can fall behind its base.

**Negative control:** restoring the pre-fix `__reduce__` makes all three
assertions fail (`pickle` / `copy` / `deepcopy` → `None`), so the tests are not
vacuous.

Runs: `tests/core` + `tests/distributions` 3418 passed, 4 skipped;
`tests/diagnostics` clean. `ruff format` and `ruff check` clean on the changed
paths. Pyright not run.

## Breaking changes

None. `copy.copy` now yields an independent annotations container where it
previously yielded a term with no annotations at all.

## Documentation

- [x] CHANGELOG entry added under `Unreleased → Fixed`

## Contract checklist

- [x] Contract of every touched abstraction was clear before coding.
- [x] Changed contracts documented in docstrings: `_restore_identity` is gone, and the reconstruction keywords are documented where they are consumed.
- [x] Code verified to obey the documented contract; tests assert it, including the type a reconstruction lands on.
- [x] `CONTRACTS.md` needs no index change — no cross-cutting contract moved.

## Checklist

- [x] Linked to the issue above.

### DIFF
--- a/probpipe/distributions/_product.py
+++ b/probpipe/distributions/_product.py
@@ -251,9 +251,17 @@ class ProductDistribution(
         self._event_template = _build_event_template(self._components)
 
     def __reduce__(self):
+        # Annotations are threaded explicitly: they are written after
+        # construction, so rebuilding from the components alone would drop them.
         return (
             _unpickle_product_distribution,
-            (dict(self._components), self._name, self._name_is_auto, self._provenance),
+            (
+                dict(self._components),
+                self._name,
+                self._name_is_auto,
+                self._provenance,
+                getattr(self, "_annotations", None),
+            ),
         )
 
     # -- Sampling (returns Record) ------------------------------------------
@@ -453,10 +461,16 @@ class ProductDistribution(
         return f"ProductDistribution({comp_str}{name_str})"
 
 
-def _unpickle_product_distribution(components, name, name_is_auto, provenance):
-    """Reconstruct a ProductDistribution (or dynamic subclass) from its components."""
+def _unpickle_product_distribution(components, name, name_is_auto, provenance, annotations=None):
+    """Reconstruct a ProductDistribution (or dynamic subclass) from its components.
+
+    ``annotations`` is optional so a pickle written before they were serialized
+    still loads.
+    """
     p = ProductDistribution(**components, name=name)
-    return p._restore_identity(name_is_auto=name_is_auto, provenance=provenance)
+    return p._restore_identity(
+        name_is_auto=name_is_auto, provenance=provenance, annotations=annotations
+    )
 
 
 # ---------------------------------------------------------------------------
--- a/probpipe/distributions/_product.py
+++ b/probpipe/distributions/_product.py
@@ -12,6 +12,7 @@ Provides:
 
 from __future__ import annotations
 
+from collections.abc import Mapping
 from types import MappingProxyType
 from typing import Any
 
@@ -217,14 +218,33 @@ class ProductDistribution(
     _sampling_cost = "low"
     _preferred_orchestration = None
 
-    def __new__(cls, *positional, name: str | None = None, **components):
+    def __new__(
+        cls,
+        *positional,
+        name: str | None = None,
+        _name_is_auto: bool | None = None,
+        _provenance: Provenance | None = None,
+        _annotations: Mapping[str, Any] | None = None,
+        **components,
+    ):
+        # ``__new__`` binds the same keywords as ``__init__``, reconstruction's
+        # included: anything left in ``**components`` is read as a component, and
+        # a stray one changes which class the probe below selects.
         components = _merge_positional_and_keyword(positional, components)
         if not components:
             return object.__new__(cls)
         actual_cls = _product_class_for_components(components)
         return object.__new__(actual_cls)
 
-    def __init__(self, *positional, name: str | None = None, **components):
+    def __init__(
+        self,
+        *positional,
+        name: str | None = None,
+        _name_is_auto: bool | None = None,
+        _provenance: Provenance | None = None,
+        _annotations: Mapping[str, Any] | None = None,
+        **components,
+    ):
         components = _merge_positional_and_keyword(positional, components)
         if not components:
             raise ValueError("ProductDistribution requires at least one component.")
@@ -247,7 +267,18 @@ class ProductDistribution(
                 )
         self._components = resolved
         name, name_is_auto = auto_name(name, "product(" + ",".join(resolved.keys()) + ")")
-        super().__init__(name=name, name_is_auto=name_is_auto)
+        # A reconstruction passes the stored flag: it always supplies *name*, so
+        # the derivation above would call a once-auto-derived name user-given.
+        # The three underscore-prefixed keywords are the reconstruction's only,
+        # and are documented on ``Distribution.__init__``.
+        if _name_is_auto is not None:
+            name_is_auto = _name_is_auto
+        super().__init__(
+            name=name,
+            name_is_auto=name_is_auto,
+            _provenance=_provenance,
+            _annotations=_annotations,
+        )
         self._event_template = _build_event_template(self._components)
 
     def __reduce__(self):
@@ -467,9 +498,12 @@ def _unpickle_product_distribution(components, name, name_is_auto, provenance, a
     ``annotations`` is optional so a pickle written before they were serialized
     still loads.
     """
-    p = ProductDistribution(**components, name=name)
-    return p._restore_identity(
-        name_is_auto=name_is_auto, provenance=provenance, annotations=annotations
+    return ProductDistribution(
+        **components,
+        name=name,
+        _name_is_auto=name_is_auto,
+        _provenance=provenance,
+        _annotations=annotations,
     )
 
 
@@ -490,8 +524,11 @@ class TFPProductDistribution(ProductDistribution):
     for interop with SBI and other TFP-dependent subsystems.
     """
 
-    def __init__(self, *positional, name: str | None = None, **components):
-        super().__init__(*positional, name=name, **components)
+    def __init__(self, *positional, **kwargs):
+        # Forwarded wholesale rather than restated: the base's keywords include
+        # the reconstruction ones, and a signature repeated here would silently
+        # swallow any it missed into ``**components``.
+        super().__init__(*positional, **kwargs)
         self._build_tfp_dist()
 
     def _build_tfp_dist(self):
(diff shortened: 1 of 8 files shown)


probpipe/distributions/_product.py
```diff
@@ -217,14 +218,33 @@ class ProductDistribution(
...
             return object.__new__(cls)
         actual_cls = _product_class_for_components(components)
         return object.__new__(actual_cls)
 
-    def __init__(self, *positional, name: str | None = None, **components):
+    def __init__(
+        self,
+        *positional,
+        name: str | None = None,
+        _name_is_auto: bool | None = None,
```
SoymilkWinsAgain, line 243: Now this and `__new__` silently consumes 3 previously valid component names, including `_name_is__autp`, `_provenance`, `_annotations`. Consequently, sth such as `ProductDistribution(_annotations=Normal(..., name="_annotations"))` now fails although the positional equivalent still works. No convention prohibits underscore-prefixed field names, yet the PR declares no breaking changes.

SoymilkWinsAgain (dismissed):

probpipe/distributions/_product.py
jhuggins: I'm choosing not to worry about this since the product distribution will get removed in the distribution overhaul

SoymilkWinsAgain: OK, then there's nothing more

SoymilkWinsAgain (approved):
AI suggests modifying the PR body: The combined CHANGELOG does not describe the final state. It contains duplicate Fixed/Changed sections and says both that old pickles do not load and that they still load. The #412 entry also says Record/NumericRecord constructors accept reconstruction keywords that #413 subsequently removes. Consolidate the entries by category and rewrite them against the top-of-stack behavior.

PR #414, opened 2026-08-13
fix(core): reading a distribution no longer modifies it

jhuggins:
## Summary

> **What a memo container is, and what it does not buy.** Each of these classes
> now holds a plain `dict` named `_memo`, assigned once by the constructor, and a
> lazy read puts its result in it under a string key — `_memo["marginal"]`,
> `_memo["components"]`, `_memo["concatenated"]` — where it used to assign an
> attribute. `NumericRecord._jax_cache` is the same pattern already in the tree.
>
> This makes the terms **pass an assignment guard**; it does not make them
> immutable. The dict's contents still change after construction, so a term's
> internal state at the second read differs from the first. It is unobservable
> through the public interface — the memoized value is deterministic, so a caller
> cannot distinguish it from recomputation — but the guard is not what prevents
> mutation here, it simply does not apply. The alternative is computing at
> construction and dropping the memo, paying the cost up front; that is the choice
> every distribution faces when the guard reaches them, and it is deliberately not
> made here, since these classes are slated to change.

Two distributions changed themselves when read. `BroadcastDistribution` assigned its marginal on the first `marginalize()`, and a backend-delegated `DistributionArray` assigned its components on the first read — so a query modified an object its caller was holding, which is what `C2` and the §V.1 operation contract rule out. Each now fills a memo container assigned at construction: computed once as before, lazily as before, with the term's own fields left as they were built.

Five more sites populated a term *after* allocating it, using plain attribute assignment:

| Site | What it builds |
|---|---|
| `DistributionArray._from_backend` | five fields on an `object.__new__`'d instance |
| `_make_distribution_array` | the declared `event_template` |
| `TFPProductDistribution._build_tfp_dist` | the combined TFP view, from `__init__` |
| `make_posterior` | the posterior's annotations store |
| `SequentialJointDistribution._condition_on_impl` | ten fields on the conditioned result |

**None of that was observable** — each wrote before the object reached a caller, so this half is not a behavior fix. It matters for two other reasons: an assignment guard on every tracked term refuses the plain form, and a reader cannot tell construction from mutation when both look identical. They write through `object.__setattr__` now, the way a constructor does. `ApproximateDistribution`'s chain concatenation joins the same memo container, though its constructor already forced it.

## A copy does not inherit the memo

Caught in review: `_shallow_copy` carried the memo dictionary itself, so `with_name` handed the rename and the original **one** memo. `marginalize` stamps the distribution's own provenance onto the marginal it builds, so whichever asked second silently received the other's lineage — decided by query order, and wrong in the rename's direction.

The memo is declared transient, and the three copy paths now agree on what that means: `_shallow_copy` reads `_transient_state` and `_decoupled_state` like the state round-trip does, so a rename, a `copy.copy`, and an unpickle produce the same thing. The declarations moved to a module-level `declared_state_names`, since they describe how a class *copies* whether or not it also refuses assignment — and only some hosts refuse it so far.

A term arriving by any of those routes therefore has no memo, so reading one rebuilds rather than assumes: `transient_memo` does that, next to the declaration that makes it necessary. This is the same correction the conversion cache needed in #413 when it was first declared transient.

All four copy paths agree, and `_shallow_copy` gets *shorter* for it — it delegates to the round-trip rather than walking attributes itself.

An earlier revision left `copy.copy` sharing the memo and argued it was harmless, on the grounds that a plain copy has its original's provenance. That was wrong: `with_provenance` attaches to the copy afterwards, so the two can differ and the shared memo hands one of them the other's lineage. Reproduced, then fixed by the ordering above.

## Linked issue(s)

Part of #395. **Stacked on #416** (the last of the stack, now that #412 and #413 have merged), which makes `TrackedTerm` inherit `Immutable`, and that order is load-bearing rather than incidental: the memo below is declared *transient*, and only a term that goes through the mixin's state round-trip honours that declaration. Reviewed in the other order, this PR's own head shared a memo across `copy`, `deepcopy`, and `pickle` while honouring the declaration in `with_name` — correct only once #416 landed on top of it. Reversed, each PR is correct where it sits.

## What this deliberately leaves alone

**`PyABCDistribution.rvs` advancing `self._key`.** #395 listed this as a tracked term whose query advances its own RNG state — the plainest violation on the list. It is not a ProbPipe term at all: it subclasses `pyabc.Distribution` to implement that library's stateful RV protocol, where `rvs()` is *required* to return a fresh draw per call. Checked directly rather than inferred from the name this time; the issue's table is corrected.

**`ProductDistribution.__reduce__`.** It is a distribution, not one of the four immutable hosts, so it keeps reconstructing through its constructor until the guard reaches distributions.

## Test plan

`tests/core/test_terms_do_not_mutate.py` — 5 tests. Each compares a term's attributes **by identity** before and after a read, excluding the memo container: a memo filled in place passes, an attribute assignment does not.

- `BroadcastDistribution.marginalize()` and the backend-delegated `DistributionArray.components` leave the term unchanged, and still return the same object on a second read;
- `TFPProductDistribution` has its combined view before any read;
- `ApproximateDistribution` concatenates during construction;
- conditioning a `SequentialJointDistribution` twice leaves the first result untouched, and the operand keeps both components;
- **a rename does not share the original's memo**, and the marginal each produces carries its own lineage in either query order.

**Negative control:** two of the five fail on #413's head, which are exactly the two observable sites. The other three pin invariants that were already true — kept because the difference is invisible from outside and easy to lose, and labelled as such in the module docstring rather than dressed up as regressions.

Two existing tests in `test_distribution_array.py` asserted on the slot that now holds a mode flag rather than the materialised tuple; they follow the memo instead.

4731 passed, 53 skipped. `ruff format` / `ruff check` clean.

## Breaking changes

None.

## Documentation

- [x] CHANGELOG entries added under `Unreleased → Fixed` and `→ Changed`

## Contract checklist

- [x] Contract of every touched abstraction was clear before coding; the ambiguity found — whether these sites were observable mutations or construction in disguise — is resolved above, per site, by negative control rather than by reading the code.
- [x] Changed contracts documented where they live: each memo says it is filled on first read and that reading does not modify the term.
- [x] Code verified to obey the documented contract; the tests assert the property (attributes unchanged across a read), not a happy-path value.
- [x] Docstrings describe the target contract — the memo containers are written as the pattern that survives the guard, not as a workaround for its absence.
- [x] `CONTRACTS.md` needs no index change; no cross-cutting contract moved.

## Checklist

### DIFF
--- a/probpipe/core/_broadcast_distributions.py
+++ b/probpipe/core/_broadcast_distributions.py
@@ -1215,9 +1215,9 @@ class BroadcastDistribution(Distribution[dict], SupportsSampling):
         name, name_is_auto = auto_name(name, "broadcast")
         super().__init__(name=name, name_is_auto=name_is_auto)
         self._approximate = True
-        # A memo, filled on first read. It is a container assigned once here
-        # rather than an attribute assigned later, so reading it does not modify
-        # the distribution.
+        # A memo, filled on first read. Reading fills it in place, which leaves
+        # the term's own attributes as construction set them — what the
+        # immutability guard sees, and what a copy drops rather than inherits.
         self._memo: dict[str, MarginalizedBroadcastDistribution] = {}
 
     # -- basic properties ---------------------------------------------------
--- a/probpipe/core/_broadcast_distributions.py
+++ b/probpipe/core/_broadcast_distributions.py
@@ -1210,7 +1210,10 @@ class BroadcastDistribution(Distribution[dict], SupportsSampling):
         name, name_is_auto = auto_name(name, "broadcast")
         super().__init__(name=name, name_is_auto=name_is_auto)
         self._approximate = True
-        self._marginal_cache: MarginalizedBroadcastDistribution | None = None
+        # A memo, filled on first read. It is a container assigned once here
+        # rather than an attribute assigned later, so reading it does not modify
+        # the distribution.
+        self._memo: dict[str, MarginalizedBroadcastDistribution] = {}
 
     # -- basic properties ---------------------------------------------------
 
@@ -1280,16 +1283,18 @@ class BroadcastDistribution(Distribution[dict], SupportsSampling):
         so the lineage is preserved without a direct reference to the
         ``BroadcastDistribution``.
         """
-        if self._marginal_cache is None:
-            self._marginal_cache = _make_marginal(
+        marginal = self._memo.get("marginal")
+        if marginal is None:
+            marginal = _make_marginal(
                 self._output_samples,
                 self._w,
                 output_distributions=self._output_distributions,
                 event_template=self._output_template,
             )
-            if self.provenance is not None and isinstance(self._marginal_cache, Distribution):
-                self._marginal_cache.with_provenance(self.provenance)
-        return self._marginal_cache
+            if self.provenance is not None and isinstance(marginal, Distribution):
+                marginal.with_provenance(self.provenance)
+            self._memo["marginal"] = marginal
+        return marginal
 
     @property
     def output(self) -> MarginalizedBroadcastDistribution:
--- a/probpipe/core/_broadcast_distributions.py
+++ b/probpipe/core/_broadcast_distributions.py
@@ -26,7 +26,7 @@ from ._empirical import (
     EmpiricalDistribution,
     RecordEmpiricalDistribution,
 )
-from ._immutable import constructing
+from ._immutable import constructing, transient_memo
 from ._numeric_record_batch import NumericRecordBatch
 from ._object_batch import _from_iterable, _is_object_array
 from ._record_batch import RecordBatch, _batch_class_for, _MappedBatchColumns
@@ -1183,6 +1183,11 @@ class BroadcastDistribution(Distribution[dict], SupportsSampling):
         Distribution name for provenance.
     """
 
+    #: The memo is not state: a copy recomputes rather than inheriting one. It
+    #: matters for more than size here, since a memoised value can carry the
+    #: provenance of the term that computed it.
+    _transient_state = ("_memo",)
+
     _sampling_cost: str = "low"
     _preferred_orchestration: str | None = None
 
@@ -1283,7 +1288,7 @@ class BroadcastDistribution(Distribution[dict], SupportsSampling):
         so the lineage is preserved without a direct reference to the
         ``BroadcastDistribution``.
         """
-        marginal = self._memo.get("marginal")
+        marginal = transient_memo(self).get("marginal")
         if marginal is None:
             marginal = _make_marginal(
                 self._output_samples,
@@ -1293,7 +1298,7 @@ class BroadcastDistribution(Distribution[dict], SupportsSampling):
             )
             if self.provenance is not None and isinstance(marginal, Distribution):
                 marginal.with_provenance(self.provenance)
-            self._memo["marginal"] = marginal
+            transient_memo(self)["marginal"] = marginal
         return marginal
 
     @property
--- a/probpipe/core/_distribution_array.py
+++ b/probpipe/core/_distribution_array.py
@@ -130,8 +130,10 @@ class DistributionArray[T](Distribution[T]):
     #: provenance of the term that computed it.
     _transient_state = ("_memo",)
 
-    # Storage slots. ``_components`` is ``None`` for backend-delegated
-    # arrays until :attr:`components` materialises the eager tuple
+    # Storage slots. ``_components`` holds the literal tuple an eagerly built
+    # array was given, and stays ``None`` for a backend-delegated one — which is
+    # what marks it as backend-delegated. :attr:`components` materialises that
+    # array's cells on first read into the memo rather than into this slot
     # lazily; the literal-array constructor sets it directly.
     # ``_backend`` is ``None`` for the literal path and set by
     # :meth:`_from_backend`.
--- a/probpipe/core/_distribution_array.py
+++ b/probpipe/core/_distribution_array.py
@@ -40,6 +40,7 @@ to pull out the i-th element of a **batch** → ``DistributionArray``.
 
 from __future__ import annotations
 
+from functools import partial
 from math import prod
 from typing import TYPE_CHECKING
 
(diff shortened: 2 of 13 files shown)


SoymilkWinsAgain (commented):
Note that `STYLE_GUIDE.md` and `CONTRIBUTING.md` still state that distributions are immutable and annotations are the only exception. The implementation explicitly permits arbitrary assignment on every Distribution. Neither convention file changes anywhere in the stack.

tests/core/test_terms_do_not_mutate.py
```diff
@@ -0,0 +1,161 @@
...
+from probpipe.core._broadcast_distributions import BroadcastDistribution
+
+
+def assigned_attributes(term) -> dict:
+    """The term's attributes, as a name -> id map; memo containers excluded.
+
+    Identity rather than value: a lazily filled memo mutates its container in
+    place, which is invisible here, while assigning any attribute is not.
+    """
+    state = object.__getstate__(term)
```
SoymilkWinsAgain, line 50: The non-mutation tests detect attribute rebinding, not general mutation. `assigned_attributes()` records only each attribute’s identity and excludes `_memo`. A query can mutate `_components`, `_chains`, or another container in place without changing its identity, and the tests will pass.
Snapshot or compare the non-memo state by value, with targeted assertions for mutable containers. Continue excluding the explicitly transient memo.

probpipe/core/tracked.py
```diff
@@ -319,20 +319,13 @@ def _shallow_copy(self) -> Self:
...
+        shared with it. All three copy paths therefore agree.
         """
         clone = object.__new__(type(self))
-        state = object.__getstate__(self)
-        instance_dict, slots = state if isinstance(state, tuple) else (state, None)
-        for attribute, value in (instance_dict or {}).items():
-            object.__setattr__(clone, attribute, value)
-        for slot, value in (slots or {}).items():
-            object.__setattr__(clone, slot, value)
+        clone.__setstate__(self.__getstate__())
```
SoymilkWinsAgain, line 328: `_shallow_copy` now decouples annotations twice. The new `clone.__setstate__(self.__getstate__())` already processes `_decoupled_state`, including `_annotations`. Afterwards:
- `with_name` copies annotations again.
- `_copy_result_term` also copies them again.

This adds redundant traversal for potentially large `DataTree` annotations and makes custom `Mapping.copy()` behavior observable twice. Remove the caller-side copies now that `_shallow_copy` owns the state-round-trip behavior.

SoymilkWinsAgain (approved):
Also, two comments are stale after the memo changes:
`_shallow_copy` still opens with “sharing all internal state,” although transient state is now omitted and decoupled stores are copied. `DistributionArray` also says `_components` remains `None` only until materialization, while the new implementation keeps it `None` permanently and caches components in `_memo`.

All these 3 are non-block issues.

jhuggins:
Both fixed in `d594d220`.

**The check** — confirmed the gap concretely: editing a term's component dict in place leaves the old identity-only comparison reporting no change. Each attribute now carries a shallow census alongside its identity — a mapping's keys and its values' identities, a sequence's length and elements'. Structural rather than by value, since a distribution's leaves are jax arrays and other terms with no cheap equality; it asserts the arrangement is untouched, not the numbers. Three tests exercise the check itself, including that it still ignores the memo.

**The double copy** — removed from both `with_name` and `_copy_result_term`; a rename now calls `decoupled_container` once, and the decoupling still holds.

tests/core/test_terms_do_not_mutate.py
```diff
@@ -0,0 +1,216 @@
...
+        # builds another result rather than editing this one.
+        before = assigned_state(conditioned)
+        again = condition_on(joint, z=jnp.asarray(3.0))
+        assert assigned_state(conditioned) == before
+        assert again is not conditioned
+        # The operand is untouched, which is what §V.1 promises.
+        assert set(joint.components) == {"z", "x"}
+
+
+class TestACopyDoesNotInheritAMemo:
```
SoymilkWinsAgain, line 246: The regression tests currently exercise only `with_name` on `BroadcastDistribution`. Generic mixin tests cover the mechanism, but would not catch removing` _transient_state` or breaking lazy reconstruction on `BroadcastDistribution`, backend `DistributionArray`, or `ApproximateDistribution`.

probpipe/core/tracked.py
SoymilkWinsAgain: I believe transient state is omitted and decoupled stores are copied

probpipe/core/_distribution_array.py
```diff
@@ -123,6 +125,11 @@ class DistributionArray[T](Distribution[T]):
       than rejecting at construction.
     """
 
+    #: The memo is not state: a copy recomputes rather than inheriting one. It
```
SoymilkWinsAgain, line 128: Note that backend arrays now keep it `None` and cache the tuple in `_memo`. The docstrings need to be changed

probpipe/core/_broadcast_distributions.py
```diff
@@ -1210,7 +1215,10 @@ def __init__(
         name, name_is_auto = auto_name(name, "broadcast")
         super().__init__(name=name, name_is_auto=name_is_auto)
         self._approximate = True
-        self._marginal_cache: MarginalizedBroadcastDistribution | None = None
+        # A memo, filled on first read. It is a container assigned once here
```
SoymilkWinsAgain: Here says that reading does not modify the distribution, while the transient dictionary is deliberately mutated. Docstrings need change

probpipe/inference/_approximate_distribution.py
```diff
@@ -174,7 +180,10 @@ def __init__(
             raise ValueError("Must provide at least one chain")
 
         self._chains = [jnp.asarray(c) for c in chains]
-        self._concatenated: Array | None = None
+        # A memo, filled on first read: a container assigned once here rather
```
SoymilkWinsAgain: Here says that reading does not modify the distribution, while the transient dictionary is deliberately mutated. Docstrings need change

SoymilkWinsAgain (dismissed):
Some non-blocking issues, mainly concerning docstrings. These can be modified and merged

PR #415, opened 2026-08-13
refactor(core): make immutability a property of being a tracked term

jhuggins:
## Summary

`TrackedTerm` inherits `Immutable`, so a record, a batch, a function, or a template is immutable **by being a term** rather than by mixing the guard in itself. `Record`, `Function`, and `Batch` drop it from their base lists, where inheriting it twice would be an MRO conflict.

Construction runs inside a per-instance window the metaclass opens, so a host's `__init__` assigns normally and **no constructor needed converting**. The metaclass already wrapped construction to check the name; it now splits `type.__call__` to open the window between allocation and initialisation, keeping both rules that split has to preserve — `__init__` runs only when `__new__` returns an instance of the class (which every host selecting a subclass from its arguments relies on), and an `__init__` returning anything but `None` raises as it would through `type.__call__`. The window closes when `__init__` returns **and when it raises**, so a half-built term left by a failure is as immutable as a finished one.

**Nothing changes for a caller.** The same four families refuse assignment as before, and a distribution still accepts it — verified side by side against #414's head. What changes is where the rule lives, and that extending it becomes a deletion rather than seventy-two edits.

## The distribution layer is exempt, deliberately

`Distribution.__setattr__` permits assignment. The documented way to build an emulator is to subclass a random function and train it in place — `GPEmulator.fit()` and `FittedPolynomial.fit()` in `docs/user_guide/10_random_functions_and_emulators.ipynb` assign `self._X_train`, `self._K_inv`, `self._weights` — and fitting has no contract yet that returns a new term instead. Enforcing here would break that pattern without offering a replacement, and the replacement is what #370 ("define a controlled fitted-mapping producer contract") is for.

It is one method with one docstring. Deleting it turns the guard on for the other **72** classes, which is 72 of the 84 tracked terms; the 12 outside the distribution branch are exactly the families already guarded. A test asserts `Distribution` is the *only* class defining its own `__setattr__`, so a second exemption cannot appear quietly.

## Linked issue(s)

Part of #395 — the third of its three changes, landing the mechanism and the structural rule while the distribution layer waits on #370. Stacked on #414.

**#395 stays open** for the exemption's removal.

## Breaking changes

None. (An earlier revision of this branch enforced the guard everywhere; that broke the emulator workflow and is what this scope replaces.)

## Test plan

`tests/core/test_immutable.py` gains 18 tests:

- **the conformance tests the rule rests on** — every `TrackedTerm` subclass, *discovered by walking the package* rather than listed, inherits `Immutable`; `Distribution` is the only class with its own `__setattr__`; a distribution accepts assignment and a batch does not, both asserted rather than assumed;
- **the window**: closes when the constructor returns; closes when it *raises*, leaving the half-built object immutable and the registry empty; covers one instance and not another built alongside it; a nested construction leaves the outer window open; an `__init__` that returns a value raises `TypeError`.

4740 passed, 53 skipped. `ruff format` / `ruff check` clean. **The `notebooks (user_guide)` job is the one that matters here** — the emulator notebook executes cleanly against this branch (exit 0), which is what the earlier revision broke.

## Documentation

- [x] CHANGELOG entry added under `Unreleased → Changed`

`TrackedTerm`'s docstring states the guarantee and names the exemption; `_immutable.py` says which layer is exempt and what removes it. `design/02` §II.4 describes the target and needs no change — per `CONTRACTS.md` directive 5, the interim state is labelled in the code rather than written into the design.

## Contract checklist

- [x] Contract of every touched abstraction was clear before coding; the one that was not — what a fitted mapping does to its term — is the reason for the exemption rather than something guessed at.
- [x] The changed contract is documented where it lives: `TrackedTerm` says a term is immutable and that construction assigns inside the window; `constructing` says what the window is and who opens it; `Distribution.__setattr__` says why it is exempt and what removes it.
- [x] Code verified to obey the documented contract; the conformance test asserts it over the whole package rather than a sample.
- [x] Docstrings describe the target contract, with the interim state labelled as such.
- [x] `CONTRACTS.md` index already carries the immutability row (added in #413).

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,34 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **Immutability is a property of being a tracked term (#395).** `TrackedTerm`
+  inherits `Immutable`, so assignment and deletion raise on a record, a batch, a
+  function, or a template once its constructor has returned — the design's `C2`
+  and the §V.1 promise that an implementer's object is never modified, enforced
+  rather than documented. Four classes enforced it individually before.
+
+  **The distribution layer is exempt for now.** `Distribution` permits
+  assignment, because the documented way to build an emulator is to subclass a
+  random function and train it in place, and fitting has no contract yet that
+  returns a new term instead. That is one method with one docstring, and deleting
+  it turns the guard on for the other seventy-two classes; a test asserts it is
+  the *only* exemption, so a second cannot appear quietly.
+
+  Construction is unaffected: it runs inside a per-instance window the
+  `TrackedTerm` metaclass opens, so a host's `__init__` assigns normally and no
+  constructor needed converting. The window closes when `__init__` returns, and
+  also when it raises, so a half-built term left behind by a failure is as
+  immutable as a finished one. Code that allocates with `object.__new__` and then
+  calls a constructor by hand opens the window itself — three sites in the
+  package do.
+
+  Nothing changes for a caller: the classes that refuse assignment are the same
+  four families as before, and a distribution still accepts it. What changes is
+  where the rule lives — in the term hierarchy rather than in four class bodies —
+  and that turning it on for the rest is now a deletion.
+
+### Changed
+
 - **Immutability is one mixin, and a term reconstructs from its state (#395).**
   Four classes spelled out the same guard — three of them hardcoding a class name,
   so `NumericRecord` reported `Record` and `NumericEventTemplate` reported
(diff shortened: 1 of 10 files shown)


jhuggins:
Superseded by #416, which is the same change with the correct base. This PR closed itself when the stack was reordered: #414's branch was rewritten to sit on top of this one, so GitHub found this commit already in its base.

The reorder is the fix for a defect review found in #414 — its memo is declared transient, and only a term that goes through `Immutable`'s state round-trip honours that declaration. This PR is what gives the distribution layer that round-trip, so it has to land first.

PR #416, opened 2026-08-13
refactor(core): make immutability a property of being a tracked term

jhuggins:
## Summary

`TrackedTerm` inherits `Immutable`, so a record, a batch, a function, or a template is immutable **by being a term** rather than by mixing the guard in itself. `Record`, `Function`, and `Batch` drop it from their base lists, where inheriting it twice would be an MRO conflict.

Construction runs inside a per-instance window the metaclass opens, so a host's `__init__` assigns normally and **no constructor needed converting**. The metaclass already wrapped construction to check the name; it now splits `type.__call__` to open the window between allocation and initialisation, keeping both rules that split has to preserve — `__init__` runs only when `__new__` returns an instance of the class (which every host selecting a subclass from its arguments relies on), and an `__init__` returning anything but `None` raises as it would through `type.__call__`. The window closes when `__init__` returns **and when it raises**, so a half-built term left by a failure is as immutable as a finished one.

**Nothing changes for a caller.** The same four families refuse assignment as before, and a distribution still accepts it — verified side by side against #414's head. What changes is where the rule lives, and that extending it becomes a deletion rather than seventy-two edits.

## The distribution layer is exempt, deliberately

`Distribution` permits assignment and deletion. The documented way to build an emulator is to subclass a random function and train it in place — `GPEmulator.fit()` and `FittedPolynomial.fit()` in `docs/user_guide/10_random_functions_and_emulators.ipynb` assign `self._X_train`, `self._K_inv`, `self._weights` — and fitting has no contract yet that returns a new term instead. Enforcing here would break that pattern without offering a replacement, and the replacement is what #370 ("define a controlled fitted-mapping producer contract") is for.

It is two overrides — `__setattr__` and `__delattr__`, since a trainer that clears what it fitted needs both. Removing **both** turns the guard on for the other **72** classes, which is 72 of the 84 tracked terms; the 12 outside the distribution branch are exactly the families already guarded, and removing only one would leave half an exemption. A test asserts `Distribution` is the *only* exempt class, so a second cannot appear quietly.

## Linked issue(s)

Part of #395, landing the mechanism and the structural rule while the distribution layer waits on #370.

**#412 and #413 have merged**, so this now targets `main` directly. It still comes *before* the memo work in #414 rather than after. That ordering is the fix for a defect review found: #414 declares its memo transient, and only a term that goes through `Immutable`'s state round-trip honours that declaration — which is what this PR gives the distribution layer. With the old order, #414's own head shared a memo across `copy`, `deepcopy`, and `pickle` while honouring the declaration in `with_name`, so it violated its own contract until this landed on top. Reversed, each PR is correct where it sits.

(Opened in place of #415, which GitHub closed when the branch rewrite put its commit inside the old base.)

**#395 stays open** for the exemption's removal.

## What a pickle needs from a class

Raised in review: the round-trip this documents cannot serialize a class built at runtime, which is what several distribution families generate. Compared against `main` — those classes already raise `PicklingError` there, since the default protocol names the class too — so the limit is pre-existing and the mixin neither causes nor fixes it. The contract wording was the defect, and now states it, along with what is unaffected: `copy` and `deepcopy` hold the class object rather than its name, and `cloudpickle` serializes it by value, which is the path Ray and Prefect take. `TestAClassBuiltAtRuntime` pins all four behaviors. Reconstruction for those families is #417.

## Breaking changes

None. (An earlier revision of this branch enforced the guard everywhere; that broke the emulator workflow and is what this scope replaces.)

## Test plan

`tests/core/test_immutable.py` gains 18 tests:

- **the conformance tests the rule rests on** — every `TrackedTerm` subclass, *discovered by walking the package* rather than listed, inherits `Immutable`; `Distribution` is the only class with its own `__setattr__`; a distribution accepts assignment and a batch does not, both asserted rather than assumed;
- **the window**: closes when the constructor returns; closes when it *raises*, leaving the half-built object immutable and the registry empty; covers one instance and not another built alongside it; a nested construction leaves the outer window open; an `__init__` that returns a value raises `TypeError`.

4740 passed, 53 skipped. `ruff format` / `ruff check` clean. **The `notebooks (user_guide)` job is the one that matters here** — the emulator notebook executes cleanly against this branch (exit 0), which is what the earlier revision broke.

## Documentation

- [x] CHANGELOG entry added under `Unreleased → Changed`

`TrackedTerm`'s docstring states the guarantee and names the exemption; `_immutable.py` says which layer is exempt and what removes it. `design/02` §II.4 describes the target and needs no change — per `CONTRACTS.md` directive 5, the interim state is labelled in the code rather than written into the design.

## Contract checklist

- [x] Contract of every touched abstraction was clear before coding; the one that was not — what a fitted mapping does to its term — is the reason for the exemption rather than something guessed at.
- [x] The changed contract is documented where it lives: `TrackedTerm` says a term is immutable and that construction assigns inside the window; `constructing` says what the window is and who opens it; `Distribution.__setattr__` says why it is exempt and what removes it.
- [x] Code verified to obey the documented contract; the conformance test asserts it over the whole package rather than a sample.
- [x] Docstrings describe the target contract, with the interim state labelled as such.
- [x] `CONTRACTS.md` index already carries the immutability row (added in #413).

## Checklist

### DIFF
--- a/probpipe/core/_distribution_base.py
+++ b/probpipe/core/_distribution_base.py
@@ -84,6 +84,22 @@ class Distribution[T](TrackedTerm, Annotated, ABC):
         If *name* is not a non-empty string.
     """
 
+    # -- Immutability: deferred for this layer ------------------------------
+
+    def __setattr__(self, name: str, value: Any) -> None:
+        """Permit assignment, which :class:`TrackedTerm` otherwise refuses.
+
+        Interim, and the only exemption from the rule that a tracked term is
+        immutable. It stands because the contract for a *fitted* mapping is not
+        settled: the documented way to build an emulator is to subclass a random
+        function and train it in place, and until fitting has a contract that
+        produces a new term instead, enforcing immutability here would break that
+        pattern without offering a replacement.
+
+        Deleting this method turns the guard on for the whole distribution layer.
+        """
+        object.__setattr__(self, name, value)
+
     def __init__(
         self,
         *,
--- a/probpipe/core/_distribution_base.py
+++ b/probpipe/core/_distribution_base.py
@@ -86,6 +86,10 @@ class Distribution[T](TrackedTerm, Annotated, ABC):
 
     # -- Immutability: deferred for this layer ------------------------------
 
+    def __delattr__(self, name: str) -> None:
+        """Permit deletion, for the reason :meth:`__setattr__` gives."""
+        object.__delattr__(self, name)
+
     def __setattr__(self, name: str, value: Any) -> None:
         """Permit assignment, which :class:`TrackedTerm` otherwise refuses.
 
--- a/probpipe/core/_distribution_base.py
+++ b/probpipe/core/_distribution_base.py
@@ -87,7 +87,10 @@ class Distribution[T](TrackedTerm, Annotated, ABC):
     # -- Immutability: deferred for this layer ------------------------------
 
     def __delattr__(self, name: str) -> None:
-        """Permit deletion, for the reason :meth:`__setattr__` gives."""
+        """Permit deletion, for the reason :meth:`__setattr__` gives.
+
+        Goes with that method: removing the exemption means removing both.
+        """
         object.__delattr__(self, name)
 
     def __setattr__(self, name: str, value: Any) -> None:
@@ -100,7 +103,9 @@ class Distribution[T](TrackedTerm, Annotated, ABC):
         produces a new term instead, enforcing immutability here would break that
         pattern without offering a replacement.
 
-        Deleting this method turns the guard on for the whole distribution layer.
+        Deleting this method **and** :meth:`__delattr__` turns the guard on for
+        the whole distribution layer. Both, or the layer keeps half an
+        exemption: a trainer that clears what it fitted would still raise.
         """
         object.__setattr__(self, name, value)
 
--- a/probpipe/core/tracked.py
+++ b/probpipe/core/tracked.py
@@ -31,7 +31,7 @@ from collections.abc import Mapping
 # exposes; the conflict-avoidance constraint itself doesn't change.
 from typing import Any, Self, _ProtocolMeta
 
-from ._immutable import decoupled_container
+from ._immutable import Immutable, constructing, decoupled_container
 from .provenance import Provenance
 
 __all__ = ["Annotated", "TrackedTerm", "auto_name"]
@@ -77,8 +77,13 @@ def _decoupled_annotations(annotations: Mapping[str, Any]) -> Mapping[str, Any]:
 
 
 class _TrackedTermMeta(_ProtocolMeta):
-    """Metaclass enforcing that every ``TrackedTerm`` instance has a
-    non-empty ``name`` set by the time construction returns.
+    """Metaclass running construction in a window, and enforcing a non-empty name.
+
+    A tracked term is immutable, so it can only be built by assigning to it
+    before anyone holds it. This runs ``__init__`` inside
+    :func:`~probpipe.core._immutable.constructing`, which is what lets a
+    constructor assign normally while assignment anywhere else raises. The window
+    is per instance and per thread, and closes even when ``__init__`` raises.
 
     The check runs after ``__init__`` so it covers every construction
     path: classes that call ``super().__init__(name=...)``, classes that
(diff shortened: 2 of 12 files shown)


probpipe/core/tracked.py
```diff
@@ -107,7 +124,7 @@ def __call__(cls, *args: Any, **kwargs: Any) -> Any:
         return instance
 
 
-class TrackedTerm(metaclass=_TrackedTermMeta):
+class TrackedTerm(Immutable, metaclass=_TrackedTermMeta):
```
SoymilkWinsAgain, line 127: `TrackedTerm` inheritance exposes a broken pickle contract for dynamic classes.

Making `TrackedTerm` inherit Immutable extends `Immutable`’s state-round-trip behavior to all tracked terms. Several public distributions use runtime-generated subclasses—`TransformedDistribution`, `SequentialJointDistribution`, and `FlattenedDistributionView`. The inherited `Immutable.__reduce__` serializes `type(self)` directly, but these runtime classes are not registered as importable module attributes. Consequently, standard `pickle.dumps()` raises `PicklingError` for these objects.

This violates the pickle round-trip contract now inherited by these classes. Please provide stable factory-based reconstruction for dynamically generated classes, or narrow the documented contract, and add standard-pickle regression tests for all three cases.

SoymilkWinsAgain (changes requested):

jhuggins:
Checked against `main`: `SequentialJointDistribution` and `FlattenedDistributionView` already raise `PicklingError` there, since the default protocol names the class too — so this is a pre-existing limit of runtime-generated classes rather than something the mixin introduces. (I couldn't build a `TransformedDistribution` on either tree, so no evidence either way for that one.)

What was wrong is the contract I wrote, which promised a pickle round-trip without saying the class must be importable by name. Fixed in `a61b5af1`, along with the regression tests you asked for: both classes refuse standard pickle, both survive `copy`/`deepcopy`, `cloudpickle` handles them, and `ProductDistribution` — which reconstructs through a module-level factory — pickles.

Filed the reconstruction question separately rather than widening this PR.

probpipe/core/_distribution_base.py
```diff
@@ -84,6 +84,22 @@ class Distribution[T](TrackedTerm, Annotated, ABC):
         If *name* is not a non-empty string.
     """
 
+    # -- Immutability: deferred for this layer ------------------------------
+
+    def __setattr__(self, name: str, value: Any) -> None:
```
SoymilkWinsAgain, line 96: The distribution exemption preserves assignment but not deletion. Distribution overrides only `__setattr__`. It therefore inherits `Immutable.__delattr__`. Consider the following behavior in main:
```python
term = Normal(0.0, 1.0, name="x")
term._trained = True
del term._trained
```
On main, deletion succeeds. On #416 it raises `AttributeError`. This contradicts the PR’s “nothing changes for a caller” claim and partially breaks the mutable-distribution exemption intended to preserve emulator subclasses.

It is recommended to add `Distribution.__delattr__` delegating to `object.__delattr__` and test both assignment and deletion. If deletion is intentionally prohibited, document it as a behavior change and apply the breaking-change label.

tests/core/test_immutable.py
```diff
@@ -251,3 +251,219 @@ def double(x: float) -> float:
...
+
+        class Returning(TrackedTerm):
+            def __init__(self):
+                self._init_tracked("t")
+                return "oops"
+
+        with pytest.raises(TypeError, match="should return None"):
+            Returning()
+
+    def test_it_closes_when_the_constructor_raises(self):
```
SoymilkWinsAgain: This manually opens `constructing()` around an Immutable host. It proves the context manager closes, but not that `_TrackedTermMeta.__call__` correctly manages a raising constructor. Use a `TrackedTerm` subclass instantiated normally and retain the half-built instance externally for the final assertion.

SoymilkWinsAgain (changes requested):
Also, authoritative convention documents still contradict the distribution exemption.
`STYLE_GUIDE.md:788` says distributions are immutable and annotations are the only carve-out. `CONTRIBUTING.md:437` says the same. #416 explicitly codifies a much broader interim exception. These convention documents should acknowledge that current exception, even if the design documents continue describing the target state.

jhuggins:
Both fixed.

**Deletion** — you're right, and I'd verified the claim too narrowly: `del term._trained` works on `main` and raised here. `Distribution` now delegates `__delattr__` to `object` alongside `__setattr__`, tests assert both operations on a distribution and both refusals on a record, and the CHANGELOG's "nothing changes for a caller" line is corrected. `e39b4476`.

**The raising-constructor test** — it now defines a `TrackedTerm` whose `__init__` raises, instantiates it normally, and keeps the half-built instance from `__new__`, so it exercises `_TrackedTermMeta.__call__` rather than the context manager (which is tested separately).

probpipe/core/_distribution_base.py
```diff
@@ -84,6 +84,31 @@ class Distribution[T](TrackedTerm, Annotated, ABC):
...
+
+    def __delattr__(self, name: str) -> None:
+        """Permit deletion, for the reason :meth:`__setattr__` gives.
+
+        Goes with that method: removing the exemption means removing both.
+        """
+        object.__delattr__(self, name)
+
+    def __setattr__(self, name: str, value: Any) -> None:
+        """Permit assignment, which :class:`TrackedTerm` otherwise refuses.
```
SoymilkWinsAgain, line 97: `Distribution` explicitly permits arbitrary assignment and deletion here. However, `STYLE_GUIDE.md` and `CONTRIBUTING.md` still say distributions are immutable and annotations are the only exception. These files describe current conventions, not only the eventual design target. Document the temporary fitted-emulator exemption there.

Can be merged after changing the documents

SoymilkWinsAgain (approved):

Issue #398, opened 2026-08-03
Give every value spec a tracked class: NumericArray and Opaque


jhuggins:
## Summary

Every value spec has a tracked class except the two raw-value ones. `RecordSpec` has `Record`, `DistributionSpec` has `Distribution`, `FunctionSpec` has `Function` — `ArraySpec` and `OpaqueSpec` have none. So an array-valued result is presented as a **single-field `Record`**, and an opaque value as a bare Python object.

This proposes `NumericArray` for `ArraySpec` and `Opaque` for `OpaqueSpec`, plus a `NumericArrayBatch`, so that every value spec has one tracked class and one batch form. It amends the target design rather than correcting the code: the single-field wrap is specified in V.1 and III.5.

## The asymmetry

V.1's `sample` contract, on what a non-empty `sample_shape` returns:

| raw type `T` | native batch form |
|---|---|
| `Array` | **an array with the `sample_shape` axes leading** |
| `Record` / `NumericRecord` | `RecordBatch` / `NumericRecordBatch` |
| `Distribution` | `DistributionBatch` |
| `ConditionalDistribution` | `ConditionalDistributionBatch` |
| function | `FunctionBatch` |
| opaque | `OpaqueBatch` |

The `Array` row is the only one whose batch form is not a class. And the last two rows have no class in the **raw type** column either — "function" and "opaque" are lowercase, because `FunctionSpec` is callable-generic and opaque values have no class at all — yet both are given batch classes, because a batch must be a tracked term.

II.2 names that second point, as the case `BatchSpec` exists for: "`OpaqueBatch`, whose elements name no kind at all, still has one." At the batch layer, kindless values already have tracked classes; at the element layer they do not, so `OpaqueBatch` is a tracked term whose elements are not tracked terms.

III.1 gives the reason the `Array` row is exempt: "Since an `ArraySpec` value batches natively, as an array with the batch axes leading, no class is needed." That is true of the storage and silent about identity, which is what the rest of the column is for.

## What the wrap costs

V.1: "A scalar law returns a single-field `Record` keyed by the distribution's name, initialized from the name and thereafter independent of it." III.5 says the same.

The wrap is then taken apart downstream so that a wrapped function receives a value it can compute with, which makes the argument's type depend on the law's field count:

| law | what a lifted function receives |
|---|---|
| `Normal(...)` | bare array |
| `EmpiricalDistribution(...)` | bare array |
| a one-field independent joint | bare array |
[...]

Issue #167, opened 2026-04-30
Consolidate single-field shims into a Scalar/Singleton wrapper


jhuggins:
## Context

ProbPipe's "uniform output wrap" rule (CONTRIBUTING.md §7) coerces every `WorkflowFunction` return into a `Record | RecordArray | Distribution` shape. To keep idioms like `float(mean(d))`, `sample(grf)(X)`, and `np.asarray(log_prob(d, v))` terse, **single-field** instances of these container types expose array-coercion shims (`__array__`, `__jax_array__`, `__float__`, `__int__`, `__bool__`, `.shape`, `.dtype`, `.ndim`, sometimes `__call__`).

The shims live on **four parallel classes**:

| Class | Single-field shim |
|---|---|
| `NumericRecord` | `__array__`, `__jax_array__`, `__float__`, `__int__`, `__bool__`, `.shape`, `.dtype`, `.ndim` |
| `Record` | `_single_numeric_leaf` (used by subclasses) |
| `RecordDistribution` | `.shape`, `.ndim` (multi-field raises) |
| `_RecordDistributionView` | inherits via the dynamic-protocol-subclass machinery |

## The problem

The architecture-overview review:

> The single-field-shim mechanism leaks everywhere. `NumericRecord`, `RecordDistribution`, `_RecordDistributionView`, and `Record` itself all carry single-field coercion shims (`__float__`, `__jax_array__`, `.shape`, `__call__`). Each individually justified; collectively they make "what does `result.shape` mean" a multi-page answer. A single concrete `Scalar`/`Singleton` wrapper that all single-field paths produce would be more findable than four parallel shim implementations.

Concrete pain:

1. **Drift across implementations.** When a new shim is needed (e.g. `__add__` for arithmetic), it has to be added to every class, and someone forgets one. The `flat_samples` work in PR-B started here.
2. **Discoverability.** A user inspecting `mean(d)` sees a `NumericRecord` and has to remember which methods are shimmed and which aren't. There's no canonical place to look.
3. **Multi-field error message.** Each class duplicates the "multiple fields; use `[fieldname]` instead" raise, with slight variations in wording.

## Proposed change

Introduce a single `Scalar[T]` (or `Singleton[T]`) wrapper that:

- Is the canonical single-field result type produced by `_wrap_as_record` when the WF return is scalar-like.
[...]

PR #418, opened 2026-08-15
docs(design): give every value spec a tracked class

jhuggins:
## Summary

`RecordSpec` has `Record`, `DistributionSpec` has `Distribution`, `FunctionSpec`
has `Function` — `ArraySpec` and `OpaqueSpec` have none. So an array-valued
result is presented as a **single-field `Record`** and an opaque value as a bare
Python object.

This amends the design to state the rule the other kinds already follow:

> Every value spec has exactly one tracked class and one batch form. An
> operation returns the tracked term of its declared kind. No kind is presented
> by wrapping it in another kind.

`NumericArraySpec` takes `NumericArray` / `NumericArrayBatch`; `OpaqueSpec`
takes `Opaque` beside the `OpaqueBatch` it already had.

**`ArraySpec` is renamed `NumericArraySpec`**, so a spec and its class agree the
way `RecordSpec`/`Record` and `OpaqueSpec`/`Opaque` do. In the design only here:
the name is public API (104 uses in `probpipe/`, 311 in `tests/`), so the code
rename lands with the class it names.

**Documentation only — no code changes.** This is stage 0 of issue 398; the
classes themselves follow.

## What the wrap costs, measured

The wrap is taken apart downstream so a lifted function receives something it can
compute with, which makes an argument's type depend on the law's field count:

| law | what a lifted function receives |
| --- | --- |
| `Normal(...)` | bare array |
| a one-field independent joint | bare array |
| a two-field independent joint | `NumericRecord` |

A function written against a one-field law stops working when the model gains a
field — a change that need not touch that call site.

Two gaps are measurable on `main` today:

```python
sample(Normal(0., 1., name="x"), sample_shape=(3,))
# NumericRecord, fields=('sample',) — no batch_shape, no level_names
```

The draw axis lives *inside* the leaf array with no level, because the `Array`
row of V.2's native-batch-form table has no class to carry one. And `log_prob`
returns a single-field `Record` for the same reason.

## Sections amended

| § | File | Change |
| --- | --- | --- |
| II.2 | `02-shared-abstractions.md` | the rule; the raw-value / term-spec split refined rather than collapsed |
| III.1 | `03-values-and-distributions.md` | reverses "Since a `NumericArraySpec` value batches natively … no class is needed"; adds the three classes; retitled |
| III.2 | `03-values-and-distributions.md` | *Single-field presentation* — the coercion shim retires |
| III.5 | `03-values-and-distributions.md` | "the draw a *user* sees" follows the declaration, not the field count |
| IV.2 | `04-functions.md` | closes the *Single-value presentation* open point |
| V.2 | `05-operations.md` | `sample`'s contract and its native-batch-form table |

II.5 needed no change: "a batch of them yields bare values" already settles what
`NumericArrayBatch[i]` is, so III.1 cites it rather than restating it.

Five further passages in V.0–V.3 asserted the old wrap in passing (the raw-host
wrap rule, the tracking-scope note, `log_prob`'s result, its batched result, and
the V.2 rationale that justified wrapping "even a scalar draw"). All are updated,
so the part does not contradict itself.

## How the pieces sit

**A raw-value spec still types what a leaf may hold.** A record's numeric leaves
stay bare arrays and `record["x"]` returns no box. The class is what an operation
*returns* — the rule III.1 already stated for the function kind, where "a
`FunctionSpec` admits any callable … the `Function` being one such, not the
required type". A `NumericRecord` consists of `NumericArray`s at the level of
*kinds*, not by boxing each leaf.

**`NumericArray` carries no batch axes.** Its `shape` is the event shape;
multiplicity lives in `NumericArrayBatch`, the same split `RecordBatch` uses with
one column instead of many.

**Selection yields the element kind**, as it does for every batch:
`NumericArrayBatch[i]` is a `NumericArray`, materialized under the derived name
and its own lineage, exactly as `RecordBatch[i]` is a `Record`. II.5's
bare-element rule keeps governing the batches whose elements really are bare — a
`FunctionBatch` of plain callables, and an `OpaqueBatch`.

**Arithmetic on a `NumericArray` yields a bare array.** Identity is attached by
operations, and arithmetic is not one.

## Linked issue(s)

Refs #398 — stage 0 of four. Refs #167 and #357, which this answers by removing
their premise rather than by choosing a coercion tier.

## Test plan

No code changes, so no tests. The claims about current behaviour above were
checked against `main` at `ebef50fd` before writing.

## Breaking changes

None here. The staged code changes will be breaking — every array-returning
operation's return type changes — which is why the design lands first.

## Documentation

This *is* the documentation change. CHANGELOG entry under `Unreleased → Changed`.


design/02-shared-abstractions.md
```diff
@@ -114,6 +114,10 @@ A **term spec** types a tracked term, one concrete class per kind. `TermSpec` su
 
 A batch is a tracked term too, so it has a term spec of its own: `BatchSpec` types the *collection*, carrying the spec its elements satisfy together with the named multiplicity. It is what keeps the storage rule below universal — a batch's own type is not its element's type, and `OpaqueBatch`, whose elements name no kind at all, still has one.
 
+**Every value spec has exactly one tracked class and one batch form**: `NumericArraySpec` has `NumericArray` and `NumericArrayBatch`, and `OpaqueSpec` has `Opaque` and `OpaqueBatch`, as `RecordSpec` has `Record` and `RecordBatch` (III.1).
```
SoymilkWinsAgain, line 117: New tracked classes cannot satisfy the constraints of current spec classes: The pull request specifies that `NumericArraySpec` -> `NumericArray` and `OpaqueSpec` -> `Opaque`, but both are still raw `ValueSpecs`. Meanwhile, `TrackedTerm.spec` is still limited to `TermSpec`. More directly, `Distribution` and `DistributionSpec.event_spec` also only accept `TermSpecs`, therefore the scalar Normal cannot declare its own draw kind using `NumericArraySpec`.

design/05-operations.md
```diff
@@ -13,17 +13,17 @@ Every operation shares the following mechanics:
...
   - *Identity is attached by derivation.* TrackedTerm terms are immutable and provenance is write-once. The default result is therefore a new term: the implementer's representation, shared, under the operation's fresh name and provenance. The implementer's object is never modified. `raw=True` returns it untouched.
-  - *A raw host wraps into its own kind.* A bare array or other raw non-mapping value becomes the single-field `Record` its `event_template` describes. A raw mapping becomes a structured `Record`. A raw callable becomes a `Function`. A backend distribution enters through its converter. A concrete family such as a `Normal` can therefore be written in plain arrays, with the family's natural constructor and no hand-written template.
+  - *A raw host wraps into its own kind.* A bare array becomes a `NumericArray` and any other raw non-mapping value an `Opaque`. A raw mapping becomes a structured `Record`. A raw callable becomes a `Function`. A backend distribution enters through its converter. A concrete family such as a `Normal` can therefore be written in plain arrays, with the family's natural constructor and no hand-written template.
```
SoymilkWinsAgain, line 18: This seems to conflict with design02 L171 and design04 L 71. On one hand, it's said that the results of arrays and opaques become `NumericArray` and `Opaque` respectively, while on the other hand, it's said that the raw-value output will be wrapped as a single-field record. This makes subsequent implementations lack a unique target.

design/03-values-and-distributions.md
```diff
@@ -75,9 +75,17 @@ def is_differentiable(x: Any, values: NamedTree | None = None) -> bool: ...
...
+
+Multiplicity lives in `NumericArrayBatch`: a `Batch` whose `element_spec` is the `NumericArraySpec` and whose storage is one array with the batch axes leading — the same split `RecordBatch` uses, with one column instead of many. A `draw` level therefore has somewhere to live for an array-valued law (V.2).
+
+`Opaque` adds identity and nothing else: no attribute forwarding, no `__call__`, one explicit accessor for the wrapped value.
```
SoymilkWinsAgain, line 84: `Opaque` breaks the lifting mechanism of regular Python functions. The design specifies that lifted functions receive a tracked draw returned by `sample`, and opaque draws are now `Opaque`s. However, `Opaque` explicitly states that "`Opaque` adds identity and nothing else: no attribute forwarding, no `__call__`, and one explicit accessor for the wrapped value." Therefore, a regular function that would normally receive a string and call `x.upper()` will now receive a wrapper and fail. This violates the fundamental contract that `Function`s receive "natural values".

design/05-operations.md
```diff
@@ -50,13 +50,13 @@ A mean is defined whenever draws can be averaged: coordinate-wise for arrays, po
...
 - A non-empty `sample_shape` prepends batch axes and returns the tracked batch type of the draw's kind, with those leading dimensions on a level named `draw`: a `RecordBatch` (`NumericRecordBatch` when numeric) for a record event, and the native batch form of the table below — a `DistributionBatch`, `FunctionBatch`, and so on — for a term-drawing law.
 - With `raw=True`, `sample` skips the identity attachment. For `sample_shape=()` it returns the bare draw type `T`; when `T` is already a tracked term (a `Record`, `Distribution`, `ConditionalDistribution`, or `Function`) the wrap is already the identity, so the only difference from the default is the identity attachment — `raw=True` returns the implementer's term itself, with its own name and provenance, while the default derives a new term from it, the same kind and representation under fresh `sample` provenance, and a raw non-term draw carries no name or provenance either way. For a non-empty `sample_shape` it returns `T`'s **native batch form**:
 
 | raw type `T` | native batch form |
 |---|---|
-| `Array` | an array with the `sample_shape` axes leading |
+| `NumericArray` | `NumericArrayBatch` |
```
SoymilkWinsAgain, line 59: This seems conflict with above

design/05-operations.md
SoymilkWinsAgain, line 36: Moment contracts have not yet migrated to declared-kind rules. Mean/variance only lists `Record` and term, quantile still consistently returns `Record`/`RecordBatch`, and cov still requires the distribution to hold `NumericEventTemplate`. None of these cover the new path of `NumericArraySpec` -> `NumericArray`.

design/package-structure.md
SoymilkWinsAgain, line 35: Needs to add `NumericArray`, `NumericArrayBatch`, `Opaque`

design/package-structure.md
```diff
@@ -9,7 +9,7 @@ This document uses the **`Function`** vocabulary. A `Function` is the universal
...
-- **A spec lives with the type it admits.** `FunctionSpec` lives in the value layer with the kind it describes, and `DistributionSpec` and `ConditionalDistributionSpec` live with the distribution classes. The rule bends only where layering forbids it: `ValueSpec` and the `TermSpec` marker are what the tracked base stores, and `ArraySpec` and `OpaqueSpec` admit no ProbPipe kind at all, so all four sit in `core/` beside `EventTemplate` and `Constraint` — `core/` cannot import the value layer. The same placement rule covers each type's batch form.
+- **A spec lives with the type it admits.** `FunctionSpec` lives in the value layer with the kind it describes, and `DistributionSpec` and `ConditionalDistributionSpec` live with the distribution classes. The rule bends only where layering forbids it: `ValueSpec` and the `TermSpec` marker are what the tracked base stores, and `NumericArraySpec` and `OpaqueSpec` admit no ProbPipe kind at all, so all four sit in `core/` beside `EventTemplate` and `Constraint` — `core/` cannot import the value layer. The same placement rule covers each type's batch form.
```
SoymilkWinsAgain, line 12: `NumericArraySpec`/`OpaqueSpec` shouldn't be "admit no ProbPipe kind" now?

SoymilkWinsAgain (changes requested):

jhuggins:
Superseded by #434. The stack grew to eleven PRs; the same commits are regrouped into four, split by concern rather than by the order they were written. No content changed — the four branches carry these commits verbatim (plus one commit recording that a batch of empty records stays refused). Closing to keep review in one place.

PR #419, opened 2026-08-15
feat(core): add NumericArray and NumericArrayBatch

jhuggins:
## Summary

Stage 1 of issue 398: the tracked class of the numeric-array kind and its batch
form, so `NumericArraySpec` has the pair every other value spec has.

**Nothing returns them yet.** The operations switch over in a later change; the
suite is unmoved apart from the new tests. Stacked on #418, which is the design
amendment these implement.

## `NumericArray`

Holds one array and carries **no batch axes**, so `shape` is the event shape;
multiplicity lives in `NumericArrayBatch`.

It carries the full array surface — arithmetic, comparison, conversion hooks —
and **arithmetic yields a bare array**:

```python
value = NumericArray(jnp.arange(3.0), name="draw")
value + 1        # Array([1., 2., 3.]) — not a NumericArray
```

Identity is attached by operations, and arithmetic is not one, so a
`NumericArray` is what an operation hands back and what a caller computes
*from*. That is what lets `Record` stay a container and carry none of it.

The ~40 operators are installed from a table rather than written out: they all
forward the same way, and a hand-written list is where one gets forgotten.
`__hash__` is dropped, since an elementwise `__eq__` would make it a false
promise.

**It is a registered pytree.** Writing the tests turned up that it could not be
passed to a traced function at all — JAX no longer converts through
`__jax_array__` during abstractification, so a value must be registered to cross
a `jit` or `vmap` boundary. Its aux carries the identity and the declared
`support`; the **spec is re-derived from the array that arrives**, because a
`NumericArray`'s spec *is* its shape and dtype, so that reading is exact rather
than the guess the batch types must refuse.

## `NumericArrayBatch`

Stores one array with the batch axes leading and splits them from the event axes
by the element spec's shape — the `RecordBatch` split with one column instead of
many. This is where a `draw` level lives for an array-valued law.

Selection yields the element kind, as it does for every batch: `batch[i]` is a
`NumericArray`, materialized under the derived name and inheriting the batch's
lineage.

**Not registered as a pytree yet**, deliberately. Doing so raises the same
question `RecordBatch` answers with its two-transformation contract, and nothing
maps over one until the operations return them.

## The rename

`ArraySpec` → `NumericArraySpec`, so the spec agrees with the class it names as
`RecordSpec`/`Record` does. Mechanical: 415 occurrences across 36 files, plus the
public export. Its own commit, so the diff reads separately from the new code.

**This is breaking** — `ArraySpec` was public API.

## Linked issue(s)

Refs #398 — stage 1 of four. Stacked on #418 (stage 0, the design amendment).

## Test plan

`tests/core/test_numeric_array.py`, 43 tests:

- what each class holds — the event/batch axis split, spec derivation, level
  tiling including one level spanning several axes;
- the identity — user vs auto names, write-once provenance, immutability, and
  that it is unhashable;
- the array surface — arithmetic parametrized over eight expressions, all
  asserted to return a bare `jax.Array`; elementwise comparison; both conversion
  hooks; scalar conversion; indexing and iteration;
- the pytree round trip, spec re-derivation under a shape-changing `tree_map`,
  and that provenance does not survive the boundary;
- selection — element kind, derived name, inherited lineage, sub-batch views;
- the refusals — no batch axis, wrong trailing shape, symbolic event dimension,
  wrong spec kind, axis groups that do not tile.

Full suite: 4739 passed, 54 skipped. `ruff check .` and `ruff format --check .`
clean from the repo root.

## Breaking changes

`ArraySpec` is renamed `NumericArraySpec`. No behaviour changes: the new classes
are not yet returned by anything.

## Documentation

CHANGELOG entries under `Unreleased → Added` and `→ Changed`. The design these
implement is #418.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,11 +86,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   later change.
 
   `NumericArray` holds one array and carries no batch axes, so its `shape` is the
-  event shape. It carries the full array surface, and **arithmetic on it yields a
-  bare array** — identity is attached by operations, and arithmetic is not one.
-  It is a registered pytree, which is what lets it cross a `jit` or `vmap`
-  boundary; its spec is re-derived from the array that arrives, since a shape and
-  a dtype state it exactly.
+  event shape. Construction **validates without converting** — the value is stored
+  in its native form and materialises at most once, at the compute boundary — the
+  rule `NumericRecord` already follows for its leaves. It carries the full array
+  surface, and **arithmetic yields a bare value of the stored type**: identity is
+  attached by operations, and arithmetic is not one. It is a registered pytree,
+  which is what lets it cross a `jit` or `vmap` boundary; the boundary presents a
+  bare array, and its spec is re-derived from what arrives, since a shape and a
+  dtype state it exactly.
 
   `NumericArrayBatch` stores one array with the batch axes leading and splits
   them from the event axes by its element spec. Selection yields a `NumericArray`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -97,7 +97,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
   `NumericArrayBatch` stores one array with the batch axes leading and splits
   them from the event axes by its element spec, which it validates the stored
-  dtype against at construction — the batch asserts that spec of every element.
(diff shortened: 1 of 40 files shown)


jhuggins:
Superseded by #434 and #435. The stack grew to eleven PRs; the same commits are regrouped into four, split by concern rather than by the order they were written. No content changed — the four branches carry these commits verbatim (plus one commit recording that a batch of empty records stays refused). Closing to keep review in one place.

PR #420, opened 2026-08-15
feat(core): add Opaque, the tracked class of the opaque kind

jhuggins:
## Summary

Stage 2 of issue 398: `Opaque`, the tracked class of the opaque kind — the last
value spec without one.

`OpaqueSpec` had a batch form already, since a collection is a tracked term
whatever its elements are; the *element* was the half that was missing. With
this, every value spec has exactly one tracked class and one batch form:

| spec | element | batch |
| --- | --- | --- |
| `NumericArraySpec` | `NumericArray` | `NumericArrayBatch` |
| `OpaqueSpec` | **`Opaque`** (new) | `OpaqueBatch` |
| `RecordSpec` | `Record` / `NumericRecord` | `RecordBatch` / `NumericRecordBatch` |
| `FunctionSpec` | `Function` | `FunctionBatch` |
| `DistributionSpec` | `Distribution` | *(later wave)* |

**Nothing returns one yet.** The operations switch over in a later change.
Stacked on #419.

## It adds identity and nothing else

No attribute forwarding, no `__call__`, no operators. The wrapped value is
reached through `.value`, explicitly:

```python
fitted = Opaque(some_model, name="posterior_sampler")
fitted.value.predict(x)     # what it affords is its own type's, once it is out
fitted.predict(x)           # AttributeError
fitted()                    # TypeError, even though the value is callable
```

A box that forwards to be convenient is indistinguishable from the value it
wraps, and then the wrapping is a surprise rather than a statement. This is the
deliberate contrast with `NumericArray`, which *does* carry a full surface —
because an array has one meaning for `+` and an arbitrary object has none.

A mapping is refused, which is the rule `OpaqueSpec.is_valid` already states:
the value layer reads a mapping as a subtree rather than a leaf.

## `OpaqueBatch` is untouched

Its elements are **stored** rather than materialized, so it keeps handing back
the caller's own object rather than wrapping each one in an `Opaque`. Two tests
pin that: `batch[0] is payloads[0]`, and a batch whose elements are themselves
`Opaque` terms hands those back under their own names.

## Linked issue(s)

Refs #398 — stage 2 of four. Stacked on #419 (stage 1).

## Test plan

`tests/core/test_opaque.py`, 24 tests: what it holds (value reachable and
unchanged, spec default and `meta`, the mapping refusal, a parametrized sweep
over admitted value types); **that it adds nothing** (no attribute forwarding,
not callable when its value is, no array surface, and the value still affords
everything once out); identity (names, write-once provenance, immutability,
copy/deepcopy/pickle round trips); and the two `OpaqueBatch` contract tests.

Full suite: 4788 passed, 54 skipped. 100% coverage on the new module.
`ruff check .` and `ruff format --check .` clean from the repo root.

## Breaking changes

None. The class is new and nothing returns it yet.

## Documentation

CHANGELOG entry under `Unreleased → Added`. The design is #418.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -102,6 +102,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Selection yields a `NumericArray` under the derived name, as `RecordBatch`
   yields a `Record`.
 
+- **`Opaque` — the tracked class of the opaque kind (#398).** What an operation
+  returns when its declared kind is an `OpaqueSpec`, completing the pair with the
+  `OpaqueBatch` that already existed. It adds identity and nothing else: no
+  attribute forwarding, no `__call__`, no operators — the wrapped value is
+  reached through `.value`, explicitly. `OpaqueBatch`'s element contract is
+  unchanged; it still hands back the object the caller put in rather than
+  wrapping it.
+
 - **`ArrayBackend.take` — positional selection for a native container.** `[]` is
   positional on a numpy-protocol container and reads *labels* on a `pandas` one,
   so a batch stored as a `DataFrame` could not address its own elements. Backends
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -110,6 +110,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   unchanged; it still hands back the object the caller put in rather than
   wrapping it.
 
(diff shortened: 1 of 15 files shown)


jhuggins:
Superseded by #435. The stack grew to eleven PRs; the same commits are regrouped into four, split by concern rather than by the order they were written. No content changed — the four branches carry these commits verbatim (plus one commit recording that a batch of empty records stays refused). Closing to keep review in one place.

PR #421, opened 2026-08-16
feat!: wrap an operation's result at its declared kind

jhuggins:
## Summary

Stage 3a of issue 398: an operation returns the tracked term of its **declared
kind**, rather than presenting every raw value as a single-field `Record`.

```python
log_prob(law, 0.0)      # NumericArray, was NumericRecord(log_prob=...)
sample(scalar_law)      # NumericArray
mean(law)               # NumericArray
some_fn(x)              # Opaque for a string, Function for a callable
```

Stacked on #420.

## One seam, not one change per operation

Every op in `ops.py` — `sample`, `log_prob`, `prob`, the unnormalized pair,
`mean`, `variance`, `cov`, `quantile`, `expectation`, `condition_on` — is
`@function`-decorated, so **none of them wraps its own result**. The wrap lives
in one function at the `Function` output boundary, whose docstring already
enumerated the two rules this replaces.

`_wrap_as_record` is renamed **`_wrap_as_term`**: it no longer wraps as a record.
Three stale descriptions around it go with the name — the summary line, the
boundary description in `_coerce_output`, and a `Returns` section that listed
`Record | RecordBatch | Distribution` as the full set.

| raw return | before | now |
| --- | --- | --- |
| numeric scalar / array | single-field `NumericRecord` | `NumericArray` |
| mapping | `Record` | `Record` |
| callable | single-field `Record` | `Function` |
| anything else | single-field `Record` | `Opaque` |

## The callable rule is the part with teeth

`sample(grf)(X)` worked because the single-field record's `__call__` **forwarded**
— not because the value had a kind. Routing callables to `Opaque`, which
deliberately forwards nothing, broke **all thirteen** random-function tests.
Wrapping them as `Function` fixes every one, and the whole file (96 tests) passes:
the result is callable because of what it *is*.

Design V.0 already stated this rule; the code had never implemented it.

## Two things this deliberately does not change

**A tracked term a body returns stays an event payload**, which #369 gates. This
needed an explicit guard: a `Function` is itself callable, so the new dispatch
would otherwise have re-wrapped a returned term as a fresh `Function`, which
`test_function_return_remains_event_payload_without_template` caught.

**The batch side.** A scalar law's `sample(sample_shape=(3,))` still returns a
`NumericRecord` with no levels — `_make_stack` is stage 3b.

## A design ambiguity fixed

The V.0 sentence added in #418 read "…any other raw non-mapping value an
`Opaque`. … A raw callable becomes a `Function`" — but a callable *is* a
non-mapping value, so the clauses contradicted. It now says the rules are ordered,
with `Opaque` the fallback the named kinds fall through to.

## Linked issue(s)

Refs #398 — stage 3a of four. Stacked on #420 (stage 2).

## Test plan

New `tests/core/test_wrap_at_the_declared_kind.py`, 12 tests: each kind in turn,
the ordering (a callable takes `Function` before the `Opaque` fallback), the
tracked-term payload guard, that the operations return their declared kind, and
that a `NumericArray` result still computes without unwrapping.

Migration: the probe's 29 failures are all resolved. They fell into three kinds —
`isinstance(result, NumericRecord)` assertions, `result["func_name"]` field reads
(now `result` itself), and a string return now read through `.value`.

Full suite: **4801 passed**, 54 skipped. `ruff check .` and
`ruff format --check .` clean from the repo root.

## Breaking changes

Yes. **A result is no longer indexable by the function's name**: `result["my_fn"]`
becomes `result`, and an opaque result is read through `.value`. Numeric results
are unaffected wherever they are consumed through conversion (`float(...)`,
`jnp.asarray(...)`, arithmetic), which is why the migration was 29 sites rather
than hundreds.

## Documentation

CHANGELOG entry under `Unreleased → Changed (breaking)` with the before/after
table. The design correction is in `design/05-operations.md`.


jhuggins:

PR #422, opened 2026-08-16
feat!: put a law's draws on a draw level

jhuggins:
## Summary

Stage 3b-i of issue 398: a non-empty `sample_shape` puts the draws on a level
named `draw`, which design V.2 has always specified and only the
record-drawing laws implemented.

```python
sample(Normal(0., 1., name="x"), sample_shape=(5,))
# before: NumericArray of shape (5,) — draws inside the value, no level
# now:    NumericArrayBatch, batch_shape=(5,), level_names=('draw',)

sample(MultivariateNormal(loc=zeros(3), cov=eye(3), name="v"), sample_shape=(5,))
# NumericArrayBatch, batch_shape=(5,), element event shape (3,), .shape == (5, 3)
```

Without this, II.3's level algebra was unavailable for the most common law kind.
Stacked on #421.

## How the split is made

`sample` assembles the batch only when the draw is a **raw array** — a law that
draws a record or a term already builds its own batch and is left alone. The
batch/event split is read off the **drawn shape** rather than asked of the law,
which stays exact for a law whose event shape is symbolic until a draw binds it,
and needs no accessor a multi-field law would refuse. A law that does not
prepend its draws where the contract says is handed back untouched rather than
having its axes mis-split.

## Two gaps this exposed in `NumericArrayBatch`

Both surfaced the moment an operation returned one, and **neither was a test
migration** — no test file changed in this PR.

**The array shim was missing.** `NumericRecordBatch` forwards `shape`, `dtype`,
`ndim`, and both conversion hooks from its sole field. `NumericArrayBatch` is
single-column *by construction*, so it forwards from its store with no "sole
field" caveat. **Forty of the 43 initial failures were that one absence.**
II.3's `batch_shape` / `batch_size` remain the unambiguous names for the batch
axes alone; `shape` is the whole store, exactly as on the sibling.

**It needed the pytree registration stage 1 deferred.** #419 says it is "not
registered yet … nothing maps over one until the operations return them" — this
is that moment. JAX no longer converts through `__jax_array__` during
abstractification, so an unregistered batch cannot reach a traced function at
all. It now follows `RecordBatch`'s **two-transformation contract**: every batch
axis preserved, or every one removed, and nothing between, because *a shape is
not a provenance*.

## One boundary change

`_wrap_as_term` retains **any `Batch`**, not only a `RecordBatch`. A batch an
operation returns is a result of its element's kind, not an event payload to
bury — without this, a returned `NumericArrayBatch` would have been wrapped as a
`Record` field by the tracked-term guard added in #421.

## Linked issue(s)

Refs #398 — stage 3b-i. Stacked on #421 (stage 3a). 3b-ii, the sweep-aggregation
half, is independent and measured at 24 failures; it is not in this PR.

## Test plan

- `tests/core/test_wrap_at_the_declared_kind.py`, 7 new: no `sample_shape` is one
  value; draws land on one `draw` level (parametrized over `(5,)` and `(2, 3)`);
  a vector law keeps its event axes out of the draw level; an element is one
  draw; a record law still draws its own batch; and a law that does not prepend
  its draws is left alone.
- `tests/core/test_numeric_array.py`, 8 new: the pytree round trip, removing
  every batch axis yielding the element, the refusal of a partial or resized
  rank, a skeleton rebuilding, reaching a traced function, and the array shim.

One of those tests is worth reading for what it documents: a `vmap` over a batch
hands the body an **element**, so returning it stacks into a `NumericArray` with
the mapped axis prepended. That value carries no levels, so re-deriving its spec
from the arriving shape is exact rather than a guess — which is the split the
two classes exist to make.

Full suite: **4816 passed**, 54 skipped. 100% on `_numeric_array_batch.py`.
`ruff check .` and `ruff format --check .` clean from the repo root.

## Breaking changes

Yes. `sample(law, sample_shape=...)` returns a `NumericArrayBatch` rather than a
flat array. Code reading `.shape`, `.dtype`, `np.asarray(...)`, or `jnp.asarray(...)`
is unaffected — the shim covers those, which is why no test needed changing —
but code that assumed the result *was* a bare array (passing it where only an
`Array` is accepted) now receives a batch.

## Documentation

CHANGELOG entry under `Unreleased → Changed (breaking)`.


jhuggins:

PR #424, opened 2026-08-17
An empty return keeps its host's kind

jhuggins:
Stacked on #426. Part of #398.

## The bug

A `Function` returning an empty mapping **raised**:

```
TypeError: Opaque holds one unstructured value, and the value layer reads a
mapping as a subtree rather than a leaf
```

`{}` fell past the record branch (guarded by `and value`) all the way to `Opaque`, which refuses mappings on principle. `[]` and `()` reached `Opaque` too — no crash, but the wrong kind.

## Why the guards were wrong

V.0 wraps a raw host **into its own kind**, and the kinds are ordered by *host type*. The empty cases broke that: `{"a": 1}` is a `Record` but `{}` was an `Opaque`, so the result type depended on **cardinality**. That is the same error `_stack_declared_columns` already rejects — "the declared kind decides the storage, not what the values happen to look like".

The `and value` guards were not a kind decision. They were a workaround for a floor of one entry in three places:

| | floor |
| --- | --- |
| `Record` | `raise ValueError("Record requires at least one named field")` |
| `EventTemplate` | `requires at least one field` |
| object batches | `requires at least one element` |

## What changes

The floors are removed rather than special-cased.

- `{}` → an empty `Record` over `EventTemplate()`
- `[]` / `()` → an `OpaqueBatch` with `batch_shape == (0,)`, on a level named for the function

A record is a finite map of fields, and the empty map is one — it is also the **identity of template composition**, which a floor of one left inexpressible.

For the sequence case no element can say what kind the batch holds, and every element spec holds vacuously of none, so the opaque one is not a narrowing. It is also the only batch whose shape stays derivable: `batch_shape` is *derived from* the store, and an object array of zero elements still reports `(0,)`, where a zero-column `RecordBatch` would have nothing to read `(0,)` from.

Two things deliberately unchanged:

- **An empty template is not promoted to `NumericEventTemplate`.** Vacuously every leaf is numeric; that is not a reason to claim it. The predicate already read `if specs and _all_numeric(...)`, so the tie-break was written.
- **A batch still requires a batch axis.** Zero elements is a multiplicity; a single object with *no axis* has no level to count along, and is refused as before.

An empty **array** was never affected — `jnp.array([])` is a well-formed array value and wraps as a `NumericArray` of shape `(0,)`.

## Blast radius

Exactly **three** tests changed — the ones asserting the floors, plus the `OpaqueBatch` sibling. Nothing else in 4824 tests depended on the invariant, which is the measure of how load-bearing it was.

## Corner cases pinned

Line coverage was already 100% on these modules, which hid the gaps: `NumericArray`'s ~40 operators are installed in a loop, so running the loop covers the lines while each operator stays unasserted. Now pinned:

- the three empty returns, and the empty array they are not
- **reflected operators** (`1.0 - arr`) agree with the forward ones
- **in-place operators** rebind to a bare array — `arr += 1` does not mutate the term and does not preserve it
- `_make_stack`'s argument guards: `batch_shape` XOR `n`, and one level name per group of axes

Suite: 4833 passed, 54 skipped. `ruff check` / `ruff format --check` clean.

## Note on reflected operators

I checked whether to drop them. The asymmetry worth knowing about is **not** theirs — it comes from `__array__`:

| left operand | `left - v` | `v - left` |
| --- | --- | --- |
| Python float / int | jax | jax |
| `np.float64` / `np.ndarray` | **numpy** | jax |
| jax array | jax | jax |

numpy coerces through `__array__` and never consults `__rsub__`, so removing the reflected methods would not fix that row — it would only break the Python-scalar and jax rows, which are correct today. The lever for the numpy row is `__array_ufunc__ = None`, which would also stop `np.sin(v)` working. Left as is; the working cases are now pinned.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   call, retain the resulting joint distribution, or materialize and reuse
   samples explicitly when a shared realization is required.
 
+- **An empty return keeps its host's kind (#398).** A `Function` returning `{}`
+  raised, and `[]` / `()` became an `Opaque`, because `Record`, `EventTemplate`,
+  and the object batches each required at least one entry. The kind now follows
+  the host's *type* whether or not anything is in it: `{}` is an empty `Record`,
+  and `[]` / `()` an `OpaqueBatch` of `batch_shape == (0,)` — no element can say
+  what kind it holds, and every element spec holds vacuously of none.
+
+  `Record()`, `EventTemplate()`, and `OpaqueBatch([], level)` are legal as a
+  result. An empty template is **not** promoted to `NumericEventTemplate`:
+  vacuously every leaf is numeric, which is not a reason to claim it. A batch
+  still requires a batch *axis* — a single object with no axis is refused as
+  before.
+
 - **A sweep over numeric rows aggregates as a `NumericArrayBatch` (#398).** A
   `Function` swept over a batch collected numeric rows into a single-field
   `NumericRecordBatch` keyed by the function's own name, so `out["my_func"]`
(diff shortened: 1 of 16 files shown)


jhuggins:

PR #426, opened 2026-08-17
Aggregate a sweep of numeric rows at their own kind

jhuggins:
> **Note.** This replaces #423, which GitHub auto-closed when the branch was
> briefly force-pushed with no commits ahead of its base, and which the API
> refuses to reopen. Same branch, same content.

Stacked on #422 (`dev/sample-draw-level`). Part of #398.

## What changes

A `Function` swept over a batch collected numeric rows into a single-field `NumericRecordBatch` keyed by the function's own name. Both array-row branches of `_make_stack` — the row-wise list path and the mapped `jnp.ndarray` path — now build a `NumericArrayBatch`.

```python
out = my_func(p=design)          # a sweep over 6 rows
out["my_func"]                   # before
out.values                       # now
```

## The batch-row branch covered only one kind

`_make_stack` has a branch for rows that are themselves batches, so a row's own levels end up behind the sweep's rather than being read as event shape. It tested for `RecordBatch` only.

That was harmless until #422, where a scalar law's `sample` began returning a `NumericArrayBatch`. Those rows then fell past the branch into the generic array handling, which read the rows' `draw` axis as event shape and dropped the level — exactly the failure the branch's own comment describes:

> which would take a single-field numeric batch for an array, read its inner batch axis as an event axis, and discard the level names with it

Concretely, `sample(dist_array, sample_shape=(7,))`:

| | before | now |
| --- | --- | --- |
| `batch_shape` | `(4,)` | `(4, 7)` |
| `level_names` | `("dist",)` | `("dist", "draw")` |
| element shape | `(7,)` | `()` |

The record-valued path was already `("dist", "draw")` over `(3, 5)`, so the two agree again. This is what design V.9 states: the outer level ranges over the laws and the inner over each law's draws, "a `RecordBatch` for a record-drawing law, and the batch form of the declared kind otherwise".

## Shared agreement checks

The all-or-nothing and row-agreement checks move into `_agreeing_batch_rows`, used by both kinds. It matches on the batch *family* rather than the exact class, so a `RecordBatch` row and a `NumericRecordBatch` row still stack together — they hold the same thing and only one says so in its name.

## Tests

24 failures, matching the measurement recorded when this stage was scoped. All were the old contract: 12 indexing the aggregate by the function's name, 12 asserting `isinstance(..., NumericRecordBatch)`. No functional gaps.

Fixing the batch-row branch resolved 3 of them before any test was touched.

New coverage in `TestBatchValuedRowAggregation`, mirroring what existed for `RecordBatch` rows:

- array-batch rows stack with the sweep level in front
- array-batch rows disagreeing on their multiplicity are refused
- mixing an array-batch row with a record-batch row is refused

Suite: 4823 passed, 54 skipped. `ruff check` / `ruff format --check` clean.

## Not in scope

`_wrap_declared_function_output` — the path taken when a `Function` carries an `output_template` — still needs the same kind dispatch. That is the next piece.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   call, retain the resulting joint distribution, or materialize and reuse
   samples explicitly when a shared realization is required.
 
+- **A sweep over numeric rows aggregates as a `NumericArrayBatch` (#398).** A
+  `Function` swept over a batch collected numeric rows into a single-field
+  `NumericRecordBatch` keyed by the function's own name, so `out["my_func"]`
+  reached the column. The aggregate is now the batch form of the rows' own kind,
+  read through `.values`.
+
+  A row that is itself a batch keeps its levels either way. Only `RecordBatch`
+  rows did before, so once a scalar law's `sample` began returning a
+  `NumericArrayBatch`, a swept `sample(dist_array, sample_shape=(7,))` read the
+  rows' `draw` axis as event shape and dropped the level. It now gives
+  `batch_shape == (n, 7)` on levels `("dist", "sample")`, as the record-valued
+  case already did.
+
 - **A non-empty `sample_shape` puts the draws on a `sample` level (#398).**
   `sample(law, sample_shape=(5,))` for an array-valued law gives a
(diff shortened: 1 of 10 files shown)


jhuggins:

PR #428, opened 2026-08-18
Pin naming across the tracked terms, batches, and operations

jhuggins:
Stacked on #424. Part of #398.

## Why

Every tracked term carries a `name` and a `name_is_auto` flag. The flag is load-bearing:
it tells a later operation whether the name is a placeholder it may replace or a
statement a caller made. Eleven classes implement that pair separately, and nothing
compared them, so a divergence between any two was invisible.

This adds one table stating the shared rules.

## What it pins

- **A given name is kept verbatim and never marked auto** — across all eleven kinds.
- **Which kinds require a name** (`Record`, `NumericRecord`, `Opaque`, `NumericArrayBatch`)
  and what the rest derive when one is omitted.
- **A derived name builds on the given one** — `posterior[draw=1]`, marked auto, so the
  lineage reads back to the batch a caller actually named.
- **An operation names its result**, and every kind the output boundary wraps takes the
  function's name — numeric, mapping, opaque, callable, sequence, and both empty forms.
- **Levels are named for what mints them** — `sample` mints `sample`, a returned
  sequence mints the function's own name.

## It found a defect

A batch of draws takes the law's name. It was marking that name user-given
unconditionally, so a law that had auto-named *itself* produced a batch claiming a name
nobody chose. It now carries the law's own flag: the name is a caller's statement
exactly when the caller's name for the law was one.

That matches how design V.2 treats a derived joint name (`lik·prior`, marked auto) —
invented names are auto, inherited ones are as inherited.

## Two rows are recorded, not endorsed

The suite documents these as they are, and says so in the test docstrings rather than
asserting they are right:

- Batches other than `NumericArrayBatch` still default their name to the class name
  (`recordbatch`, `opaquebatch`). `NumericArrayBatch` requires one. Aligning them is
  separate work.
- A record-drawing law names its result for the **law**; a scalar law names it for the
  **operation**, because the record law builds its own already-named value and the
  boundary keeps a tracked term as it is.

Both are visible now as two rows of one table, which is the point.

Suite: 4894 passed, 54 skipped (52 new).

### DIFF
--- a/tests/core/test_naming.py
+++ b/tests/core/test_naming.py
@@ -1,0 +1,278 @@
+"""Naming across the tracked terms, the batches, and the operations.
+
+Every tracked term carries a name and a flag saying whether the name was given
+by a caller or derived for it. The flag is what tells a later operation whether
+it may rename: a derived name is a placeholder, a given one is a statement. The
+rules are spread over the classes that implement them, so they are pinned here
+in one place, where a divergence between two kinds is visible as two rows of the
+same table rather than as two files that never meet.
+
+The invariant that ties them together: **a name a caller did not give is marked
+auto, and a name a caller gave is never overwritten.**
+"""
+
+from __future__ import annotations
+
+import jax
+import jax.numpy as jnp
+import pytest
+
+from probpipe import (
+    Function,
+    FunctionBatch,
+    Normal,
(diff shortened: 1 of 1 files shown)


jhuggins:

PR #431, opened 2026-08-18
fix(core): every batch is a workflow operand

jhuggins:
Stacked on #428. Addresses the "generic `Batch` planning and JAX parity" item of the stack review.

## The gap

`build_broadcast_plan` recognised only `RecordBatch` and `DistributionArray`, so a
`NumericArrayBatch`, `OpaqueBatch`, or `FunctionBatch` argument was handed to a body
**whole**. A body written for one element received the whole collection, and the levels
collapsed into the value's shape on the way out.

What makes a value sweepable is that it holds a multiplicity on named levels — the `Batch`
contract, not anything specific to records. The test is now `isinstance(value, Batch)`.
`DistributionArray` is a `Distribution` and not a `Batch`, so it stays separate.

| argument | body receives | result |
| --- | --- | --- |
| `NumericArrayBatch` | a `NumericArray` element | batch on the argument's level |
| `OpaqueBatch` | the stored object | batch on the argument's level |
| `FunctionBatch` | the callable | batch on the argument's level |

## The same assumption, three more places

Widening one site is what exposed the rest, which is the argument for the registry below.

**A mapping transform had no carrier for a single-store batch.** A row returning a
`RecordBatch` crosses inside `_MappedBatchColumns`, an inert carrier whose unflatten
rebuilds it verbatim — which is what lets it survive the added axis that `RecordBatch`
unflattening must refuse, since a shape is not a provenance. `_MappedBatchStore` is that
carrier for a batch holding one store, and lives beside the class it carries.

**The probe read a non-record batch as a law.** It classified any non-`RecordBatch`
argument as a *draw*, so a `NumericArrayBatch` argument failed with an internal contract
error about a missing stochastic plan. It is a swept source, and the probe now says so.

**An operation dropped its operand's levels.** `log_prob(law, sample(law, sample_shape=(3,)))`
returned a bare `(3,)` array. The vectorized path is right — it is the fused
implementation design V.9 allows above the elementwise map — but V.9 also says the batch
axes are *preserved*. The levels the operand stated are now restated on the result, split
by the **operand's** rank so anything the operation added belongs to the element.

## The registry

The spec → tracked class → batch form correspondence was open-coded at six sites, each
knowing a different subset. That is precisely why widening one left the others behind.

`core/_kinds.py` holds the table; each kind registers itself at the bottom of the module
defining its classes, so a new kind becomes known by being imported rather than by being
added to a list elsewhere. Registration refuses a contradicting pair rather than letting
the last writer win, and resolution walks the MRO so a spec refinement inherits its base's
kind. `_column_as_batch` is the first consumer routed through it, replacing a three-way
`isinstance` switch.

`RecordSpec` keeps its own resolver: the choice between the plain and numeric record batch
depends on the template's leaves, not on the spec type.

## Deliberately not finished

Full JAX **vectorization** of a single-store batch. The mapped row body is built from one
leaf per field, which such a batch does not have, so widening it means changing the row
body and the probe's synthesis together. Rather than mis-read the batch, an explicit
`dispatch='jax'` now declines with that reason; `auto` and `sequential` sweep it
correctly, verified to agree on both levels and values.

## Tests

Suite: 5640 passed, 54 skipped. The planner widening changed **no existing test** — the
gap was in what the planner accepted, not in what anything asserted, which is why it
survived this long. New coverage: batches as operands across all three kinds, dispatch
agreement for batch-returning rows, the decline message, the kind table including its
contradiction guard and MRO resolution, and level preservation through `log_prob`.

### DIFF
--- a/probpipe/core/_broadcast_distributions.py
+++ b/probpipe/core/_broadcast_distributions.py
@@ -28,7 +28,7 @@ from ._empirical import (
 )
 from ._function_batch import FunctionBatch
 from ._immutable import constructing, transient_memo
-from ._numeric_array_batch import NumericArrayBatch
+from ._numeric_array_batch import NumericArrayBatch, _MappedBatchStore
 from ._numeric_record_batch import NumericRecordBatch
 from ._object_batch import _from_iterable, _is_object_array
 from ._opaque_batch import OpaqueBatch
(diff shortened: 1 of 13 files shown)


jhuggins:

PR #432, opened 2026-08-18
fix(core): each swept row takes its own kind, and the aggregate is named for its function

jhuggins:
Stacked on #431. Addresses the "per-point host-kind wrapping" and "aggregate naming" items of the stack review.

## Each swept row takes its own kind

A row is one call's return, so the rule that names a single return's kind (design V.0) is
the rule that names a row's. Rows reached the aggregation raw, and the branches there read
them as whatever they happened to stack as. Four wrong answers, not four missing features:

| a swept body returning | was | now |
| --- | --- | --- |
| `{"y": ...}` | `TypeError: cannot aggregate outputs of types ['dict']` | `RecordBatch` |
| `OrderedDict(y=...)` | same, via `Opaque` refusing mappings | `RecordBatch` |
| `[a, b]` | multiplicity became **event shape** | `(3, 2)` on `("row", "f")` |
| ragged numeric rows | silently an `OpaqueBatch` | refused, naming the shapes |

Only the two *structural* hosts are converted per row. A numeric, opaque, or callable row
already reaches a branch that batches it at its kind, and converting it early would cost
the vectorized path for no gain.

The single-value boundary now tests `Mapping` rather than `dict`, so a mapping subclass is
a tree there too.

## The swallowed exception is gone

`except (TypeError, ValueError): pass` around the returned-sequence stack covered nothing —
the line never executed in the whole suite — while turning a genuine schema disagreement
into one opaque value. The stack now has a batch form for every element kind, so what
reaches it and raises is the rows disagreeing, which is the caller's to see.

## An aggregate is named for its function

An aggregate is built by the boundary, not a caller, so its name is derived and must be
marked auto — otherwise a later operation reads it as a caller's statement and declines to
rename it. Three paths disagreed:

- the undeclared record aggregate kept `stack`'s class-name default, so a swept mapping came
  back named `numericrecordbatch`;
- the scalar, opaque, callable, and declared paths marked the derived name as user-given;
- the one-level early return skipped the construction that names it at all. **Removed rather
  than patched** — the reshape it avoided is an identity in that case, so the general path
  already covered it.

| swept body | name | `name_is_auto` |
| --- | --- | --- |
| numeric / mapping / opaque / callable / sequence | `double` | `True` |
| declared `output_template` | `double` | `True` |
| multi-axis sweep | `double` | `True` |

## Tests

Suite: 5649 passed, 54 skipped. New coverage pins each row kind's aggregate, the two
refusals, `Mapping` at the single-value boundary, and the naming table widened across
declared/undeclared, single/multi-axis, and every result kind — the form that would have
shown the divergence in the first place.

### DIFF
--- a/probpipe/core/_broadcast_distributions.py
+++ b/probpipe/core/_broadcast_distributions.py
@@ -20,7 +20,7 @@ import numpy as np
 
 from .._weights import Weights
 from ..custom_types import Array
-from ._array_backend import _to_jax_array
+from ._array_backend import _event_shape_of, _is_numeric_leaf, _to_jax_array
 from ._distribution_base import Distribution
 from ._empirical import (
     EmpiricalDistribution,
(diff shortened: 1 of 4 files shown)


jhuggins:

PR #433, opened 2026-08-18
fix(core): a native-backed batch crosses a JAX boundary

jhuggins:
Stacked on #432. Addresses the "native pytree conversion / validation" items of the stack review.

## A native-backed batch could not enter a trace

`NumericArrayBatch`'s flatten handed JAX the native container. Abstractification fails
there, before `__jax_array__` is ever consulted, so a pandas- or xarray-backed batch could
not be traced at all — while the single-value `NumericArray` beside it could, because its
flatten already converted.

```python
b = NumericArrayBatch(pd.Series([1.0, 2.0, 3.0]), "row", element_spec=..., name="b")
jax.jit(lambda x: x)(b)
# TypeError: Argument '0  1.0 ...' of type <class 'pandas.Series'> is not a valid JAX type
```

The boundary now presents a bare array, as `NumericArray`'s flatten does, and the
conversion is **memoised on the instance** rather than repeated per call. The batch keeps
its native store — converting at the boundary is not materialising the batch:

| | after `jit` |
| --- | --- |
| the traced value | `ArrayImpl` |
| `batch.values` | still `pd.Series` |
| `as_jax()` twice | the same object |

## Unflatten did not check the element it rebuilt

It compared the surviving **rank** and let the trailing axes be anything, so a transform
that resized the event shape produced a batch whose `element_spec` was a false statement
about its own store. It built successfully and failed later, at the first selection:

```
rebuilt.values.shape   == (4, 5)
rebuilt.element_spec   == NumericArraySpec(shape=(3,))   # not what it holds
rebuilt[0]             -> ValueError, one layer too late
```

The element's own axes are not the transform's to change. Changing what an element *is*
rebuilds the batch where the new spec is known, so this refuses with that reason. The two
transformations the contract does allow are unaffected — every batch axis preserved, or
every one removed, which still yields a `NumericArray`.

## Tests

Suite: 5656 passed, 54 skipped. New coverage: a native store reaching a trace, the store
surviving it, conversion memoised, the resized-event refusal, and both permitted round
trips.

### DIFF
--- a/probpipe/core/_numeric_array_batch.py
+++ b/probpipe/core/_numeric_array_batch.py
@@ -86,7 +86,10 @@ class NumericArrayBatch(Batch[NumericArray]):
 
     _values: Any
 
-    __slots__ = ("_values",)
+    __slots__ = ("_jax_cache", "_values")
+
+    #: Derived from the store rather than transported, as for ``NumericArray``.
+    _transient_state = ("_jax_cache",)
 
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


jhuggins:

PR #434, opened 2026-08-18
refactor(core)!: rename ArraySpec to NumericArraySpec

jhuggins:
## Summary

- Hard-renames the public `ArraySpec` API to `NumericArraySpec`. No compatibility alias is provided; the bare backend-array type remains `Array`.
- Aligns design Parts II–V and the package layout with the `raw value → tracked term → tracked batch` hierarchy finalized in #443.
- Updates contracts, contributor guidance, the Unreleased changelog, test names, and public-export coverage to use the final API consistently.

## Migration

Replace imports and type references:

```python
from probpipe import ArraySpec
```

with:

```python
from probpipe import NumericArraySpec
```

`probpipe.ArraySpec` is intentionally unavailable and is not listed in `probpipe.__all__`.

## Design baseline

The design now specifies:

- kind-directed wrapping: array → `NumericArray`, mapping → `Record`, callable → `Function`, other object → `Opaque`;
- `Batch.raw()` as a shared view of concrete backing storage, including object arrays for function/opaque batches and nested raw-column mappings for record batches;
- tracked element results for batch indexing, preserving stored tracked identity and deriving identity when wrapping a stored raw value;
- operation-level `raw=True` as returning the raw value held by the result kind, including the wrapped callable for `Function`;
- non-empty `sample_shape` raw results as the tracked batch storage view;
- `NumericArray.raw` as the underlying-array accessor.

## Non-goals

- No runtime implementation of `NumericArray`, `Opaque`, `Batch.raw()`, or operation-level `raw=True`; those remain in #435, #436, and #439.
- No adoption of the next-stack single `TermSpec` family, `EventTemplate`/`RecordSpec` merge, `InputSpec`/`OutputSpec`, tracked record access, numeric protocol, restructuring, or provenance-tier changes from #443.
- No concrete storage format is invented here for distribution batch types.
- `Function.apply()` remains the raw point-evaluation boundary for ordinary functions; this PR does not add a generic `raw=True` call control.

## Validation

- Focused core tests: `945 passed`
- Full suite: `5575 passed, 48 skipped, 1 xpassed`
- `uv run ruff check .`
- `uv run ruff format --check .`
- `uv run mkdocs build --strict`
- `git diff --check`

All validation commands completed successfully.


SoymilkWinsAgain:
@jhuggins There are two things I'd like to clarify. First, #398 states that `NumericArrayBatch[i]` returns a raw array, but the design document written in #434 states that it returns a `NumericArray` with derived name/lineage. Which one are we following?

The second issue is that design 05 says `raw=True` should return a raw `T`; however, the following table lists the raw type as `NumericArray`, and the batch form as `NumericArrayBatch`. `NumericArray` itself is a tracked class, not a raw array. And the current public signature doesn't even have `raw=`; the raw boundary is `Function.apply()`. This needs clarification on:
- Revising the design document to match `.apply()`;
- Or implementing `raw=True`;
- And whether the raw result of a non-empty `sample_shape` is a raw array or a tracked batch.

SoymilkWinsAgain:
#443 is now finalized. This PR follows that decision by keeping #434 to the hard rename and design baseline, while specifying the raw value -> tracked term -> tracked batch hierarchy, `Batch.raw()` storage views, and tracked batch indexing.

PR #435, opened 2026-08-18
feat(core): give the numeric-array and opaque kinds their tracked classes

jhuggins:
Stacked on #434. Part of #398.

**Purely additive**: new classes and their tests. Nothing that already existed changes
behaviour, so this can be read on its own terms.

## What it adds

Every value spec should have one tracked class and one batch form. Two kinds had neither:

| spec | tracked class | batch form |
| --- | --- | --- |
| `NumericArraySpec` | **`NumericArray`** | **`NumericArrayBatch`** |
| `OpaqueSpec` | **`Opaque`** | `OpaqueBatch` (already existed) |

`NumericArray` holds one array with identity. Its `shape` is the **event** shape — it
carries no batch axes; multiplicity lives in `NumericArrayBatch`. It carries the full array
surface (arithmetic, comparison, conversion hooks) because with no fields and no field
count `arr + 1` has exactly one meaning, which is what lets `Record` stay a container.
**Arithmetic yields a bare array**: identity is attached by operations, and arithmetic is
not one.

`Opaque` adds identity and nothing else — no attribute forwarding, no `__call__`, one
explicit accessor. A box that forwards is indistinguishable from what it wraps.

## Points worth review

- **Native-form storage.** Construction validates against container metadata and does not
  convert, so a lazy or disk-backed value is not materialised merely to be named.
  Conversion happens at the compute boundary through a set-once cache.
- **The pytree aux carries the declaration.** It is not re-read off the arriving array:
  `is_valid` admits a same-kind cast, so a float32 value under a float64 declaration would
  otherwise come back declaring float32.
- **`NumericArrayBatch` requires a name.** A batch is what an operation hands back, and the
  name says which one it is; a class-name default would name every batch in a pipeline
  alike. The other batch classes do not yet follow this — that is the fifth PR.
- **`OpaqueSpec` moves** into `_opaque.py`, beside the class it names.

Suite: 5534 passed, 54 skipped.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -111,6 +111,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Selection yields a `NumericArray` under the derived name, as `RecordBatch`
   yields a `Record`.
 
+- **`Opaque` — the tracked class of the opaque kind (#398).** What an operation
+  returns when its declared kind is an `OpaqueSpec`, completing the pair with the
+  `OpaqueBatch` that already existed. It adds identity and nothing else: no
+  attribute forwarding, no `__call__`, no operators — the wrapped value is
+  reached through `.value`, explicitly. `OpaqueBatch`'s element contract is
+  unchanged; it still hands back the object the caller put in rather than
+  wrapping it.
+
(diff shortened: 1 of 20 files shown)


SoymilkWinsAgain:
Resolved conflict in `CHANGELOG.md`

PR #440, opened 2026-08-18
docs: say what the boundary and the batch layer now do

jhuggins:
Stacked on #439. Part of #398. Documentation only — no behaviour changes.

The prose still described the model these PRs replaced.

## `docs/api/records.md`

Said **a numeric array needs no class**. That stopped being true when native storage and
identity were separated: the store batches natively, and `NumericArrayBatch` is what
carries a level name, a specification, and provenance.

Said an object batch **needs at least one element**. The rule that survives is about
*axes* — zero is a count a level can carry, and what has nowhere to count is a single
object with no axis at all.

`NumericArray`, `NumericArrayBatch`, and `Opaque` are added to the page; they were absent.

## `CONTRIBUTING.md`

Described the old `Record | RecordBatch | Distribution` contract, and the single-field
`NumericRecord` keyed by the function's name. Replaced with the rule now in force: a
return becomes the tracked term of the kind it already is.

## Docstrings

`_workflow_result`'s module docstring described the same superseded contract, and both it
and `_function_contract` deferred to a carve-out that has since been removed.

`_make_stack`'s header table listed a fallback that no longer happens — opaque values do
not become a plain list with an error, they become an `OpaqueBatch`. The table now names
every kind and is seven lines shorter.

## Also

Merges the two consecutive `### Changed` headings the CHANGELOG picked up while these
branches were being merged together.

`mkdocs build --strict` clean. Suite: 5677 passed, 54 skipped.


jhuggins:
Folded into the PRs that own the code it documents, so a reviewer reads a change and its documentation together: the class-identity prose and the `NumericArray`/`NumericArrayBatch`/`Opaque` API entries into #435, the boundary rule and the result-contract docstring into #436, and the `_make_stack` header into #437. Nothing is left over.

PR #442, opened 2026-08-19
fix(core)!: one answer per computation, whatever ran it

jhuggins:
Stacked on #439. Part of #398. Closes the gaps a second external review of the stack found, plus the ones measuring them turned up.

Every fix here is the same shape: an answer that depended on something it should not have — which executor ran the call, which operation was named, which kind the value happened to be. Each is now stated once and read the same way everywhere.

## What changed

**One answer per computation, whichever executor runs it.** The row-wise sweep gave each row the kind of its own return; the mapped path did not, so a body returning a mapping raised `_make_stack cannot aggregate output of type dict` under `dispatch="auto"` and returned a `NumericRecordBatch` under `"sequential"`. A sequence row raised a bogus count error. The mapped path now applies the same per-row rule and carries a record row across the map the way it already carried a batch one. A declared `output_template` still wins, as it does row-wise. The tests for this are a parametrised dispatch matrix rather than a second test, since agreement is the property being asserted.

**Every batch is read the way the `Batch` contract states it.** Lifting a function's declaration against a batched operand read `event_template` — the convenience view only a batch of records has — so a `NumericArrayBatch`, `OpaqueBatch`, or `FunctionBatch` operand could not be swept by a declared function at all, however its declaration was written. It reads `element_spec` now, which the ABC states at every kind; the unification pass already had the spec-against-spec path this needs.

That view was also unsound. It unwrapped a single-field element to that field, so `EventTemplate(v=())` — a declaration naming a bare array — accepted an operand whose elements are records. Reading the element spec refuses it and names both specs.

A distribution operand is deliberately left alone: it is lifted by being sampled, and a scalar law reports its event as one field named after itself. Restating that belongs with the `Distribution`-side work.

**Every density op keeps its operand's levels.** `log_prob` restated them; `prob`, `unnormalized_log_prob`, and `unnormalized_prob` returned a bare array, so the axes survived only as event shape. The same draws scored as a batch over `("chain", "draw")` under one op and as one value of shape `(2, 3)` under another. All four share the helper the `log_prob` work introduced for exactly this.

**A row of unstackable elements keeps its level.** The row-stacking path knew two of the four batch forms, so a row returning a sequence of opaque objects or callables had its own batch stored whole as one element: the row's multiplicity vanished, and a row of callables came back as an `OpaqueBatch`. Object batches now get the stacking path their storage calls for. An empty-sequence row is a batch of nothing on its own level, matching how a single return already reads one — which settles the last dispatch disagreement. Two nested anonymous levels stay refused: both would take the function's name, and a clash is not resolved by suffixing it.

**`sample` mints its level for every kind of draw.** *Breaking.* The boundary assumed a law that assembles its own draws also names what they range over. An empirical does not: `sample_shape=(3,)` came back as one `NumericRecord` whose fields had grown an axis, or — for record and opaque atoms — as a single `Opaque` holding the whole object array. All three now give the batch form of the draw's kind over a `sample` level, matching what a plain law already gave.

The empirical's own comment called returning a record "a deliberate deviation from the uniform output wrap contract", because a batch inside a `vmap` is refused. That reason holds for `_sample` and is now stated as such: the level is minted at the boundary, outside the trace, so the deviation is gone without putting a named multiplicity where `vmap` would reject it.

**The kind table answers where it already can.** `RecordBatch` construction listed the admissible field kinds inline while the reading end asked the registry — the split the registry was introduced to close.

## Breaking changes

- `sample(law, sample_shape=...)` on an empirical returns a batch. Code that read `drawn["field"]` now reads a column of a batch rather than a field of a record, and `"field" in drawn` becomes `"field" in drawn.event_template`. A single draw is unchanged.
- A batch of one-field records no longer satisfies a declaration naming a bare array.
- A row of callables aggregates to a `FunctionBatch`, not an `OpaqueBatch`.

## Documentation

Updated alongside the code rather than separately. Requiring a name on every batch had left three notebooks constructing one without a name — they would have raised when CI executed them — four docstrings offering "defaults to the class name lowercased", and the empty-batch example written as a call that no longer constructs. The `_sample` protocol table promised a `NumericRecordBatch` for a batched structured draw, which no implementation returns and none should; it now says the columns come back, and why.

## Verification

- 5720 passing, 52 skipped; `ruff format` and `ruff check` clean; `mkdocs build --strict` clean.
- `docs/user_guide/02_records.ipynb` and `docs/user_guide/12_diagnostics_workflow.ipynb` execute clean under `nbconvert`. `docs/tutorials/getting_started.ipynb` gets past its batch-construction cell and stops later on a locally missing `pymc`, which the tutorial CI leg installs.

## Not addressed

`quantile(dist, q)` drops the levels of a batched `q` for the same reason the density ops did, but its result is a per-field record rather than an array, so restating them needs the record-valued form of the helper. Left out rather than half-done.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -258,6 +258,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **A declared function can be swept over any kind of batch (#398).** Lifting a
+  declaration against a batched operand read `event_template`, the view only a
+  batch of records has, so a `NumericArrayBatch`, `OpaqueBatch`, or `FunctionBatch`
+  operand raised `does not expose an authoritative event_template for lifting`
+  however its declaration was written. It now reads `element_spec`, which the
+  `Batch` contract states at every kind. Two consequences: every batch kind is
+  read the same way, and a batch of one-field records no longer satisfies a
+  declaration that named a bare array — the record-only view unwrapped a
+  single-field element to its field, so `EventTemplate(v=())` accepted an operand
+  whose elements are records. A distribution operand is unchanged: it is lifted by
+  being sampled, and its event template remains what the draw is checked against.
+
 - **A swept row's kind no longer depends on which executor ran it (#398).** The
   row-wise path gave each row the tracked class of its own kind; the mapped
   (`jax.vmap`) path handed its rows to the aggregation raw, so a body returning a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -106,7 +106,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   and `[]` / `()` an `OpaqueBatch` of `batch_shape == (0,)` — no element can say
   what kind it holds, and every element spec holds vacuously of none.
 
-  `Record()`, `EventTemplate()`, and `OpaqueBatch([], level)` are legal as a
+  `Record()`, `EventTemplate()`, and `OpaqueBatch([], level, name=...)` are legal as a
   result. A *batch* of empty records is not: a batch reads its multiplicity off
   a column, and a zero-field element supplies none, so `RecordBatch` still
   requires at least one field. An empty template is **not** promoted to `NumericEventTemplate`:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -258,6 +258,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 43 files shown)


jhuggins:
Folded into the PRs that own the code each fix touches, rather than sitting behind them as a sixth PR: the mapped-sweep, lifting, density-op, unstackable-row, and kind-table fixes into #437, whose code they correct; `sample` minting its level into #436, which introduced that boundary helper; and the batch-name documentation into #439, which introduced the rule. Nothing is left over — the folded #439 tip is content-identical to this branch, differing only in which CHANGELOG section the `sample` entry sits in (it is breaking, so it moved up).

Issue #443, opened 2026-08-21
Design proposal: one spec family, InputSpec/OutputSpec symmetry, and tracked-by-default access


jhuggins:
**Status: agreed direction — drafted into the design reference on #444; not yet in the code.** Ten changes to the value and operation layers of the design reference. The first four simplify the spec layer — one spec family, one schema class, and a symmetric pair of input/output declarations shared by functions and kernels; the fifth and sixth make tracked identity the default at every access point, with `raw` as the single escape hatch; the seventh unifies the numeric kinds behind one protocol; the eighth lets realignment restructure; the ninth binds provenance cost to one mode; the tenth gives kernel mixing its own operation. Capabilities, dispatch and registries, and `Batch` and named levels are unchanged. *(Revised after a design-review pass: items 3, 6, and 7 sharpened; items 8 and 9 added; presentation note at the end. Second revision, with the reference draft on #444: item 10 and the `LinOp` accessor consequence in item 7 added; #444 also documents the workflow RNG system as shipped — scope-based seeding, structural event keys, replay. Third revision, after the review pass on #444: item 5 sharpened to the storage/source split and the one `raw` semantics — as-produced results, `raw()` on the tracked base — and item 7 gained the `NumericSpec` mixin.)*

Section references (II.2, V.0, V.2, …) are to the design reference as it stood on the open consolidated stack (#434–#437, #439), against which these changes were argued — that is also where the kind table (`register_kind`, `probpipe/core/_kinds.py`) lands. #444 (branch `docs/one-spec-family`) carries the revised reference text, under its renumbered sections.

## 1. One spec family: `TermSpec` becomes the base

[...]

PR #444, opened 2026-08-23
docs(design): one spec family, uniform access, scoped randomness, and an operation definition

jhuggins:
## Summary

Revises the design reference to the core simplification of #443, and to the decisions taken while writing it. Documentation only.

The scope grew past #443's spec-layer changes because review, chiefly #446, showed that the reference could not be implemented from as it stood. The problems were structural rather than local, and each fix is stated once and pointed to everywhere else. The reference is now seven parts: Part III is Term Kinds, the new Part IV is Distributions, and Functions, Operations, and the Catalog are Parts V through VII, with files renamed to match.

## What was wrong, and what changed

- **The call path had no executable shape.** VI.0 stated four laws and V.1 described the engine in a paragraph; neither said what runs first, what each stage requires, or where a call fails. V.1 now states one **stack** of eight steps, configure, bind, normalize, classify the lift, plan, resolve, execute, return, and Part V has one section per step in execution order, each ending with its requirement and its failure. Bind and admit were split into what they are: controls resolved through `with_options` first, then binding, then wrap, plan conversion, admit.
- **The design's structure did not match the code's.** The sequence sat in VI.0 and was assigned to `operations/_operation.py`, but the engine that runs it lives below the operations and runs for every `Function`. The sequence moved to Part V, VI.0 shrank to what an operation adds at three points, roles by kind, a result rule with its applicability conditions, and routes, the evaluation-rule registry moved from VI.1 to the resolve step, and package-structure names which module holds which step and where every existing module goes.
- **Failures were unnamed.** "Fails applicability" and "resolves to nothing and raises" were phrases. `ApplicabilityError` and `ResolutionError`, the latter shared by every II.7 registry, give each step an assertable outcome and settle the regression behind #446 A2, where a bare object with a `_sample` method crashed on a missing attribute instead of being refused.
- **One word meant several things.** "Raw" meant four mechanisms, "route" meant both an operation's realization and a dispatch mode, and "admit" bundled three operations. Raw is now the representation layer detached, route is what the engine selects, dispatch mode is how points run, and normalization names its sub-steps.
- **Two sorts of name were one word.** A term's label and the component name a consumer matches on were both "name", so what an omitted declaration was named had no single statement and a rename could plausibly change what a joint contained. C5 now separates them as a principle: a label has no mathematical meaning, and a component name is required for each mathematically distinguishable component. A record's components are its fields; any other term has one component, which defaults to its label captured once at construction, so `with_name` never moves it. The explicit forms are `component_name` on the family constructors, `OutputSpec(mean=None)` on a function, and a function's `output_name` for the label its results receive.
- **Rules were stated in many places and drifted.** Twenty patterns were restated across parts with small differences, and the consistency defects review found were the visible ones. Every cross-cutting rule now has one owner, and a check of headers, cross-references, and principle citations runs before each commit.
- **Sections were out of dependency order.** Parts II, III, and V introduce each object after what it depends on, and forward references are down to a handful of deliberate ones.
- **Existing public API had no stated fate.** Seventy exported names never appeared in the design. Each now has a placement, a rename, or an explicit retirement, and the correspondence table covers every module with a design contract.
- **The prose buried the contract.** Compression and voice passes cut about a tenth of the text, and every section leads with a contract that can be lifted into code.

## What changes, by part

**Part I** gains a *three layers* section — semantic, representation, workflow — derived from the principles rather than added as a fourth one. It defines a value's two presentations by layer: the **tracked** form is the semantic layer's presentation of a value and the **raw** form the representation layer's, the tracked form carrying the raw form together with the workflow record. The layers then fix a new principle family, the **boundary principles**: B1, either presentation in, the declared kind directing the wrap; B2, representations only inside, so the representation layer never depends on the workflow layer; B3, tracked forms out by default; B4, no copying at boundaries. Each replaces a rule that had been restated per site — fifteen "this accepts raw or tracked" statements, several "without copying" claims — and Rationales cite them as `B1 – …`, like the C's and D's. C5 is restated as a principle rather than a description: an object's label has no mathematical meaning, and a component name is required for each mathematically distinguishable component.

**Part II** is generic and type-agnostic throughout, and ordered by dependency, opening with the type layer everything else is typed by: II.1 `TermSpec` · II.2 the `InputSpec`/`OutputSpec` declarations · II.3 `Numeric`/`NumericSpec`/`Constraint` · II.4 identity · II.5 `Batch` · II.6 `NamedTree` · II.7 dispatch. `NamedTree` moves late because nothing in Part II builds on it.

- **One spec family.** `TermSpec` is the base; the raw-value/term-spec split retires with `ValueSpec`, and *value* stays reserved for the mathematical kind. The kind rule is scoped to the six **base kinds** — one spec, one tracked class, one batch form each — with the batch kind stated as its own case: a batch of batches is a batch carrying an added level.
- **The declarations** are their own section: `InputSpec`, a flat mapping of independently bindable slots, and `OutputSpec`, which declares the returned term's kind together with the named **components** it exposes: `OutputSpec(beta=spec)` is one whole term named `beta`, `OutputSpec(RecordSpec(...))` exposes the record's fields, and several keywords declare such a record. Object names and component names are independent, so composition never matches on a label. `OutputSpec(mean=None)` is a **type hole**, the component `mean` declared with its spec pending, filled by a constructor from its parameters or by the engine from the return. II.2 defines the declarations only; how an omitted or bare declaration is completed is stated where it happens, III.3 and V.6 for functions, III.7 for laws, and II.5 for batches. Neither declaration is a `TermSpec`.
- **`Numeric` and `NumericSpec`** are the value- and spec-side halves of one flat-vector interface, both abstract bases. II.3 states the rule that governs the choice: an interface a closed set of ProbPipe kinds implements is a base, while an open claim any object may make is a structural protocol. A `NumericSpec` has two roles: as a type it identifies numeric structure from declarations alone, which is how record promotion, the numeric-distribution marker, and differentiability claims read it, and as an interface it adds `vector_size`, so the flat dimension is known from the spec's shapes before any value exists; `to_vector` and `from_vector` belong to the kind.
- **One identity base.** `Annotated` merges into `TrackedTerm`, so every tracked term carries name, spec, provenance, and annotations, and a batch of draws is annotatable like any record.
- **Access and raw are one rule, stated once.** Reaching into any container returns a **view** named from the accessor, marked `name_is_auto`, its provenance recording the container and any supplied source; `raw()` is the single door out, returning the value **detached** — without provenance, annotations, or container references, keeping its spec and name. II.4 owns this and the immutability convention (every `with_*` returns a new term sharing the representation); the per-kind restatements are gone, each kind saying only what it alone can.
- **Registries have one mechanics**, in II.7. Selection order is fidelity, then priority within the tier, then specificity, then registration order, in every registry. A method *declares* its fidelity where it is registered, fixed for the method's life; `priority: int | None` ranks within the tier, and `None` — the default — is opt-in-only, so registering a method never silently changes what runs; `set_priorities` re-ranks within a tier and cannot change a fidelity. `Fidelity` is one two-tier enum, exact and approximate, shared by converters, evaluation rules, inference methods, and routes; sampling is a property of a method rather than a third tier, so a finite MCMC posterior is approximate, and `ConversionMethod` retires into it. A `check` may report **unresolved** when the declarations it needs are not yet available, and a known mathematical nonexistence raises `MathematicalDomainError`, kept distinct from `ResolutionError`. Each registry section keeps only its key type, its entries, and its own semantics.
- II.6 gives `with_path_names` path-valued targets, so realignment can restructure, and fixes the edge semantics (snapshot simultaneity, collision and cycle raises, deterministic ordering). II.4 states boundary-attached identity and binds the fingerprint tier to the provenance mode. II.1 owns symbolic dimensions: names form one scope wherever specs meet, within a schema, between a batch and its element, across a map's slots, and across the operands of a composition, so a name two factors share is one dimension, bound at its first occurrence with later ones required to agree, and two dimensions that are different quantities are renamed apart with `with_dim_names`, added to the protocol and to both law classes. II.5's `BatchSpec` carries the element **declaration** rather than a bare spec, so a name key on any batch addresses a component of every element, the field column for a record element and the batch itself for a whole term.

**Part III**, now **Term Kinds**, gives each kind its own section, in dependency order: III.1 `NumericArray`, III.2 `Opaque`, III.3 `Function`, III.4 `LinOp` — beside the base it subclasses — then III.5 records, III.6 record batches, III.7 `Distribution`, III.8 its capabilities, III.9 `ConditionalDistribution`, and III.10 the distribution batches. Each section states only what its kind adds; everything else is Part II's contract, so the per-kind batch class blocks that re-declared `spec` and `__getitem__` are gone. Storage and access are separated in the record contract: storage retains the representation and the supplied term as the field's **source**, while access returns a field-named view. The draw type `T` is defined once, derived from the event spec's kind. **Components and fields**: a law's produced components are its `event_spec.components`, only a record-valued draw has fields, and `with_path_names` renames components or event fields without changing the event's kind. `Distribution.event_spec` is an `OutputSpec`; a kernel's given side is an `InputSpec`; composition runs on produced slots with same-named unmet givens unifying. `event_spec` and `output_spec` are the same declaration type named apart deliberately: a function's output is what one call returns, a law's event the space its draws inhabit. III.3 separates a function's three names: `name` labels the callable, `output_name` labels its results and defaults to the initial `name`, and a whole-term result's component defaults to `output_name`, with `OutputSpec(mean=None)` naming it otherwise. III.3 also defines the **capabilities** a `Function` can claim, `SupportsInverse`, `SupportsLogDetJacobian`, and `is_invertible`, parallel to III.8 for distributions, and III.4 states the operator's claim guarded by squareness. In III.7 a bare term spec completes at construction, a record to its fields and anything else to a whole-term event whose component is the law's `name`, captured once, with the families' `component_name` overriding; `d[name]` addresses a component whatever the packaging, the field marginal for an exposed record and the law itself for a whole term.

**Part IV — Distributions** is new and carries what Part III held as III.11–III.15: what the distribution kinds add beyond their definitions. IV.1 factored distributions, IV.2 composition, and IV.3 cross-type conversion. Composition shares one dimension scope across its operands (II.1). IV.3 opens with the converter and the registry, states what a conversion preserves, the event declaration and the law up to the recorded fidelity, and leaves planning and execution to Part V. The former hierarchy section is dropped: draw structure is declared in `event_spec` and factor access by `SupportsFactors`, and the families are Part VII's catalog. Distributions and functions now each have a part for what they add beyond their definitions.

**Part V** documents the **workflow RNG system as it shipped** (V.8): randomness is scoped rather than carried, so no object holds a seed. `workflow_run(seed)` opens a scope, every ProbPipe-caused draw is a workflow-owned random event keyed by structural identity, no call takes a key, and `replay_run(provenance)` re-executes one recorded call, validating rather than approximating. Caching is stated as a workflow-layer option with no semantic effect. The `seed` control is gone from `Function`, matching the code. Part V is now organized around the engine's **stack**: V.1 states the `Function`, the three declarations the engine reads from it, what each parameter accepts, the result declaration, and the realization, and the eight steps in order, and each step then has its own section with what it requires and what happens on failure. Controls are step 1, resolved from the framework's defaults, the decorator, and a `with_options` view (V.2); binding is step 2 (V.3); normalization is step 3 in three sub-steps, wrapping raw hosts into their kinds through the kind-directed table and the array-backend registry, planning the conversion of a distribution whose class differs from the one its parameter names through the converter registry under per-parameter `conversions` controls, the conversion itself constructed at execution so that `check` never samples, and admitting each argument against what its parameter accepts (V.4); then lifting (V.5), planning (V.6), resolution with the evaluation-rule registry moved here from VI.1 (V.7), randomness (V.8), execution under a dispatch mode (V.9), return (V.10), the differentiability claims (V.11), and constraint reparameterization, moved from the distribution layer since a bijector is a `Function` (V.12); the inverse and Jacobian claims it relies on are defined on the `Function` kind (III.3), where the change-of-variables rule of V.7 finds them. Two failures are named because a caller can act on them: `ApplicabilityError` and `ResolutionError`, the latter shared with the II.7 registries; a wrong returned kind and a violated declaration raise `ResultKindError` and `ResultSchemaError`, which are route defects. A backend object entering at a parameter carries no event declaration, so the parameter's declared spec supplies it (V.4). Planning computes the result declaration as far as the declarations determine it and defers to return what only the return can settle, so `check` reports each route as feasible, infeasible, or unresolved.

**Part VI** replaces the operation checklist with a definition. An operation is a `Function` that declares **operand roles**, **applicability conditions**, and a **result rule** computing the result's full spec before execution; it is realized by registered **routes** or, when derived, by an identity over other operations. VI.0 states what an operation adds at the three points the engine's sequence reads (V.1): operand roles by kind at admission, applicability conditions and a result rule at planning, and routes at resolution and execution. Admission refuses a bare object at a distribution role whatever methods it carries, a capability route's check is protocol membership plus the capability's guard where support is partial, and an implementation tests nothing about its operand. Routes come from four sources (structural, capability, registry, fallback) in any combination, so a capability route names the operand it dispatches on and no operation needs a single distinguished subject. The stored fallback status is retired, since a route's presence is a fact about the registry at call time. `raw` is a `with_options` control defined as the result detached, so it cannot select a different route or a different draw. Controls are derived rather than authored: a parameter is authored exactly when the result rule reads it. `op.check(...)` runs the sequence through resolution without executing, reusing the name the converters and inference methods already use.

It now also says how an operation is *written*. The `@operation` decorator takes the result rule and registers the operation; a primitive's body is empty, a derived operation's body is its identity, and a derived operation may carry direct routes besides, its identity being the floor. Routes register against the operation at import through four construction helpers over one `OperationRoute` shape, bound to a `BoundCall` whose `specs` are what `check` reads and whose `operands` only `execute` reads. Operand roles are declared by kind — the spec *class* — not by an `InputSpec`, since an operation is generic over a kind. The operations are themselves a registry: `operation_registry.list()` returns typed summaries (operands by accepted spec class, routes with source and fidelity) and `describe()` renders them, cataloged beside the other registries.

Part VI is ordered functions before distributions, each operation after what its routes need: VI.1 `evaluate` · VI.2 `inverse` · VI.3 `sample` · VI.4 `log_prob` · VI.5 distribution functionals · VI.6 `condition_on` · VI.7 `joint` · VI.8 `marginal`/`factor` · VI.9 `mixture` · VI.10 `convert` · VI.11 batched. `mixture(K, μ)` is split out of `condition_on`, defined as the kernel's reconstructed output projected from the composed joint — the operation formerly called `predictive`, renamed because the result *is* the mixture, which covers predictive, compound, and state-propagation uses without privileging one. VI.6 ranks inference methods on the shared two-tier scale, a finite MCMC posterior being approximate, in place of integer tiers, and states its three cases, currying an exogenous given, slicing a produced field exactly, and Bayes' rule through the inference registry, by whether binding leaves a likelihood over unconditioned variables rather than by graph position. Per-operation restatements of the Boundary law are cut; each section keeps what only it can say.

**package-structure** follows: the RNG modules are mapped, `values/` covers the value layer, `core/_numeric.py` holds the `Numeric` pair, `functions/` holds the call sequence step by step, `operations/_operation.py` the decorator, route protocol, route resolution, and registry, and the module list is ordered to match Part VI. `functions/_reparameterization.py` holds the bijector factory, and the function capabilities sit with the `Function` base in `values/_function_base.py`.

## Consolidation

A pass over the whole reference found twenty patterns where one rule was stated repeatedly for slightly different kinds or parts of the system. Each now has one owner — the view and naming rule and boundary-attached identity (II.4), the batch contract and level naming (II.5), unification (II.1), registry mechanics and fidelity (II.7), record-shaped positions (III.5), the draw type `T` (III.7), the return step of the call sequence (VI.0) — and every other site either says only what is specific to it or points. Forward references are down to four deliberate ones.

A second pass compressed the long blocks: sentences that restated a rule owned elsewhere, code comments repeating the prose above them, and rationale that restated its contract are gone, and the eight files are about a tenth shorter overall. A third pass rewrote Parts II–VI in plain declarative prose — word connectors rather than appositive commas, parentheticals folded in or split, two examples where a list only illustrated — and a vocabulary sweep replaced metaphor with the precise term throughout (a route has the *interface* of a dispatch method; advertised support *matches actual support*; `raw()` is the *access point* to the representation layer; `shape` now means an array shape only). Two conventions from these passes are recorded in README: a code comment states what the signature does not, never the sentence above the block; a Rationale names the principle and the reason, never the contract.

## Notes for review

- **Name reuse:** `TermSpec` today names the marker subfamily; here it names the base. The rename is atomic with the marker's removal in this diff, so the name never carries two meanings at once — the same must hold in the implementation.
- **`LinOp` drops `input_template` / `output_template`.** Its domain and codomain schemas are read through the declared sides it inherits as a `Function`, so *template* survives only in `differentiable_template`.
- **Decisions from the stack are carried over**, not reverted: `FunctionBatch` is `Batch[Function]` and `OpaqueBatch` is `Batch[Opaque]`, `from_vector` takes `axes_per_level`, `stack` may derive its name from the first record, and a `RecordBatch` field column is the tracked batch rather than a bare array.
- **`set_priorities` encoding.** `core/_registry.py` and its tests still encode tiers as integers (above 50 exact, 1–50 inexact, 0 opt-in-only). The design declares a two-tier `Fidelity` per method, exact or approximate, with `priority: int | None` ranking within the tier; the capability matches, the encoding does not.
- **Also placed:** the flat view of a numeric law is `evaluate(to_vector, d)` (III.7), retiring the flat distribution classes; the array-backend registry has its contract in II.3; designs get their own package, generalized to any element spec, with their section still to be written. A `convert` target may be a capability protocol (IV.3, VI.10), and `expectation`'s `return_dist` with its global default is retired in favor of the explicit bootstrap (VII.2).
- **Today's public API, placed.** `prob`, `unnormalized_prob`, `random_log_prob`, and `random_unnormalized_log_prob` join VI.4, the first two as derived operations; `from_distribution` becomes the `convert` operation (VI.10) with the converter registry as its route; inference-method parameters are controls (VI.6); Stan and PyMC models are program-defined models, a data-bound parameter target as a `Distribution` or an unbound-data kernel as a `ConditionalDistribution` (VII.9); the likelihood protocols and the simple model classes are retired; `Module` and `workflow_method` are experimental with placement to be decided.
- **Implementation note for `diagnostics/` and `validation/`.** Numeric exports read the declared component names and packaging (II.2), not object labels or a hard-coded `x`.
- **New citation family.** `B1`–`B4` are cited in Rationales as `B# – Name`, like the C's and D's, and never in a Contract.
- **Issue #446, Part B.** The design side of B1–B3 is carried by II.1 (the kind table the resolver implements), II.4 and II.5 (the per-kind raw contract and the storage view), II.5 (a field column is a view with its container's levels and the field's spec), V.4 and VI.0 (admission refuses a bare object at a distribution role, and a capability route's check is protocol membership plus its guard), V.6 (the swept batch's `element_spec`), and V.4 with III.2 (a list, tuple, or set return is opaque). `ApplicabilityError` and `ResolutionError` are target-state names with no counterpart in the code yet.
- Target-state reference per `design/README.md`'s convention: the code intentionally lags until the implementation stacks land.

Closes the design-doc side of #443.


jhuggins:
Restructured per review: Part II now holds only generic, type-agnostic machinery — every concrete spec is defined beside the kind it describes (the doc-level mirror of the package rule that a spec lives with the type it admits), `InputSpec`/`OutputSpec` stay introduced in II.2, and the old schema section is dissolved (Identity II.3, `Batch` + `BatchSpec` II.4, dispatch II.5). Part III gives each base kind its own section — III.1 `NumericArray`, III.2 `Opaque`, III.3 `Function`/`SupportsDifferentiation` — with records following at III.4 (`RecordSpec`, `Record`, `NumericRecord`); downstream sections renumber through III.15, and all cross-references across the reference and package structure are swept.

design/03-values-and-distributions.md
```diff
@@ -272,51 +320,52 @@ class LinOp(Function, ABC):        # the linear subtype of the III.1 base
...
 
-A `NumericDistribution` is a `Distribution` whose draws are numeric, so it carries a `NumericEventTemplate` and can use the flat-vector machinery.
+A `NumericDistribution` is a `Distribution` whose draws are numeric, so it carries a `NumericRecordSpec` and can use the flat-vector machinery.
```
SoymilkWinsAgain: `NumericDistribution`, `LinOp`, `cov`, etc. only accept `NumericRecordSpec`. However, the spec of `Normal` is infact `NumericArraySpec`:

```text
Normal.spec -> DistributionSpec -> event_spec: OutputSpec -> spec: NumericArraySpec
```

Thus, we should not automatically assume that the type is `NumericDistribution` is equivalent as the event spec is exactly `NumericRecordSpec`.

design/05-operations.md
```diff
@@ -34,8 +34,8 @@ The operation model is where the core principles become mechanical. Capability d
...
+- `cov(d)` requires a numeric draw and returns a covariance operator over the *flattened* draw, a `(vector_size, vector_size)` `LinOp` rather than an event-typed value, since covariance couples distinct coordinates. Its input and output schemas are both the distribution's `NumericRecordSpec`, so it applies directly to draws.
 - `quantile(d, q)` requires a numeric draw. It takes a level `q ∈ [0, 1]` or array of such levels and returns the quantile for each, computed per coordinate for a multivariate draw. If a single level is provided it returns a `Record`; for multiple levels, it returns a `RecordBatch`. Both require a numeric draw, so the record form is the only one in question here.
```
SoymilkWinsAgain, line 40: `quantile` has the same problem as I commit in 03/L345

design/03-values-and-distributions.md
```diff
@@ -2,215 +2,269 @@
...
     def select_all(self) -> dict[str, Any]: ...   # every top-level field, ready to splat
 ```
 
-`select` resolves each argument with `at_path`, so a key reaches a leaf and a partial path a subtree view, and returns a plain `dict` carrying no schema; its purpose is `**`-splatting a value's parts into a `Function` call (Part IV), with `select_all` the whole-record form over the top-level children.
+`select` resolves each argument with `at_path`, so a key reaches a leaf and a partial path a subtree view, and returns a plain `dict` of tracked values carrying no schema; its purpose is `**`-splatting a value's parts into a `Function` call, with `select_all` the whole-record form over the top-level children.
+
+**Access is tracked.** `record[path]` returns the field as the tracked term of its kind — a `NumericArray` for an array field, the stored term itself for a term-valued field — under the field key as its name, marked `name_is_auto`, with view provenance; an interior path yields a sub-`Record` view as before. `record.raw(path)` returns the stored representation, and `record.raw()` the whole record's nested mapping of raw leaves — the record kind's raw host.
```
SoymilkWinsAgain: We should clarify whether `record[path]` returns the stored tracked term or a newly created field view.

These requirements cannot all hold for an immutable term. Line 163 says that a supplied tracked term is stored with its identity—including its name and provenance—preserved. This paragraph then says that access returns “the stored term itself,” while also renaming it to the field key, setting `name_is_auto=True`, and attaching view provenance. Returning the same object cannot change that metadata; returning a copy or view would no longer be returning the stored term itself.

The same ambiguity affects `NumericArray`: a supplied tracked array is said to retain its identity, but line 210 requires numeric leaves to be stored in native form, which implies unwrapping it.

We should therefore define storage and access separately: what is retained internally, whether access creates a new contextual view, and how the original term’s provenance remains reachable. One coherent model would be to preserve the supplied term as the source, while `record[path]` returns a new field-named view whose provenance references both the record and that source.

design/04-functions.md
```diff
@@ -109,34 +129,62 @@ This is `C4 – Function lifting via pushforward` realized: replacing any argume
 
 - *Default sample count.* How many draws a broadcast takes by default (a speed-versus-accuracy ceiling, with an explicit per-call override always available) is unsettled. The default should signal "rough estimate," not "tuned."
 
-## IV.3 — Controls vs. arguments
+## IV.3 — Randomness: workflow scopes and structural keys
+
+### Contract
+
+Randomness is scoped, not carried: no `Function` and no operation holds a seed. A **workflow scope** owns every ProbPipe-caused draw inside it, and each such **workflow-owned random event** receives a key derived structurally from the scope's root seed — never from a counter, ambient state, or the order of execution.
```
SoymilkWinsAgain, line 136: This description is inaccurate: when the same call is executed repeatedly or recursively within the same `workflow_run`, the occurrence path/source/logical unit listed in the documentation may be exactly the same, contradicting the statement "no key is drawn twice". Current implementations actually use canonical occurrence ordinals; therefore, the text should explicitly state that it "depends on deterministic logical ordinal numbers, but not on thread completion order or scheduling order."

A more detailed update to the RNG system design documentation can be added later by me. For now, only this sentence needs to be corrected.

design/03-values-and-distributions.md
```diff
@@ -272,51 +320,52 @@ class LinOp(Function, ABC):        # the linear subtype of the III.1 base
...
.... Its single stored source of type is its `DistributionSpec`, whose `event_spec`, the output declaration of one draw (an `OutputSpec`, II.2), it exposes as a view. Construction accepts a bare term spec or nested mapping data as a convenience, normalizing to an `OutputSpec` whose name defaults from the law's own constructor `name`, and the stored spec's class fixes the draw kind. It is `TrackedTerm` and `Annotated`. It declares the operations it supports as **capabilities**, so operational support is decoupled from the class. The draw a *user* sees is the tracked term of the declared kind: a `Record` for a record-valued law, and a `NumericArray` for a scalar one (III.1). It is not wrapped in another kind to make it uniform, so a draw's type follows the law's declaration rather than its field count. Fields are renamed or moved with `with_path_names` — a path-valued target restructures the event (II.1) — returning the same law under the canonical relabeling of its event space; on a factored joint the result is a relabeling view over the stored factors. A distribution whose declaration is polymorphic is legal, binding exactly as II.2 fixes — by value, or explicitly through `with_dims`.
```
SoymilkWinsAgain: The non-Record `OutputSpec.name` lacks a complete interface definition. It's a produced slot in composition, but it's stated here that term-drawing laws don't have fields; also, `with_path_names` only describes "event fields". It needs to be clarified whether the scalar output name can be renamed by `with_path_names`, whether it can be accessed by `d[name]`/`marginal`, and whether it's actually called a field or a slot.

design/02-shared-abstractions.md
```diff
@@ -76,178 +75,90 @@ Using named paths is necessary to satisfy `C5 – Naming for unambiguous meaning
...
 - *Abstract substrate.* `NamedTree` provides no constructor logic and is not directly instantiable; each concrete family's constructor owns storage and validation policy.
-- *Bare-name reference.* `with_path_names` renames by `old="new"` pairs, where a key may be a bare name instead of a full path: a bare name resolves to the unique node so named and raises when the tree contains it more than once. Keyword pairs therefore cover the common case, while the positional mapping form addresses any path, as in `with_path_names({"group1/mu": "loc"})`. The `Mapping` interface itself is untouched, with `[]` and `keys()` keyed by full path.
+- *Bare-name reference.* `with_path_names` renames by `old="new"` pairs, where a key may be a bare name instead of a full path: a bare name resolves to the unique node so named and raises when the tree contains it more than once. Keyword pairs therefore cover the common case, while the positional mapping form addresses any path, as in `with_path_names({"group1/mu": "loc"})`.
```
SoymilkWinsAgain: The `with_path_names` does not define swapping and renaming, moving ancestor and descendant nodes simultaneously, moving a node into its own descendant, multiple targets causing prefix conflicts, or the canonical order after a move. These must be fixed before implementation.

SoymilkWinsAgain (changes requested):

SoymilkWinsAgain:
Another complex problem is that the end2end definition of `raw` contract needs to be clarified. V.0 currently uses `raw=True` to mean both “return the implementer result untouched” and “return the raw side of the result kind.” These coincide only when the implementer already returns a bare array, mapping, or callable. If it returns a tracked `Function`, `Distribution`, or `LinOp`, “untouched” means returning that tracked term, whereas the kind-based rule requires unwrapping it to a callable, backend object, or “bare operator.” That unwrapping is not currently available: `Function` deliberately keeps its callable private, `NumericArray` and `LinOp` declare no raw accessor, and there is no common protocol across kinds. There is also a second-layer problem because every operation is itself a `Function`: even if its dispatcher returns a bare result for `raw=True`, the enclosing `Function.__call__` engine will wrap it again unless the engine explicitly intercepts this option, which its contract does not currently specify.

In short, 4 "raw"s are mixed up:

|Concept| Meaning|
|-- | --|
|implementer raw type |Methods such as `_mean` and `_sample` handle natural Python/backend types.|
|Function.apply | Calling wrapped callable, but no lifting, tracking, and provenance|
|raw=True |The user requested that the operation skip the final tracking|
|term.raw() |Retrieve the underlying storage from an existing tracked term|

Please choose one semantic and define the complete call path. AI recommends defining `raw=True` as “skip operation-created wrapping and identity attachment, returning the implementer result as produced,” so an already tracked result remains unchanged; if strict unwrapping is intended instead, every tracked kind needs a common extraction contract and a defined raw representation, including `Function`, `Distribution`, and lazy `LinOp`.

design/03-values-and-distributions.md
jhuggins: I like your proposal

design/04-functions.md
jhuggins: Great. I was hoping you could help with that.

design/02-shared-abstractions.md
```diff
@@ -60,194 +58,114 @@ class NamedTree[L]:
...
+```
 
-The class is what an operation **returns**, not what a leaf holds. `record["x"]` is a bare array, because a `NumericArraySpec` leaf holds one. An operation declared to return a `NumericArraySpec` returns a `NumericArray`.
+A spec accepts a value in that's either the raw representation or the tracked term of the spec's kind, with construction normalizing to the stored form. The rule is library-wide, since every accepting position is typed by a spec: a bare array is accepted at an array-spec position, a mapping at a record-shaped one, a callable at a function-shaped one, and a backend distribution at a distribution-shaped one, entering through its registered converter, while a tracked term is stored or handed on with its identity kept. The rule runs in both directions — wherever a raw value is accepted, its tracked form is accepted too — with one deliberate narrowing: an implementer method's *arguments* arrive already normalized to the implementer type, so an implementation handles exactly one presentation. What an implementer *returns* may be either presentation: the boundary keeps the kind and mints the result's identity afresh either way.
+d
```
arob5, line 103: Errant d?

SoymilkWinsAgain:
Please see #446 part B

design/04-distributions.md
SoymilkWinsAgain, line 17: The preceding text has been updated, so this section should be updated accordingly. Consider the following code:

```python
a = Normal("prior", loc=0.0, scale=1.0, component_name="x")
b = Normal("prior", loc=0.0, scale=1.0, component_name="y")

j = a * b
factor(j, "prior")
```

The component names for the two distributions are "x" and "y" respectively; there are no connection conflicts, and they can perfectly well form an independent joint distribution. However, both share the label "prior":
- If combination is disallowed, simply renaming "b" would enable it; thus, the label remains the deciding factor for whether combination can occur.
- If combination is allowed, the "unique factor name" condition is violated, and `factor(j, "prior")` cannot uniquely determine which factor to return.

If the "factor name" here is intended to be independent of the object label, the documentation currently fails to define the source or rules for such names.

I suggest adhering to the new principle that "labels do not affect combination" and consistently revising the requirements for factor uniqueness and the factor access interface. At the very least, we need to clarify how a specific factor is selected and how to handle cases where labels share the same name. While the specific generation of internal IDs can be deferred to the implementation phase, the method for user access to factors must be determined now.

design/05-functions.md
```diff
@@ -0,0 +1,313 @@
...
+
+```python
+@function(output_spec=OutputSpec(rate=NumericArraySpec(("obs",), float32, positive)))
+def rate(x):
+    return jnp.exp(x)
+# inference alone would read the support as real; the declaration carries support=positive
+# and binds "obs" to the actual output length on each call
+```
+
+**The result of a lift.** A distributional lift plans a `DistributionSpec` over the function's output declaration, not an empirical family. An exact rule may return a Gaussian or transformed law; the sampling route constructs an empirical approximation. With no output declaration the planned law's whole-term component is the function's `output_name`, and its type is completed at return. A sweep plans a batch of the declared output kind, retaining the swept levels. With no output declaration, row results must agree on one inferred spec; an empty sweep requires an explicit declaration sufficient to construct the result, so a name-only type hole is insufficient. A nested sweep returns a batch of laws. An operation's result rule follows the same distinction between declared type and chosen representation (VI.0).
```
SoymilkWinsAgain, line 142: Earlier, it was stated that in the absence of an output declaration, the output is inferred from the return value: records expose their individual fields, whereas other types use `output_name` for the entire value. However, the current description suggests that without an output declaration, the distributed result of lifting uses `output_name` as the component name for the whole, with the specific type structure resolved only at the return stage.

It is recommended to standardize this: function lifting should follow the same output inference rules as standard calls. When there is no declaration and the return structure is not yet known, the planning stage treats the output as unknown; the return stage then determines whether to expose record fields or declare a single "whole-term component."

SoymilkWinsAgain (approved):

Issue #446, opened 2026-09-05
Track regressions and unfinished contracts from #435–#437 and #439


SoymilkWinsAgain:
## One-line summary

Repair regressions and incomplete contracts from #435–#437 and #439, separating independent runtime fixes from design decisions and implementation work carried by #444.

## Motivation and audit baseline

The merged stack is partially complete. Historical execution identified **three existing workflows that start failing in #436**, additional defects in the new kinds and aggregation, and two accepted contracts that were never implemented.

Acceptance was checked against the commitments at merge time, using #434 as the accepted baseline:

| Boundary | Commit | Audit conclusion |
|---|---|---|
| #434, accepted baseline | d96411fbcde3ba63853d17f79f1bd13f878fbcae | Context only; not re-audited as part of this issue |
| #435 | b59fe2c7891a2142acd3b6a380f3841799e961ad | New kinds largely work; native-scalar storage is defective |
| #436 | 8a9d1dfe1d98b3cf0717e78880cd4af1b86b59e0 | Kind-directed outputs implemented, with three confirmed regressions |
| #437 | 405dd68c738c672322f58fc00d38c3f10f58656a | Many integration gaps fixed; executor consistency and declaration preservation remain incomplete |
| #439, audited endpoint | 55af8497fb5692c35180f93c96316535a178778f | Scoped naming/constructor/axis-partition migration passed the checks; no independent new regression confirmed |

Each merge was examined relative to its first parent, followed by the combined stack. Public behavioral probes ran against all five package snapshots with the same dependencies, adapting constructor spelling where necessary.

The [#434 design baseline](https://github.com/TARPS-group/prob-pipe/pull/434) and the [closing comment on #398](https://github.com/TARPS-group/prob-pipe/issues/398#issuecomment-5387759834) explicitly include raw access and tracked batch selection. Their absence is an unfinished commitment, not a newly introduced legacy regression. The later target design in #444 changes some of those semantics; the implementation should follow the finalized new contract rather than first rebuilding an obsolete intermediate API.

## Dependencies and work split

- **A: Independent runtime fixes.** These can be addressed against the audited runtime without waiting for #444 or adopting its new architecture. A1 and A5 overlap future execution/spec work, but the underlying correctness requirements already apply today.
- **B: Design work and deferred implementation through #444.** Finalize the relevant contracts in #444 and implement those contracts in follow-up code PRs. Some requirements are already decided; they need an explicit implementation handoff rather than another design decision.
[...]

PR #447, opened 2026-09-08
fix(core): Bug fixed & completion of #435-#439

SoymilkWinsAgain:
## Summary

Fix the independent regressions and contract gaps identified in A1–A6 of #446 after #435–#437 and #439:

- Preserve numeric sweep results, declared event specs, and named batch levels across `auto`, `sequential`, and supported `jax` dispatch; prevent traced conversions from leaking into native-array caches (A1).
- Support structural samplers without `name` or `name_is_auto` attributes (A2).
- Preserve complete opaque events when aggregating batched samples, including structured events and empty draws (A3).
- Retain native NumPy scalar storage and precision in `NumericArray` (A4).
- Preserve agreed numeric, opaque, and function declarations during row aggregation, and reject incompatible declarations (A5).
- Report the same actionable `TypeError` for mixed batch/non-batch rows in either order (A6).

Also move the numeric-array predicate into its consuming modules and remove the unused root `_utils.py`, updating import-graph test fixtures accordingly.

## Linked issues

Part A of #446. Part B requires completion of #444 first.

## Test plan

- [x] Add regression coverage for dispatch equivalence, numeric declarations and named levels, conversion-cache safety, unnamed samplers, structured/empty sample events, native scalar precision, object-batch declarations and downstream use, and mixed-row validation in both orders.
- [x] Run the latest focused A4–A6 tests with `pytest -n 4 --no-cov`: **40 passed**. Against the pre-fix snapshot, **14 cases fail as expected**, confirming that the tests detect the original defects.
- [x] Run the full suite locally with **4 workers**: **6,036 passed, 57 skipped, 1 XPASS**, plus one JAX worker abort. The affected test passes in isolation; the abort is consistent with local memory pressure. The crashed worker did not return coverage data, so the incomplete **80.70%** result does not verify the CI coverage threshold of 88%.
- [x] Run `ruff check .`, `ruff format --check .`, `pre-commit validate-config`, and `git diff --check`: all pass.
- [x] Complete CI validation across Python 3.12–3.14, the BayesFlow and Stan jobs, and notebook execution. BridgeStan is unavailable locally; these CI checks were still running when this description was updated.

## Documentation

## Checklist


SoymilkWinsAgain:
Fixed A1.

SoymilkWinsAgain:
Fixed A2

SoymilkWinsAgain:
Fixed A3-A6

probpipe/core/_empirical.py
```diff
@@ -108,17 +120,17 @@ def _index_record(record_data: Record, idx) -> NumericRecord:
     )
 
 
-def _fieldwise_op(record_data: Record, op: Callable) -> NumericRecord:
-    """Apply *op* to each leaf of a Record, returning a :class:`NumericRecord`.
+def _is_numeric_array(x: object) -> bool:
```
jhuggins, line 123: It seems like this should be in `_array_utils.py`

probpipe/distributions/_joint_empirical.py
```diff
@@ -54,6 +56,20 @@
 __all__ = ["JointEmpirical", "NumericJointEmpirical"]
 
 
+# --- Shared helpers ---
+def _is_numeric_array(x: object) -> bool:
```
jhuggins, line 60: second copy

jhuggins (changes requested):
Overall looks good. AI found a few things to fix before merge: 

## 1. The A1 tracer guard covers one of three identical cache sites (reproduced)

`NumericArrayBatch.as_jax` (`probpipe/core/_numeric_array_batch.py:214`) and `NumericRecord._child_field_as_jax` (`probpipe/core/_numeric_record.py:311`) still do the unguarded `object.__setattr__(self, "_jax_cache", cached)`. Build `NumericArrayBatch("draws", pd.Series([1.,2.,3.]), "draw", element_spec=NumericArraySpec((), dtype=np.float64))`, call `as_jax()` inside a `jax.jit` body, and afterwards `_jax_cache` holds a `DynamicJaxprTracer`, so every later `as_jax()` raises `TracerArrayConversionError`. A `NumericRecord` poisons per leaf the same way. This is reachable from our own code — `_broadcast_distributions.py:920` routes every aggregated row through that `as_jax()`. The docstring at `_numeric_array_batch.py:208` also now claims parity with `NumericArray` that no longer holds.

## 2. A5 adopts a declaration no row collectively made (reproduced)

`_broadcast_distributions.py:1036` skips non-`NumericArray` rows, so one tracked row's spec becomes the whole aggregate's `element_spec` — and `NumericArraySpec.is_valid` does not check `support` (`event_template.py:355`), so the adopted constraint is never verified. A sequential sweep returning `-1.0` for row 0 and `NumericArray("held", 2.0, spec=NumericArraySpec((), np.float64, support=positive))` for row 1 produces a batch declaring `support=positive` over the store `[-1., 2.]`. `test_a_raw_numeric_row_does_not_erase_a_tracked_rows_declaration` uses a positive raw row, so the hole is untested rather than intended. Adopting only when every row declared the spec would also dissolve the first item below.

## Worth fixing in the same pass

- **The same block rejects mixes that worked on main** (reproduced): because A4 now preserves native dtypes, an `np.float64`-backed row and a jax-backed row carry different spec dtypes and raise `ValueError`; an `int32`-declared row mixed with a float row fails inside `NumericArrayBatch` construction with a message naming neither the sweep nor the field. No test mixes native- and jax-backed rows.
- **A2 reports a caller-given name as auto-derived** (reproduced): at `ops.py:119,125` a sampler that sets `.name` without `name_is_auto` gets `True`, so aggregation may rename it. `getattr(dist, "name_is_auto", not hasattr(dist, "name"))` keeps A2 and preserves the rule. `test_ops.py:90` versus `:97` encodes the current behavior, so the test needs correcting too — a *present* name is dropped when `sample_shape=()`.
- **The A3 object branch lost its siblings' guard**: `ops.py:192` reshapes unconditionally where the numeric (`:207`) and `Record` (`:173`) branches return `drawn` unchanged on an axis mismatch, so an object array of shape `(5,)` with `sample_shape=(3,)` now raises a bare numpy reshape error.
- **`_is_numeric_array` was duplicated rather than moved**: byte-identical bodies at `_empirical.py:123` and `_joint_empirical.py:60`, with no test at either site. `probpipe/_array_utils.py` already holds `_is_array` and `_is_numpy_scalar` and has `tests/test_array_utils.py`.
- **`CHANGELOG.md:288` claims a fix that was not made**: the declaration agreement added here is in the numeric branch of `_make_stack`; the object/function/record agreement in `_agreeing_batch_rows:398` is pre-existing and untouched — which is also why `test_object_batch_rows_with_conflicting_declarations_are_refused` passes both before and after. Entries 1 and 4 describe the same block twice.
- **A4's guarantee is construction-time only**: storage keeps float64, but `as_jax()` still returns float32 `inf` for `1e100`, and `np.float64(0.1)` reads back as `0.10000000149011612`. Either document that scope or pin the conversion. Worth recording why `type(value) in (...)` replaced `isinstance` — it relies on `np.float64` subclassing `float` — and deciding whether `int`/`float` subclasses should still normalise, since they no longer do.

## Smaller

Four of six A4 parametrized cases don't discriminate, since only `np.float64` and `np.complex128` were ever converted pre-fix, and `with jax.enable_x64(False)` at `test_numeric_array.py:96` is a no-op. `test_workflow_sweep.py:508` pins the declaration but not the store's dtype. `_make_stack`'s `Raises` omits the new `ValueError`. Stale text at `_numeric_array_batch.py:373` ("Take *batch* apart" after the parameter rename), the `# --- Shared helpers ---` banner at `_joint_empirical.py:59`, and `CONTRIBUTING.md:390` still listing `_utils.py`. Two commit subjects break `<type>(<scope>): <subject>`, one of which misspells the deleted file as `ultils.py`.

SoymilkWinsAgain:
Fixed a very small unrelated place: type declaration is not rigorous (non-blocking nature) in `probpipe/core/_workflow_broker.py`.

SoymilkWinsAgain:
Fixed another type error

probpipe/core/_empirical.py
SoymilkWinsAgain: Moved to `_array_utils.py`

SoymilkWinsAgain:
All bugs fixed

probpipe/core/ops.py
```diff
@@ -158,9 +158,19 @@ def _drawn_at_its_batch_form(
...
-    if not sample_shape or isinstance(drawn, Batch):
+    if isinstance(drawn, Batch):
         return drawn
 
+    if not sample_shape:
+        if isinstance(drawn, TrackedTerm):
+            return drawn
+        from ._workflow_result import _wrap_as_term
+
+        result = _wrap_as_term(drawn, SAMPLE_LEVEL)
```
jhuggins: name should be set here, rather than using `object.__set_attr__` on the next line. So _wrap_as_term will need to take `name_is_auto` as an argument too.

jhuggins (approved):

jhuggins (approved):

Issue #448, opened 2026-09-15
Design-alignment workstreams: implementation tracks toward the design reference (supersedes #350)


jhuggins:
# Design-alignment workstreams — implementation tracks toward the design reference

> **📌 Standing directive.** Read [`CONTRACTS.md`](../blob/main/CONTRACTS.md) before implementing any item here and document every touched contract in its docstrings. Where an older issue disagrees with the design reference, **the reference ([`design/`](../tree/main/design)) wins.**

**Supersedes #350.** The reference merged in #444 restructured itself into seven parts and settled rules that change most open items there, and the #434–#439 stack, #400, and #447 changed the ground they stand on. This issue restates every remaining item against the reference as merged and renumbers the work in **implementation order** so that two or three people can work on disjoint files. #350 stays as the record of Waves 0 and 1 and of W3.1 and W3.2, all merged; its remaining ids retire, with the map in §4.

Ground rules carried over: everything targets **0.2.0** as breaking-change work with no compatibility shims; ids are planning anchors for this issue only, so **name PRs by the behavior they change** and link them to the item they complete; each item carries its design section, a size, and the items that must be on `main` first. Design sections are cited as `II.2`, `IV.1`, `V.7`, referring to [`02-shared-abstractions.md`](../blob/main/design/02-shared-abstractions.md), [`03-term-kinds.md`](../blob/main/design/03-term-kinds.md), [`04-distributions.md`](../blob/main/design/04-distributions.md), [`05-functions.md`](../blob/main/design/05-functions.md), [`06-operations.md`](../blob/main/design/06-operations.md), [`07-distribution-catalog.md`](../blob/main/design/07-distribution-catalog.md), and [`package-structure.md`](../blob/main/design/package-structure.md).

## 1. What changed in the reference since #350's last revision (2026-08-19)

**Structure.** The reference is seven parts. Part III, *Term Kinds*, holds only the kind definitions, III.1–III.10. Part IV, *Distributions*, is new and carries factored distributions, composition, and cross-type conversion as IV.1–IV.3. Functions, Operations, and the Catalog are Parts V, VI, and VII. Every citation in #350 is written against the six-part scheme and needs the map in §6.

**Rules that change open items.** Each is stated once in the reference; the section is given.

| change | where | items affected |
|---|---|---|
[...]

PR #449, opened 2026-09-15
[G1] refactor(all): remove `name_is_auto`

SoymilkWinsAgain:
## Summary

This is G1 in #448 

Modified the `factor` contract design: it now queries the component name instead of the label. Removed all `name_is_auto`. The detailed plan has been provided in #444, and the acceptance criteria in #448.

## Linked issue(s)

Refs: #444, #448.

## Test plan

No test added. But most of them changed due to the change of protocol.

## Breaking changes

- Remove the `name_is_auto` and `_name_is_auto` attributes, storage slots, constructor parameters, invocation arguments, and associated conditional branches.
- Retain `auto_name(name, default)` but modify it to return only the resolved string, and update all callers.
- Clean up markers related to copying, pickling, pytree helper data, workflow wrappers, and temporary containers, without introducing replacement markers.
- Remove the constructor's reservation of this name as a metadata keyword; containers with a field-keyword interface will now be treated according to standard field rules.

## Documentation

Modified the `factor` contract design following #444: it now queries the component name instead of the label. 

## Checklist


probpipe/core/_batch.py
```diff
@@ -897,7 +894,13 @@ def _at_axes(
 
         selection = self._compose_selection(normalized)
         label = _render_index(self._root_spec, selection)
-        name = f"{self._root_name}[{label}]" if label else self._root_name
+        name = (
```
jhuggins: this is pretty obscure; I would prefer a simpler if/else statement structure

probpipe/core/_batch.py
```diff
@@ -1009,10 +1006,6 @@ def _with_level_names(self, level_names: tuple[str, ...]) -> Self:
         renamed = self._shallow_copy()
         object.__setattr__(renamed, "_spec", replace(self._spec, level_names=level_names))
         object.__setattr__(renamed, "_root_spec", replace(self._root_spec, level_names=root_names))
```
jhuggins, line 1009: since renaming is no longer possible here, I would propose changing views to have the same name as the object they are a view into: 

`post[0:2].with_level_names(chain="ch")` keeps the name `post[chain=0:2]` while its levels are `('ch','iter')`

Instead, I would propose using provenance to track the parent of the view and what was selected (as otherwise the metadata would be identical for the object and its view).

probpipe/core/_batch.py
jhuggins, line 1001: the rename-refusal message still explains itself via a view "deriving its name from" its levels, a mechanism this PR removed.

jhuggins (approved):

jhuggins (changes requested):
Meant to request changes

probpipe/core/_batch.py
SoymilkWinsAgain: Using provenance to record selections is a sound design choice, but a direct implementation suffers from significant performance issues. Because the existing mechanism recalculates the content fingerprints of the entire parent batch every time. This is solvable but requires another PR...

SoymilkWinsAgain:
1, 3 chnaged. Left 2 because of performance issues

jhuggins (approved):

PR #450, opened 2026-09-16
[C1] refactor(core): exact-first dispatch with Feasibility reports; ResolutionError and MathematicalDomainError

jhuggins:
## Summary

Make dispatch exact-first and replace the integer priority tiers with a declared boolean. `probpipe/core/_registry.py` becomes `probpipe/core/_dispatch.py`, implementing II.7 of the design reference as revised here. Every method declares `name`, `exact`, `priority`, and `supported_types()`; a registry reads those once at registration and selects by exactness, then priority, then type specificity, then registration order. For example, `condition_on(model, y, exact_only=True)` refuses every built-in inference method, since each declares `exact = False`, and `inference_method_registry.check(model, y)` names the method that would run together with its declared exactness.

- Declare `exact: bool` on every method, fixed for its life; `priority: int | None` ranks methods of the same exactness, and `None`, the default, is opt-in-only: skipped by automatic selection, reachable through `method="..."`.
- Break priority ties by type specificity: the method whose declared types are closest to the arguments' classes in method-resolution order wins; a type that admits without appearing in the MRO, such as a registered virtual subclass, counts as least specific; a binary registry sums the two distances. `list_methods()` ranks by the other three criteria, since specificity depends on the call.
- Make `BaseDispatchMethod[SupportedTypesT]` state the whole method contract, generic over the shape of `supported_types`; `UnaryDispatchMethod` and `BinaryDispatchMethod` fix the shape through the exported `UnarySupportedTypes` and `BinarySupportedTypes` and add nothing else. Registration validates the runtime shape, which Python does not enforce and `issubclass` would otherwise admit silently.
- Snapshot the four declarations at registration and validate all of them before the registry changes: name a non-empty `str`, `exact` a `bool`, `priority` an `int` or `None` with `bool` rejected, `supported_types()` of the arity's shape. A method mutated afterwards changes nothing. `set_priorities` takes a positional mapping as well as keywords, validates every name and value before recording any override, cannot change exactness, and warns on a move into or out of opt-in-only.
- Split the reports: a method's `check` returns a `Feasibility` (`feasible: bool | None`, `description`, `pending`), where `None` means required declarations are unavailable and `pending` names them; the registry's `check` returns a `MethodInfo`, a `Feasibility` with `method_name` and `exact` set from the registration. `feasible` must be a `bool` or `None`, since registries compare it by identity. A feasible or unresolved `MethodInfo` names its method; only the infeasible report that lists every method tried names none.
- Raise `ResolutionError(Exception)` for no available implementation under the requested controls: no feasible method, an unresolved first candidate, or a `method=` name that is infeasible, excluded by `exact_only`, or not registered. It is not a `TypeError`, because well-typed arguments can still have no applicable method, and not a `LookupError`, whose meaning is a missing key or index; `get_method` and `set_priorities`, which look a name up, keep `KeyError`. `MathematicalDomainError(ValueError)` is defined beside it for known mathematical nonexistence and is never raised by the registry itself. Fewer positional arguments than the arity requires, none included and whether or not `method=` is given, is a `TypeError`. Both exceptions are exported from `probpipe`.
- Split the conditioning capability in two. `SupportsConditioning` becomes `SupportsExactConditioning` and `SupportsApproximateConditioning`, so a distribution with a built-in conditioning path declares whether that path returns the conditional law or a stand-in for it, and `condition_on` has one capability route for each. Both are abstract base classes rather than runtime-checkable protocols: a class claims one by inheriting it, and a class that merely defines `_condition_on` claims neither. Exactness is a claim about the result rather than a fact about the method, so no structural check can read it, and two protocols declaring the same `_condition_on` would match the same classes.
- Make `exact_only` a control of `condition_on`, a keyword-only parameter beside `method` that the operation consumes and resolves before selecting a route. It previously fell through to `**kwargs` and was forwarded into `_condition_on` as if it were observed data, so conditioning a `ProductDistribution` with `exact_only=True` raised `KeyError` and an amortized posterior dropped the control and answered approximately. The route order is a named method, then the exact conditioning capability, then the approximate one, then registry auto-selection. Exactness is compared across route sources, not only within the registry: the approximate capability runs only when the registry has no feasible exact method for the call, and `exact_only` skips it altogether.
- Remove `JointEmpirical`'s conditioning route. Its `_condition_on` dropped the named fields and kept every atom and weight, so it ignored the value conditioned on and `condition_on(je, x=0)` and `condition_on(je, x=1)` returned the same distribution. That is marginalization rather than conditioning of either kind, so the route is removed rather than relabeled, and the class docstring says how to build the marginal directly.
- Declare `exact = False` once for every built-in inference method through `InferenceMethod`, now a subclass of `UnaryDispatchMethod` rather than an alias. The former priority-0 methods (`blackjax_hmc`, `blackjax_sghmc`, `pymc_advi`, `tfp_nuts`, `tfp_hmc`) are `None` and the other ranks are unchanged, so auto-selection is unchanged.

Scope: this PR supplies the dispatch contract and moves the inference registry onto it. The converter registry is not a dispatch registry and is unchanged; it keeps its `TypeError` and `ConversionMethod` until its move to `distributions/_conversion.py`. Protocol checks in `ops.py` keep raising `TypeError`: `log_prob("a string")` still does, because a `str` does not support it, and only the registry's "nothing applicable" failure moved.

The design edits are in the same PR. They replace the two-member `Fidelity` enum with the boolean throughout Parts II, IV, V, VI, and the package structure, on the grounds that exactness is a fact about a returned representation rather than a scale: the only candidate for a third level, a consistent sampler against a biased approximation, is a method guarantee the reference already records as such. `exact_only` replaces the `min_fidelity` control.

## Review guide

**The two new files and the design section carry the substance.**

| File(s) | What to review |
| --- | --- |
| **New** `core/_dispatch.py` | The whole module, including that the type cache is dropped when `abc.get_cache_token()` moves, since registering a virtual subclass changes what `issubclass` answers with no registry call: the generic method base and the two alias types; `Feasibility` and `MethodInfo` with their construction invariants; `_Registration` and the validation order in `register` and `set_priorities`; `_rank`, `_find_methods`, and the `_distance` hook that implements specificity; `check` and `execute`, in particular that the dispatch key is computed before the `method=` branch; `ResolutionError` and `MathematicalDomainError`. |
| **Deleted** `core/_registry.py` | Replaced, not shimmed. Every importer moves to `core._dispatch`. |
| `design/02-shared-abstractions.md` (II.7) | The code block and the prose are the contract the module implements: selection order with the specificity, snapshot, and arity rules; the two report classes; the exception bases. `tests/core/test_dispatch.py::TestDesignAgreement` reads the block and compares declared classes and type aliases, exception bases, report fields, and registry signatures with their defaults and positional-only marker against the implementation, so the two cannot drift silently. |
| `inference/_registry.py` | `InferenceMethod` is a subclass declaring `exact = False`; the module re-exports `Feasibility`. |
| `inference/__init__.py` | Imports move to `core._dispatch`; the two exceptions and the two alias types are re-exported; the TFP registrations pass `None`. |
| `core/protocols.py` | The two conditioning capabilities, and why they are claimed by inheriting rather than structurally. |
| `core/ops.py` | `condition_on` resolves `exact_only` before selecting a route, so a control never reaches `_condition_on`; the docstring documents the four routes and gains `Returns` and `Raises` sections. |
| `pyrightconfig.json`, **New** `tests/typecheck/dispatch_shapes.py` | pyright reads the fixture, pytest does not: a unary method returning the binary shape is rejected, and `reportUnnecessaryTypeIgnoreComment` fails the file if pyright ever stops rejecting it. The typecheck job is advisory; the runtime shape check at registration is what gates. |

Focused tests:

- **New `tests/core/test_dispatch.py`**, each invariant parametrized over the unary and binary registries where the shape allows: exactness declared once and checked as a real boolean; `supported_types` abstract on the base and the arity classes adding no abstract members; a seeded property test that every exact method precedes every approximate one for any ranks; specificity in either registration order, below priority and exactness, per call, rebuilt after re-ranking, a virtual subclass least specific, the binary distance the sum; a seeded property test that no `set_priorities` sequence lifts an approximate method above an exact one; the mapping and keyword forms, their conflict rule, a non-identifier name, and non-integer values rejected before anything is stored; one warning per crossing of opt-in and none from an aborted override; `exact_only` in automatic and named dispatch; `check` returning the first feasible or unresolved candidate with its registered name and exactness, naming every method tried when none is, never executing, and unable to be overridden by a `check` that returns a contradicting `MethodInfo`; `execute` raising `ResolutionError` on nothing feasible, an unresolved first candidate, a named infeasible method, and an unknown name, and propagating a method's own exception with no second attempt; `TypeError` for too few positional arguments in named and automatic dispatch alike, without the method's `check` running; `MathematicalDomainError` never wrapped; the construction invariants of both report classes, malformed `feasible` values included; a rejected registration or override leaving the registry as it was; a mutated method changing nothing, admission on a fresh key included; caches; the pre-filter; the design agreement.
- **`tests/core/test_protocols.py`**: each implementation claims the capability it should and not the other, and a class that defines `_condition_on` without inheriting claims neither, which is what pins the capability as nominal.
- **`tests/core/test_ops.py`**: an exact-only call keeps the exact capability route; the controls never reach `_condition_on`; an exact-only call skips the approximate route without running it.
- **`tests/inference/test_inference_registry.py`**: every registered method is approximate; the opt-in set is exactly the expected one; the built-in ranks match the documented anchors; an unknown `method=` raises `ResolutionError`. Three classes that tested the generic registry through a unary fake are folded into the suite above.
- **Deleted `tests/test_binary_registry.py`**: folded into the parametrized suite.

Small additional changes that are **not just renames**:

- The inference-method docstrings lose their tier ranges; each leads with its registered name, priority or opt-in status, and applicability, with the rank rationale in a Notes section. `_blackjax_mcmc.py` and `_blackjax_sgmcmc.py` pass `None` for the opt-in methods where they passed `0`, and their `priority` types widen to `int | None`.
- `docs/api/extending.md` replaces the tier tables with "Exactness, then rank", the contributor-facing criteria for ranking a new method against the built-in anchors; `docs/api/inference.md` and `CONTRIBUTING.md` state the selection criteria once each and point there; `design/04`, `05`, `06`, and `package-structure` carry the `Fidelity` removal and the `Feasibility` return type; `docs/user_guide/05_external_backends.ipynb` no longer sorts by `-priority`, catches `ResolutionError`, and describes the opt-in methods as having no priority.
- `tests/core/test_ops.py`, `tests/modeling/test_simple_model.py`, `tests/inference/test_blackjax_mcmc.py`, and `tests/inference/test_blackjax_sgmcmc.py` expect `ResolutionError` and `None` where they expected `TypeError`, `KeyError`, and `0`.
- `.github/labeler.yml` names `core/_dispatch.py`.

Outside these, the consumer edits are import migrations from `core._registry` to `core._dispatch` and from `MethodInfo` to `Feasibility` in each method's `check`, with `method_name=self.name` dropped since the registry supplies it.

## Linked issue(s)

Part of #448 — implements C1 (the `exact` flag and the registry errors), and the specificity tie-break that was listed under C4.

## Breaking changes

- Rename `probpipe.core._registry` to `probpipe.core._dispatch`; no shim.
- `BaseDispatchMethod` is generic over the shape of `supported_types`, which it declares as abstract; `exact` is abstract; `priority` is `int | None`, default `None`; `OPT_IN_ONLY_PRIORITY` is removed and `0` is an ordinary rank. Registration rejects a non-string name, a non-boolean `exact`, a non-integer or boolean `priority`, and a `supported_types()` value of the wrong shape; `set_priorities` rejects a non-integer or boolean value.
- A method's `check` returns a `Feasibility`, not a `MethodInfo`; `MethodInfo(Feasibility)` adds `method_name` and `exact`, which the registry sets. `feasible` must be a `bool` or `None`.
- `BaseDispatchRegistry.execute` and `check` raise `ResolutionError` where they raised `TypeError` for no feasible method and `KeyError` for an unknown `method=` name; code that caught `TypeError` from `condition_on` to mean "no method for this model" must catch `ResolutionError`. `check()` with no positional arguments raises `TypeError` instead of returning an infeasible report, and too few positional arguments raise `TypeError` in named dispatch too.
- Among methods of equal exactness and priority, type specificity decides before registration order.
- `JointEmpirical` offers no conditioning route, so `condition_on` on one raises `ResolutionError`. Build the marginal directly, by constructing a `JointEmpirical` from the fields to keep with the same `weights`.
- `SupportsConditioning` is replaced by `SupportsExactConditioning` and `SupportsApproximateConditioning`, which are abstract base classes rather than protocols, so an implementation claims one by inheriting it. `ProductDistribution`, `JointGaussian`, `SequentialJointDistribution`, and `JointEmpirical` claim the exact capability; `BayesFlowModel` claims the approximate one.
- `condition_on` takes `exact_only` as a keyword-only parameter, so it is no longer forwarded to `_condition_on` or to a registered method as an inference kwarg.
- `InferenceMethod` is a subclass of `UnaryDispatchMethod` rather than an alias, so a third-party method subclassing `UnaryDispatchMethod` directly must declare `exact` itself.
- `set_priorities(mapping | None, /, **kwargs)`; `check(..., exact_only=False)`; `execute(..., exact_only=False)`.

## Documentation

- [x] CHANGELOG updated

## Checklist


probpipe/core/ops.py
```diff
@@ -646,7 +664,7 @@ def condition_on(
     if isinstance(dist, SupportsConditioning):
```
SoymilkWinsAgain: `exact_only` doesn't cover the direct conditioning path if these branches here. The `SupportsConditioning` branch bypasses the registry and forwards this parameter directly to `_condition_on`. 

Reproduced: `BayesFlowModel` still returns an approximate posterior with `exact_only=True`. An ordinary `ProductDistribution` treats it as an observation field and raises ``KeyError`.

It is recommended to handle this control before selecting the route, allowing exact conditioning while restricting approximate paths

CONTRIBUTING.md
```diff
@@ -600,7 +600,7 @@ uv build packaging/probpipe   # probpipe (metapackage)
...
 | `Module` | Stateful workflow-aware base class (see `@workflow_method`) |
 | Protocols | `SupportsSampling`, `SupportsLogProb`, `SupportsMean`, `SupportsConditioning`, etc.; dynamic inclusion on `ProductDistribution` and `TransformedDistribution` |
-| `BaseDispatchRegistry` | Abstract base for priority-based registries: holds registration, priority management (incl. opt-in-only sentinel + override warnings), and the `check`/`execute` loop. Arity-specific subclasses override `_cache_key`, `_find_methods`, and `_format_key`. |
+| `BaseDispatchRegistry` | Abstract base for the dispatch registries: holds registration, exactness-then-rank ordering, opt-in filtering (`priority=None`) with override warnings, and the `check`/`execute` loop. Arity-specific subclasses override `_cache_key`, `_find_methods`, and `_format_key`. |
```
SoymilkWinsAgain: The document needs to update. The document still tells subclasses to implement `_find_methods`. The new implementation requires `_cache_key`, `_validate_supported_types`, `_distance`, and `_format_key`. A subclass following the current instructions cannot be instantiated because abstract methods remain unimplemented.

SoymilkWinsAgain (changes requested):

SoymilkWinsAgain (approved):
All bugs are fixed. Everything's fine

PR #453, opened 2026-09-17
[G2] feat(core): unify specs and add input/output declarations

### LINKED ISSUE #167: Consolidate single-field shims into a Scalar/Singleton wrapper

SoymilkWinsAgain:
## Summary

Unify record schemas and kind specifications, and add explicit input/output declarations. `RecordSpec` is now the schema itself instead of wrapping an `EventTemplate`. `InputSpec` describes named input slots; `OutputSpec` distinguishes a named whole term from exposed record components. For example, `OutputSpec(x=NumericArraySpec((3,)))` declares an array, while `OutputSpec(RecordSpec(x=(3,)))` declares a one-field record. Neither form implicitly converts into the other.

- Support all three output forms, single-keyword type holes (`OutputSpec(x=None)`), and derived `spec`/`components` views.
- Use one `TermSpec` protocol, with `NumericSpec`/`NumericRecordSpec`, partial `with_dims`, simultaneous `with_dim_names`, and binding from either values or declarations.
- Share dimension bindings across nested records, batches, callable/distribution specs, and input slots. Validate actual record field shapes and dtypes; do not substitute an inferred schema for the actual data. Collect bindings before constructing the result, so boolean validation does not build discarded schemas.
- Store input slots in a copied private dictionary for constant-time lookup. Use frozen slotted dataclasses for `InputSpec` and `OutputSpec`.
- Tighten numeric projection and vector reconstruction: prune empty subtrees, preserve numeric leaf metadata/order, propagate layout errors, and reject unsupported non-array leaves at reconstruction.

Scope: this supplies the shared declarations. Live `Function` input/output-template and distribution event-template APIs retain their current signatures and packaging rules; their full declaration migration belongs to the subsequent workstreams. Existing sampling/callable adapters remain at those boundaries. Strict spec-to-spec matching does not unwrap a one-field record.

Note: The current PR deliberately adds a temporary fallback `OpaqueSpec` branch to correctly handle `Distribution` objects that lack an `event_template`. This should be removed once the final design is implemented.

## Review guide

The large file count comes mainly from replacing `EventTemplate` -> `RecordSpec`, `NumericEventTemplate` -> `NumericRecordSpec`, `ValueSpec` -> `TermSpec`, and updating imports.

**Only the files listed blow are important.**

| File(s) | What to review |
| --- | --- |
| **New** `core/_specs.py` | `InputSpec` and the three `OutputSpec` forms; component validation, holes, derived views, mapping equality/hash, shared input dimension scope, and slots. |
| **New** `core/_spec_base.py` | The shared `TermSpec`/`NumericSpec` protocol and array/opaque specs; one spec-unification entry point; shape/dtype checks, dimension substitution and renaming. |
| **New** `core/_record_spec.py` | The merged record schema/kind spec, numeric promotion and projection, canonical leaf layout, inference, recursive value validation and dimension binding. Much is moved from the deleted `core/event_template.py`; focus on the changed contracts rather than treating every added line as new logic. |
| **New** `core/_kind_specs.py` | Distribution/callable declarations and recursive binding of their declared sides. Missing callable declarations remain unspecified; `FunctionSpec.is_valid` still checks callability without executing the function. |
| `core/_batch.py` | Batch dimension hooks now participate in the common binding/substitution protocol. |
| `core/_function_contract.py` | Use `lifted_names` to distinguish actual-value validation from lifted element-spec matching, sharing one binding scope. Keep legacy distribution packaging adaptation outside strict unification. |
| `core/record.py`, `core/_record_batch.py`, `core/_workflow_result.py` | Store the record schema directly, remove wrapper normalization, and preserve the declaration through reconstruction, pytree and workflow result wrapping. |
| `core/_numeric_record.py`, `core/_numeric_record_batch.py` | Direct schema storage and canonical leaf traversal for vector reconstruction, including explicit rejection of non-array numeric leaves and dtype preservation. |
| Deleted `core/event_template.py`; `core/_opaque.py` | Implementations move into the four spec modules above; the duplicate record wrapper disappears and `OpaqueSpec` moves to `_spec_base.py`. |

Focused tests:

- **New `tests/core/test_specs.py`**: design II.2 examples; kinds and type holes; no implicit record unwrapping; mapping isolation/equality/hash; direct/nested binding consistency; dtype validation; slots, copy/pickle and JIT.
- **`tests/core/test_event_template.py`**: migrated schema contracts plus numeric projection, empty subtrees, metadata preservation, dimension binding and numeric-layout regressions.
- **`tests/core/test_first_class_function.py`**: workflow integration, particularly mixed actual values and lifted inputs sharing dimensions.

Small additional changes that are **not just renames**:

- Runtime `assert` statements are removed. Missing conversion `sample_shape` now raises `RuntimeError` in `converters/_scipy.py` and `_tfp.py`; missing distribution storage/backend raises `RuntimeError` in `core/_distribution_array.py`.
- `core/_broadcast_distributions.py` simplifies the mutually exclusive `n`/`batch_shape` check and adopts the binding hooks. `core/_empirical.py` simplifies access to a constructor-validated sole field. `core/_function_batch.py`, `_numeric_array_batch.py`, `_opaque_batch.py`, and `inference/_approximate_distribution.py` use type narrowing for invariants already established at construction. The main files above also remove redundant assertions.
- `docs/api/records.md`, `CONTRACTS.md`, `CHANGELOG.md`, and the two edited `design/` files describe the declarations, partial binding and the implementation boundary.

Outside these changes, the consumer, documentation and test edits are mechanical name/import migrations, removal of obsolete template-wrapper access, and corresponding expected-schema updates. They can be reviewed as a migration sweep rather than independent feature implementations.

## Linked issue(s)

Part of #448 — implements G2 (the declarations).

Closes #167 and #357 by the design decision: there is no single-field wrap in the new declarations, rather than introducing a Scalar/Singleton wrapper. These issues are already closed; this PR implements that decision at the shared declaration layer.

## Breaking changes

- Remove `EventTemplate`, `NumericEventTemplate`, `ValueSpec`, and `core.event_template`; use `RecordSpec`, `NumericRecordSpec`, and `TermSpec` through the public exports.
- `RecordSpec` is the schema directly; consumers must remove the former wrapper's `.event_template` indirection.
- Public dimension-binding methods return a refined spec. `with_dims` now permits partial substitution. Custom specs use the shared binding/substitution hooks.
- Numeric projection drops empty branches and rejects an empty result; vector reconstruction requires `NumericArraySpec` leaves.
- No deprecated aliases or legacy-pickle compatibility layer is added.

## Documentation

- [x] CHANGELOG updated

## Checklist


SoymilkWinsAgain:
Note: The current PR deliberately adds a temporary fallback `OpaqueSpec` branch to correctly handle `Distribution` objects that lack an `event_template`. This should be removed once the final design is implemented.

probpipe/core/_spec_base.py
```diff
@@ -0,0 +1,485 @@
...
+    Specs are immutable, hashable, and compare by declaration. Symbolic
+    dimensions share a scope through nested schemas and declarations.
+    """
+
+    @property
+    def is_concrete(self) -> bool:
+        """Whether no symbolic dimensions remain."""
+        return not self.free_dims
+
+    def with_dims(self, **sizes: int) -> Self:
```
jhuggins, line 40: Should we rename to `with_dim_sizes` for clarity and symmetry with `with_dim_names`?

probpipe/core/_spec_base.py
```diff
@@ -0,0 +1,485 @@
...
+        """
+
+
+class NumericSpec(TermSpec):
+    """A spec whose values have a flat numeric layout; this adds no kind."""
+
+    @property
+    @abstractmethod
+    def vector_size(self) -> int:
+        """The number of scalar coordinates, defined only for concrete specs."""
```
jhuggins, line 247: Implement the desired behavior here for when the spec isn't concrete. Implementations would then implement `_vector_size` with the guarantee it's only called if the spec is concrete.

probpipe/core/_spec_base.py
```diff
@@ -0,0 +1,485 @@
...
+
+def _require_hashable(value: Any, *, context: str) -> None:
+    """Fail at construction when a schema component cannot be hashed."""
+    try:
+        hash(value)
+    except TypeError as error:
+        raise TypeError(f"{context} must be hashable: {error}") from None
+
+
+class TermSpec(ABC):
```
jhuggins: Should TermSpec implement `Immutable`?

probpipe/core/_record_spec.py
```diff
@@ -0,0 +1,715 @@
...
+            elif isinstance(spec, NumericArraySpec):
+                result[name] = spec.shape
+            else:
+                result[name] = (spec.vector_size,)
+        return result
+
+    def _compute_vector_size(self) -> int:
+        """Total scalar count across all numeric leaves."""
+        if self.free_dims:
+            return 0
```
jhuggins, line 620: I would suggest `None` rather than `0`

probpipe/core/_spec_base.py
```diff
@@ -0,0 +1,485 @@
...
+class NumericSpec(TermSpec):
+    """A spec whose values have a flat numeric layout; this adds no kind."""
+
+    @property
+    @abstractmethod
+    def vector_size(self) -> int:
+        """The number of scalar coordinates, defined only for concrete specs."""
+
+
+@dataclass(frozen=True, eq=False, init=False)
```
jhuggins, line 250: Should include `slots=True`

probpipe/core/_spec_base.py
```diff
@@ -0,0 +1,485 @@
...
+            if previous != actual_dimension:
+                raise ValueError(
+                    f"{path} binds symbolic dimension {declared_dimension!r} to "
+                    f"{actual_dimension}, but it is already bound to {previous}"
+                )
+        concrete.append(actual_dimension)
+    return tuple(concrete)
+
+
+@dataclass(frozen=True)
```
jhuggins, line 455: should include `slots=True`

jhuggins (changes requested):
Some additional recommendations from AI: 

## Must fix

**1. Restore the concrete-spec guard in `DistributionSpec._bind_dims_from_value`** — [[_kind_specs.py:50](https://claude.ai/epitaxy/probpipe/core/_kind_specs.py:50)](probpipe/core/_kind_specs.py:50). Confirmed regression: pre-PR `_unify_spec_with_value` gated the lenient path on `spec.free_dims` and ran `is_valid` on concrete specs. Today `RecordSpec.is_valid` returns `True` for a value that `Record` construction then rejects (dtype widening, field reordering, `support` present-vs-absent). Match the base class at `_spec_base.py:193` and `FunctionSpec`. Add coverage while there — [[_kind_specs.py:56](https://claude.ai/epitaxy/probpipe/core/_kind_specs.py:56)](probpipe/core/_kind_specs.py:56) `DistributionSpec._bind_dims_from_spec` is entirely untested across all of `tests/core`, which is why this got through.

**2. `len(template)` → `len(template.children)`** — [[_function_contract.py:236](https://claude.ai/epitaxy/probpipe/core/_function_contract.py:236)](probpipe/core/_function_contract.py:236), repeated at [[_kind_specs.py:194](https://claude.ai/epitaxy/probpipe/core/_kind_specs.py:194)](probpipe/core/_kind_specs.py:194). The sampling-lift adapter is declared in scope and its `not isinstance(expected, RecordSpec)` guard correctly keeps strict matching from unwrapping — the bug is only that the condition counts leaves where the docstring says "a sole field", so `RecordSpec(outer=RecordSpec(inner=(3,)))` satisfies a declared `NumericArraySpec((3,))`. Worth giving the "Interim implementation detail" comment an issue number at the same time.

**3. Fix empty-child numeric promotion** — [[_record_spec.py:81](https://claude.ai/epitaxy/probpipe/core/_record_spec.py:81)](probpipe/core/_record_spec.py:81), together with the `_substitute_dims` face of the same bug at `:223`/`:330`. `RecordSpec(theta=(3,), aux={})` and `RecordSpec(theta=(3,), aux=NumericRecordSpec())` compare equal with equal hashes and both report `is_numeric`, but only one has `vector_size`; and `NumericRecordSpec(theta=("d",), aux=NumericRecordSpec()).with_dims(d=3)` demotes to plain `RecordSpec`. Either promote on an empty numeric child or refuse an empty mapping child. The `__init__` face is new in this PR; the `with_dims` face is pre-existing (`event_template.py`'s `_replace_template_dimensions` had it too), but they share a root cause and should be fixed together.

**4. Two edits to `docs/user_guide/02_records.ipynb` cell 36.** Drop the sentence "`record.spec` is a `RecordSpec` wrapping the template, and `record.event_template` is a view on it — so the two cannot disagree", which describes the model this PR deletes, and fix "An `RecordSpec`". `mkdocs.yml:54` is `execute: false`, so this publishes regardless of CI. The rest of the cell is correctly migrated.

## Should fix

**5. Catch `TypeError` in the `infer_from` distribution branch** — [[_record_spec.py:503](https://claude.ai/epitaxy/probpipe/core/_record_spec.py:503)](probpipe/core/_record_spec.py:503) and `:697`. `getattr(val, "event_template", None)` swallows only `AttributeError`, but `_numeric_record_distribution.py:270-280` raises `TypeError` when `_name` is unset or `event_shape` isn't derivable — the condition `DistributionSpec.is_valid` documents and catches. New in this PR. The branch is provisional by the PR's own note, so the bar is lower, but it should fall back rather than raise.

**6. Restore coverage of the unbound-dimensions error.** `_concretize_event_template` raises at [[_record_spec.py:687](https://claude.ai/epitaxy/probpipe/core/_record_spec.py:687)](probpipe/core/_record_spec.py:687), live-called from [[_function_contract.py:285](https://claude.ai/epitaxy/probpipe/core/_function_contract.py:285)](probpipe/core/_function_contract.py:285) and [[node.py:780](https://claude.ai/epitaxy/probpipe/core/node.py:780)](probpipe/core/node.py:780); `grep -rn "unbound symbolic" tests/` returns nothing. The old test asserted the message and the rewrite replaced it with a weaker `free_dims` check — the one contract genuinely lost in an otherwise faithful migration.

**7. Document `_kind_specs.py`.** It's absent from `design/package-structure.md`, whose line 29 still attributes `TermSpec`/`NumericArraySpec`/`OpaqueSpec` to `_specs.py` (now a re-export) and line 160 still describes a different four-way split. `CONTRACTS.md:96-98` dropped the owner rows for `DistributionSpec`, `FunctionSpec`, `NumericArraySpec` and `OpaqueSpec` — and the two kind specs are where docstrings shrank most (`DistributionSpec` 821→226 chars, `FunctionSpec` 1832→769). Neither has an mkdocstrings entry.

**8. Rename `tests/core/test_event_template.py`.** 96 KB of tests for three new modules under the name of the module this PR deletes, with a docstring pointing at `probpipe.core.record.RecordSpec`, which doesn't hold it. STYLE_GUIDE §1.12 and §8.1. Splitting into `test_record_spec.py` / `test_spec_base.py` / `test_kind_specs.py` is the tidier option but a rename plus a corrected docstring clears the convention.

**9. Repair three tests that stopped discriminating.** [[test_event_template.py:584](https://claude.ai/epitaxy/tests/core/test_event_template.py:584)](tests/core/test_event_template.py:584) `test_specs_are_frozen` writes `.shape`, which `RecordSpec`/`DistributionSpec`/`FunctionSpec` don't own, so it passes on the absent attribute alone — write a real field, as `test_specs.py:108` already does. (Its widening from `FrozenInstanceError` to `AttributeError` was *correct* and should stay: `RecordSpec` raises the base class.) [[:1954](https://claude.ai/epitaxy/tests/core/test_event_template.py:1954)](tests/core/test_event_template.py:1954)'s `match=r"does not conform|expected named fields"` accepts either message for both loop cases — parametrize per case. [[test_specs.py:191](https://claude.ai/epitaxy/tests/core/test_specs.py:191)](tests/core/test_specs.py:191) covers only array-expected/record-supplied; assert the reverse direction too.

**10. Correct `InputSpec`'s `Raises` section** — [[_specs.py:58](https://claude.ai/epitaxy/probpipe/core/_specs.py:58)](probpipe/core/_specs.py:58). Two independent `TypeError` conditions are written as one conjunction, and "type holes are not admitted" is filed under `ValueError` when `_check_component` raises `TypeError`.

**11. Fix the stale "a view on" prose** — `record.py:583` and `CONTRACTS.md:101`. Keeping `Record.event_template` is declared in scope, so the dual name itself is your call and not a finding; describing it as a view on `spec` when both return the same object is simply wrong now.

**12. Decide the converter exception class.** [[_scipy.py:295](https://claude.ai/epitaxy/probpipe/converters/_scipy.py:295)](probpipe/converters/_scipy.py:295) and [[_tfp.py:290](https://claude.ai/epitaxy/probpipe/converters/_tfp.py:290)](probpipe/converters/_tfp.py:290) raising `RuntimeError` is a declared change, so this isn't an oversight — but it now disagrees with `converters/_registry.py:99`, which raises `TypeError` for the identical `plan.sample_shape is None`. Either align or accept the split knowingly.

## Minor

One CHANGELOG line for the fingerprint change: `_fingerprint.py:346,374` hash `type(spec).__module__`, which moved, so pre-PR persisted provenance fingerprints no longer match — the breaking-change list covers pickles but not this. The 22 deleted `__all__` grouping comments in `probpipe/__init__.py`. `docs/api/records.md:51` names `bind_dims_from_value`/`bind_dims_from_spec` on declarations that only have `with_dims`/`with_dim_names`. Unreachable `TypeError`s at `_specs.py:208,217` (two lenses agreed). Stale `event_template` vocabulary in `_unify_event_template_with_value` and `_concretize_event_template`, neither of which reads that attribute (two lenses agreed). `_record_spec.py:70`'s message reads "an RecordSpec" and offers a redundant alternative, since `issubclass(RecordSpec, TermSpec)` is true; same article error at `tests/core/test_record.py:524`. Missing `Raises` on the public `components` at `_distribution_array.py:392`. Redundant tests: `test_event_template.py:1503/1566/1599` assert one contract three times, and `:1530-1544` duplicates `:1611-1619` verbatim; `:2242+` asserts private `_bind_dims_*` against STYLE_GUIDE §8. `RecordSpec(**{"": (3,)})` constructs but every dimension operation then raises (pre-existing). Uncovered: `_kind_specs.py:210`, `_spec_base.py:142,323,437`, `_specs.py:117,130`.

SoymilkWinsAgain:
In order to save memory. I added `slots=True` to `NumericArraySpec` and `OpaqueSpec`. To preserve the old behavior, I also added `__slots__ = ()` to `NumericSpec` and `__slots__ = ("__weakref__"),` to `TermSpec`.

probpipe/core/_record_spec.py
SoymilkWinsAgain: Yes. This is a good idea

probpipe/core/_spec_base.py
SoymilkWinsAgain: Agree

probpipe/core/_spec_base.py
SoymilkWinsAgain: Agree

jhuggins (approved):

